PR #27, opened 2022-02-08
adding tests, Make file and interfaces

aronchick:
There's a lot here - happy to walk through if you'd like. I've started creating interfaces to allow for dependency injection for tests, added a makefile, and a bunch of mocking.


lukemarsden:
We started looking at this but it wasn't clear why the "networker" interface touches cobra.

We agree that interfaces are a good idea. Maybe better to start with a design doc, get consensus with us on the interfaces, then implement it?

aronchick:
ok, so the direct answer to your question is that in order to test cobra, you need to (as near as i can tell) both get the cobra.command object and the arguments the function was called with. That's why i have getter and setters for both - so that i can use them in both live and tests.

I'm happy to work on a design doc - do you have a format you'd like?

lukemarsden:
I think just sketch out the interfaces in a doc so we can iterate on those,
thanks!

aronchick:
Lol, genuinely not sure what you are looking for here (I assume you don't want me architecting all objects eventually supported) - here's what I wrote as a starter - https://github.com/filecoin-project/bacalhau/wiki/Interfaces

cc @binocarlos

lukemarsden:
I think this is out of date now

PR #55, opened 2022-03-14
Reformatting logs for multi threading

aronchick:
Moved over to zerolog, and merged the various PRs together to attempt to clean up.


binocarlos:
`make test` is passing locally - for some reason I can't get an action to trigger to check that CI is also passing so let's merge and see

Issue #228, opened 2022-06-08
Otel support in REST


lukemarsden:
Support OpenTelemetry in our REST interactions (across client/server).

Refactoring the codebase to use contexts more idiomatically (at the moment we pass them into nodes as part of their constructor). And then the plan is to thread the trace IDs around using contexts, which should magically work

Goal: be able to view traces from our production network in honeycomb.

lukemarsden:
See also: https://github.com/filecoin-project/bacalhau/issues/97

PR #235, opened 2022-06-10
Refactor Transport interface to use contexts for all APIs, fix IPFS daemon leakage bug.

Bubblyworld:
See #228. I've implemented a CleanupManager that the various nodes can use to
register clean-up routines when the main goroutine exits, fixing the leakage
bug in a (hopefully) more robust way. Every API in the Transport interface now
takes contexts as an argument so we can thread traces.


lukemarsden (approved):
nice, I like the cleanup manager!

PR #242, opened 2022-06-11
Added a whole bunch of endpoints for checking healthy servers.

aronchick:


binocarlos (approved):
this is great :+1: 

there are a few placeholders that we can fill in later and I've moved the health functions into their own file - at some point we can think about auth

having a lighttpd server that can handle shelling out into the system without needing to concern itself with the main golang is a great idea

LGTM merging and then will attempt a deploy using it

PR #270, opened 2022-06-22
run -> docker run

lukemarsden:
Move docker run-esque command to 'bacalhau docker run' (plumbing) to make room for language specific commands under `bacalhau run` (porcelain)

See https://docs.google.com/document/d/1PElwrngdbmlktqFI8Ukl5Qnk4kS2YF5jcy3RkGbZkL4/edit#


lukemarsden:
M1 runner seems to be down, but everything else is green.

PR #273, opened 2022-06-23
Implement 'bacalhau get' for fetching job results without an IPFS binary.

### LINKED ISSUE #255: `bacalhau get`
Currently you need to have an ipfs binary installed to be able to view the results of your jobs - it would be nice if the `bacalhau` binary did this for us to avoid the UX problem of failed ipfs installations

This also means you can use a bacalhau job id as the input to the `get` command

```bash
bacalhau get <jobid> --target /tmp/results
```

This should be possible by vendoring some parts of the go-ipfs codebase into `bacalhau`

Bubblyworld:
Fixes #255.


Bubblyworld:
See `pkg/ipfs/client.go` for details of how we can cut dependencies on an `ipfs` binary. Still to do is replace the binary-based code that is already in the codebase.
Screenshot 2022-06-23 at 13 46 14

Issue #253, opened 2022-06-17
Need a way to identify submitting user


aronchick:
We need to identify what user submitted a job in some unique way, so that in the future they can query based on that.

lukemarsden:
how are we going to stop malicious clients impersonating random users on the network? will we require that REST API calls are signed, so that everyone can verify that only they could have that public identity?

lukemarsden:
oh, cool, that is exactly what @Bubblyworld is doing in https://github.com/filecoin-project/bacalhau/pull/315 :+1:

Bubblyworld:
@lukemarsden Yeah, a user's identity is going to be defined by their configured keypair 👌 one interesting problem is how the client is going to share their public key with the server so that the signature can be verified. Probably the easiest option for now is just attaching it to each API call, don't think it's going to be a big bottleneck.

Bubblyworld:
Libp2p automatically signs messages in the gossip network right? So once we're passed the api server boundary then message authenticity shouldn't be a problem.

lukemarsden:
Sounds right to me

Bubblyworld:
Last thing is that currently `system/config.go` only supports RSA keys, but we can support ECDSA keypairs fairly easily if we want to. I've left a detailed TODO for it in the code. Generally I'd guess that users will just use the autogenerated key, so it shouldn't be a big deal for now.

PR #315, opened 2022-06-30
Add config initialisation code and auto-generate a unique client RSA keypair to use as their id/auth in API calls.

Bubblyworld:
See #253. The plan is to use the hash of the client's auto-generated public id
key as their client ID in API calls, and we can sign messages with the client's
private key to verify that it is them. The code for all of this is done, just
need to wire it up to the API now.


lukemarsden (approved):
nice and clean

PR #318, opened 2022-07-01
Sign and verify job submission requests in the public api client/server using client's ID key.

### LINKED ISSUE #253: Need a way to identify submitting user

Bubblyworld:
Fixes #253. The flow of authenticate public api stuff is:
1. Client creates a JobSpec/JobDeal and signs the data using their private ID key.
2. Server receives the data, signature and the client's public key.
3. Server verifies the signature, and if it is valid submits the job to the network.


Bubblyworld:
Something we could do later is factor the auth stuff into headers and make it a generic requirement for all api routes, but keeping it simple for now (since there _is_ only one api route needing auth right now).

Issue #324, opened 2022-07-03
Use `go-ipfs` as a library rather than shelling out to `ipfs`


Bubblyworld:
Currently we shell out to `ipfs` in `bacalhau serve` and `bacalhau devstack`. I've started work on using `go-ipfs` as a library instead (have a look at the `bacalhau get` implementation), but to completely cut ties with the `ipfs` binary we'll need to implement uploading/downloading tar files.

Functionality can go in `pkg/ipfs/client.go`.

Bubblyworld:
Related, allowing `go-ipfs` to work with devstack: #303

PR #334, opened 2022-07-04
Improve in-process IPFS client so we can upload/download files/dirs, and run local IPFS networks for testing.

### LINKED ISSUE #303: `bacalhau get` doesn't work for devstack
Currently `bacalhau get` connects to the public IPFS network, but devstack runs its own isolated local one. We need to add a flag for setting the network somehow.

https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1656424550962629

Bubblyworld:
Fixes #303, see #324. Adds a bunch of support for new stuff in the in-process IPFS client:
1. you can now run in `ModeLocal` to avoid using the public IPFS network for test environments
2. you can now upload and download files and directories using the client
3. `bacalhau get` now supports an optional `--ipfs-swarm-addrs` flag to connect to a specific IPFS swarm node (such as devstack)

Should now be possible to replace the IPFS shell code in devstack with this client.

### DIFF
--- a/cmd/bacalhau/get.go
+++ b/cmd/bacalhau/get.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/filecoin-project/bacalhau/pkg/ipfs"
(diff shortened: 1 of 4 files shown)


Bubblyworld:
It's short lived in `bacalhau get`, but we don't do uploading there - in `bacalhau devstack` the client will live as long as the devstack process is going. Basically the behaviour doesn't change.

Bubblyworld:
Mmm, there's an argument for splitting the in-process node and client though, conflating them a little here.

lukemarsden:
ok cool. Just FYI, I also wired up the requestor node to upload to IPFS here in the recent language frontend / python_wasm work: https://github.com/filecoin-project/bacalhau/blob/main/pkg/requestornode/requesternode.go#L130

Bubblyworld:
Cool, noted, seems like there's still a bit of complexity to unwind... I'll have a think about the best way to do this, maybe what I should do is have an in-process node implementation for running a node locally, and a separate API client (like the httpclient we already have), maybe with some util code for getting a client that points to your local node easily.

Should just be a matter of reorganising stuff I think.

Issue #356, opened 2022-07-07
removal of container already in progress errors


binocarlos:
We seem to be getting [these errors](https://app.circleci.com/pipelines/github/filecoin-project/bacalhau/214/workflows/61891860-8d5e-466b-88b9-0da74b3ba68b/jobs/737) that are failing tests quite often

The error perhaps is not an error more of a:

> don't worry - I am already doing this

Let's try to make this not fail our tests

binocarlos:
^ with the caveat that we check if it IS actually an error

Bubblyworld:
Yeah, I think this is a real error. It originates in the docker executor when it tries to fetch stdout/stderr from the container to send results back to the requester node. If it errors we can't send any results, which is a case we'd probably consider to be a real error, right?

binocarlos:
Right so this issue is about tracking down why a container is not there when we would expect it to be

something is thinking "this container has completed and we want to get the logs" <---- i.e. I expect the container to be there

something else is removing the container <----- i.e. it thinks something has gone wrong

let's discover the cause of this dis-agreement and bring harmony to this small corner of the world

binocarlos:
> Must be some kind of race condition :thinking_face: although it's a race against the docker daemon, which is external.
> (I think between the job finishing and the results directory being prepared, the docker daemon starts to clean up the container? maybe)
> Or do we trigger that in cleanup code? I'll have to do a bit more digging haha :sweat_smile:

Bubblyworld:
So a summary of the issues that eventually led to a fix:

1. The docker SDK is quite low level, and there are a lot of race conditions you need to handle that are already handled by the `docker` CLI tool. So I've swapped out the log streaming code with a shell out to `docker logs -f <container>` which is much more robust.
2. `go test` runs tests for different packages in parallel. This would be fine, except in the executor cleanup code we remove all containers that have a label matching that executor's ID. All of our tests use the same IDs, `devstack0`, `devstack1` etcetera, so if cleanup for another package is happening while your test is running, your container gets deleted in the middle of execution! This was causing a lot of really strange bugs in the test runner. Also why it works locally when you test a single package with `-count 10`!
3. We fixed this by making executor IDs more unique in tests, appending the ipfs node ID to them which guarantees (basically) that they'll never clobber each other, even when running in parallel.

PR #359, opened 2022-07-07
Add docker util for streaming logs while container is running.

Bubblyworld:
See #356. There's a race between the docker daemon cleaning up the job container
and us fetching logs from it. This adds a utility that can stream logs while
the container is running and cache them, so we don't have to poll the logs
once the container stops.

### DIFF
--- a/pkg/docker/logs.go
+++ b/pkg/docker/logs.go
@@ -1,0 +1,102 @@
+package docker
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/docker/docker/api/types"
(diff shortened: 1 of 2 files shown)


binocarlos:
I've plugged this into the docker executor and have also the docker executor return results (via the verifier) even if the container run failed - this will help users debug failures with their jobs

The results are sent back to the user in the same way as a succesful container run (i.e. via the verifier) - this opens an interesting question - does verification still count for failed jobs?

This PR makes that at least possible if not desireable - certainly the ability to see logs from failed jobs is handy

binocarlos:
sorry the PR has grown to a large number of files changed - it's mostly fixing lint errors from merging main

PR #398, opened 2022-07-21
feat(url-download): add --input-urls argument to cli

enricorotundo:
With CLI arguments:

```
go run . --api-port=$BACALHAU_API_PORT_0 --api-host=localhost docker run \
    --input-urls https://gist.githubusercontent.com/enricorotundo/990f0ad01a50d08dfb580e4ad404870e/raw/aa6934257351a0da93f1e740c72f27128590cebc/foo_data.txt:/app/foo_data_1.txt,https://gist.githubusercontent.com/enricorotundo/f2b122743fd46cc86a06eabcd97e4b5f/raw/64d38fcb53324348320d30de32c6adfca59b9d7c/pymongo_connect.py:/app/foo_data_2.txt,https://gist.githubusercontent.com/enricorotundo/f2b122743fd46cc86a06eabcd97e4b5f/raw/64d38fcb53324348320d30de32c6adfca59b9d7c/pymongo_connect.py:/app/foo_data_3.txt \
    ubuntu ls /app
```

With YAML job spec:

```
go run . --api-port=$BACALHAU_API_PORT_0 --api-host=localhost apply -f ./testdata/job-url.yaml
```

**Make sure the job selection data locality policy can accept data from Anywhere**

Check -> https://github.com/filecoin-project/bacalhau/issues/373


pkg/storage/types.go
```diff
@@ -31,6 +31,9 @@ type StorageSpec struct {
 	// IPFS storage spec this will be the data's CID).
 	Cid string `json:"cid" yaml:"cid"`
 
+	// Source URL of the data
+	URL string `json:"url"`
```
enricorotundo, line 35: @enricorotundo add support for `apply`

enricorotundo: ✅

pkg/storage/url/urldownload/storage.go
```diff
@@ -0,0 +1,123 @@
...
+func (sp *StorageProvider) IsInstalled(ctx context.Context) (bool, error) {
+	_, span := newSpan(ctx, "IsInstalled")
+	defer span.End()
+	return true, nil
+}
+
+func (sp *StorageProvider) HasStorageLocally(ctx context.Context, volume storage.StorageSpec) (bool, error) {
+	_, span := newSpan(ctx, "HasStorageLocally")
+	defer span.End()
+	return false, nil
```
binocarlos, line 57: I was about to comment "this should be true otherwise job selection policy won't pick up the job"

but then it struck me that the datra will ALWAYS be remote because it's a URL so good job :-)

binocarlos, line 57: I was about to comment "this should be true otherwise job selection policy won't pick up the job"

but then it struck me that the datra will ALWAYS be remote because it's a URL so good job :-)

pkg/job/job.go
```diff
@@ -36,6 +38,23 @@ func ConstructDockerJob(
 	jobInputs := []storage.StorageSpec{}
 	jobOutputs := []storage.StorageSpec{}
 
+	for _, inputURL := range inputUrls {
+		// split using LastIndex to support port numbers in URL
+		lastInd := strings.LastIndex(inputURL, ":")
+		rawURL := inputURL[:lastInd]
+		path := inputURL[lastInd+1:]
+		// should loop through all available storage providers?
```
binocarlos, line 46: So there might be a future when we have S3 URLs and use a different (somehow authenticated) storage driver that can download them - I suppose at that point we would switch on the engine used - but at present we assume ipfs for inputVolumes and urldownload for inputUrls and this is OK :-)

pkg/storage/constants.go
```diff
@@ -10,6 +10,7 @@ type StorageSourceType int
 const (
 	storageSourceUnknown StorageSourceType = iota // must be first
 	StorageSourceIPFS
+	StorageSourceURLDownload
```
binocarlos, line 13: Good job juggling my PR and getting everything lined up with it

pkg/storage/url/urldownload/storage.go
```diff
@@ -0,0 +1,123 @@
...
+		Type:   storage.StorageVolumeConnectorBind,
+		Source: outputPath + "/file",
+		Target: storageSpec.Path,
+	}
+
+	return volume, nil
+}
+
+// nolint:lll // Exception to the long rule
+func (sp *StorageProvider) CleanupStorage(ctx context.Context, storageSpec storage.StorageSpec, volume storage.StorageVolume) error {
```
binocarlos, line 97: Just a general code style comment - I always tend toward the following style to get around long line length complaints from the linter:

```go
func (sp *StorageProvider) CleanupStorage(
  ctx context.Context,
  storageSpec storage.StorageSpec,
  volume storage.StorageVolume,
) error {
```

binocarlos (approved):
LGTM - really great job :+1:

lukemarsden (approved):

cmd/bacalhau/docker_run.go
```diff
@@ -41,7 +42,10 @@ func init() { // nolint:gochecknoinits // Using init in cobra command is idomati
 		&jobInputs, "inputs", "i", []string{},
 		`CIDs to use on the job. Mounts them at '/inputs' in the execution.`,
 	)
-
+	dockerRunCmd.PersistentFlags().StringSliceVarP(
+		&jobInputUrls, "input-urls", "u", []string{},
+		`URL:path of the input data volumes downloaded from a URL source. Mounts data at 'path' (e.g. '-u http://foo.com/bar.tar.gz:/app/bar.tar.gz' mounts 'http://foo.com/bar.tar.gz' at '/app/bar.tar.gz').`, // nolint:lll // Documentation, ok if long.
```
lukemarsden, line 47: what happens if there is a `:` in the URL?

pkg/storage/url/urldownload/storage.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return storage.StorageVolume{}, err
+	}
+
+	sp.HTTPClient.SetTimeout(config.GetDownloadURLRequestTimeout())
+	_, err = sp.HTTPClient.R().
+		SetOutput(outputPath + "/file").
+		Get(storageSpec.URL)
+	if err != nil {
+		return storage.StorageVolume{}, err
```
lukemarsden, line 84: should we do retries? maybe make a follow-up issue for it.

also: do we currently download all URLs in parallel, because they are separate storage driver instances? this is probably ok for now, but if user provides millions of URLs this might overload the system and we'd want to put them in some sort of capped parallelism queue

pkg/storage/url/urldownload/storage.go
enricorotundo: Fixed thanks!

cmd/bacalhau/docker_run.go
enricorotundo: That case should work because unlike the other storage providers here I split URL & Path by. the position of the right most `:`

https://github.com/filecoin-project/bacalhau/blob/c8488515e8ff1000356dbb4ebf81d8be54476fc1/pkg/job/job.go#L43

This way you can use a port number in URL:

```
go run . --api-port=$BACALHAU_API_PORT_0 --api-host=localhost docker run \
    --input-urls https://gist.githubusercontent.com:443/enricorotundo/990f0ad01a50d08dfb580e4ad404870e/raw/aa6934257351a0da93f1e740c72f27128590cebc/foo_data.txt:/app/foo_data.txt \
    ubuntu cat /app/foo_data.txt
```

enricorotundo: I've expanded the description string to clarify that 👍

pkg/storage/url/urldownload/storage.go
enricorotundo: Good points. Noted https://github.com/filecoin-project/bacalhau/issues/411

testdata/job-url.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - -c
+    - cat /app/foo_data_1.txt
+  env: []
+resources:
+  cpu: ''
+  gpu: ''
+  memory: ''
+  disk: ''
+inputs:
+  - engine_name: ipfs
```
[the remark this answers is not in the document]
enricorotundo: Good catch. This sparked a good discussion with Kay regarding how we manage constant values . Updated storage provider to use the "new" approach (i.e. stringer tool)

PR #518, opened 2022-08-23
220 publisher interface

binocarlos:
This PR introduces the concept of "publishers" and clearly differentiates between "verifying" and "publishing" a job.

The publisher interface is:

```go
// Publisher is the interface for publishing results of a job
// The job spec will choose which publisher(s) it wants to use
// (there can be multiple publishers configured)
type Publisher interface {
	// tells you if the required software is installed on this machine
	IsInstalled(context.Context) (bool, error)

	// compute node
	//
	// once the results have been verified we publish them
	// this will result in a "publish" event that will keep track
	// of the details of the storage spec where the results live
	// the returned storage spec might be nill as jobs
	// can have multiple publishers and some publisher
	// implementations don't concern themselves with storage
	// (e.g. notify slack)
	PublishShardResult(
		ctx context.Context,
		jobID string,
		hostID string,
		shardIndex int,
		shardResultPath string,
	) (storage.StorageSpec, error)

	// return a slice of storage specs that when re-assembled
	// constitutes a complete set of results for a job
	// this is basically picking a valid shard result
	// and then re-assembling the storage spec it has associated
	// with it
	// this is NOT actually downloading results simply giving you
	// enough information to do so if you wanted to
	ComposeResultReferences(
		ctx context.Context,
		jobID string,
	) ([]storage.StorageSpec, error)
}
```

The important change here is the sequence of events that a job will transition through (in the happy path):

 * JobEventCreated
 * JobEventBid
 * JobEventBidAccepted
 * JobEventResultsProposed
 * JobEventResultsAccepted
 * JobEventResultsPublished
 
Also that when an executor has completed it's execution - it no longer uploads the results via the verifier (hijacking the verifier to publish was a temporary solution).

Instead - there is now a `JobEventResultsProposed` event which contains a `VerificationProposal` field of type `[]byte`

The requester node will await enough proposed verification shards before deciding that it has enough to make a decision.

Then - for each shard, the requester node will broadcast one of:

 * `JobEventResultsAccepted`
 * `JobEventResultsRejected`

In both cases - the compute node will call the publishers `PublishShardResult` method to upload the results and record that in the `PublishedResult` field of the associated event (which will in turn update the shard state)

Finally - we have the `ComposeResultReferences` method of the publisher which will return an array of storage specs that point at the complete result set ready for re-composition by the client.


lukemarsden:
:shipit: 
❤️ 🚀

lukemarsden (approved):

Issue #487, opened 2022-08-15
control loops


binocarlos:
There is a growing need for both compute nodes and requester nodes to deal with job state **over time**

Here is an example of things that have cropped up recently:

 * how we deal with `stale` bids for both compute & requester nodes
   * the compute node bid but didn't hear back - so has reserved capacity that could be used for other bids
   * the requester node accepted a bid but didn't hear back - so needs to ask for more bids
 * how we wait for the "best" bids
   * the requester node currently does first come first serve with bids
   * this means it doesn't optimize which bids to accept in any way
   * it should probably wait a period of time after hearing of a bid before accepting it
 * re-broadcasting result
   * if the requester node didn't hear about a "complete" event then the job gets stuck in a void
   
There will be many more examples of similar time based logic as we progress.

So we need control loops in both the compute node and requester node.  It should be pretty easy to write a new controller and so we need to introduce a framework for them to exist.

We already have the concept of `local events` which means these controllers can keep local state between ticks.  For example this is how the compute node marks jobs as `selected` without re-running the job selection policy each time.  Also - we have an existing [control loop setup](https://github.com/filecoin-project/bacalhau/blob/0c1ebd0f66e4b259c160d953cee5658143f49ba7/pkg/computenode/computenode.go#L111) in the compute node

wdbaruni:
Most of the issues mentioned here are covered in https://github.com/filecoin-project/bacalhau/pull/1061

Note that requester node will fail the job instead of retrying or asking for more bids if the selected compute nodes are no longer responsive. That will require more significant changes that we an revisit in the future if there is a demand for it.

Also improving the bid selection process is not covered by the PR. Should be covered by the tasks derived from https://www.notion.so/pl-strflt/Bacalhau-Orchestration-8cf6f483414c4457954b5abfcc4bf509

PR #522, opened 2022-08-24
Add compute node state machine

wdbaruni:
Introducing compute node state machine at the shard level. In addition to simplifying managing the lifecycle of the job execution, this PR also fixes a concurrency problem were previously all jobs were executed serially as the controller blocks for event subscribers to complete processing the event ([here](https://github.com/filecoin-project/bacalhau/blob/main/pkg/controller/controller.go#L554)), and the compute node used to run the job in the event handler goroutine. 

The fix now is a dedicated goroutine for each compute shard that will update its state as events arrive, execute the job and publish the result to the verifier and requester node.

These are the remaining items related to this task, which I'll link/create their separate follow up issues:
1. Add unit tests for the state machine #523 
2. Add state machine to the requester node #524 
3. Add timeouts to the shard/job states and fail if a state is taking too long to move forward. (e.g. stale bid) #487


lukemarsden:
Thanks, this looks great! Please can you placate the linter?

PR #532, opened 2022-08-27
Stop calling controlLoopBidOnJobs outside of the ticker

wdbaruni:
We have a bimodal behaviour on how compute nodes trigger the job bidding loop:
1. We have a periodic job that runs every 10 minutes
2. We trigger the loop when a bid is rejected as capacity is released
3. We trigger the loop when a job is completed 

These are over optimizations in our aim to bid for a job as soon as possible, but it adds extra complexity and can make debugging issues more challenging.

The proposal here is to only trigger the loop inside the periodic job, and reduce the ticks from 10 minutes to 100ms. I ran benchmarks using 3, 100 and 200 nodes and I didn't see any degradation. In contrary, I am getting less deviated numbers across retries with this change, but I don't have metrics to explain why

### DIFF
--- a/pkg/computenode/computenode.go
+++ b/pkg/computenode/computenode.go
@@ -24,7 +24,7 @@ import (
 
 const DefaultJobCPU = "100m"
 const DefaultJobMemory = "100Mb"
-const ControlLoopIntervalMinutes = 10
+const ControlLoopIntervalMillis = 10
 const DelayBeforeBidMillisecondRange = 100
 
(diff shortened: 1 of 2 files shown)


binocarlos (approved):
you are entirely correct that the approach of having multiple callers to `controlLoopBidOnJobs` was an optimisation but that it made things harder to reason about so I'm happy that reducing the tick timer to 100ms and making it the only source of calling the bid function is a far simpler approach than having multiple callers

Issue #463, opened 2022-08-09
remove import cycle errors


binocarlos:
The `executor` package has become the defacto home for core job spec types...

It also references:

```
import (
	"context"
	"time"

	"github.com/filecoin-project/bacalhau/pkg/capacitymanager"
	"github.com/filecoin-project/bacalhau/pkg/storage"
	"github.com/filecoin-project/bacalhau/pkg/verifier"
)
```

The problem is to reference core system types like `JobState` - we have to say `executor.JobState` - which, if used from any of:

 * capacitymanager
 * storage
 * verifier
 
Causes an import cycle error.

This cropped up when I wanted part of the verifier interface to accept a `executor.JobState` argument - boom can't do it

We could move core types to a `types` package that does not reference any other bacalhau packages.  The problem with that is the `JobSpec` mentions `Engine` - which points at the executor enumeration of engine types - which by default should live in the executor package.

We could move enumerations of different engine types (such as in executor, storage, verifier and publisher) - also to the types package.  Or we could solve it another way - either way we should be able to mention core data types from any package without import cycle loops.

aronchick:
i was just thinking this as well. 

qq - i think we should have a "cli parameters" job type, btw. That would make reseting after tests much more sane, and feels a lot more elegant than a million private variables at the top of each command. Thoughts?

PR #534, opened 2022-08-28
Extract models into a new pkg, and introduce JobShard type

wdbaruni:
This refactor became large and got out of hand, but the goal is to reduce the cross pkg dependencies that we have today, which is mainly because of our main models/types reside in the `executor` pkg (e.g. `Job` and `JobSpec`), which also depend on types that exist in multiple other pkgs, such as `storage`, `verifier`, `capacitymanager`, ....

This resulted in a circular dependency as soon as I tried to introduce a `JobShard` type that holds `Job` and `shardIndex` info instead of passing them individually, and having to flatten and break them in different places.

This PR moves all the *core* models into a separate package that doesn't have any other dependency and don't have any business logic. While there might be better ways to organize things, hopefully this should be a starting step and should take us closer to extract interfaces and help with mocking our tests


binocarlos:
this deals with #463

PR #536, opened 2022-08-29
Bacalhau command jobspec jsonyaml generate dryrun

js-ts:
Commands

Convert the command into a yaml jobspec and submit to the network
`bacalhau --output-jobspec job-hello.yml ubuntu echo hello`

Convert the command into a yaml jobspec, without submitting to the network
`bacalhau --output-jobspec job-hello.yml --dry-run ubuntu echo hello`

Convert the command into a json jobspec and submit to the network
`bacalhau --output-jobspec job-hello.json ubuntu echo hello`

Convert the command into a json jobspec, without submitting to the network
`bacalhau --output-jobspec job-hello.json --dry-run ubuntu echo hello`


cmd/bacalhau/docker_run.go
```diff
@@ -112,7 +117,10 @@ func NewDockerRunOptions() *DockerRunOptions {
 
 func init() { //nolint:gochecknoinits,funlen // Using init in cobra command is idomatic
 	dockerCmd.AddCommand(dockerRunCmd)
-
+	dockerRunCmd.PersistentFlags().BoolVar(
+		&ODR.DryRun, "dry-run", ODR.DryRun,
+		`Run a Job withtout Submitting it to the network`,
```
aronchick, line 122: This isn't correct - you're not running a job - you're simulating running a job, right? And outputting what is expected?

cmd/bacalhau/docker_run.go
```diff
@@ -142,6 +150,10 @@ func init() { //nolint:gochecknoinits,funlen // Using init in cobra command is i
 		&ODR.OutputVolumes, "output-volumes", "o", ODR.OutputVolumes,
 		`name:path of the output data volumes. 'outputs:/outputs' is always added.`,
 	)
+	dockerRunCmd.PersistentFlags().StringVar(
```
aronchick, line 153: Please remove. This is not required. You should just print the jobspec to stdout.

cmd/bacalhau/docker_run.go
```diff
@@ -315,20 +327,30 @@ var dockerRunCmd = &cobra.Command{
...
-		err = ExecuteJob(ctx,
-			cm,
-			cmd,
-			jobSpec,
-			jobDeal,
-			ODR.IsLocal,
-			ODR.WaitForJobToFinish,
-			ODR.DockerRunDownloadFlags)
-
+		if !ODR.DryRun {
```
aronchick, line 330: Don't do this. Just see if it's a DryRun, and if it is, print out the jobspec, and return.

cmd/bacalhau/docker_run.go
```diff
@@ -315,20 +327,30 @@ var dockerRunCmd = &cobra.Command{
...
+				jobSpec,
+				jobDeal,
+				ODR.IsLocal,
+				ODR.WaitForJobToFinish,
+				ODR.DockerRunDownloadFlags)
+		}
 		if err != nil {
 			return fmt.Errorf("error executing job: %s", err)
 		}
+		if ODR.OutputFileName != "" {
```
aronchick, line 343: Delete all this

cmd/bacalhau/utils.go
```diff
@@ -189,6 +194,61 @@ func setupDownloadFlags(cmd *cobra.Command, settings *ipfs.IPFSDownloadSettings)
 		settings.IPFSSwarmAddrs, "Comma-separated list of IPFS nodes to connect to.")
 }
 
+//nolint
+func ConvertJobspecToFile(
```
aronchick, line 198: Delete all this. Please reuse the functionality in Describe to unmarshall the object into a YAML/JSON file. We don't need to support both - just use YAML.

aronchick (changes requested):
Thanks for working on this - please make these changes.

Also, your test case should be using `bacalhau create -f` the file you outputted - and see if the job is created.

cmd/bacalhau/docker_run.go
```diff
@@ -315,20 +327,30 @@ var dockerRunCmd = &cobra.Command{
...
+				cmd,
+				jobSpec,
+				jobDeal,
+				ODR.IsLocal,
+				ODR.WaitForJobToFinish,
+				ODR.DockerRunDownloadFlags)
+		}
 		if err != nil {
 			return fmt.Errorf("error executing job: %s", err)
 		}
```
binocarlos, line 342: We need to move this error block inside the `!ODR.DryRun` otherwise this is refering to the error before and so is having no affect

cmd/bacalhau/docker_run.go
```diff
@@ -315,20 +327,30 @@ var dockerRunCmd = &cobra.Command{
...
+		if !ODR.DryRun {
+			err = ExecuteJob(ctx,
+				cm,
+				cmd,
+				jobSpec,
+				jobDeal,
+				ODR.IsLocal,
+				ODR.WaitForJobToFinish,
+				ODR.DockerRunDownloadFlags)
+		}
```
binocarlos, line 339: We need to move this error catching code inside the !ODR.DryRun block otherwise this is refering to the error before and so is having no affect

binocarlos (commented):
I think it might be nice to also run `VerifyJob` as part of doing a dry run

We run `VerifyJob` from inside the api server and because this request is never getting sent to the server we miss out on all of it's goodness (there will be occasions when `VerifyJob` will catch silly errors and so makes sense to run that on the client)

^ there is probably the case from running verify job ALWAYS in the CLI - as in we should run it even if dry run is not activated

Here is an example:

```go
err := job.VerifyJob(jobSpec, jobDeal)
if err != nil {
  return fmt.Errorf("error validating job: %s", err)
}
```

PR #542, opened 2022-08-30
Integration tests refactoring

enricorotundo:
https://github.com/filecoin-project/bacalhau/issues/493

- [x] create a test tooling package (move existing funcs)
- [x] sync with main
- [x] discover and update tests to use the new test tooling package
- [x] make sure `sudo` is not needed to run tests (both locally and in CI)
- [x] rm pkg/test/verifier ?
- [ ] ~separate tests (e.g. https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html) AND add testing docs~


pkg/storage/ipfs_apicopy/storage.go
```diff
@@ -106,8 +106,12 @@ func (dockerIPFS *StorageProvider) PrepareStorage(ctx context.Context, storageSp
 
 //nolint:lll // Exception to the long rule
 func (dockerIPFS *StorageProvider) CleanupStorage(ctx context.Context, storageSpec model.StorageSpec, volume storage.StorageVolume) error {
-	return system.RunCommand("sudo", []string{
-		"rm", "-rf", fmt.Sprintf("%s/%s", dockerIPFS.LocalDir, storageSpec.Cid),
+	// TODO make sure sudo is not needed here
```
enricorotundo, line 109: @binocarlos Thoughts RE: using `sudo` here? It kinda breaks running tests locally because it prompts the user for sudo pwd. And overall it should not be needed because this func removes a temp-dir. I dropped it and all tests pass anyway :)

pkg/storage/url/urldownload/storage.go
```diff
@@ -102,8 +102,12 @@ func (sp *StorageProvider) CleanupStorage(
 ) error {
 	pathToCleanup := filepath.Dir(volume.Source)
 	log.Debug().Msgf("Cleaning up: %s", pathToCleanup)
-	return system.RunCommand("sudo", []string{
-		"rm", "-rf", pathToCleanup,
+	// TODO make sure sudo is not needed here
```
enricorotundo, line 105: @binocarlos Same as above :)

pkg/storage/ipfs_apicopy/storage.go
aronchick: WFM - feels well constrained to just test cleanup.

pkg/storage/url/urldownload/storage.go
aronchick: Same as above :)

pkg/storage/ipfs_apicopy/storage.go
enricorotundo: 👍

binocarlos (approved):
LGTM

very nice to have de-duped the various "give me a stack" functions into a sensible package

we can easily extend these functions now when the need for different test scenarios crop up :+1:

philwinder:
Very minor suggestion. From an outside dev perspective, I’d like to see a tiny bit more of an explanation in `stack.go` that explains why you’d want to pick that particular stack, and what that stack does. Like, why shouldn’t I just use NewDockerIpfsStackMultiNode all the time?

PR #601, opened 2022-09-11
finished refactoring from err to RunOutput

aronchick:
This refactor is to add additional return information from anywhere a job is run. In the return object, instead of just passing back error, we pass back STDOUT, STDERR and exit code, in addition to any error.

This should allow for clearer descriptions/logs/etc.


lukemarsden:
I'm concerned about the security (DoS) implications of putting unbounded user specified strings in memory and onto libp2p. Better would be to read these values from ipfs where needed to improve the UX.

Would you be up for tackling log streaming instead? That would be a huge UX win! Happy to provide design guidance.

lukemarsden:
Kai is going to do https://github.com/filecoin-project/bacalhau/issues/624 instead which will make this change unnecessary. I see there are some other unrelated changes in the PR, could you pull them out into a separate PR please?

pkg/executor/docker/executor_test.go
```diff
@@ -0,0 +1,36 @@
+package docker
```
lukemarsden, line 1: Is this empty?

lukemarsden (approved):
Lgtm! Thanks for doing this!

pkg/computenode/shard_fsm.go
```diff
@@ -358,7 +358,11 @@ func runningState(ctx context.Context, m *shardStateMachine) StateFn {
 	// we get a "proposal" from this method which is not the results
 	// but what the compute node verifier wants to pass to the requester
 	// node verifier
-	proposal, err := m.node.RunShard(ctx, m.Shard)
+
+	// Need to do something with RunOutput
```
lukemarsden, line 362: Is this still true? Does the stdout and stderr make it to describe output?

aronchick: no not yet sadly - need to work with @wdbaruni to understand how to get into state machine

pkg/computenode/computenode.go
```diff
@@ -567,29 +567,33 @@ func (n *ComputeNode) BidOnJob(ctx context.Context, shard model.JobShard) error
...
 	if err != nil {
-		return shardProposal, err
+		runOutput.Error = err
+		return shardProposal, runOutput, err
 	}
 	resultFolder, err := verifier.GetShardResultPath(ctx, shard)
 	if err != nil {
-		return shardProposal, err
+		runOutput.Error = err
+		return shardProposal, runOutput, err
```
wdbaruni, line 591: I am not a big fan that `RunShard` now returns two types of errors that callers would have to check for, even though _most_ of the time the errors are identical. I know that I suggested over slack to keep returning an `err` value, but I didn't expect that we would still have to encapsulate the error inside `runOutput` instance as well.

While I am still new to golang, but comparing to Java, I am assuming the best practice is to return a dedicate `err` value as callers would know they need to handle it compared to encapsulating the error in another type. However, if we believe in this specific case that encapsulating the error inside `runOutput` would make things more readable and manageable, then I am happy with that as long as we drop the `err` value and only return an error inside `runOutput`

wdbaruni: Yeah it makes sense given that `runOutput.Error` is just for reference and callers wouldn't need to handle it. Maybe using `errorMsg` of type `string` would avoid future confusion instead of `error` type, but both are fine. Thanks for the clarification :)

PR #602, opened 2022-09-11
bacalhau output jobspec, create template

js-ts:
discussion doc https://docs.google.com/document/d/11yp7wwVoWqWenQIsXAf_Wq-7wxCM__dxjBDvITmq-3w/edit#

`bacalhau docker run --output-jobspec ubuntu echo`
print the jobspec to stdout and submit the job to the network

`bacalhau docker run --output-jobspec --dry-run ubuntu echo`
print the jobspec to stdout without submitting the job to the network

`bacalhau create template`
print a ubuntu echo hello jobspec to stdout


aronchick:
I don’t think it makes sense to do it this way for docker run - I think we want describe to be able to output a job as a yaml that can be input into create

js-ts:
> I don’t think it makes sense to do it this way for docker run - I think we want describe to be able to output a job as a yaml that can be input into create

that doesn't Work and doing it this way doesn't make any sense

js-ts:
@aronchick I created the doc you asked me to 
https://docs.google.com/document/d/11yp7wwVoWqWenQIsXAf_Wq-7wxCM__dxjBDvITmq-3w/edit

cmd/bacalhau/job-test.yaml
```diff
@@ -0,0 +1,60 @@
+# jobspec for the bacalhau command bacalhau docker run ubuntu -- /bin/bash -c 'echo "simple job output"'
+apiVersion: v1alpha1
+engine: 2
```
philwinder, line 3: Might want to add a `kind` to this, to allow for extension of other things later. E.g. a "dag of job specs" or whatever.

cmd/bacalhau/create.go
```diff
@@ -168,3 +182,39 @@ var createCmd = &cobra.Command{
 
 	},
 }
+
+var templateCmd = &cobra.Command{
```
philwinder, line 186: I don't see the need for this command when you have an `--output-template` parameter on the actual docker run?

cmd/bacalhau/docker_run.go
```diff
@@ -241,20 +254,48 @@ var dockerRunCmd = &cobra.Command{
...
+			if err != nil {
+				log.Error().Msgf("Failure marshaling job spec '%s'", err)
+				return err
+			}
+			var jobspectmp *model.JobSpec
+			_ = yaml.Unmarshal(bytes, &jobspectmp)
+
+			jobspectmp.APIVersion = "v1alpha1"
+			jobspectmp.EngineName = "docker"
+			jobspectmp.VerifierName = "noop"
```
philwinder, line 273: Ok for these to be hard coded?

enricorotundo:
Why writing a whole new google doc when you can use the PR description at the top? 👀

aronchick:
There are too many issues to discuss here. Please withdraw the PR and we'll discuss in the doc.

aronchick:
> > I don’t think it makes sense to do it this way for docker run - I think we want describe to be able to output a job as a yaml that can be input into create
> 
> that doesn't Work and doing it this way doesn't make any sense

What way doesn't work? 
What doesn't make sense?

Issue #673, opened 2022-09-26
Windows support


lukemarsden:
Finish this up and land it

simonwo:
Actual support landed in #681, but keeping this issue around until we've also done documentation as well. Documentation should include:

- How to install and run on Windows (just what to do with the exe)
- What is supported and what is not (client is supported, nodes may work, resource limits unsupported)

simonwo:
Documentation updated in https://github.com/bacalhau-project/docs.bacalhau.org/pull/30.

PR #681, opened 2022-09-27
#673 Native Windows support

simonwo:
This branch adds support for native Windows builds to Bacalhau.

What does that mean? Well, it is now possible to check out, build, test and release the entire repository [natively on Windows](https://app.circleci.com/pipelines/github/filecoin-project/bacalhau/1176/workflows/37e3fe08-b863-43de-be89-30fefacd1c69/jobs/5869) (i.e. not through WSL), including running the client and devstack. Windows builds will be included in future releases.

### Things that don't work
There are some constraints that need to be documented:

- Docker resource limits don't work on Windows (because they have to applied differently based on WSL or Hyper-V type hosts)
- ARM64 builds are not supported because IPFS doesn't produce `windows-arm64` builds.
- Tests that use Docker don't work on CircleCI, but do work locally (because you can't run Linux-based containers on Windows hosts in CircleCI) – some of this work has been about removing Docker as the default way to run jobs where it's not necessary.

### Things we'll need to do differently if we merge this
Writing for cross-platform requires us to behave a little differently (all good ways really):

- Make more use of noop jobs to avoid using Docker if we're not actually testing Docker.
- Using e.g `filepath.Join` instead of sprintf-style path manipulations.
- Less shelling out to coreutil commands – so more using golang library functions where appropriate, and no `sudo`.
- Judicious use of build tags when adding new POSIX-only features (like FUSE?).

As per [this discussion](https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1662976230450479), I'm open to not merging this because we don't want those behavioural changes. The top of this branch has Windows-specific stuff, the bottom half has general stuff that would be good to merge anyway for code hygiene and tech debt, so if we don't want to commit to Windows then we can just strip off the top of the branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,10 +3,16 @@
 version: 2.1
 
 executors:
-  linux:
+  linux: &linux
     machine:
       image: ubuntu-2204:2022.07.1
-    resource_class: xlarge
+      resource_class: xlarge
+  darwin: *linux
+  windows:
+    machine:
+      image: windows-server-2022-gui:current
+      resource_class: windows.xlarge
+      shell: bash.exe
 
 # Define a job to be invoked later in a workflow.
 # See: https://circleci.com/docs/2.0/configuration-reference/#jobs
@@ -23,7 +29,7 @@ jobs:
       CGO: 0
       GCS_TEST_RESULTS_BUCKET: bacalhau-global-storage/test-results
     working_directory: ~/repo
-    executor: linux
+    executor: << parameters.target_os >>
     parameters:
       target_arch:
         type: string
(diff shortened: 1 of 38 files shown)


.circleci/config.yml
```diff
@@ -284,12 +324,13 @@ workflows:
...
             parameters:
-              target_os: ["linux", "darwin"]
+              target_os: ["linux", "darwin", "windows"]
               target_arch: ["amd64", "arm64"]
               run_tests: [true]
             exclude:
-              - target_os: "linux"
+              - target_os: "windows"
                 target_arch: "arm64"
+                run_tests: true
```
simonwo, line 333: Previously this looks like it tried to exclude `linux-arm64` builds, but it didn't work (because all parameters need to be specified) and those builds have been running anyway. So I've removed that and only excluded `windows-arm64` builds – but we can put it back if it really was meant to be excluded.

lukemarsden (approved):
Nice, thanks! :shipit:

PR #689, opened 2022-09-28
replace controller with separate event handlers

wdbaruni:
## Problem Statement
As I started working on introducing state machine for the requester node, I realized a lot of requester node related logic resides in the `controller` instead of `requesternode` type, and that the `controller` is a very large type with different and unrelated responsibilities, including:
1. Acts as a proxy to publish and consume events from the `transport` layer
2. Acts as a proxy to access the underlying `localDB`
3. Mutates `localDB` based on the events received 
4. Adds lifecycle events to otel tracer
5. Creates and manages the lifecycle of context used to trace the job execution 
6. Handles job creation and submitting jobs to the compute nodes, which is supposed to be the responsibility of the requesternode

This makes testing the code very challenging. In fact, we don't have a unit tests for the controller.

## Proposed Fix
This PR improves things by introducing a *chain* of isolated event handlers, each concerned with a very specific functionality that used to be owned by the `controller`. The event handlers are broken into the following categories:

### 1. JobEventConsumers:
Responsible for handling `JobEvents` received from the transport layer, which include events published by other nodes and events published by the current node. Currently we have the following handlers:
1. `lifecycleEventHandler`: add tracing metadata to the context about the read event
2. `tracerContextProvider`: ends the span for the job if received a terminal event
3. `localDBEventHandler`: update the job state in the local DB
4. `requesterNode`: handles bid and result proposals 
5. `computeNode`: handles job execution

### 2. JobEventPublishers:
Responsible for handling  newly generated `JobEvents` that were created by the current node (e.g. `requesternode` or `computenode`). Currently we have the following handlers:
1. `transport`: publish events to the network
2. `lifecycleEventHandler`: add tracing metadata to the context about the published event

### 3. LocalEventConsumers:
Responsible for handling  newly generated `JobLocalEvents` that were created by the current node. Currently we have the following handlers:
1. `localDBEventHandler`: adds the local event to the node's local state

### 4. ContextProvider

In addition to event handlers, the PR introduces `ContextProvider` which takes care of setting the context along with tracing metadata at the beginning of the chain to avoid setting the `nodeID` and `jobID` in different places of the code.

## Next Steps
- [ ] Add unit tests to the newly introduced event handlers. I wanted to reduce the gap and merge conflicts I had to resolve, so I just made sure existing tests are passing but didn't improve the test coverage just yet
- [ ] Update architecture [doc](https://docs.bacalhau.org/about-bacalhau/architecture), and preferably introduce a layer architecture diagram similar to the one from [IPFS](https://docs.google.com/drawings/d/1OVpBT2q-NtSJqlPX3buvjYhOnWfdzb85YEsM_njesME/edit)


wdbaruni:
This is a major refactor, so I am hesitant from merging without getting some feedback first in case I am missing some context or motivation behind the current structure of the controller. Otherwise, please squash and merge

cc: @lukemarsden @binocarlos @aronchick

lukemarsden:
This looks great, I'm going to squash and merge and in parallel ask @binocarlos to take a look. Thank you Walid, nice work.

PR #864, opened 2022-10-11
624 publish on error

binocarlos:
It turns out we are actually publishing results to IPFS (or whatever the publisher is configured as) in the case of contianer exit code >= 1

I've added a test for this case and we also now have `WaitOptions` which we can pass to the wait function so we can configure aspects of the wait function without bloating the input params

### DIFF
--- a/pkg/job/state.go
+++ b/pkg/job/state.go
@@ -135,12 +135,36 @@ func (resolver *StateResolver) Wait(
 	totalShards int,
 	checkJobStateFunctions ...CheckStatesFunction,
 ) error {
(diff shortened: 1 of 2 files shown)


lukemarsden:
ok but what's the user experience of getting those logs?

binocarlos:
The test does basically the same as a `bacalhau get <job-id>` - i.e. it's the same UX as getting results for a successful job

binocarlos:
We could at some point think about `bacalhau logs <job-id>` which wraps get and `cat stdout`

PR #891, opened 2022-10-14
889 adding some precommit hooks

aronchick:


Makefile
```diff
@@ -25,7 +25,7 @@ endif
 export GO111MODULE = on
 export GO = go
 export CGO_ENABLED = 0
-export PYTHON = python3
+export PYTHON = python3ƒ
```
simonwo, line 28: Typo?

.circleci/config.yml
```diff
@@ -80,9 +80,29 @@ jobs:
...
+            poetry run python --version --version >> pre-commit-cache-key.txt
+
+      - restore_cache:
+          keys:
+            - v1-pc-cache-{{ checksum "pre-commit-cache-key.txt" }}
+
       - run:
           name: Build
-          command: make build
+          command: make build-ci
```
simonwo, line 105: Where do we actually run pre-commit in CI?

.circleci/config.yml
```diff
@@ -80,9 +80,29 @@ jobs:
...
+      - run:
+          name: Cache Precommit
+          command: |
+            cp .pre-commit-config.yaml pre-commit-cache-key.txt
+            poetry run python --version --version >> pre-commit-cache-key.txt
+
+      - restore_cache:
+          keys:
+            - v1-pc-cache-{{ checksum "pre-commit-cache-key.txt" }}
+
```
simonwo, line 102: Can we run pre-commit in the `lint` CI job? Because we presumably don't need to run it on every GOOS_GOARCH combo, and the `lint` job is currently quite short so doing it there won't increase overall build times.

Issue #697, opened 2022-09-29
Estuary publisher should be more resilient


enricorotundo:
## Problem statement

Current Job submission relies on Estuary endpoint availability. Unfortunately, when [downtimes occur](https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1664457853868299?thread_ts=1664448185.071059&cid=C02RLM3JHUY) Bacalhau fails to [publish the job's data to Filecoin](https://github.com/filecoin-project/bacalhau/blob/main/pkg/publisher/estuary/publisher.go) and consequently fails the job. Specifically, when the endpoint is unreachable we timeout and the job state ends up being "Error".
```
Sep 29 11:05:17 bacalhau-vm-production-1 bash[6993]: 11:05:17.546 | ERR computenode/shard_fsm.go:445 > [QmXaXu9N] shard: b18d6f64-1bb4-42c1-ab58-08c9df947636:0 at state: Error error completing job due to Post "https://shuttle-6.estuary.tech/content/add-car": dial tcp 147.28.156.33:443: i/o timeout
```

Thread: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1664448185071059

## Possible solution

Estuary has a viewer endpoint that lists down all the available shuttles: `/viewer`. They are working on a long-term solution of having a proxy and an API gateway. In the meantime our publisher could try shuttles until it finds one that works, assuming N of them exist with some naming scheme.

PR #904, opened 2022-10-17
Add resilience to Estuary publishing

### LINKED ISSUE #696: Need a Publisher failover mechanism
## Problem statement

When a publisher service is down bacalhau fails the job instead of failing over to alternative publishers.
For instance, if the Estuary publisher fails we could fail over to IPFS publisher.
Here's an instance of the failure due to a lack of resiliency throughout this process: https://github.com/filecoin-project/bacalhau/issues/697

simonwo:
We now [fallback to IPFS storage if Estuary is unavailable](https://github.com/filecoin-project/bacalhau/commit/2c02079f3f2faca313313fd1cfd35bfcb0052c33) and [query the Estuary API for correct submission URLs](https://github.com/filecoin-project/bacalhau/commit/db7e3d969018b67ab51c02635adc3af5429229f1). This means we are able to retry the Estuary publication with multiple hosts and just write to IPFS if none work, so Estuary jobs should stop failing if their service goes down or one of the nodes in their network is affected.

Resolves #696 and #697.

### DIFF
--- a/pkg/publisher/fallback/publisher.go
+++ b/pkg/publisher/fallback/publisher.go
@@ -1,0 +1,78 @@
+// fallback implements a publisher.Publisher that will try multiple other
+// Publishers in order until one suceeeds.
+//
+// The Publishers are tried in the order specified in the call to
+// NewFallbackPublisher. If and only if all of the Publishers return an error
+// result, the fallback publisher will also return an error result. Otherwise,
+// it will return the result of the first Publisher to succeed and will swallow
+// any errors. Subsequent Publishers will not be attempted after one succeeds.
+
+package fallback
+
+import (
+	"context"
+
+	"github.com/filecoin-project/bacalhau/pkg/model"
+	"github.com/filecoin-project/bacalhau/pkg/publisher"
+	"github.com/rs/zerolog/log"
+	"go.uber.org/multierr"
+)
+
+type fallbackPublisher struct {
+	publishers []publisher.Publisher
+}
+
+func NewFallbackPublisher(publishers ...publisher.Publisher) *fallbackPublisher {
+	return &fallbackPublisher{
+		publishers: publishers,
+	}
+}
+
+// fallback accepts a slice of publisher objects and passes them to the supplied
+// function in order, until one does not return an error value and then returns
+// that result. If all publishers return an error value, a composite error is
+// returned.
+func fallback[T any, P any](ctx context.Context, publishers []P, method func(P) (T, error)) (T, error) {
+	var anyErr error
+	for _, publisher := range publishers {
+		result, err := method(publisher)
+		if err == nil {
+			return result, nil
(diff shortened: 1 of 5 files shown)


pkg/publisher/fallback/publisher.go
```diff
@@ -0,0 +1,78 @@
...
+	return &fallbackPublisher{
+		publishers: publishers,
+	}
+}
+
+// fallback accepts a slice of publisher objects and passes them to the supplied
+// function in order, until one does not return an error value and then returns
+// that result. If all publishers return an error value, a composite error is
+// returned.
+func fallback[T any, P any](ctx context.Context, publishers []P, method func(P) (T, error)) (T, error) {
```
binocarlos, line 35: ohhhh nice! generics - I'm assuming this won't introduce any problems with CI or compiling anything?

I'm sure the answer is no but felt compelled to comment because a part of my brain hasn't yet lept into "you can use generics now in go" territory :-)

pkg/publisher/fallback/publisher.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+
+	var zeroResult T
+	return zeroResult, anyErr
+}
+
+// ComposeResultReferences implements publisher.Publisher
+func (f *fallbackPublisher) ComposeResultReferences(ctx context.Context, jobID string) ([]model.StorageSpec, error) {
+	return fallback(ctx, f.publishers, func(p publisher.Publisher) ([]model.StorageSpec, error) {
```
binocarlos, line 53: what if we succeed with `PublishShardResult` using publisher A and then when we try to `ComposeResultReferences` and publisher A fails - are we now trying to `ComposeResultReferences` using publisher B but for results that were generated by publisher A?

binocarlos (approved):
This is awesome! I like how the fallback publisher works with generics and also that you've got rid of the hard-coded dependency on the estuary endpoints

I don't think the question about composing results with a different publisher that published them is a blocker to merging this because it feels like an edge case so as far as I'm concerned LGTM :+1:

simonwo:
OK, given the above, I'm going to merge this and raise a tech-debt issue for revisiting the ComposeResultReferences method.

pkg/publisher/fallback/publisher.go
binocarlos: > This perhaps means ComposeResultReferences shouldn't be a part of this interface? And is worth a refactor?

I was just thinking about this and I think actually it's OK because the publish result is a `StorageSpec` - and those should be capable of downloading results from whereever they live (it's the point of a storage spec after all)

So - I'm not sure we actually do need to know "which pubisher published" - as long as the act of publishing returns a storage spec that can be resolved then it's OK

For example - in the Estuary fallback case - it's either on IPFS or on IPFS :-)

PR #953, opened 2022-10-26
876 Improvements to downloader

binocarlos:
## overview

This PR improves the download experience of both `bacalhau get` and `bacalhau docker run`

In both cases - if no `--output-dir` flag was provided (which let's the user dictate where we should download results) - we now create a folder in `$PWD` named after the job (e.g. `job-12345678`) and place the downloaded results into that newly created folder.  This prevents the download from splatting files all over the users current working directory

We have also removed the `ComposeResultReferences` from the publisher interface and are instead using the new `PublishedResults` struct to vastly simplify the amount of concerns the publisher has and enable easy reconstruction of the result set by the client. 

## new output structure

The results folder before looked like this:

image

and now it looks like this

image

## features

 * If no `--output-dir` flag is given - auto create a folder for the results (e.g. `job-123456ab`)
 * Add a `--download` flag to `docker run` so we can download the results automatically
 * Add a new `PublishedResults` struct that contains host and shard info for a given storage spec
 * Remove the need for `ComposeResultReferences` function in publishers
   * This was only present because of the lack of a `PublishedResults` struct
 * Change the structure of the results folder to be easier to read
   * The `shards` folder now has a folder for each shard named after it's index (e.g. `0`)
   * All log files from different hosts are written (flat) to the shard folder
   * We prepend the host id before each log file


cmd/bacalhau/docker_run.go
```diff
@@ -222,6 +215,14 @@ func init() { //nolint:gochecknoinits,funlen // Using init in cobra command is i
 		&ODR.IDOnly, "id-only", ODR.IDOnly, "Print out only the Job ID on successful submission.",
 	)
 
+	dockerRunCmd.PersistentFlags().BoolVar(
+		&ODR.RunTimeSettings.AutoDownloadResults,
+		"download",
+		ODR.RunTimeSettings.AutoDownloadResults,
+		"Should we download the results once the job is complete?",
+	)
```
simonwo, line 223: Why not add this to `setupRunTimeFlags`? (Disclousre: I've just spent time adding the common flags to WASM so if you don't I probably will :D)

pkg/model/storage_spec.go
```diff
@@ -29,3 +29,12 @@ type StorageSpec struct {
 	// Additional properties specific to each driver
 	Metadata map[string]string `json:"Metadata,omitempty"`
 }
+
+// PublishedStorageSpec is a wrapper for a StorageSpec that has been published
+// by a compute provider - it keeps info about the hos, job and shard that
```
simonwo, line 34: Typo 😅

simonwo: But otherwise ++++++1 for this, StorageSpec has felt like a janky way of representing this for a long time

pkg/publisher/types.go
```diff
@@ -33,16 +33,4 @@ type Publisher interface {
...
-	// constitutes a complete set of results for a job
-	// this is basically picking a valid shard result
-	// and then re-assembling the storage spec it has associated
-	// with it
-	// this is NOT actually downloading results simply giving you
-	// enough information to do so if you wanted to
-	ComposeResultReferences(
-		ctx context.Context,
-		jobID string,
-	) ([]model.StorageSpec, error)
```
simonwo, line 47: 💯💯💯💯💯💯

simonwo:
> All log files from different hosts are written (flat) to the shard folder
> We prepend the host id before each log file

So can multiple hosts run 1 shard? Or if not, why add the node id?

cmd/bacalhau/docker_run.go
binocarlos: Great point :+1:

pkg/model/storage_spec.go
binocarlos: lol - what even is a `hos` :-)

binocarlos:
> So can multiple hosts run 1 shard? Or if not, why add the node id?

Indeed they can - as soon as we use `--concurrency` > 1 then multiple hosts will all run a single shard

having said that - I just ran that locally and realised that it's not actually writing those files and so you caught a bug with your comment :+1: 

will write a test and fix now !

binocarlos:
improvements made to the downloader:

 * downloads the cids of the actual results -> cids
   * deterministic jobs have the same results cid
   * the cid will only be downloaded & stored once
 * reconstructs the shards folder based on nodes running shards
   * each shard folder is made from hard links so we are not doubling the disk space
   * the name of the shard folder is more easy to understand
 * merges the volumes folder from all shards / nodes
   * this means that deterministic, sharded jobs have a complete result set in one folder
   * the files are also hardlinks meaning none of this eats more disk space

in conclusion - we should now be able to download concurrent, sharded, deterministic and non-deterministic jobs using the least amount of bandwidth and disk space possible

image

here is proof it's hardlinks

image

lukemarsden (approved):

simonwo:
Do we have any tests that run on Darwin and also use hard links? It was my understanding that hard links don't work on Apple platforms anymore.

binocarlos:
Darn it - I was checking to see if  that was the case and came the conclusion that they seemed supported - the copy function we use falls back to doing a literal copy but that is obviously not very desirable when we are copying things wuth abandon!
 
Perhaps we can find a neat way to support this on Darwin and/or add some flags that control how much copying we do

@simonwo do you have a Darwin machine you can test this branch on?

binocarlos:
> Do we have any tests that run on Darwin and also use hard links?

^ this is a good point I will add some and see what CI says :-)

lukemarsden:
```
luke@chunky:~/hardlinks$ echo hello > world
luke@chunky:~/hardlinks$ ln world world2
luke@chunky:~/hardlinks$ ls -alh
total 16
drwxr-xr-x    4 luke  staff   128B 28 Oct 11:00 .
drwxr-x---+ 171 luke  staff   5.3K 28 Oct 10:59 ..
-rw-r--r--    2 luke  staff     6B 28 Oct 10:59 world
-rw-r--r--    2 luke  staff     6B 28 Oct 10:59 world2
```

hardlinks are working fine for me here, on Darwin chunky 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64

simonwo:
Ah, I have partially misremembered... you can create links as long as they are only to regular files, not directories:
```
simon@Simons-MacBook-Pro ~ % mkdir hello
simon@Simons-MacBook-Pro ~ % ln hello hello2
ln: hello: Is a directory
```
So we might be ok as long as we are doing it file by file?

lukemarsden:
Yes that's what Kai did, file by file. Cool

PR #954, opened 2022-10-26
inter-connect nodes with internal ips

binocarlos:
This connects terraform nodes with their internal ip addresses

It does not affect public users

It is a quick fix for a broken production and we should find a better way to do this


lukemarsden:
This seems to work in practice, nice job. Do the internal IPs get reused reliably then?

binocarlos:
I'm hoping it's a case of at the time we say "use this fixed address" - the address is free because previously we have destroyed the vm that was claiming it previously.

I have not tested that theory - the above PR was a "quick fix production" thing :smile:

PR #973, opened 2022-10-27
Aggressive reconnect

lukemarsden:
* Use direct peering arrangements to form the network backbone, rather than one-off connections that won't retry/reconnect on failure
* Reconnect disconnected libp2p peers in 10 seconds, not 5 minutes


lukemarsden:
linux tests passed, calling it

PR #995, opened 2022-10-31
Tf-controller implementation 

cmharden:


lukemarsden:
@philwinder you were looking at Flux stuff, does this look reasonable to you? @cmharden if you could provide a description in the PR describing the behavior change that would be really helpful please

cmharden:
@lukemarsden, the `gitops` directory structure was flattened and reorganized to be more consistent with the `gitops-root` workflow.

Specifically,
* Unnecessary sub-directories were removed
* Secrets were moved to a dedicated directory
* Kustomize overlays were added for the `Terraform` objects to manage environment specific configurations
* Workspaces are managed in the `Terraform` object now
* `internal_ip_addresses` was added to `env-vars.tfvars` 
  * Note to follow up that, this index starts at 1 (is this correct).  
* `public_ip_address` will now be output as a `secret` 

The functional behavior shouldn't change, the changes implemented were meant for ease of maintenance and clarity.  I'm completely open to further modifications/rollbacks if that help the team.

PR #1002, opened 2022-11-01
Pin results to Estuary after publishing to IPFS

### LINKED ISSUE #1001: Queue and pin to Estuary in the background
Our Estuary publisher is disabled again because sending data to Estuary or retrieving again immediately afterwards is sometimes too slow.

Instead, Estuary has an alternative API that allows pinning an existing CID. So instead, we can use the normal IPFS publisher to publish results to our own IPFS node and then get Estuary to pin them afterwards. We can do this out of band and ensure jobs remain fast.

simonwo:
Our Estuary publisher is disabled again because sending data to Estuary or retrieving again immediately afterwards is sometimes too slow.

Instead, Estuary has an alternative API that allows pinning an existing CID. So instead, we now use the normal IPFS publisher to publish results to our own IPFS node and then get Estuary to pin them afterwards. We do this out of band and ensure jobs remain fast.

Resolves #1001.


lukemarsden:
@simonwo please could you rebase this on main?

simonwo:
Yes – I am also now awaiting the results of https://github.com/application-research/estuary-clients/pull/1

lukemarsden:
looks like upstream merged!

simonwo:
Alas, the code that is generated is broken! So I'm talking with them – they either need to upgrade the code generator to v3 or modify a doc signature to refer to a proper type. Still working through it!

PR #1038, opened 2022-11-07
Fix running locally docs to use binary name

iand:
Instructions say to build the binary so we should use that for following examples


lukemarsden:
LGTM, seems that `go build` indeed places the binary in `.`

Issue #1032, opened 2022-11-07
Remove all logging information during test runs and/or devstack


aronchick:
ETA: 2022Q4

When running locally, the logs are virtually unusable. Just a wall of meaningless text. Need to improve.

Further, we do a lot of OpenTelemetry stuff - we should detect and turn that off (with no errors) when running DevStack.

lukemarsden:
I don't find the logging too hard to interpret, can you give some specific examples of things that we should change please?

simonwo:
@aronchick What would improve this for you? The logging seems to be useful in debugging situations, but I agree there is a lot of it. Would you prefer logs to only be shown if tests fail?

simonwo:
I'm closing this as I don't think we've reached a consensus on whether a change is required. Please re-open with more details if necessary!

PR #1039, opened 2022-11-07
Suppress logging output in tests

iand:
Part of [#1032](https://github.com/filecoin-project/bacalhau/issues/1032)

Also some gofmt noise unfortunately

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -9,11 +9,12 @@ import (
 	"strings"
 	"time"
 
-	ipfslog1 "github.com/ipfs/go-log"
 	ipfslog2 "github.com/ipfs/go-log/v2"
 	"github.com/mattn/go-isatty"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 type JobEvent struct {
@@ -49,6 +50,7 @@ func ConfigureTestLogging(t tTesting) {
 	t.Cleanup(func() {
 		log.Logger = oldLogger
 		zerolog.DefaultContextLogger = oldContextLogger
+		configureIpfsLogging(log.Logger)
 	})
 }
 
@@ -141,6 +143,8 @@ func configureLogging(loggingOptions ...func(w *zerolog.ConsoleWriter)) {
 	// While the normal flow will use ContextWithNodeIDLogger, this won't be so for tests.
 	// Tests will use the DefaultContextLogger instead
 	zerolog.DefaultContextLogger = &log.Logger
+
+	configureIpfsLogging(log.Logger)
 }
 
 func LoggerWithRuntimeInfo(runtimeInfo string) zerolog.Logger {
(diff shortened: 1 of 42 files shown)


pkg/logger/logger.go
```diff
@@ -155,3 +159,13 @@ func ContextWithNodeIDLogger(ctx context.Context, nodeID string) context.Context
 	l := loggerWithNodeID(nodeID)
 	return l.WithContext(ctx)
 }
+
+// Suppress suppresses log output unless the BACALHAU_UNSUPPRESS_LOGS environment variable is set.
+func Suppress() {
```
iand, line 164: This is the Suppress function that does the log suppression

wjam:
I'm not sure how valuable this change is, as it'll make debugging any failure in CI difficult

lukemarsden:
Yeah, I think it might be better to go through the log messages and move some of them down a log-level, i.e. from Debug to Trace or whatever, rather than silencing them entirely.

iand:
There's no need to emit logs for tests that have passed, but it's useful for debugging failures. This approach allows tests to be re-run with the env variable set but I admit that's rather a brute force approach. I'll look at whether I can buffer logs via `ConfigureTestLogging` and emit on failure.

Downgrading log levels isn't very helpful. Many of the tests pass but emit noisy logs at ERROR level or similar.

iand:
Oh, I think it's only the ipfs logging that is not being diverted in `ConfigureTestLogging`

aronchick:
My take is only that any transport stuff is just brutal. I don't want to turn it off or hide! Even if we just shunt it to a file or endpoint in CI or devstack that's be valuable

lukemarsden:
Yeah, let's investigate how to make only failed tests show verbose logs. (Not showing logs for failed tests would be bad, because they're flaky, it's hard to re-run them and get the same behavior, also it's hard to set env vars in CI).

If passing tests are emitting noisy error logging, let's move the noise down to Trace level?

wjam:
The current behaviour of the tests is to only output the failed test(s) - apart from a bunch of noise from [system/environment.go](https://github.com/filecoin-project/bacalhau/blob/e59bfb79e16a96e1a179762e7dc3bafbc0658f71/pkg/system/environment.go#L38) and [system/tracer.go](https://github.com/filecoin-project/bacalhau/blob/e59bfb79e16a96e1a179762e7dc3bafbc0658f71/pkg/system/tracer.go#L46) that is outputted before we've had a chance to redirect the logging output.

A failing test does have the tendency to be very verbose, especially as some tests don't correctly use parameterised tests and so includes the output from other tests.

aronchick:
I mean let's start by just having people run devstack locally and seeing how brutal it is. Two birds with one stone?

CI tests are already useless now. Every failure spins until timeout which causes a totally lack of debugging information due to it running off the end of the logs - https://app.circleci.com/pipelines/github/filecoin-project/bacalhau/1753/workflows/a8cec7c3-0da7-4201-97c6-8734e651f0cb/jobs/10833/parallel-runs/0/steps/0-109

iand:
Here's an alternate implementation that configures the iipfs logging package to use zerolog. Since `ConfigureTestLogging` redirects logging to Go's test logger this has the effect of suppressing logs for tests that pass (unless verbose testing is requested).

Before this change, a failing test looked like:

```bash
15:27 $ go test -run TestComputeNodeResourceLimitsSuite
2022-11-07T15:28:22.692Z	ERROR	flatfs	go-ds-flatfs@v0.5.1/flatfs.go:997	could not store final value of disk usage to file, future estimates may be inaccurate
2022-11-07T15:28:23.001Z	ERROR	flatfs	go-ds-flatfs@v0.5.1/flatfs.go:997	could not store final value of disk usage to file, future estimates may be inaccurate
2022-11-07T15:28:23.284Z	ERROR	flatfs	go-ds-flatfs@v0.5.1/flatfs.go:997	could not store final value of disk usage to file, future estimates may be inaccurate
2022-11-07T15:28:23.990Z	ERROR	flatfs	go-ds-flatfs@v0.5.1/flatfs.go:997	could not store final value of disk usage to file, future estimates may be inaccurate
2022-11-07T15:28:24.530Z	ERROR	flatfs	go-ds-flatfs@v0.5.1/flatfs.go:997	could not store final value of disk usage to file, future estimates may be inaccurate
2022-11-07T15:28:24.739Z	ERROR	flatfs	go-ds-flatfs@v0.5.1/flatfs.go:997	could not store final value of disk usage to file, future estimates may be inaccurate
--- FAIL: TestComputeNodeResourceLimitsSuite (12.27s)
    --- FAIL: TestComputeNodeResourceLimitsSuite/TestJobResourceLimits (0.27s)
        resourcelimits_test.go:83: 
            	Error Trace:	/home/iand/ws/cod/bacalhau/pkg/test/computenode/resourcelimits_test.go:83
            	            				/home/iand/ws/cod/bacalhau/pkg/test/computenode/resourcelimits_test.go:87
            	Error:      	Not equal: 
            	            	expected: 1
            	            	actual  : 2
            	Test:       	TestComputeNodeResourceLimitsSuite/TestJobResourceLimits
FAIL
exit status 1
FAIL	github.com/filecoin-project/bacalhau/pkg/test/computenode	12.288s
```

After change:

```bash
15:28 $ go test -run TestComputeNodeResourceLimitsSuite
--- FAIL: TestComputeNodeResourceLimitsSuite (12.52s)
    --- FAIL: TestComputeNodeResourceLimitsSuite/TestJobResourceLimits (0.27s)
        logger.go:174: 15:28:47.472 | ??? go-ds-flatfs@v0.5.1/flatfs.go:997 > ERROR flatfs could not store final value of disk usage to file, future estimates may be inaccurate
        resourcelimits_test.go:83: 
            	Error Trace:	/home/iand/ws/cod/bacalhau/pkg/test/computenode/resourcelimits_test.go:83
            	            				/home/iand/ws/cod/bacalhau/pkg/test/computenode/resourcelimits_test.go:87
            	Error:      	Not equal: 
            	            	expected: 1
            	            	actual  : 2
            	Test:       	TestComputeNodeResourceLimitsSuite/TestJobResourceLimits
FAIL
exit status 1
FAIL	github.com/filecoin-project/bacalhau/pkg/test/computenode	12.534s

```

lukemarsden:
This looks like a good change now, thanks! Please merge

lukemarsden:
Since this was an external PR, looks like our tests didn't actually run. We have #1040 to fix that, just going to merge for now and we'll see how the test suite feels about it momentarily!

lukemarsden:
Thanks @iand - if you'd like merge permissions, please ping me on Slack :)

PR #1069, opened 2022-11-10
Use Scenario-based tests to remove Docker and streamline test code

### LINKED ISSUE #1036: Remove use of Docker for things that don't test Docker

simonwo:
This PR gives a whole lot more power to the Scenario-based declarative test system. In particular, it:

- Allows Scenarios to declare more environmental configuration, so that all the setup needed for a test can be declared inline as data.
- Allows Scenarios to declare more post-conditions, so that all the checks that are done at the end of tests can be declared inline as data too.
- Introduces a ScenarioRunner that can take any scenario and spin up an appropriate devstack to execute the scenario.
- Fixes a number of small bugs with the WASM executor, which could previously not handle single file mounts and did not download inputs in parallel.

The primary motivation for this was also to get fewer jobs using Docker when they didn't need to, which is also achieved.

There's a bunch of other cool stuff in here, such as a WASM program which acts similar to POSIX `cat` 😻. 

TODO is running the old jobs that did run Docker under _both_ WASM and Docker, as we have lost a bit of test coverage of Docker jobs.

Resolves #1036.


lukemarsden (approved):

lukemarsden:
This change seems to have regressed the windows tests, please could you take a look?

PR #1077, opened 2022-11-11
Issue 880: Load WASM import modules from IPFS

lsp1138:
Modified wasm_run.go, executor.go and job.go to enable WASM modules import in the executor as per issue 880.

### DIFF
--- a/cmd/bacalhau/wasm_run.go
+++ b/cmd/bacalhau/wasm_run.go
@@ -97,6 +97,16 @@ func init() { //nolint:gochecknoinits // idiomatic for cobra commands
 		EnvVarMapFlag(&wasmJob.Spec.Wasm.EnvironmentVariables), "env", "e",
 		`The environment variables to supply to the job (e.g. --env FOO=bar --env BAR=baz)`,
 	)
+	runWasmCommand.PersistentFlags().VarP(
+		NewURLStorageSpecArrayFlag(&wasmJob.Spec.Wasm.ImportModules), "import-module-urls", "mu",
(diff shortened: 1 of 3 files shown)


simonwo:
For testing, we will need a WASM module that exposes a simple exported function and then a WASM module that requires that imported function.

Most of our WASM test data is built in Rust right now because it has good WASM support. It all lives in `testdata/wasm`. 

## Simple exported function
I think you can do this by:
- creating a Rust module in library mode (`cargo init --lib`)
- writing a function in the code that is annotated as `extern` and `#[no_mangle]` – more info here: https://www.pwnthebox.net/rust/2020/11/01/deciphering-no-mangle.html
- compiling it using `cargo build --target wasm32-unknown-unknown --release` (add a Makefile target for it)
That should give you a WASM module with the function exported. You can look at the corresponding `.wat` file to see if the export is there.

## WASM module that requires it
Now we just need a normal WASM module that requires this import...
- create a Rust module in normal mode (`cargo init`)
- at the top add the function from the other WASM module as an external reference e.g. as in the top example here, https://doc.rust-lang.org/std/keyword.extern.html, but you don't need the link annotation
- write a main function that calls that function and prints some result of it, so we know it works
- build it (`make myexample/main.wasm myexample/main.wat`) and look at the wat file to check that it requires the function as an import

## Test scenario
Take a look at `pkg/test/scenario/test_scenario.go` as examples of how to set up a WASM job. You'll need to load both WASM modules onto IPFS and then call the main one, and check the result.

lsp1138:
Thanks for the comments Simon, I have added the validation step as suggested. I will get on with building the test run with the WASM import next.

simonwo:
Merged in https://github.com/filecoin-project/bacalhau/pull/1144 🎉

PR #1083, opened 2022-11-11
Split tests out into unit vs integration, and run each type in parallel

lukemarsden:


wjam:
What's the expected gain from running unit tests separately but in parallel to the integration tests? I can see the advantage of running the unit tests _before_ the integration tests, to give a quick sanity check before running the slower tests, but I'm not sure what the benefit if everything is going to be done in parallel.

I also have concerns that people are going to forget to tag their tests correctly. I'm not sure if there is a good way to do this in Go - may require playing with Bash or something? IMO, the best way to organise it would be the Rust/Java way where you can say all of the tests in _this_ directory are unit tests and all of the tests in _that_ directory are integration tests.

I also had a message pop up at some point complaining about concurrency limit in CircleCI? I'm not sure if this was a genuine problem, historical problem, or something else.

lukemarsden:
Thanks for the review @wjam !

> What's the expected gain from running unit tests separately but in parallel to the integration tests? I can see the advantage of running the unit tests before the integration tests, to give a quick sanity check before running the slower tests, but I'm not sure what the benefit if everything is going to be done in parallel.

a) it will be faster overall due to more parallelism
b) less to re-run on failure
c) you get feedback on your PR from CI faster

> I also have concerns that people are going to forget to tag their tests correctly. I'm not sure if there is a good way to do this in Go - may require playing with Bash or something? IMO, the best way to organise it would be the Rust/Java way where you can say all of the tests in this directory are unit tests and all of the tests in that directory are integration tests.

The tags are `integration` or `!integration` - the latter isn't required, if developers forget to tag their test files, all that will happen is that the tests will run in _both_ the unit and integration suites (I think). That's my read from https://mickey.dev/posts/go-build-tags-testing/ anyway, but please check my reasoning!

Hopefully the concurrency limit in CircleCI is something we can throw more money at them to fix 🤷

wjam:
But if the unit tests are so fast, is it worth the extra time needed to spin up more boxes to download/compile the same packages? How much code coverage do we get out of just the unit tests? (I you want to copy/paste the work I was doing on getting the coverage - https://github.com/filecoin-project/bacalhau/compare/coverage - feel free)

lukemarsden:
ah please land the coverage separately! That's awesome to have

The unittests aren't _that_ fast, it's 45 seconds to a minute, I think there's some benefit here.

PR #1142, opened 2022-11-14
Avoid flaky port collisions in tests by assign node ports up front

simonwo:
Calling `freeport.GetFreePort()` will return a port that is currently free. However, as we do not actually set up our API or metrics endpoints before the next call to `GetFreePort()` the previously returned values are actually still free and so sometimes randomly get returned.

This commit changes this to allocate all ports up front, and then dole them out as required to the services that need them.


lukemarsden:
Yay thanks

PR #1158, opened 2022-11-15
Try 2: Update/improve progressive UX and text

lukemarsden:
re-doing #1081 after revert
Now main has better CI behavior, should be easier to see what this breaks in the tests (if anything)


lukemarsden:
@aronchick if it's ok I'm going to leave it with you to land this, is that alright?

aronchick:
Replaced with https://github.com/filecoin-project/bacalhau/pull/1529

PR #1273, opened 2022-11-21
Avoid common test flake of IPFS failing to start

wjam:
A common test failure is to do with attempting to start up the devstack but failing with an error saying `failed to create ipfs node` due to `bind: address already in use`.

This error occurs because IPFS must be told which ports it is going to be used before the listener on that port can be started. This introduces a significant time delay after discovering a free port which allows other processes to start using that port.


wjam:
This is admittedly a band-aid solution, but it's the only thing available as IPFS expects configuration as a file and doesn't start up all listeners at the same time.

Issue #1274, opened 2022-11-22
Tests fail due log in APIServer after test has completed


wjam:
`APIServer.ListenAndServe` logs messages when the listener has closed, which may occur _after_ the test has finished. We need to make sure that we block the test until the listener has closed.

```
github.com/filecoin-project/bacalhau/pkg/publicapi.(*APIServer).ListenAndServe(0xc000305650, {0x2a1a7d0, 0xc0000561b0}, 0xc002e000f0?)
	/home/circleci/repo/pkg/publicapi/server.go:197 +0xaff
github.com/filecoin-project/bacalhau/pkg/publicapi.SetupRequesterNodeForTestsWithPortAndConfig.func1()
	/home/circleci/repo/pkg/publicapi/util.go:154 +0x3c
created by github.com/filecoin-project/bacalhau/pkg/publicapi.SetupRequesterNodeForTestsWithPortAndConfig
	/home/circleci/repo/pkg/publicapi/util.go:153 +0x128b
```

PR #1275, opened 2022-11-22
Test Go packages in parallel

### LINKED ISSUE #1034: Implement parallel?
ETA: 2022Q4

Go test suites offer built in parallelization - we should do our best to use.

wjam:
Reduce the CI time by testing Go packages in parallel. All CI instances have at least 8 CPUs (Windows has 16), so we set the number of parallel packages to _less_ than the number of CPUs to ensure that there's capacity to run additional Docker containers, devstacks, etc.

Also fix a test flake where APIServer was able to log _after_ the test as finished.

Fixes #1034 #1274

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@
 version: 2.1
 
 executors:
+  # Available resource classes - https://circleci.com/product/features/resource-classes/
   linux-amd64:
     machine:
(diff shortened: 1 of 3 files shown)


wjam:
Seems like this roughly halves the time spent running `dev_branches`, although this number is now heavily influenced by how long it takes to be scheduled onto an instance.

PR #1296, opened 2022-11-23
Set Timeout to 3600

js-ts:


lukemarsden:
Please just set the timeout param in examples that need it. Dave has been clear that we don't want to invite excess usage of free compute on the platform.

Issue #1120, opened 2022-11-13
M3: Dockerize Bacalhau


wdbaruni:
Deploying Bacalhau to production is more than just running a binary. We have dependencies that must be installed and running on the servers, including IPFS, Prometheus, log publishing agent (missing today), health checker, Postgres client, ...

Node preparation complexity is visible in [install_node.sh](https://github.com/filecoin-project/bacalhau/blob/b2120012ea/ops/terraform/remote_files/scripts/install-node.sh) and #802

This task is about exploring shipping Bacalhau as a docker image along with all the mentioned dependencies. This is part of #1119 as it can open the door to run Bacalhau on ECS, which can also simplify periodic benchmarking of the network by adding hundreds or thousands of nodes on fewer machines

ETA: 2023Q2

aronchick:
Fascinating! What if we went the other direction? Would it be possible to pack all dependencies into a single binary?

philwinder:
@wdbaruni Just had a chat on slack: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1669280791045329

And cc this PR: https://github.com/filecoin-project/bacalhau/pull/1364

Tasks:

* [ ] Develop Bacalhau docker container (see #1364)
* [ ] Add CI to build images for all systems
* [ ] Add instructions to compute provider docs and refactor a bit to show the different installation methods
* [ ] Add to the user docs to show people how to run a private node for testing purposes

SvenDowideit:
oh boy, yup, I'll go read the install_node.sh etc, and work on #1364 to add that to both the quickstart Docker image, and some words for users that want to setup a node manually

philwinder:
Added to docs: https://github.com/bacalhau-project/docs.bacalhau.org/pull/81

A bit rough, and a few edge cases, but mostly working.

PR #1364, opened 2022-11-24
[WIP] I'd prefer to try things out using Docker

SvenDowideit:
for #1120
matching docs PR: https://github.com/bacalhau-project/docs.bacalhau.org/pull/40

I was reading https://github.com/bacalhau-project/docs.bacalhau.org/blob/main/docs/running-node/quick-start.md and decided that having installed Docker, I wouldn't want to run ipfs and bacalhau in separate terminals and have them stop when I lost my ssh connection.

so I've started following the docs, but making containers - which is what this PR shows (and i'm sure there are things that should be done differently) - ideally, these images would be auto-built for each release (or combined into one `bacalhauproject/quickstart:latest` image (each image is ~220MB, but a combined image seems about 320MB)

I'd love some feedback on things that would suite the project's goals better.

Assuming this work is desired, I'll also make a PR to the docs project.... (the README.md in this PR should mostly go there)
* the https://docs.bacalhau.org/running-node/test-network docs don't make it clear - how is this different to `bacalhau serve` - is a devstack not connected to the bigger network?


docker/quick-start/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+# To see that bacalhau is running your docker job locally, you can run 'docker events` in another terminal, and you should see the container 'create', 'start', 'die', and 'destroy' events for a container with a name starting with 'bacalhau'
+
+docker exec -it \
+    --env BACALHAU_API_HOST=127.0.0.1 \
+    --env BACALHAU_API_PORT=1234 \
+    bacalhau \
+     bacalhau docker run ubuntu echo hello
+
+# Interestingly, because your bacalhau node is a part of the bootstrap-cluster, you can get info about your job directly from your node, or via the <<<insert dns..name>>>
```
SvenDowideit, line 54: this actually makes me uncomfortable - I _think_ I'd rather start with a local, not joined to the global bacalhau network - so as not to worry new users / their IP management team

SvenDowideit:
ah, and then there's https://github.com/filecoin-project/bacalhau/blob/main/docs/running_locally.md which seems similar to https://github.com/bacalhau-project/docs.bacalhau.org/blob/main/docs/running-node/test-network.md

docker/quick-start/README.md
SvenDowideit: in the slack thread, @philwinder suggested quickstart would be for a user adding a temporary node to the global bacalhau network, and then the 'make a totally safe temporary? private bacalhau network' would be a separate doc.

This is a more realistic goal for my current knowledge :) and I can start on the other in a follow on PR

SvenDowideit:
oh wow, so the current doc uses ( https://docs.bacalhau.org/running-node/quick-start#ensure-ipfs-is-running ) :

`export IPFS_VERSION=$(wget -q -O - https://raw.githubusercontent.com/filecoin-project/bacalhau/main/ops/terraform/production.tfvars | grep --color=never ipfs_version | awk -F'"' '{print $2}')`

to get what release of kubo to use, and the production tf says v0.12.2 - from April 2022 - they're up to 0.17 atm - and there isn't an `ipfs/kubo:v0.12.2` image. So I guess one question is - @enricorotundo is the prod/stable ipfs_version likley to change soon? or should I pick some other place (and I can update the docs)?

it looks like https://hub.docker.com/r/ipfs/kubo/tags?page=6 only goes back to July2022 :)

https://github.com/ipfs/kubo/releases?page=2

__OR__ for the changes I'm proposing, is there a public IPFS node address we can default to?

docker/quick-start/Dockerfile
```diff
@@ -0,0 +1,27 @@
+
+# this container should match the manual steps in 
+# https://docs.bacalhau.org/running-node/quick-start
+
+# the idea of the quickstart is to help someone bring up the simplest global cluster node
+
+# TODO: build-args for what version of bacalhau?
+
+FROM ubuntu:22.04 as bacalhau
```
philwinder, line 9: I think I'd prefer to use a proper multistage golang/scratch container here, instead of ubuntu. We're inside the repo at this point, so we can actually make nightly builds for every commit.

WDYT?

docker/quick-start/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# crib from, or use https://github.com/filecoin-project/bacalhau/blob/b2120012ea/ops/terraform/remote_files/scripts/install-node.sh
+RUN apt-get update -y && apt-get install -y \
+      ca-certificates \
+      curl \
+      gnupg \
+      lsb-release
+
+RUN curl -sL https://get.bacalhau.org/install.sh | bash
+
+RUN cd /tmp \
```
philwinder, line 20: Why are we installing docker here? I thought we were going to leverage the host's docker? (I'll double check the other comments in a second)

docker/quick-start/bacalhau-entrypoint.sh
```diff
@@ -0,0 +1,40 @@
...
+  if ! docker inspect bacalhau-ipfs > /dev/null ; then
+    # TODO: can we put the desired version into the bacalhau binary?
+    export IPFS_VERSION=$(wget -q -O - https://raw.githubusercontent.com/filecoin-project/bacalhau/main/ops/terraform/production.tfvars | grep --color=never ipfs_version | awk -F'"' '{print $2}')
+    #ignoring the repo version - they're all too old atm
+    export IPFS_VERSION=latest
+    # run IPFS (this should eventually be run in the bacalhau entrypoint)
+    # TODO: run container as readonly fs
+
+    echo "Starting the backalhau-ipfs ${IPFS_VERSION} container"
+    docker run -d \
```
philwinder, line 17: This doesn't feel right. Starting a dependent container in the entrypoint script of another container? Surely it's easier to start it separately (or use docker compose or something?)

Alternatively, is it possible to bundle the kubo binary at build time, then run a forked process to run kubo?

(Again, I'll check the notes in a second)

docker/quick-start/bacalhau-entrypoint.sh
```diff
@@ -0,0 +1,40 @@
...
+        --name bacalhau-ipfs \
+        -v bacalhau_ipfs_staging:/export \
+        -v bacalhau_ipfs_data:/data/ipfs \
+        -p 4001:4001 \
+        -p 4001:4001/udp \
+        -p 127.0.0.1:8080:8080 \
+        -p 127.0.0.1:5001:5001 \
+            ipfs/kubo:${IPFS_VERSION}
+
+    # TODO: how do we test if the image has pulled, and we're started enough to continue?
```
philwinder, line 28: Yeah exactly, not great.

philwinder (changes requested):
In general, I think we're going down the wrong path here. Depending on what you're trying to do, I see two options:
1. If you're trying to get something up and running as quick as possible, then we should use devstack. That should be as simple as a normal multistage go build and an entrypoint script calling bacalhau devstack.
2. If you're trying to give pro users a dockerised experience, then we should do the above, but dockerise `bacalhau serve` and leave the IPFS node to a separate container. If you want to make that process nicer, then maybe you could bundle the IPFS binary. Or use docker-compose/kubernetes to allow a more distributed setup.

WDYT?

SvenDowideit:
first up - excellent - thanks for the feedback - at this point, I don't know enough of the existing thinking, and so don't even know what arguments I'm interested in :)

Initially, I'm interested in (1), because that's the stage of play I'm at - and running a devstack is not what I'm after (tho I did that, because there doesn't currently seem to be something closer to what I have in my head).

I'm really hoping for a (1) that is as few commands as possible, that would give me a fully containerised, similar to a minimal production config, that can give me some feel for what I can expect (load, traffic, performance and utility (to me)), before I then think about teasing those components into my existing landscape.

So in many ways, I'm splitting up the path to doing (2) as my time isn't available enough to pretend I can get that done in one chunk :)

Pulling in everything :+1: I'm going to head towards making that multi-stage image, and a compose file, and likely a bacalhau:quickstart that just contains the compose file to be a `install docker, then docker run quickstart`

docker/quick-start/Dockerfile
SvenDowideit: happy to - tho I'm guessing there also needs to be a release Dockerfile that makes the tagged image from the released `bacalhau` binary

docker/quick-start/Dockerfile
SvenDowideit: this is installing the docker commandline - I didn't check, I presumed that `bacalhau` exec's `docker`? (if not YAY)

docker/quick-start/bacalhau-entrypoint.sh
SvenDowideit: For cleanup reasons (and because i was stuffing with different versions), I wanted to have kubo in a separate container, with separate volumes - I'm going to head further down that route as that fits the compose route.

nicely, the making of `bacalhauproject/bacalhau:<commit sha>` and `bacalhauproject/bacalhau:v0.17.0` for use in a compose file, isn't incompatible with using that same image as `docker run --dir -v dockersockt bacalhauproject/bacalhau:<commit sha> --devstack`

docker/quick-start/bacalhau-entrypoint.sh
SvenDowideit: grin - that's why the RancherOS installer container pattern became the other `docker run -it -v dockersocket complexservice:installer` which then could do the interactive things a user needed to answer, and finish by running `docker compose up` - and exit.

docker/quick-start/Dockerfile
SvenDowideit: oh, mind you, I went with ubuntu:22.04 because the docs all suggest that's what is tested - going deeper into the formal deployment images is really a matter of dev time, finding the corner cases (and worth it over time)

docker/quick-start/Dockerfile
SvenDowideit: It looks like most docker things are done via the APIClient, but there's still `https://github.com/filecoin-project/bacalhau/blob/main/pkg/executor/docker/executor.go#L287`

docker/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+if [[ ! -e /var/run/docker.sock ]]; then
+  echo "ERROR: no docker socket mounted into the container. Please re-run with:"
+  echo
+  echo "docker run -dit -v /var/run/docker.sock:/var/run/docker.sock --publish 1234:1234 bacalhauproject/bacalhau:dev"
+  return 1
+fi
+
+if [[ "${1:-}" == "devstack"  ]]; then
+  LOG_LEVEL=debug bacalhau \
+    --api-host ${BACALHAU_API_HOST} \
```
SvenDowideit, line 19: sadly, this doesn't work - see #1501

docker/quick-start/bacalhau-entrypoint.sh
SvenDowideit: humm, looking into what `bacalhau devstack` does - it's also starting dependant containers from what will be the bacalhau container - for example the `LOTUS_TEST_IMAGE`

docker/quick-start/Dockerfile
aronchick: Have we looked at Chainguard images? https://www.chainguard.dev/chainguard-images

philwinder:
Closing in favour of #1748

PR #1378, opened 2022-11-25
Initial trivial simulator framework & "smart contract" mock

lukemarsden:
* A new simulator server which can behave as both a transport (via websockets) and have "smart contract" style logic attached to it, including an initial concept of wallet balances (but no mutating of wallet balances or enforcement yet)
* A transport implementation that uses the simulator server instead of libp2p (for simulations only! not for production!)
* A test which demonstrates that the rest of the Bacalhau infrastructure still basically works when using the simulator transport (centralized communication & smart contract) instead of libp2p


lukemarsden:
<3 props to @binocarlos for the main part of this 🎉

PR #1383, opened 2022-11-25
Auto-generate clients in many languages from our swagger.json

simonwo:
Now that we publish a swagger.json file for our API, we can generate clients in many languages automatically.

This commit automates the generation of clients and ties it in with the build process of the swagger JSON.


simonwo:
Are we sure we want them all here?

enricorotundo:
Update: will start by generating a python client first (see https://github.com/filecoin-project/bacalhau/issues/1502), and then we'll add other languages

Issue #1402, opened 2022-11-28
Test and Investigate IPFS Issues


philwinder:
When working with examples that have a large number of concurrent IPFS gets/puts (approx more than 100 ish) bac grids to a halt. Gets don't work. Jobs move into an error state. But it doesn't seem to happen precisely at the same time. Sometimes only a few jobs fail. Sometimes most of them.

The best way I have at the moment of recreating this is to run the ethereum example, which has approx 150 jobs of 50MB IPFS reads, and 1KB writes.

If it is IPFS related, I should be able to reproduce on devstack by stressing IPFS file access.

* add a canary job that gets and stores IPFS content
* Using devstack, develop load test that recreates observed IPFS issues
* Check that opentracing spans surround the IPFS storage tasks
* If issues are observed, track down the source

philwinder:
I wrote a script to test this and found no major issues (other than the bac get timeouts that everyone already knows about). I turned this into an example so it's not wasted time: https://docs.bacalhau.org/examples/workload-onboarding/python-script/

No idea why this script didn't work first time around. Possibly it was a bug that has already been fixed.

philwinder:
Found the trace in honeycomb. Example: https://ui.honeycomb.io/protocol-labs/datasets/bacalhau_production/result/9sfVX9fMqYJ?tab=traces

Basically you want to do a query with a `WHERE` of `name = pkg/ipfs.Get`. That will show some spikes if you look over 30 days.

This shows that some get's have taken many hours (36 for the longest one)...

![Screenshot 2022-11-30 at 15 02 47](https://user-images.githubusercontent.com/8793723/204832927-b5e5cde2-1044-4632-bad1-b9ff171c48ea.png)

PR #1425, opened 2022-11-29
feat(canary): add scenario that leverages IPFS

philwinder:
Related to #1402


ops/aws/canary/lib/canary-stack.ts
```diff
@@ -40,6 +40,7 @@ export class CanaryStack extends cdk.Stack {
         this.createLambdaScenarioFunc({action: "submitAndGet", timeoutMinutes: 1, rateMinutes: 2, memorySize: 1024});
         this.createLambdaScenarioFunc({action: "submitAndDescribe", timeoutMinutes: 1, rateMinutes: 2, memorySize: 256});
         this.createLambdaScenarioFunc({action: "submitWithConcurrency", timeoutMinutes: 1, rateMinutes: 2, memorySize: 256});
+        this.createLambdaScenarioFunc({action: "submitDockerIPFSJobAndGet", timeoutMinutes: 1, rateMinutes: 2, memorySize: 256});
```
wdbaruni, line 43: What I found out when running `submitAndGet` is that more memory is required, or else reading from IPFS sometimes hangs with no explanation or logs, and the scenario fails eventually after 1 minute timeout. Thats the reason submitAndGet uses 1GB to be stable, while other scenarios are okay with just 256MB. 

This scenario will need a minimum of 1GB as well, and might require more as it is downloading  and uploading a larger file compared to submitAndGet. We can start with 1GB and increase it later if we see issues.

Other than that, the PR looks good to me :)

PR #1430, opened 2022-11-30
Refactor model.Job and `/submit` payload + make canary tests work locally

enricorotundo:


ops/aws/canary/lambda/pkg/scenarios/list.go
```diff
@@ -2,15 +2,14 @@ package scenarios
...
 
-	"github.com/filecoin-project/bacalhau/cmd/bacalhau"
+	"github.com/rs/zerolog/log"
 )
 
 func List(ctx context.Context) error {
 	// intentionally delay creation of the client so a new client is created for each
 	// scenario to mimic the behavior of bacalhau cli.
-	client := bacalhau.GetAPIClient()
+	client := getClient()
```
enricorotundo, line 12: `bacalhau.GetAPIClient()` would set `apiHost, apiPort` to production values.
`getClient()` will use test values if applicable

ops/aws/canary/lambda/pkg/scenarios/submitDockerIPFSJobAndGet.go
```diff
@@ -6,32 +6,36 @@ import (
...
 // This test submits a job that uses the Docker executor with an IPFS input.
-//
 func SubmitDockerIPFSJobAndGet(ctx context.Context) error {
-	client := bacalhau.GetAPIClient()
+	client := getClient()
 
 	cm := system.NewCleanupManager()
 	j := getSampleDockerIPFSJob()
+
+	// In test we use the cid of the file we uploaded in scenarios_test.go
```
enricorotundo, line 21: In tests, we push the test file to devstack ipfs

ops/aws/canary/lambda/pkg/test/scenarios_test.go
```diff
@@ -4,18 +4,35 @@ import (
...
+	t.Logf("BACALHAU_ENVIRONMENT: %s", os.Getenv("BACALHAU_ENVIRONMENT"))
+
+	swarmAddresses, err := stack.Nodes[0].IPFSClient.SwarmAddresses(context.Background())
+	require.NoError(t, err)
+	// Need to set the swarm addresses for getIPFSDownloadSettings() to work in test
+	os.Setenv("BACALHAU_IPFS_SWARM_ADDRESSES", strings.Join(swarmAddresses, ","))
+	t.Logf("BACALHAU_IPFS_SWARM_ADDRESSES: %s", os.Getenv("BACALHAU_IPFS_SWARM_ADDRESSES"))
+
+	// Add file to devstack IPFS
+	cid, err := ipfs.AddFileToNodes(context.Background(), "data.tar.gz", stack.IPFSClients()...)
```
enricorotundo, line 32: We need to add this test file manually because devstack, for some reason, wouldn't pull it from web3.storage

pkg/model/job.go
```diff
@@ -44,10 +47,8 @@ type Job struct {
 
 	// All local events associated with the job
 	LocalEvents []JobLocalEvent `json:"LocalJobEvents,omitempty"`
-}
 
-func (job Job) String() string {
-	return job.ID
```
enricorotundo, line 50: Removed this method because it'd prevent me from printing the full Job object in logs. Plus, it had zero references.

ops/aws/canary/lambda/Makefile
```diff
@@ -24,7 +24,10 @@ build: ${BIN_TARGETS}
 .PHONY: test
 test: unittests.xml
 
-unittests.xml: $(shell find ./pkg ./cmd -type f)
+pkg/test/data.tar.gz:
+	curl https://w3s.link/ipfs/bafybeihxutvxg3bw7fbwohq4gvncrk3hngkisrtkp52cu7qu7tfcuvktnq -L -o $@
+
+unittests.xml: pkg/test/data.tar.gz $(shell find ./pkg ./cmd -type f)
```
simonwo, line 30: Do we actually need this if we're now just populating with a different CID? Could we just use `cid, err := ipfs.AddTextToNodes("some random text")` like the other tests?

enricorotundo: Good point. I'll have to add an extra if-then in the checksum step because they'll differ but I guess that's still better than depending on the internet when testing

simonwo:
LGTM!

My future work will now probably just involve splitting the `Spec` three ways, into an IPVM `Task`, IPVM `Config` and Bacalhau-specific Deal and Sharding stuff.

PR #1492, opened 2022-12-12
Support submitting UCAN Invocations as jobs

simonwo:
This commit adds support for submitting UCAN Invocations in JSON form as defined by the draft spec using the Bacalhau client.

Submitted invocations are converted into Bacalhau job specs on the client side and submitted to the requestor node API as usual. Users can do this using the `bacalhau create <invocation.json>` command.

Note that this means there is no support for sharding, concurrency or minimum bidding for these jobs as these are not supported in the invocation spec. (This is because they are about defining *how many* jobs are done and how the executor is selected, which is not in scope for the Invocation spec which only focuses on execution itself).

Support has been built into the client rather than the server as an MVP because it's much easier to add support there quickly and without changing any existing APIs. The same technique of parsing an invocation into a job spec could also be applied on the server but will require a new API endpoint or e.g. content-type detection.

Unfortunately this conversion is not as straight-forward as we would like:

* UCAN models storage specs as just URIs, so we have to manually convert these into our own StorageSpec objects
* because the Go IPLD library does not currently support inline unions, we have to do some of manual re-parsing ourselves.

Some of this pain would go away if we just wholesale adopted the UCAN objects as our internal data model, but this is quite a big change to make. And of course, would require using these objects on our API as well.


enricorotundo:
re: UCAN objects. Are they stable/definitive enough to be used as our internal data model? Also, I think it'd make sense to adopt them directly at some point; are other CoD platforms keen?

binocarlos:
I agree that using UCAN internally would be a good step and I also agree that it's a rather large job so this is an awesome way to support it right away

simonwo:
> re: UCAN objects. Are they stable/definitive enough to be used as our internal data model? Also, I think it'd make sense to adopt them directly at some point; are other CoD platforms keen?

Indeeed, if we want to stablise our API ahead of our own 1.0 release, we presumably want to settle on something that isn't going to change much very quickly! The spec is still very much in draft, here: https://github.com/ucan-wg/invocation/pull/1. So I think it is open to change (probably not in significant ways, mostly around how it handles config and metadata AFAIU) and therefore maybe not something we should internally just yet.

Other CoD platforms like DAG House seem keen, and there are a bunch of people in the IPVM group who are also keen.

simonwo:
OK, I'm going to merge this PR (as I've just demo'd it to the IPVM WG, and it works!) I'm still keen for feedback on the how and when, and will be creating some tickets to cover that future work.

Issue #1138, opened 2022-11-14
Download using Estuary API for results published to Estuary


aronchick:

pyropy:
Currently job publisher starts publishing it's job result concurrently to both IPFS and Estuary. 
Given that publisher publishes data to local IPFS node its usually faster then upload to Estuary, causing only IPFS result (CID) to be returned as job result. 

Possible improvement is for publisher would to wait for few seconds before returning job results to ensure that there's a bigger chance for Estuary publisher to finish the publish process.

In case Estuary publisher finishes in time, URL for Estuary download should be added to return [value](https://github.com/filecoin-project/bacalhau/blob/7a107b35f5ad80eb34c48018b60fe2a0fa0ce339/pkg/publisher/estuary/publisher.go#L94).

Bacalhau CLI `get` command should also be changed to download from Estuary when possible and fallback to IPFS if there is no results for Estuary.

pyropy:
@simonwo I guess we can close this issue?

simonwo:
Yes! Merged in #1527.

PR #1514, opened 2022-12-14
Add priority fanout publisher

pyropy:
This PR relates to issue #1138 

- Added preferred (priority) publisher to fanout publisher. Since we can only return one result from publisher we might want to have higher priority for some publisher over the others (e.g. we might want estuary publisher result over the IPFS publisher) 
- Result from prioritised fanout publisher is not returned immediately if publisher is not of highest prioritised. Once we get first result from publisher with lower priority timeout is triggered. Once timeout is reached, all results are checked for each publisher in provided order and the one from publisher with highest priority is returned (first result found).


pkg/publisher/combo/fanout.go
```diff
@@ -102,13 +116,49 @@ func (f *fanoutPublisher) PublishShardResult(
...
-		return value, nil
-	case err := <-errorChannel:
-		return model.StorageSpec{}, err
+	type result struct {
+		Sender publisher.Publisher
+		Value  model.StorageSpec
 	}
+
+	timeoutChannel := make(chan bool, 1)
+	results := make([]result, 0)
```
simonwo, line 125: A very minor nitpick is that this could probably be `make([]result, 0, len(publishers))` as we are expecting a result from every publsiher.

pkg/publisher/combo/fanout.go
```diff
@@ -102,13 +116,49 @@ func (f *fanoutPublisher) PublishShardResult(
...
+			}
+
+			// start timeout for other results when first result is returned
+			if len(results) == 1 {
+				go func() {
+					// TODO: Make timeout value configurable
+					time.Sleep(time.Second * 2)
+					timeoutChannel <- true
+				}()
+			}
```
simonwo, line 144: If the result we got was from the preferred publisher, we don't need to start a timeout and can return immediately. Similarly, we can break out of the loop before everyone returned if we got a result from the preferred publisher because we know we are going to use it right away.

pkg/publisher/combo/fanout.go
```diff
@@ -102,13 +116,49 @@ func (f *fanoutPublisher) PublishShardResult(
...
+
+	// search for preferred result
+	for _, result := range results {
+		if result.Sender == f.preferred {
+			return result.Value, nil
+		}
+	}
+
+	// return first result if preferred not found
+	return results[0].Value, nil
```
simonwo, line 161: Would a `map[Publisher]StorageSpec` make more sense here? And then for the default case just use `maps.Values`?

simonwo (commented):
Looks great so far!!

pkg/publisher/combo/fanout.go
pyropy: Makes sense!

pkg/publisher/combo/fanout.go
pyropy: Great feedback!

pkg/publisher/combo/fanout.go
pyropy: Yup, makes more sense 👍🏻

pkg/publisher/combo/fanout_test.go
```diff
@@ -24,11 +24,12 @@ var uninstalledPublisher = mockPublisher{
...
-		"noone is installed":    {NewFanoutPublisher(&uninstalledPublisher), uninstalledPublisher},
-		"waits for good value":  {NewFanoutPublisher(&errorPublisher, &sleepyPublisher), sleepyPublisher},
-		"returns error for all": {NewFanoutPublisher(&errorPublisher, &errorPublisher), errorPublisher},
+		"single publisher":                 {NewFanoutPublisher(&healthyPublisher), healthyPublisher},
+		"takes first value":                {NewFanoutPublisher(&healthyPublisher, &sleepyPublisher), healthyPublisher},
+		"waits for installed":              {NewFanoutPublisher(&uninstalledPublisher, &sleepyPublisher), sleepyPublisher},
+		"noone is installed":               {NewFanoutPublisher(&uninstalledPublisher), uninstalledPublisher},
+		"waits for good value":             {NewFanoutPublisher(&errorPublisher, &sleepyPublisher), sleepyPublisher},
+		"returns error for all":            {NewFanoutPublisher(&errorPublisher, &errorPublisher), errorPublisher},
+		"waits for highest priority value": {NewPrioritizedFanoutPublisher(&sleepyPublisher, &healthyPublisher), sleepyPublisher},
```
simonwo, line 33: Could we add a test that it definitely only waits for the maximum time? So if the max is 2s then it clearly only waits for that length. E.g. something like

```go
"only waits for max time": {NewPrioritizedFanoutPublisher(20 * time.Millisecond, &sleepyPublisher, &healthyPublisher), healthyPublisher}
```

And also a test for what happens if the timeout ends before anyone has returned?

simonwo (commented):
Changes look good!

Would you also be able to update the config to select the new prioritized publisher? https://github.com/filecoin-project/bacalhau/blob/fad620b9ec7c3cd3bf5e2749efc3207b2c936f2d/pkg/publisher/util/utils.go#L34-L37

pyropy:
> Changes look good!
> 
> Would you also be able to update the config to select the new prioritized publisher?
> 
> https://github.com/filecoin-project/bacalhau/blob/fad620b9ec7c3cd3bf5e2749efc3207b2c936f2d/pkg/publisher/util/utils.go#L34-L37

Sure thing! :D

pyropy:
I was thinking about splitting up downloading estuary results into new PR. The way I thught to do it is to create a new package named `downloader` or `results_downloader` that would hold all downloader implementations similar to how `publisher` package holds all publishers. If we would do it that we it would cause quite a lot of changes and would result in big PR. 

@simonwo  What do your thoughts on that?

simonwo:
> I was thinking about splitting up downloading estuary results into new PR.

Yes, I agree! This PR is looking great now; let's get it merged.

Can you target `dev` branch instead of `main`? We are taking a feature freeze over Christmas whilst people aren't available to urgently fix any issues with new code, and we'll merge `dev` early in the new year.

PR #1527, opened 2022-12-17
Add multiple downloader implementations

pyropy:
This PR relates to discussion #1524 and issue #1138

- Move IPFS downloader to new `downloader` package
- Add Estuary and HTTP downloader
- Add mapped downloader provider that maps storage source type to downloader
- Upgrade Estuary package
- Add URL to Estuary Publisher result (old results without URL will fallback to IPFS download by using CID)
- Add integration test for Estuary downloader


pkg/model/downloader.go
```diff
@@ -0,0 +1,22 @@
+package model
+
+const (
+	DownloadFilenameStdout   = "stdout"
+	DownloadFilenameStderr   = "stderr"
+	DownloadFilenameExitCode = "exitCode"
+)
+
+type DownloaderSettings struct {
+	TimeoutSecs    int
```
simonwo, line 10: Given that we're refactoring, shall we turn this into a `time.Duration`?

pkg/downloader/types.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"context"
+
+	"github.com/filecoin-project/bacalhau/pkg/model"
+)
+
+type Downloader interface {
+	// FetchResult fetches result contained in PublishedShardDownloadContext
+	FetchResult(ctx context.Context, shardCidContext model.PublishedShardDownloadContext) error
```
simonwo, line 11: Is `PublishedShardDownloadContext` definitely the right type to be passing to a Downloader? The current implementations only ever seem to access the `Result` and the `CIDDownloadDir`. Would it be better to only pass those things?

One motivation for this is that we currently also have `pkg/storage/url/urldownload/storage.go` which downloads things from URLs for _input_. That currently takes just a StorageSpec. I can see a world where we (in the future) unify the storage providers and the downloader providers, so that you only have to implement one of them to make it work on both server and client. So having a similar interface would make that easier.

simonwo (commented):
This is really great so far! And great to see an HTTP downloader appear too :)

It would be really nice to have some tests for the various downloaders. The Estuary one can contact Estuary directly if you use an `intergration` test.

pkg/downloader/ipfs/downloader.go
```diff
@@ -0,0 +1,83 @@
...
+	Client   *ipfs.Client
+}
+
+func NewIPFSDownloader(ctx context.Context, cm *system.CleanupManager, settings *model.DownloaderSettings) (*Downloader, error) {
+	// NOTE: we have to spin up a temporary IPFS node as we don't
+	// generally have direct access to a remote node's API server.
+	n, err := spinUpIPFSNode(ctx, cm, settings.IPFSSwarmAddrs)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
pyropy: I think downloader only lives until downloading is finished. Generally we could possibly do this in the `FetchResult` function for the downloader and then just kill the node once it's done downloading.

pyropy:
@simonwo I've done some fixes you've suggested and added integration test for Estuary and canary test scenario where job using docker as runtime and Estuary as publisher is submitted. I'm unsure should I add tests for http downloader since we're basically testing it with Estuary integration test, but if necessary I would be happy to add it. 

I would appriciate if you could take another look at the code and give me suggestions for potential improvements 🙏🏻

pkg/downloader/http/downloader.go
```diff
@@ -0,0 +1,79 @@
...
+	defer span.End()
+
+	err := func() error {
+		log.Ctx(ctx).Debug().Msgf(
+			"Downloading result URL %s '%s' to '%s'...",
+			result.Data.Name,
+			result.Data.URL, downloadDir,
+		)
+
+		filepath := fmt.Sprintf("%s/%s", downloadDir, result.Data.Name)
```
simonwo, line 38: Please use `filepath.Join` for Windows portability!

pkg/downloader/estuary/downloader.go
```diff
@@ -0,0 +1,34 @@
...
+		httpDownloader: httpDownloader,
+	}, nil
+}
+
+func (downloader *Downloader) FetchResult(ctx context.Context, result model.PublishedResult, downloadDir string) error {
+	ctx, span := system.GetTracer().Start(ctx, "pkg/downloader.estuary.FetchResult")
+	defer span.End()
+
+	return downloader.httpDownloader.FetchResult(ctx, result, downloadDir)
+}
```
simonwo, line 34: Great! ...Can we have any fallback to IPFS behaviour if the StorageSpec doesn't have a URL? I.e. all of our existing `StorageSourceEstuary` specs will only have a CID... and it would be nice not to break downloads for old jobs.

pkg/downloader/util/utils.go
```diff
@@ -0,0 +1,61 @@
...
+
+	estuaryDownloader, err := estuary.NewEstuaryDownloader(settings)
+	if err != nil {
+		return nil, err
+	}
+
+	return downloader.NewMappedDownloaderProvider(map[model.StorageSourceType]downloader.Downloader{
+		model.StorageSourceIPFS:    ipfsDownloader,
+		model.StorageSourceEstuary: estuaryDownloader,
+	}), nil
```
simonwo, line 60: Could we also add the HTTP downloader in here? Seems like we are then ready to have an HTTP-based publisher for free (which I am writing next!)

pkg/downloader/ipfs/downloader.go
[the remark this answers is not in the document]
simonwo: Yes, I think that would be better – we don't want these nodes sticking around for a long time.

ops/aws/canary/lambda/pkg/scenarios/submitDockerEstuaryJobAndGet.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/filecoin-project/bacalhau/pkg/downloader"
+	"github.com/filecoin-project/bacalhau/pkg/downloader/util"
+	"github.com/filecoin-project/bacalhau/pkg/model"
+	"github.com/filecoin-project/bacalhau/pkg/system"
+	"github.com/rs/zerolog/log"
+)
+
+// This test submits a job that uses the Docker executor with an IPFS input.
+func SubmitDockerEstuaryJobAndGet(ctx context.Context) error {
```
simonwo, line 18: We don't want issues with Estuary to be triggering our alerting when there's nothing we can do about it. Writing and reading from Estuary should always be on a "best effort" basis with a fallback if Estuary is not available.

...I think this canary job can't actually test this? Because the Estuary write will sometimes/often return an IPFS result from our own node, and then the Estuary download should also have an IPFS back up?

pkg/downloader/util/utils.go
```diff
@@ -0,0 +1,61 @@
...
+			settings.IPFSSwarmAddrs = os.Getenv("BACALHAU_IPFS_SWARM_ADDRESSES")
+		}
+	case system.EnvironmentStaging:
+		log.Warn().Msg("Staging environment has no IPFS swarm addresses attached")
+	}
+
+	return &settings
+}
+
+func NewIPFSDownloaders(
```
simonwo, line 43: Hm, not sure of the name – one of these is using the HTTP downloader after all! Why not just `NewStandardDownloaders` like we do elsewhere?

pkg/test/devstack/combodriver_test.go
```diff
@@ -33,7 +32,8 @@ func TestComboDriverSuite(t *testing.T) {
 const exampleText = "hello world"
 
 var testcase scenario.Scenario = scenario.Scenario{
-	ResultsChecker: scenario.FileEquals(ipfs.DownloadFilenameStdout, exampleText),
+	ResultsChecker: scenario.FileEquals(model.DownloadFilenameStdout, exampleText),
+	Contexts:       scenario.CatFileToStdout.Contexts,
```
simonwo, line 36: You can remove this line now that WASM executor doens't use `Contexts` any more.

ops/aws/canary/lambda/go.mod
```diff
@@ -311,3 +311,5 @@ require (
 	lukechampine.com/blake3 v1.1.7 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace github.com/filecoin-project/bacalhau => ../../../../
```
simonwo, line 315: Don't think we want this?

simonwo (commented):
Really great stuff. Just a few little tweaks and I think we are there!

ops/aws/canary/lambda/pkg/scenarios/submitDockerEstuaryJobAndGet.go
simonwo: Ignore me – I think this is valuable. Jobs that explicitly target Estuary should not be failing under any circumstances.

pkg/test/devstack/combodriver_test.go
simonwo:
```suggestion
```

ops/aws/canary/lambda/go.mod
pyropy: Woops, totally forgot about that one 😬

pkg/downloader/estuary/downloader.go
pyropy: Yeah that makes sense!

pyropy:
@simonwo I've added IPFS downloader to Estuary downloader used as a fallback in case we use Estuary result without `URL` property. Canary tests will fail until we upgrade bacalhau package version over there.

simonwo (approved):
Looks good!

So I'm not sure what the status is with the canary right now and how we are making changes to it (cc @wdbaruni). Maybe make your canary changes can go onto a different branch for now, and we can do them afterwards?

pyropy:
> Looks good!
> 
> So I'm not sure what the status is with the canary right now and how we are making changes to it (cc @wdbaruni). Maybe make your canary changes can go onto a different branch for now, and we can do them afterwards?

Sounds good!

PR #1530, opened 2022-12-19
fix printingUpdateForEvent deadlock

wdbaruni:
The deadlock will happen when a test case runs multiple jobs, such as :
https://github.com/filecoin-project/bacalhau/blob/improve_branch_new/cmd/bacalhau/docker_run_test.go#L854

The first job will hold the lock, exit early [here](https://github.com/filecoin-project/bacalhau/blob/improve_branch_new/cmd/bacalhau/utils.go#L811) without releasing the lock, which will block the second job.


lukemarsden:
This PR should be into `dev`

PR #1531, opened 2022-12-19
Spinners and moving fish + fix printingUpdateForEvent deadlock (to dev)

lukemarsden:
version of #1530 targeting the correct branch


lukemarsden:
replaced with #1531 which includes Walid's mutex fix

PR #1580, opened 2023-01-02
refactor: Use inmemory Kubo API if the local node is used.

### LINKED ISSUE #1523: [Refactor] Investigate the use of the "direct" IPFS API for `get` and `devstack` commands
Context: https://filecoinproject.slack.com/archives/CRLC5H4J0/p1671198847619879?thread_ts=1671195709.622489&cid=CRLC5H4J0

Discussion: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1671199237553819

**Description**
Currently we use the http API to communicate with IPFS. This is required if you want to communicate with an external IPFS node. But for the `bac get` and `bac devstack` commands we use an in-memory IPFS node.

Also, we've been observing lots of timeouts doing a `bac get` in the monitoring. We've tracked it down to the final `http.Get` inside the IPFS HTTP API, but then we can't see what happens next because it's buried deep inside the in-memory IPFS node. I.e. we can't trace what happens next without breaking open the IPFS node.

Joporro suggested that at the very least, the direct API will tell you exactly what's going on if it fails. But the direct API might also be more stable and faster because you're removing the HTTP overhead.

Unfortunately we still need the HTTP API for external IPFS nodes.

**Task**

* [ ] Investigate the feasibility of ripping out the current IPFS client and replacing it with the direct one for the `bac get` and `bac devstack` commands: https://github.com/ipfs/kubo/blob/master/core/coreapi/unixfs.go
* [ ] If feasible, do it, experiment with it, test it. You should be able to do some `bac get`s against production with it. See whether it's better/faster/whatever.
* [ ] Make sure all the tests still pass

Jorropo:
Fixes #1523


philwinder (approved):
@Jorropo Thank you so much for doing this. I've reviewed the code and it all makes sense.

First I'll rebase this, then I'm just going to do some manual tests. I'll merge it shortly.

Thank you!

philwinder:
See #1615

PR #1651, opened 2023-01-14
Merge dev into main

### LINKED ISSUE #1595: merge main -> dev
currently the dashboard depends on hearing all events via the websocket endpoint of the api server

the `dev` branch introduces our libp2p point to point protocol - we need to merge the main branch into dev before we can deploy it

wdbaruni:
Merges dev changes into main branch.
Closes #1595


wdbaruni:
The main change in behaviour which breaks the dashboard is a change in the schema of the node info streamed over websockets, and a change in the websocket uri from `/node/websocket` to `/requester/node/websocket`

@binocarlos what are your thoughts about the current `NodeInfo` found [here](https://github.com/filecoin-project/bacalhau/blob/6a1b343f834794c3702b2ae1159805520e0b5e2e/pkg/model/nodeinfo.go)? Is there any additional data you want me to add? We just need to be mindful of publishing data that can get large, such as list of executions and connected peers :)

wdbaruni:
replaced with #1703

PR #1662, opened 2023-01-17
1149 centralized logging loki

enricorotundo:
Implements https://github.com/filecoin-project/bacalhau/issues/1149

This branch has been deployed to Prod already.

Loki dashboard for aggregated logs (requires credentials): https://protocollabs.grafana.net/goto/lKmGkWT4z?orgId=1
Screenshot 2023-01-17 at 17 43 27


wjam:
Shouldn't we filter the JSON logs so that we avoid over-paying for logging, such as only having `INFO` and above?

enricorotundo:
> Should we filter the JSON logs so that we avoid over-paying for logging, such as only having `INFO` and above?

@wjam Fair point 👍

enricorotundo:
Are there other systemd services (ipfs, system, etc) or local logs we'd like to collect?

enricorotundo:
Note: https://github.com/filecoin-project/bacalhau/issues/1693 has been hot-patched in production (ie. edit promtail config, restart service - no cluster redeploy to avoid downtime)

PR #1679, opened 2023-01-18
Allow HTTP job networking using an allowlist

simonwo:
We have previously added support for HTTP networking that enforces domains are restricted to a certain allowlist. The idea is for compute providers to somehow decide what jobs they are willing to run using the domain list in the job spec and bid appropriately.

This commit adds the simplest possible decision-making: an allowlist. There is now a probe to be used with the "job selection exec probe" that will compare what is on the allowlist to what is provided in the spec. If there is anything in the spec not on the allowlist, the job is rejected (or if the job is asking for full networking).

This is intended as configuration for the "Bacalhau compute provider" rather than a new core feature for all compute providers. We are still learning about what compute providers will do with networking features and expect them to want more sophisticated strategies in the future.

### DIFF
--- a/ops/terraform/main.tf
+++ b/ops/terraform/main.tf
@@ -91,6 +91,14 @@ sudo tee /terraform_node/start-bacalhau.sh > /dev/null <<'EOI'
 ${file("${path.module}/remote_files/scripts/start-bacalhau.sh")}
 EOI
 
+sudo tee /terraform_node/apply-http-allowlist.sh > /dev/null <<'EOI'
+${file("${path.module}/remote_files/scripts/apply-http-allowlist.sh")}
+EOI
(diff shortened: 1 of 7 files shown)


enricorotundo:
What about adding, in the future, a hot reloading/editing of the whitelist via, say, the cli, to relieve the burden or re-deploying the whole cluster?

simonwo:
> What about adding, in the future, a hot reloading/editing of the whitelist via, say, the cli, to relieve the burden or re-deploying the whole cluster?

Yes, agree that if we see a lot of contention over the allowlist in the future, easier upgrading is definitely possible! We could also switch to using the HTTP probe instead and deploy the allowlist as a separate app.

PR #1718, opened 2023-01-19
Stop Bacalhau from crashing when Docker times out

### LINKED ISSUE #1694: Bacalhau crashes when running docker containers if `--timeout` is reach
The current implementation of the Docker executor waits for the container to stop running, fetches the container, then fetches the logs for the container before parsing the logs.

The problem with this is if the `context.Context` times out before the container stops running, then `ContainerWait` call will return an error, fetching the container, fetching the logs will also then fail, and then parsing the logs will panic as it assumes the logs object to be populated.

Ideally, Bacalhau should return all of the logs up to that point or at least not crash.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x10182bd44]

goroutine 421 [running]:
github.com/filecoin-project/bacalhau/pkg/executor.writeOutputResult({0xc003e54b00, 0x72}, {{0x0, 0x0}, {0x1022174ac, 0x6}, 0x40000000, 0xc002764370, 0x800, 0xc002764380})
	pkg/executor/results.go:34 +0x84
github.com/filecoin-project/bacalhau/pkg/executor.WriteJobResults.func1()
	pkg/executor/results.go:113 +0x45
go.ptx.dk/multierrgroup.(*Group).Go.func1()
	go.ptx.dk/multierrgroup@v0.0.2/multierrgroup.go:50 +0x67
created by go.ptx.dk/multierrgroup.(*Group).Go
	go.ptx.dk/multierrgroup@v0.0.2/multierrgroup.go:48 +0x8d
```

Command to reproduce the error:
`bacalhau docker run ubuntu:22.04 --id-only --timeout=10 --wait=false -- sleep 9223372036`

wjam:
Bacalhau was panicking when a Docker job reached the `timeout` value, as the container's logs couldn't be fetched with a cancelled context.

Fixes #1694

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -276,7 +276,7 @@ devstack:
 
 .PHONY: devstack-one
 devstack-one:
-	IGNORE_PORT_FILES=true PREDICTABLE_API_PORT=1 go run . devstack --requester-nodes 0 --compute-nodes 0 --hybrid-nodes 1
+	IGNORE_PID_AND_PORT_FILES=true PREDICTABLE_API_PORT=1 go run . devstack --requester-nodes 0 --compute-nodes 0 --hybrid-nodes 1
 
 .PHONY: devstack-100
 devstack-100:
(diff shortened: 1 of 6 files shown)


wjam:
Blocked by failing canary

PR #1719, opened 2023-01-19
Avoid timeouts so large they become negative

wjam:
Ensure that timeouts are small enough to fit into a `time.Duration`


wjam:
Blocked by failing canary

PR #1746, opened 2023-01-23
Start IPFS node when `ipfs-connect` is not defined

### LINKED ISSUE #1231: Default for Server Setup takes no parameters to join default network
ETA: 2022Q4

### LINKED ISSUE #1745: Change `bacalhau serve` so it can be run without any arguments
> Would be nice if `bacalhau serve` just worked everywhere with no args. Then we could have a simple k3s-like installer (maybe with nicer UX 😉).

wjam:
Start up an internal IPFS node when `ipfs-connect` isn't specified for the `serve` command so that the command can be run as `bacalhau serve`.

Fixes #1231

### DIFF
--- a/cmd/bacalhau/serve.go
+++ b/cmd/bacalhau/serve.go
@@ -1,6 +1,7 @@
 package bacalhau
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


wjam:
#1745 might be a duplicate of #1231?

wjam:
Change tested on a separate node running in AWS without IPFS installed (base Ubuntu 22.04 image)

PR #1755, opened 2023-01-23
canary pipeline integration test

wdbaruni:
A new integration test stage has been added to canary pipeline that verifies that latest commit is compatible with production endpoint before deploying a new version of the canary. This change also enables `check-canary` cricle-ci workflow to test locally against the latest commit instead of against production endpoint.


wdbaruni:
cc: @simonwo

PR #1786, opened 2023-01-27
use an allow list for local directory storage driver

binocarlos:
This removes the idea of a base folder for the local directory storage driver and reverts control to an "allow list" of local directories that is configured via an environment variable.

This allows compute providers to be specific about which local folders they are willing to offer to jobs.


simonwo:
We now have this in a slightly different form! :)

PR #1795, opened 2023-01-29
publish libp2p resource manager metrics

wdbaruni:


wdbaruni:
Resource manager dashboard is ALIVE!
https://protocollabs.grafana.net/goto/aPl18i0Vk?orgId=1

PR #1865, opened 2023-02-02
break circle ci tests into separate workflows

wdbaruni:
our tests are very flaky and takes too long to finish only to realize some of the jobs failed. When you retry, new jobs might fail while previous ones succeed, and the loop continues.

This is not a long term solution, but a mitigation to minimize the jobs that needs to be retried by breaking `dev_branch` workflow into smaller jobs. Each workflow still contains both unit and integration tests to be able to generate a meaningful test coverage report

This change also includes:
1. Isolating lotus node tests into a separate optional workflow as it is the most flaky test
2. Stop pulling unnecessary docker images before running the tests


wdbaruni:
ref https://github.com/filecoin-project/bacalhau/pull/1865

Issue #1607, opened 2023-01-07
Allow changing the format of `bacalhau describe` to JSON or YAML


aronchick:
We have a command `bacalhau describe` that outputs the spec and status of a job. This command currently prints the spec in YAML format.

Our messaging actually serialises to JSON, which is generally regarded as a safer and easier data format to work with. We should support a flag to `describe` like `--json` to output only in JSON format.

KalindaC-muSpace:
Hello, I would like to contribute by solving this issue. Could you provide background describing the case and some directions (it is my first time working on an open-source project)? Thanks!

KalindaC-muSpace:
@aronchick @simonwo Please help :)

simonwo:
Hi @KalindaC-muSpace! 👋 I've added some comments to the description. 

Start by cloning the repository and making sure you can run the code locally by e.g. running `go run . docker run ubuntu date` followed by `go run . describe`. Then have a look in `cmd/bacalhau/describe.go` for how the describe command works. 

You can see that it already accepts some flags, to which you'll need to add a `--json` flag: https://github.com/filecoin-project/bacalhau/blob/72c15eb626c80f5d57447f5519043f76787f99fd/cmd/bacalhau/describe.go#L61-L64

And then in fact it already parses to JSON but converts that to YAML, so when the `--json` flag is set you only need to skip that last step! https://github.com/filecoin-project/bacalhau/blob/72c15eb626c80f5d57447f5519043f76787f99fd/cmd/bacalhau/describe.go#L139-L143

Let me know if you have any questions or concerns :)

KalindaC-muSpace:
This was very helpful, I'll get to work on it right away.

PR #1872, opened 2023-02-03
#1607 Fixed

KalindaC-muSpace:
I have added the `--json` flag to output the description as JSON text with an indent of 4 units. Without the flag, it will automatically output the job description as YAML.


cmd/bacalhau/describe.go
```diff
@@ -131,18 +141,24 @@ func describe(cmd *cobra.Command, cmdArgs []string, OD *DescribeOptions) error {
 		jobDesc.Status.LocalEvents = localEvents
 	}
 
-	b, err := model.JSONMarshalWithMax(jobDesc)
+	b, err := model.JSONMarshalIndentWithMax(jobDesc,3)
```
simonwo, line 144: Your PR message said you're using 4 spaces but this seems to output 3... did you mean to use 3 or 4?

simonwo (commented):
The fix looks great, thanks! Just one minor question about the indent size.

simonwo:
Please can you also check that the output of `make lint` is empty? To ensure the code is styled correctly. You will need to [install `golangci-lint`](https://golangci-lint.run/usage/install/).

cmd/bacalhau/describe.go
KalindaC-muSpace: Yes, I reduced it to 3 after. I thought it looked better

simonwo:
Hi, looks like you still need to run ` go fmt cmd/bacalhau/describe.go `.

PR #1893, opened 2023-02-06
introduce lite ipfs node for downloading results

wdbaruni:
A kubo node configured with low power and limited functionality for the purpose of downloading results from IPFS. In addition to that, the downloader now is reusing the node instead of spinning up a fresh node for each download request

### DIFF
--- a/cmd/bacalhau/utils.go
+++ b/cmd/bacalhau/utils.go
@@ -470,7 +470,7 @@ func downloadResultsHandler(
 		return err
 	}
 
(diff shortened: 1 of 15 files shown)


wdbaruni:
it seems there are better options using https://github.com/ipfs/go-libipfs library
Thread: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1675705603221589?thread_ts=1675682783.855449&cid=C02RLM3JHUY

PR #1917, opened 2023-02-08
Remove 5 minute IPFS timeout

wjam:
As an input coming from IPFS could take a while to download, especially if it is coming from a remote machine, we should remove the 5-minute limit. Removing this limit does not affect how IPFS downloads are cancelled as the original context already has a timeout applied to it.

### DIFF
--- a/pkg/storage/ipfs/storage.go
+++ b/pkg/storage/ipfs/storage.go
@@ -21,8 +21,8 @@ import (
 // a job to run - it will remove the folder/file once complete
 
 type StorageProvider struct {
-	LocalDir   string
-	IPFSClient ipfs.Client
+	localDir   string
+	ipfsClient ipfs.Client
 }
 
 func NewStorage(cm *system.CleanupManager, cl ipfs.Client) (*StorageProvider, error) {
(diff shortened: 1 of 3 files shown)


pkg/storage/ipfs/storage.go
```diff
@@ -40,40 +40,36 @@ func NewStorage(cm *system.CleanupManager, cl ipfs.Client) (*StorageProvider, er
...
-func (dockerIPFS *StorageProvider) PrepareStorage(ctx context.Context, storageSpec model.StorageSpec) (storage.StorageVolume, error) {
+func (s *StorageProvider) PrepareStorage(ctx context.Context, storageSpec model.StorageSpec) (storage.StorageVolume, error) {
 	ctx, span := system.GetTracer().Start(ctx, "storage/ipfs/StorageProvider.PrepareStorage")
 	defer span.End()
 
-	var cancel context.CancelFunc
-	ctx, cancel = context.WithTimeout(ctx, config.GetDownloadCidRequestTimeout(ctx))
-	defer cancel()
-
-	stat, err := dockerIPFS.IPFSClient.Stat(ctx, storageSpec.CID)
```
wdbaruni, line 76: this is where the magic is happening :)

Issue #872, opened 2022-10-11
Support for mounting git repos in the -r flag


js-ts:
I think it will great to have -u flag supporting cloning/downloading .git repo from github and huggingface

So currently if you want mount a model from  https://huggingface.co/openai/clip-vit-large-patch14, you need to clone it locally and then upload it IPFS and then you can mount the CID

```
bacalhau docker run \
jsacex/stable-diffusion \
-v bafybeie3pocunkoigla4q6iyxq32tflaykx6vsspk5iwmppsnbk6bymqbu:/clip-vit-large-patch14-336
-- python demo.py --prompt "gouda being eaten by a spaceman on a horse" --output ../outputs/gouda.png --tokenizer ../clip-vit-large-patch14-336 
```

so I think it would make things lot easier for the users if they could directly mount their .git repos

```
bacalhau docker run \
jsacex/stable-diffusion \
 -u https://huggingface.co/openai/clip-vit-large-patch14-336
-- python demo.py --prompt "gouda being eaten by a spaceman on a horse" --output ../outputs/gouda.png --tokenizer ../inputs/clip-vit-large-patch14-336 
```

Suggestion from luke
> That would be cool, especially if you can refer to a specific commit, like pip can

simonwo:
Sounds like @js-ts os working on this??

js-ts:
yes

PR #1936, opened 2023-02-11
Repo mount #872

js-ts:
`bacalhau docker run -r https://github.com/filecoin-project/bacalhau.git ubuntu ls /inputs`
or schedule only to nodes that have git-lfs installed
`bacalhau docker run -i gitlfs://github.com/filecoin-project/bacalhau.git ubuntu ls /inputs`

Repo will be mounted at /inputs/filecoin-project/bacalhau

Since the approach of cloning a git repo every time a job is submitted is not scalable 
future mounts will utilize IPFS. We query the SHA hash of the repository against a key-value store hosted at kv.bacalhau.org, which stores a key-value pair where the key is the SHA1 hash and the value is a CID. If the CID exists in the store, we mount the CID, otherwise, we clone the repository and add its value to the key-value store.

The KV store has two endpoints (for only read and write no update/delete)
get kv.bacalhau.org/<SHA1> returns a {"CID":"<CID>"}
post kv.bacalhau.org is where you can add in a key value (SHA1:CID) pair


js-ts:
We need to install git-lfs on every node for this to work

aronchick:
I thought we got rid of git-lfs? This will never work if that is required. Never.

js-ts:
Git-lfs is required for cloning huggingface repos

aronchick:
The please rewrite it without gitlfs. Or use a go based one so it doesn't require installing it on the node.

Or failing that, write very nice error messages that detect if the repo is already mounted and if not try to schedule the pull to an LFS enabled node and if nothing is available, please warn the user this job will never succeed.

We cannot be dictating what is required to run bacalhau.

js-ts:
It can work without git-lfs eg github, gitlab etc. but it wouldn't clone git-lfs repos eg which hugging face uses

pkg/model/v1alpha1/storage_source.go
```diff
@@ -15,6 +15,7 @@ type StorageSourceType int
 const (
 	storageSourceUnknown StorageSourceType = iota // must be first
 	StorageSourceIPFS
+	StorageSourceRepoClone
```
wdbaruni, line 18: we should not update previous versions of our schema. They must remain the same to maintain backward compatibility with jobs created with older schema

pkg/clone/clone.go
```diff
@@ -0,0 +1,152 @@
...
+		return "", fmt.Errorf("failed to execute script: %s", err)
+	}
+	fmt.Println()
+	// Print the output
+	fmt.Printf("%s\n", output)
+	return "", nil
+}
+
+func (cl *Clone) IfNotInstalledInstallingGitlfs() {
+	_, err := filepath.Abs("./install-lfs.sh")
```
wdbaruni, line 76: where are you reading these scripts from? Are we trying to install dependencies at runtime on behalf of the user?

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,331 @@
...
+		key = envkey
+	} else if envkey != "" && spkey != "" {
+		key = spkey
+	} else {
+		key = ""
+	}
+	if key != "" {
+		log1.Println("Pinning to Estuary...")
+		//nolint:govet,noctx
+		err := estuary.PinToIPFSViaEstuary(ctx, key, CID)
```
wdbaruni, line 144: why are we cloning the repo to IPFS? I thought we just needed to mount the repo locally as a job input. Maybe adding some description to the PR will help me understand the feature more

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,331 @@
...
+	if u.Scheme != "https" && u.Scheme != "http" && u.Scheme != "ssh" {
+		return nil, fmt.Errorf("URL must use HTTPS, HTTP, or SSH scheme")
+	}
+	if !strings.HasSuffix(u.Path, ".git") {
+		return nil, fmt.Errorf("URL must use .git file extension")
+	}
+	return u, nil
+}
+
+// func runShellScript(scriptPath string, args []string) (string, error) {
```
wdbaruni, line 258: commented out code

wdbaruni:
I agree with Dave's point. Since gitlfs is written in go, can we look into the part of the code where they download pointers into actual files and reuse that? If that is too complicated, then we should have the compute nodes start advertising the supported storage engines as part of their `NodeInfo` message to allow the requester to route jobs to the only the nodes that have the library installed, or fail fast.

We should also start looking into defining the inputs as URIs (`ipfs://, http://, git://, s3://...`) instead of different input arguments, but that can be a separate change

pkg/clone/clone.go
```diff
@@ -0,0 +1,152 @@
...
+		last := scriptArgs[len(scriptArgs)-1]
+		path, err := ioutil.TempDir("", path.Base(last))
+		scriptArgs[len(scriptArgs)-1] = path
+		if err != nil {
+			fmt.Println(err)
+		}
+		return path, nil
+	}
+
+	x := fmt.Sprintf("/bin/bash %v", scriptArgs)
```
simonwo, line 50: We can't assume Bash will be installed on all compute nodes.

pkg/job/factory.go
```diff
@@ -75,7 +78,22 @@ func ConstructDockerJob( //nolint:funlen
...
+		// }
+		if repoCID != "" {
+			inputRepos = clone.RemoveFromSlice(inputRepos, url)
+			repoCIDPATH := repoCID + ":/inputs"
+
+			SHAtoCID := []string{}
+			SHAtoCID = append(SHAtoCID, repoCIDPATH)
+			inputVolumes = append(inputVolumes, SHAtoCID...)
+		}
+	}
```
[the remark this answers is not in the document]
simonwo: Also, this is only for Docker jobs. Why not WASM jobs or other types?

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func RemoveFromSlice(arr []string, item string) []string {
+	newArr := []string{}
+	for _, s := range arr {
+		if s != item {
+			newArr = append(newArr, s)
+		}
+	}
+	return newArr
```
simonwo, line 322: Also duplicate of clone.go?

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,331 @@
...
+	return storageHandler, nil
+}
+
+func (sp *StorageProvider) IsInstalled(context.Context) (bool, error) {
+	err := checkGitLFS()
+	if err != nil {
+		return false, err
+	} else {
+		return true, nil
+	}
```
simonwo, line 75:
```suggestion
	return err == nil, err
```

simonwo (changes requested):
Thanks for this work and all the effort you have put in so far!

I am seeing a useful feature here – one that I have needed! But I think you are conflating two separate features, and should focus on the first one before trying to deliver the second.

The first feature is git repository mounting. I do think this is a useful feature and it makes easier a number of use cases, partiuclarly around distributed CI which plenty of people are mentioning. I think you are along the right lines in implementing a StorageProvider to handle that, and having a new StorageSourceType to cover a repository.

However the provider seems to be referring to IPFS and/or Estuary as well, e.g. the `Upload` method just uploads to IPFS. The StorageProvider should have [one single responsibility](https://en.wikipedia.org/wiki/Single-responsibility_principle) and this is cloning and mounting a git repo (and maybe pushing back to a git repo, if we think anyone wants to do that, but it's hard to do without authentication). Anything outside of that responsibility should sit elsewhere.

I think the second feature you are trying to implement is caching of git repositories on IPFS. This is certainly reasonable but it can perhaps come later – there is no major problem with having compute nodes download git repositories, just like they would download URLs. So I think you should focus on the first feature and come back to this. 

When you come back to it, it should sit outside of the storage provider and look more like a job transformation, e.g. we have something [here which just moves inline specs to IPFS](https://github.com/filecoin-project/bacalhau/blob/main/pkg/requester/jobtransform/inline_spec_storage.go); that area would be the correct place to implement something similar for git repositories.

What is this `kv.bacalhau.org`? Where has it come from? Where is the code for it?

More general comments:
- Don't comment out gosec lints unless you are aboslutely positive they are not a problem. If so, leave a descriptive comment about why. Currently, there are some serious security bugs that are being hidden.
- Running shell scripts really shouldn't not be necessary. It's bad because it introduces a ton of security issues. Are we making appropriate use of the go git libraries? What do we need shell scripts for?
- Anywhere that does an external network request (e.g. git, or HTTP) should take a context. We need those things to be cancellable if they are taking too long or if the node is shutting down.
- The functions are also very noisy, and are doing a lot of printing to Stdout – you should avoid `fmt.Println` and use `log.Ctx(ctx).Debug()` or `log.Ctx(ctx).Error()` for that sort of thing. Make sure you are using `zerolog` and not the standard library `log` package.

pkg/storage/repo/storage.go
js-ts: Sorry I didn't add much description here but the point with saving it to IPFS is that subsequent mounts would be done from IPFS instead of cloning the repo every time a person wants to mount a git repo


There is a key value (Key=SHA1hash, Value=CID) store hosted at kv.bacalhau.org where we make queries against the SHA hash of a git repo and check it if the CID exists in the store if it does we mount the CID instead of the repo, if the CID isn't there we clone the repo and post its's value in the kv store

pkg/storage/repo/storage.go
js-ts: I will remove that

pkg/clone/clone.go
js-ts: That's unused code my linter didn't catch it I think
I had that previously but since sudo didn't work with go 

Will remove that

pkg/model/v1alpha1/storage_source.go
js-ts: Ok will remove that

js-ts:
> then we should have the compute nodes start advertising the supported storage engines as part of their NodeInfo message to allow the requester to route jobs to the only the nodes that have the library installed, or fail fast.


@wdbaruni That's a good idea I will implement it and thanks for elaborating on @aronchick point it's clear now

> failing that, write very nice error messages that detect if the repo is already mounted and if not try to schedule the pull to an LFS enabled node and if nothing is available, please warn the user this job will never succeed.

js-ts:
> We should also start looking into defining the inputs as URIs (ipfs://, http://, git://, s3://...) instead of different input arguments, but that can be a separate change

💯% agree there are a lot of flags now we could use a single flag '-m' or anything else

pkg/clone/clone.go
```diff
@@ -0,0 +1,152 @@
...
+	args := ScriptStruct{
+		path: "pkg/clone/install-lfs.sh", arguments: []string{}}
+
+	if _, err := cl.RunShellScript(args.path, args.arguments); err != nil {
+		fmt.Println(err)
+	}
+}
+
+func (cl *Clone) CloneRepo(repoURL *url.URL, Path string) (string, error) {
+	_, err := filepath.Abs("./clone.sh")
```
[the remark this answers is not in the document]
js-ts: @simonwo I tried using go-git it doesn't support git-lfs or cloning from sites other than github like huggingfacehub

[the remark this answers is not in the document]
js-ts: > The functions are also very noisy, and are doing a lot of printing to Stdout – you should avoid fmt.Println and use log.Ctx(ctx).Debug() or log.Ctx(ctx).Error() for that sort of thing. Make sure you are using zerolog and not the standard library log package.

👍

pkg/job/factory.go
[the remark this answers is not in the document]
js-ts: I will add those

pkg/clone/clone.go
```diff
@@ -0,0 +1,152 @@
...
+
+func RepoExistsOnIPFSGivenURL(urlStr string) (string, error) {
+	cmd := exec.Command("git", "ls-remote", urlStr)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		fmt.Println(err)
+		return "", err
+	}
+	x := "http://kv.bacalhau.org/" + fmt.Sprintf("%v", string(output)[:40])
+	//nolint:gosec,noctx
```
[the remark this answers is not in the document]
js-ts: > We should also start looking into defining the inputs as URIs (ipfs://, http://, git://, s3://...) instead of different input arguments, but that can be a separate change
I think just having one flag is better as Walid stated but that's a separate thing

pkg/model/storage_spec.go
```diff
@@ -20,6 +20,9 @@ type StorageSpec struct {
 	// Source URL of the data
 	URL string `json:"URL,omitempty"`
 
+	// URL of the git Repo to clone
+	Repo string `json:"Repo,omitempty"`
+
```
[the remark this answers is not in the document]
js-ts: > We should also start looking into defining the inputs as URIs (ipfs://, http://, git://, s3://...) instead of different input arguments, but that can be a separate change
I think just having one flag is better as Walid stated but that's a separate thing

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,331 @@
...
+	// Check if the URL is a Git repository URL
+	if u.Scheme != "https" && u.Scheme != "http" && u.Scheme != "ssh" {
+		return nil, fmt.Errorf("URL must use HTTPS, HTTP, or SSH scheme")
+	}
+	if !strings.HasSuffix(u.Path, ".git") {
+		return nil, fmt.Errorf("URL must use .git file extension")
+	}
+	return u, nil
+}
```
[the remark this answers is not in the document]
js-ts: Sorry My linter didn't catch it I will delete it

pkg/clone/clone.go
js-ts:
```
package main

import (
	"fmt"
	"os/exec"
	"runtime"
	"strings"
)

func main() {
	var cmd *exec.Cmd
	if runtime.GOOS == "windows" {
		cmd = exec.Command("cmd", "/C", "echo %COMSPEC%")
	} else {
		cmd = exec.Command("sh", "-c", "echo $SHELL")
	}
	out, err := cmd.Output()
	if err != nil {
		fmt.Println("Error:", err)
		return
	}
	shell := strings.TrimSpace(string(out))
	fmt.Println("Current shell:", shell)
}

```

will this work ?

pkg/job/factory.go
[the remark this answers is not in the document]
js-ts: > Factory methods like this should be fast, but RepoExistsOnIPFSGivenURL is making an HTTP request and is therefore slow. You should implement the cache on the requestor node side as a [jobtransform](https://github.com/filecoin-project/bacalhau/tree/main/pkg/requester/jobtransform).

@simonwo I'm not sure about how it should be done can you please help in figuring that out?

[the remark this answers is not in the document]
js-ts: > Also, this is only for Docker jobs. Why not WASM jobs or other types?
I want to test how well does this work in production for docker if it works well then add for other executors too

js-ts:
@simonwo I applied all the changes you wanted 
except for this one https://github.com/bacalhau-project/bacalhau/pull/1936#discussion_r1108453239

js-ts:
Merging this by today itself, can't wait this long for it to be reviewed

Issue #1626, opened 2023-01-10
M3: Retry with other nodes if execution fails


wdbaruni:
ETA: 2023Q2

Currently once bids were accepted, we don't attempt to find or retry with other compute nodes if execution fails. Improving the detection of failed nodes will be a pre-requirement to avoid retrying bad jobs #1627. I would also avoid retrying indefinitely until the job succeed to mitigate situations where a malicious job found a hole and is the one taking the nodes down

Issue #1100, opened 2022-11-13
M3: Requester node detection of stale jobs and update their state


wdbaruni:
A background job in requester node to detect jobs that are not making progress or have been stuck for a while. Will require leader election if the requester is made of a fleet of nodes

ETA: 2023Q2

PR #1937, opened 2023-02-11
new job store with fsm support, and explicit job state

### LINKED ISSUE #1624: M2: Requester FSM Based on JobStore
ETA: 2023Q1

Requester node currently uses an in-memory state machine to track the lifecycle of jobs and issue requests/updates to compute nodes. The problem is the requester node loses all this state when the host restarts. We already switched to a persisted store for JobStore (LocalDB->JobStore), but we are not using that for job lifecycle management.

The recommendation is to discard the in-memory state machine, and just rely on JobStore state to track the lifecycle of jobs, similar to how compute FSM was replaced in favour of ExecutionStore.

Once this is done, it will be easy to build background jobs to track stale/stuck jobs after a host was restarted, which is tracked in #1100

### LINKED ISSUE #1647: M3: Fail fast when possible instead of waiting for `--timeout` to expire
ETA: 2023Q2

As it stands today, if a compute node fails the job execution, the requester node doesn't mark the job as failed (or retry) right away. Instead it waits for the `timeout` window to expire which is not the best user experience and can take hours to report a failure that was spotted in the first few minutes

wdbaruni:
A major change in `JobStore` (aka `localdb`) state management where the state is now updated directly by the requester node instead of incrementally by consuming `JobEvents`. This enabled conditional updates and stronger state consistency, which allowed early failure when bids are rejected or execute fails instead of having to wait for timeout window, and will simplify the next work of retrying execution with other nodes.

`localdb` or the ability to build the state from events will continue to exist as it is needed by the dashboard, and will be needed by Octostore as well. A follow up task will be to merge both states instead of the forking them.

## New Features
1. Jobs and shards now have an explicit state that marks if the job is in progress, failed or completed successfully. Previously the state is _inferred_ through the state of nodes that are executing the jobs, which adds complexity to clients and to users
2. Replace `shard_fsm` with using `JobStore` conditional updates, which allows having a single source of truth for job state, and enables Requester Node to recover and manage in-progress jobs after reboots
3. A new Housekeeping background job that tracks stale/stuck jobs. Currently only fails jobs if timeout window is breached, but can be extended in the future.
4. Requester node can fail early as soon as bids are rejected instead of having to wait for negotiation timeout to expire
5. Changes in `JobState` schema to have better view of shard state, and become Job -> Shards -> Executions (Nodes), instead of Job -> Nodes -> Shards
6. New `JobHistory` type that is built from mutations to the job state. Similar to JobEvent, but built by the `JobStore` itself

## Issues/Merge Blockers
1. The new Job schema is not compatible with the Postgres schema that the dashboard is using. A schema converter needs to be implemented and tested, which I couldn't find the time to include in this PR. I am also hoping to revisit the Job schema more and move fields that don't uniquely identify the job spec outside of the spec, such as resource usage and timeout. A dirty solution I did for now is I defined the dashboard as a separate module the depends on an older version of Bacalhau, though I might've broken some stuff. What are your thoughts @binocarlos?
2. Requester nodes no longer expose JobEvents through a websocket. The Dashboard should consume JobEvents directly from GossipSub, which the Requester Nodes are still publishing to.

## Next Actions:

- #1938 
- #1939
- #1940
- #1941
- #1942

Closes #1647
Closes #1624
Enables #1100
Enables #1626


wdbaruni:
I've decided to merge into `dev-job-store` branch, and will merge into `main` once the dashboard issues are fixed

Issue #1925, opened 2023-02-09
Improve the quality of distributed traces


wjam:
The traces currently produced by Bacalhau suffers from a number of problems:
- [x] Not all spans associated with running a job have useful tags on them, such as `jobid` or `nodeid`
- [x] Some parts of the code base don't seem to be associated with a span at all, such as the code that calls an executor
- [x] Span names aren't consistent - some refer to the code being called, some refer to the event happening, and some include information from the tags. E.g. `pkg/pubsub/libp2p.Publish`, `service/JobLifecycle`, or `service/JobLifecycle-QmWg7m5G`
- [ ] ~Traces for executors or publishers are disconnected from everything - e.g. one trace for `pkg/executor/docker.RunShard` and a completely separate one for `pkg/publisher/ipfs.PublishShardResult`.~ moved to #1991
- [ ] In a traditional synchronous environment, a single trace would cover all of the work performed for a single request. It is less clear for an asynchronous environment, as tracing is newer here, but work can either be tracked as a single trace or multiple traces linked together. Linking multiple traces does run into problems when trying to find child traces from the parent one. Neither of these are done in Bacalhau - there are multiple traces and they're not all linked together.
- [ ] All traces are under a single service - `bacalhau`. There should probably be multiple services to help identify work pass between areas of the system - such as `requester`, `compute` & `devstack`?

Also:
- [x] Consider using https://github.com/XSAM/otelsql for auto-magically tracing SQL queries (does it capture SQL queries and are we correctly using parameterised queries?)
- [x] The Docker executor is currently a single massive block - we should have multiple spans to split this up (pulling, creating, running?)
- [ ] ~Trace running WASM~ moved to #1990
- [x] Trace running IPFS remotely rather than in-process

wjam:
Closing now as _most_ of the work identified has been done. Remaining work has been split out into other tickets

PR #1957, opened 2023-02-14
Improve the naming of the distributed traces

wjam:
This changes the tracing so that:
* Node & job IDs are attributes of as many spans as possible
* Names of non-HTTP spans are standardised on `sub/package.Struct.Method`, which is similar to what Go can auto-generate
* HTTP server spans now follow the recommended naming pattern of `{http.method} {http.route}`
* Publisher and storage services have tracing handled by automatically by be wrapped by a tracing publisher/storage service
* Ensure job ID is present on the critical areas of a job, so that searching for `jobid=...` brings up the submission, running and publishing spans.

Part of #1925

### DIFF
--- a/cmd/bacalhau/create.go
+++ b/cmd/bacalhau/create.go
@@ -87,7 +87,7 @@ func create(cmd *cobra.Command, cmdArgs []string, OC *CreateOptions) error { //n
 	defer cm.Cleanup()
 	ctx := cmd.Context()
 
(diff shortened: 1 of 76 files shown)


wjam:
OTEL HTTP trace naming - https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#name

You can get Go to produce a similar style using `reflect.ValueOf().Type().Name()` function.

PR #1958, opened 2023-02-14
Shares devstack environment vars via a dotenv file

rossjones:
Each time a developer runs the bacalhau devstack, they are presented with a few environment variables that they need to set in another window to configure the bacalhau command to connect to the local devstack.  It is possible to keep a persistent port, but the IPFS Swarm addresses do change.  If work there is a need to restart the devstack, the details have to be set again.

Bacalhau already has support for loading `.env` files, and this PR extends the functionality to allow for it to load a .env file created by the devstack.

When the devstack is launched, if it is able to find a value for the environment variable `DEVSTACK_ENV_FILE` then it will write the required environment files into the file described within.  This file will be cleaned up when the devstack is shut down.

If `DEVSTACK_ENV_FILE` is present when running a normal bacalhau command, it will use the details found within.  This means that if the developer were to define something like `DEVSTACK_ENV_FILE=.devstack.env` then there is no further need to keep copying and pasting environment variables between terminals.

The filename `.devstack.env` is recommended in the running_locally documentation as this is now ignored by git to avoid accidentally committing it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ dist/*.sha256
 dist/*.tar.gz
 env.*
 .env
+.devstack.env
 *.swp
 *.swo
 bin/*/*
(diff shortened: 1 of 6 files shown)


aronchick:
Looks great!

PR #2071, opened 2023-02-27
Ensure long running jobs have own context

rossjones:
The scheduler has several notify* functions that spawn a new goroutine before returning on the current thread of execution - currently each has a context created from a parent that is a HTTP request, so all of the goroutines created from that context are closed when the request is finished.

To avoid this, each of the notify* functions that create a new goroutine, do so with a DetachedContext which although it provides access to the parent's values, it does not response to the parents cancel() or timeout.

Although this will stop goroutines being killed when the request has finished, it does mean that the original context is kept around for lifetime of the new context, resulting in more garbage.


simonwo:
What happens if the node is shutdown during a long-running HTTP request?

rossjones:
> What happens if the node is shutdown during a long-running HTTP request?

If the node is shutdown via a signal (SIGINT or SIGTERM) then the notify* goroutines in the scheduler state machine will not be cancelled cleanly, they'll just stop executing when the process exits. Similarly if the `cleanup` function is called, although that doesn't explicitly do anything with the context it is passed other than pass it to other components to do their own cleanup. 

If the node is shutdown via SIGKILL then no cancellations will happen from any contexts, because, well, it's SIGKILL.

Issue #1780, opened 2023-01-26
Output job metrics in JSON via HTTP


simonwo:
Station wants to track our job metrics via a JSON/HTTP API:

> A module exposes its stats via HTTP(S), on an address communicated via stdout (see above API:). The response will be a JSON object with at least the following fields:
>
>     {
>       "jobsCompleted": 1234
>     }
>
> The number of jobs completed is expected to be a monotonically increasing number. It's the responsibility of the module to persist this number across process restarts, somewhere in ROOT_DIR.
>
> The module may include additional fields in the JSON response, although these will be ignored by Station.

We already track our job numbers via Prometheus: https://github.com/filecoin-project/bacalhau/blob/acd06e32588450f04cef465c7f459a86a3c644da/pkg/compute/metrics.go#L26-L32

So the tasks are to:

- [ ] Understand if the Prometheus metrics are persisted and acessible
- [ ] Or otherwise find a way to record those numbers whilst also making them available to Prometheus
- [ ] Serve an HTTP API containing the associated metrics in addition to and/or unified with our current metrics API
- [ ] Write the metrics API endpoint to Stdout when booting up in "Station mode"

PR #2076, opened 2023-02-27
Add JobsCompleted metric to API for tracking completed executions on a

olgibbons:
single compute node.

A job is defined, by this metric, as an 'execution' and any completed executions will be counted for the compute node, whilst it is running. This count is not currently persisted across sessions, but should be implemented in the future via a database.


pkg/compute/sensors/completed_jobs.go
```diff
@@ -0,0 +1,29 @@
...
+
+func NewCompletedJobs(e store.ExecutionStore) *CompletedJobs {
+	return &CompletedJobs{ExecutionStore: e}
+}
+
+// GetDebugInfo implements model.DebugInfoProvider
+func (c *CompletedJobs) GetDebugInfo(ctx context.Context) (model.DebugInfo, error) {
+	jobcounts := c.ExecutionStore.GetExecutionCount(ctx)
+	return model.DebugInfo{
+		Component: "completed jobs:",
```
simonwo, line 22: For #1780, this should be `"jobsCompleted"`.

pkg/compute/sensors/completed_jobs.go
```diff
@@ -0,0 +1,29 @@
+package sensors
+
+import (
+	"context"
+
+	"github.com/filecoin-project/bacalhau/pkg/compute/store"
+	"github.com/filecoin-project/bacalhau/pkg/model"
+)
+
+type CompletedJobs struct {
```
simonwo, line 10: Naming: this should probably be `CompletedJobProvider` to match the naming scheme of things being "debug info providers".

pkg/compute/store/types.go
```diff
@@ -91,4 +91,7 @@ type ExecutionStore interface {
 	UpdateExecutionState(ctx context.Context, request UpdateExecutionStateRequest) error
 	// DeleteExecution deletes an execution
 	DeleteExecution(ctx context.Context, id string) error
+	// GetExecutionCount returns a count of all executions that completed
+	// successfully on this compute node
+	GetExecutionCount(ctx context.Context) int
```
simonwo, line 96: You could argue that this should return a `uint` to more accurately describe the count – it shouldn't go below zero, right?

pkg/compute/sensors/completed_jobs.go
```diff
@@ -0,0 +1,29 @@
+package sensors
+
+import (
+	"context"
+
+	"github.com/filecoin-project/bacalhau/pkg/compute/store"
+	"github.com/filecoin-project/bacalhau/pkg/model"
```
simonwo, line 7: Your build errors are because these need to be `bacalhau-project/bacalhau` now.

PR #2130, opened 2023-03-03
Reconnect devstack peers to avoid flakes

wjam:
Improve test reliability by getting nodes within a devstack to reconnect to each other rather than assuming they connect and stay connected. Also reduce the dial timeout time to only a second rather than the default of 60 seconds.


wjam:
See run attempts - https://app.circleci.com/pipelines/github/bacalhau-project/bacalhau/4406

PR #2151, opened 2023-03-09
Fix dashboard password

binocarlos:
Restore the .htaccess setup for the dashboard so we have a password set for https://dashboard.bacalhau.org/


frrist:
@binocarlos is this fix still needed? if so can you rebase and merge, otherwise we can probably close it.

Issue #1971, opened 2023-02-16
Defining storage specs as URIs (ipfs://, http://, git://, s3://...) instead of different arguments


wdbaruni:

js-ts:
@wdbaruni @simonwo I think it will make things much simpler for the users
since most of the other shorthand flags are already being used '-m' seems the best option


| Storage type      | command |
| ----------- | ----------- |
| ipfs | bacalhau docker run -m ipfs:// |
| url | bacalhau docker run -m  http || https://    |
|   inline | bacalhau docker run -m    file://   |
|   repo | bacalhau docker run -m   git://    |

js-ts:
If I'm not wrong the plan is to deprecate all the flags

wdbaruni:
Thanks for picking this up @js-ts. I don't think we need a new flag for this. The goal is to have existing `--inputs` and `--outputs` accept URIs, which should allow us to deprecate `--input-urls`, and I am not sure why we have a separate `--input-volumes`. Also I am not sure I understand your last comment about deprecating all flags

js-ts:
> -outputs accept URIs
how could output volumes accept URIs ?

 > --inputs accept URIs
I wanted to keep all the existing flags working so decided on a new flag

> I am not sure I understand your last comment about deprecating all flags
ok I thought we wanted to deprecate those flags and them with a single flag

PR #2155, opened 2023-03-09
Uri storage spec

js-ts:
Instead of having separate flags for each storage type we use a single flag

Since currently we only support mounting CIDs, URLs  (very easy to add other types also) this PR for now  will just support these two and will add this to more executors if the way this is implemented gets approved

example commands

Mount CID
```
bacalhau docker run -i ipfs://Qmc5gCcjYypU7y28oCALwfSvxCBskLuPKWpK4qpterKC7z ubuntu ls /inputs/
```
Mount URL

```
bacalhau docker run -i https://gateway.ipfs.io/ipfs/Qmc5gCcjYypU7y28oCALwfSvxCBskLuPKWpK4qpterKC7z ubuntu ls /inputs/
```
..others will be added later

### DIFF
--- a/cmd/bacalhau/docker_run.go
+++ b/cmd/bacalhau/docker_run.go
@@ -3,6 +3,8 @@ package bacalhau
 import (
 	"context"
 	"fmt"
+	"log"
+	"net/url"
 	"strings"
 
 	"github.com/bacalhau-project/bacalhau/pkg/bacerrors"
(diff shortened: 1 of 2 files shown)


js-ts:
https://github.com/bacalhau-project/bacalhau/issues/1971

js-ts:
Get this merged before merging https://github.com/bacalhau-project/bacalhau/pull/1936

js-ts:
@lukemarsden I think that it wouldn't break -v

Issue #448, opened 2022-08-06
DX Improvement: implement log streaming (--follow, much bigger project)


js-ts:
ETA: 2023Q2

step 2 of  https://github.com/filecoin-project/bacalhau/issues/314

Is this very useful for now ?

aronchick:
this is interesting - and useful! but why wouldn't this just be an overload of --local?

lukemarsden:
i don't see how streaming logs from a running job has anything to do with --local

anyway, here's one way to implement it:
* include node addresses in the job metadata for a running job
* CLI can make a websocket connection or libp2p bitswap connection directly to one or more of the servers that are running a job
* when the connection is made initially, the backend streams the contents of `docker logs -f <containerid>` over the websocket/bitswap stream

aronchick:
Yeah I think I put the local comment on the wrong issue. Deleting

lukemarsden:
this is a big project, we should write a design doc for it - also, it's not particularly high priority

wesfloyd:
Sharing an update that this feature will also be helpful for our partners LabDAO.

Goal: the request is for a way to pipe the stdout/stderr logs to a logging service or the user as the job is running. 

Benefit: to improve user feedback for longer running jobs. The attached image shows an example for a molecular docking tool that fills in a progress bar as the program runs. A little more helpful than the generic job running that bacalhau shows.
![image](https://user-images.githubusercontent.com/260568/219493706-ad76000a-59fc-4643-94b0-65c1f7de946b.png)

rossjones:
First pass of logstreams is complete, and merged. There are a couple more tickets to improve what is there, but currently you can 

```
$ bacalhau docker run -f myimage
# or 
$ bacalhau logs <jobid>
$ bacalhau logs -f <jobid>
```

PR #2159, opened 2023-03-10
Implement a way to get output from a running job

rossjones:
As we require a mechanism to capture and transport an execution's stdout and stderr, this commit introduces a new method to the compute Executor interface for executors to implement.

Initially we support only docker, with support for wasm coming at a future time.  In addition, we re-use the framing that docker uses when multiplexing stdout and stderr across a single stream, something we can re-use for wasm and transport across the network to a requester node. The format of the data frame is documented in
pkg/util/logstream/dataframe.go.

### DIFF
--- a/pkg/docker/docker.go
+++ b/pkg/docker/docker.go
@@ -104,18 +104,35 @@ func (c *Client) RemoveObjectsWithLabel(ctx context.Context, labelName, labelVal
 	return multierr.Combine(containerErr, networkErr)
 }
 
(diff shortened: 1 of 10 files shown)


wesfloyd:
Would this PR close out issue #448 ? (I'm thinking yes, but curious to hear your take)

lukemarsden:
@wesfloyd it's the start of what will do that, but not the whole thing

lukemarsden (approved):

PR #2161, opened 2023-03-13
Stream docker logs to the CLI

rossjones:
Adds a new cli command, `logs` and an `--follow/-f` option to `docker run` - both of which will show users the execution logs from a currently running job. This currently only supports docker based jobs and will error for those that run wasm.

This is achieved by the cli `logs` function creating a websocket connection to the requester node.
Once the requester node has received the request, and determined that it is valid, it will 
call the relevant compute node and ask it for an address where it can connect for a log stream,
without using the usual request/response pattern of bprotocol. 

The address provided by the compute node is intended to allow for the use of a different connection (not network.Stream) than is used for the connection for bprotocol.  Currently this will use the same connection and multiplex a stream (at the libp2p layer) - this may eventually limit the number of logs that can be streamed between any given requester/compute pair, although most of the heavy lifting on the compute node is handled by docker and `io.Copy`.  

Once the address has been retrieved, the requester node will connect to that address using the new 
protocol ID `/bacalhau/compute/logs/1.0.0`. Once connected it will continue to receive dataframes from the compute node, translating them to small structures for the websocket. This will continue until either the client disconnects or the log stream ends.

``` shell
# View the logs from now, in real-time, from job 12345
$ bacalhau logs 12345

# View the logs from job 12345, including those we may have missed
$ bacalhau logs -a 12345

# Run the docker image specified, and then show the logs as soon as 
# they are available
$ bacalhau docker run image --follow 
```

## Lifetime of docker logs edge-case

When docker writes logs, by default it writes them into a single JSONL file in a format that looks like the following (but with newlines stripped).

```json
{
    "log": "\u001b[32mThe Project Gutenberg EBook of The Cosmic Computer, by Henry Beam Piper\n",
    "stream": "stdout",
    "time": "2023-03-17T11:37:08.764363458Z"
}
```
This is stored in a folder named after the currently running container.  When that container stops, it is cleaned up and the logs disappear.  This will happen if there is an error when the docker container runs (e.g. the contained program crashes).  To handle this edge-case, if a request is sent for logs for a job where the container has terminated, we will instead return the stdout/stderr from the job state and stream it as normal.

In future we might consider supporting other docker logging drivers for more flexibility in serving logs (e.g. using syslog/fluentd etc).


lukemarsden:
Is it worth changing the logs command to be consistent with both docker and kubectl whereby logs alone fetches all logs until now and logs -f does that and attaches to the stream, blocking when the execution completes?

rossjones:
> Is it worth changing the logs command to be consistent with both docker and kubectl whereby logs alone fetches all logs until now and logs -f does that and attaches to the stream, blocking when the execution completes?

Probably, it'll probably be quite glaring if we do the opposite.  I'll get that done Weds morning and then aim for the release after lunch.

PR #2216, opened 2023-03-21
Repo mount

js-ts:
`bacalhau docker run -r https://github.com/filecoin-project/bacalhau.git ubuntu ls /inputs`

mounting by using the git URI
`bacalhau docker run -i git://github.com/filecoin-project/bacalhau.git ubuntu ls /inputs`

or schedule only to nodes that have git-lfs installed
`bacalhau docker run -i gitlfs://huggingface.co/bigscience/bloomz-560m.git ubuntu ls /inputs`

Repo will be mounted at /inputs/filecoin-project/bacalhau

Since the approach of cloning a git repo every time a job is submitted is not scalable 
future mounts will utilize IPFS. We query the SHA hash of the repository against a key-value store hosted at kv.bacalhau.org, which stores a key-value pair where the key is the SHA1 hash and the value is a CID. If the CID exists in the store, we mount the CID, otherwise, we clone the repository and add its value to the key-value store.

The KV store has two endpoints (for only read and write no update/delete)
get kv.bacalhau.org/<SHA1> returns a {"CID":"<CID>"}
post kv.bacalhau.org is where you can add in a key value (SHA1:CID) pair


cmd/bacalhau/docker_run.go
```diff
@@ -360,15 +370,16 @@ func CreateJob(ctx context.Context, cmdArgs []string, odr *DockerRunOptions) (*m
 				odr.InputUrls = append(odr.InputUrls, i)
 			case "ipfs":
 				odr.InputVolumes = append(odr.InputVolumes, fmt.Sprintf("%s:/inputs", scheme.Host))
+			case "git":
+				odr.InputUrls = append(odr.InputRepos, fmt.Sprintf("https://%s", scheme.Host+scheme.Path))
```
simonwo, line 374: What about git repositories that are accessed only over insecure HTTP, or via SSH?

pkg/clone/clone.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func RepoExistsOnIPFSGivenURL(urlStr string) (string, error) {
+	output, err := GetLatestCommitHash(urlStr)
+	if err != nil {
+		fmt.Println(err)
+		return "", err
+	}
+	url := baseURL + output
+	ctx := context.Background()
```
simonwo, line 49: This func needs to take a context.Context rather than using Background so that it can do proper logging and be cancelled.

pkg/clone/clone.go
```diff
@@ -0,0 +1,127 @@
...
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		fmt.Println("Error creating request:", err)
```
simonwo, line 52: This should use `log.Err(err).Error()` instead of printing directly.

pkg/job/factory.go
```diff
@@ -41,7 +43,22 @@ func ConstructDockerJob( //nolint:funlen
 		GPU:    gpu,
 	}
 
-	jobInputs, err := buildJobInputs(inputVolumes, inputUrls)
+	for _, url := range inputRepos {
+		repoCID, _ := clone.RepoExistsOnIPFSGivenURL(url)
+		// if err != nil {
+		// 	fmt.Print(err)
+		// }
```
simonwo, line 50: You need to do this in a function that is expected to be long-running.

As per the past comment, implement a [job transform](https://github.com/bacalhau-project/bacalhau/tree/main/pkg/requester/jobtransform) to turn the repo into an IPFS CID if necessary. 

Also, the error should not be ignored.

pkg/model/storage_spec.go
```diff
@@ -20,6 +20,9 @@ type StorageSpec struct {
 	// Source URL of the data
 	URL string `json:"URL,omitempty"`
 
+	// URL of the git Repo to clone
+	Repo string `json:"Repo,omitempty"`
+
```
simonwo, line 25: Why not use the URL field for this?

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	repoURL := storageSpec.Repo
+	var err error
+	_, err = clone.IsValidGitRepoURL(repoURL)
+	// fmt.Printf("%+v", storageSpec)
+	if err != nil {
+		return storage.StorageVolume{}, err
+	}
+	// # create a tmp directory
+	outputPath, err := os.MkdirTemp(sp.LocalDir, "*")
+	fmt.Printf("Output Path: %s\n", outputPath)
```
simonwo, line 92: Use `log.Debug().Str("Path", outputPath).Send()`, don't print directly to stdout.

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	var key string
+	if spkey != "" && envkey == "" {
+		key = spkey
+	} else if envkey != "" && spkey == "" {
+		key = envkey
+	} else if envkey != "" && spkey != "" {
+		key = spkey
+	} else {
+		key = ""
+	}
```
simonwo, line 144: You don't need to do this, the Estuary API key will already have been taken from the ENV and passed to the storage provider.

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	if key != "" {
+		log1.Println("Pinning to Estuary...")
+		//nolint:govet,noctx
+		err := estuary.PinToIPFSViaEstuary(ctx, key, CID)
```
simonwo, line 148: The responsibility of this class is to allow users to use git repos directly in Bacalhau. Pinning repos to Estuary is a separate job to be done. This class should be usable even on nodes that aren't connected to Estuary (and we currently aren't even configuring Estuary on our nodes, so it won't work in prod). 

You need to take the Estuary pinning out and put it in a separate part, I'd suggest part of the job transform discussed above that looks at whether the repo is already on IPFS.

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+		if err != nil {
+			return storage.StorageVolume{}, err
+		}
+		log1.Println("Successfully Pinned to Estuary...")
```
simonwo, line 152: Don't use the built-in log librray, use zerolog only.

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	cid, err := sp.IPFSClient.Upload(ctx, localPath)
+	fmt.Print(cid)
+	if err != nil {
+		fmt.Print(err)
+		return model.StorageSpec{}, err
+	}
+	return model.StorageSpec{
+		StorageSource: model.StorageSourceIPFS,
+		CID:           cid.CID,
+	}, nil
```
simonwo, line 187: As I said in the previous comments, this class should not be uploading to IPFS, but should be pushing back to a git repo. If you don't want to implement that, just `panic`.

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+			StorageSource: model.StorageSourceRepoClone,
+			Path:          spec.Path,
+			Repo:          spec.Repo,
+		},
+	}, nil
+}
+
+func CreateSHA1CIDPair(data url.Values) error {
+	//nolint:gomnd
+	time.Sleep(100 * time.Millisecond)
```
simonwo, line 216: Why?

pkg/storage/repo/storage.go
```diff
@@ -0,0 +1,269 @@
...
+	//nolint:noctx
+	resp, err := http.PostForm("http://kv.bacalhau.org", data)
```
simonwo, line 218: What is this service and where does the code for it live?

pkg/storage/repo/storage_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	type repostruct struct {
+		Site     string
+		URL      string
+		repoName string
+	}
+	// Rewrite this test replacing it with the clone part
+	filetypeCases := []repostruct{
+		{Site: "github", URL: "https://github.com/bacalhau-project/bacalhau.git",
+			repoName: "filecoin-project/bacalhau",
```
simonwo, line 141: This has changed

simonwo (changes requested):

dashboard/api/pkg/server/utils.go
```diff
@@ -34,6 +34,21 @@ func GetRequestBody[T any](w http.ResponseWriter, r *http.Request) (*T, error) {
...
+		return data, err
+	}
+	err = json.NewDecoder(resp.Body).Decode(&data)
+	if err != nil {
+		return data, err
+	}
+	resp.Body.Close()
+	return data, nil
+}
+
```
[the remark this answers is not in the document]
js-ts: I didn't write that, that's the code from the dashboard

pkg/storage/repo/storage.go
js-ts: https://github.com/bacalhau-project/kv.bacalhau.org

pkg/storage/repo/storage.go
js-ts: > This class should be usable even on nodes that aren't connected to Estuary (and we currently aren't even configuring Estuary on our nodes, so it won't work in prod).

I think it will not execute on node that aren't configuring estuary

cmd/bacalhau/docker_run.go
js-ts: Added HTTP support

IDK about SSH I think it may work

pkg/storage/repo/storage.go
js-ts: If we don't upload to IPFS we won't get a CID, If no CID then we can't create the kv pair for subsequent mounting of the repo

pkg/model/storage_spec.go
js-ts: I tried using the URL field for it but failed, but I think if we use the URL field it can't be easily distinguished whether an URL is an URL that is to be used to clone the repo

pkg/storage/repo/storage.go
js-ts: We will implement push functionality later

js-ts:
It doesn't take this amount of time even for government approval and I think all the relevant changes are made and tested

simonwo:
@js-ts Please do not keep closing  and re-opening pull requests to get around the review system. Being able to merge PRs is a privilege and if necessary it can be removed, but I would prefer to work with you constructively to achieve the highest quality code we can.

I understand that it’s frustrating to have code waiting, but can you explain why this needed to be rushed?

PR #2225, opened 2023-03-22
Enable docker style logging for wasm

rossjones:
Docker currently writes its logs to disk, enabling us to pick them up again during the lifetime of the container.  This PR implements some of the same functionality for wasm so that we can stream its logs in the same fashion as docker.


rossjones:
Closed in favour of #2301 which takes a slightly different approach in how the data is written to a buffer before syncing to disk.

PR #2234, opened 2023-03-27
Add nodeid to path of the result folder

rossjones:
Currently the result folder only contains the jobid.  This PR adds the current node's ID to the path, meaning that if a single machine is running multiple nodes it is able to create multiple result folders for a given job.  This may be useful when retries are needed.


rossjones:
This needs to be the execution id rather than the node id, there's a possibility that a node could be asked to process a job again.

PR #2243, opened 2023-03-28
Allow authenticated pulls from default docker hub

rossjones:
Allows the compute node to do an authenticate image pull when pulling from the default (rate-limited) docker hub - so images of the form `image` or `user/image`, but not `registry/user/image`.

By setting the environment variables `DOCKER_USERNAME` and `DOCKER_PASSWORD`, any images pulled will authenticate when it is using the docker.com hub.  Rather than use an actual password for the DOCKER_PASSWORD field, it is better to get a read-only access token from https://hub.docker.com/settings/security


pkg/docker/docker.go
```diff
@@ -325,3 +341,29 @@ func logImagePullStatus(ctx context.Context, m *sync.Map) {
...
+			encodedJSON, err := json.Marshal(authConfig)
+			if err != nil {
+				log.Ctx(ctx).Err(err).Msg("failed to encode docker credentials")
+			} else {
+				log.Ctx(ctx).Info().Msg("authenticated inspect from docker registry")
+				return base64.URLEncoding.EncodeToString(encodedJSON)
+			}
+		} else {
+			log.Ctx(ctx).Info().Msg("cannot authenticate for custom registry")
+		}
```
simonwo, line 365: Do custom registries require a different auth method??

simonwo (approved):
LGTM!

pkg/docker/docker.go
rossjones: They use the same mechanism, but the structure that is jsonified and then base64-ified has different fields, including `ServerAddress`.  And in addition to username/password there are fields for other authentication credentials.  

We'd need a map of server -> creds for each value docker sees in the server part of `server/user/image` and that's probably not feasible.  

I think caching will be a much friendlier way of doing this for manifest info.

PR #2244, opened 2023-03-28
Added Install GIt and Git-lfs function in install-node.sh

js-ts:
To install git and git-lfs on all of our nodes.

This is to support the repo mount feature, since cloning git-lfs repos requires  us to install git-lfs

### DIFF
--- a/ops/terraform/remote_files/scripts/install-node.sh
+++ b/ops/terraform/remote_files/scripts/install-node.sh
@@ -39,7 +39,7 @@ function install-git() {
 function install-git-lfs() {
   echo "Installing Git Large File Storage (LFS)..."
   # Update package lists
-  sudo apt-get update
+  sudo apt-get update -y
   # Add git-lfs package repository
   curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
   # Install git-lfs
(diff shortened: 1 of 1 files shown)


ops/terraform/remote_files/scripts/install-node.sh
```diff
@@ -29,6 +29,26 @@ function install-docker() {
...
+  echo "Installing Git..."
+  sudo apt-get update -y
+  sudo apt-get install -y git
+  echo "...Git installation is complete!"
+}
+
+function install-git-lfs() {
+  echo "Installing Git Large File Storage (LFS)..."
+  # Update package lists
+  sudo apt-get update
```
enricorotundo, line 42: I'd add `-y` here as well

ops/terraform/remote_files/scripts/install-node.sh
```diff
@@ -29,6 +29,26 @@ function install-docker() {
...
+  # Add git-lfs package repository
+  curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
+  # Install git-lfs
+  sudo apt-get install -y git-lfs
```
enricorotundo, line 46: are you sure you don't want to pin a specific version here?

js-ts: No, they don't specify any version and as we as we are just goin to use the clone functionality

PR #2248, opened 2023-03-29
Update http-domain-allowlist.txt add github.com

wesfloyd:
Can we add github.com to the domain? Need it for [yolo pipeline work with prelinger](https://github.com/wesfloyd/prelinger-bacalhau/blob/115be0073ccc826a94a1d19a3c595e07bbd27b0d/README.md#L80)


js-ts:
@wesfloyd Do you still need this ?

wesfloyd:
@js-ts I'd like to have this merged and added to the allowlist. 

Otherwise, is there a reason for why we should not include github.com in our allowlist?

frrist (approved):
will wait for an approval from @simonwo then this is good to go, thanks Wes!

rossjones:
It's probably best to move any assets to IPFS, or load into the image at build time, but opening up github puts us at risk of being unpopular with them.

PR #2261, opened 2023-03-31
chore: requester node dictates ExecutionID/ComputeReference

frrist:
- This removes the possibility of ExecutionID.ID() returning an incomplete ID as it does now. This change sets up future work of making the JobStore persistable.

### DIFF
--- a/pkg/compute/endpoint.go
+++ b/pkg/compute/endpoint.go
@@ -4,15 +4,15 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/rs/zerolog/log"
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/bacalhau-project/bacalhau/pkg/bidstrategy"
 	"github.com/bacalhau-project/bacalhau/pkg/compute/capacity"
 	"github.com/bacalhau-project/bacalhau/pkg/compute/logstream"
 	"github.com/bacalhau-project/bacalhau/pkg/compute/store"
 	"github.com/bacalhau-project/bacalhau/pkg/model"
 	"github.com/bacalhau-project/bacalhau/pkg/system"
-	"github.com/google/uuid"
-	"github.com/rs/zerolog/log"
-	"go.opentelemetry.io/otel/trace"
 )
 
 type BaseEndpointParams struct {
@@ -104,7 +104,7 @@ func (s BaseEndpoint) prepareAskForBidResponse(
 	}
 
 	execution := *store.NewExecution(
-		"e-"+uuid.NewString(),
+		request.ExecutionID,
 		request.Job,
 		request.SourcePeerID,
 		resourceUsage,
--- a/pkg/requester/scheduler.go
+++ b/pkg/requester/scheduler.go
@@ -4,6 +4,14 @@ import (
 	"context"
 	"time"
 
+	sync "github.com/bacalhau-project/golang-mutex-tracer"
+	"github.com/google/uuid"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/bacalhau-project/bacalhau/pkg/compute"
 	"github.com/bacalhau-project/bacalhau/pkg/jobstore"
 	"github.com/bacalhau-project/bacalhau/pkg/logger"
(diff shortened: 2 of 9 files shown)


pkg/compute/endpoint.go
```diff
@@ -104,7 +104,7 @@ func (s BaseEndpoint) prepareAskForBidResponse(
 	}
 
 	execution := *store.NewExecution(
-		"e-"+uuid.NewString(),
+		request.ExecutionID,
```
wdbaruni, line 107: The executionID is intended to be assigned by the compute node and not passed by the requester node. It is expected to be unique only within the scope of a compute node. Think of it as a server side (compute side) request id.

wdbaruni (changes requested):

pkg/compute/endpoint.go
frrist: My motivation for this changes stems from the fact that the JobStore of the requester doesn't have a "complete" executionID when it creates an execution state. When the ExecutionState is created on the Requester node here:
https://github.com/bacalhau-project/bacalhau/blob/69f9e65eb157a5b2416bd0d1b3550022a5b9de7d/pkg/requester/scheduler.go#L140-L148
The ComputeReference is empty meaning the calculated ExecutionID used here:
https://github.com/bacalhau-project/bacalhau/blob/69f9e65eb157a5b2416bd0d1b3550022a5b9de7d/pkg/jobstore/inmemory/inmemory.go#L277 isn't a complete execution ID.


Once a bid is responded to, we must reference an ExecutionState by constructing another "incomplete" executionID to update its state:
https://github.com/bacalhau-project/bacalhau/blob/69f9e65eb157a5b2416bd0d1b3550022a5b9de7d/pkg/requester/scheduler.go#L373-L391
Internally to the JobStore we mutate the ExecutionState to include a ComputeReference:
https://github.com/bacalhau-project/bacalhau/blob/69f9e65eb157a5b2416bd0d1b3550022a5b9de7d/pkg/jobstore/inmemory/inmemory.go#L338
So that subsequent `UpdateExecution` calls can reference the ExecutionState using a "full" executionID: one that includes the `ComputeReference` in the output.

This flow makes it hard to reason about the various states an `ExecutionState` can be in since the ID we use to reference an Execution depends on what stage a request is in between a requester and compute node.

I'd like to simply this by ensuring we have a "complete" executionID at the time an execution is created in the JobStore of a Requester node. We can achieve this by having the Requester node dictate the ExecutionID of an Execution when it sends a bid to a compute node. The compute node may use its own internally generated ID to track the execution locally, but should use the executionID created by the Requester node when communicating with it over the bprotocol.

wdbaruni: yeah the only place we keep track of an-incomplete ExecutionID is when the RequesterNode is asking a compute for a bid or the ask was rejected. It should still be unique enough as we don't have parallel AskForBid requests to the same node, and we don't retry if the ask was rejected . However, I do agree that it would be nice if we don't have this bi-modal behaviour on how we represent the execution, but the change is a bit bigger than what is proposed:

1. Requestor nodes can start generating unique executionID, which as you've mentioned should be used for compute node callbacks. The executionID should be unique at the RequesterNode level and we don't need to generate the ExecutionLongID anymore.
1. Compute nodes will generate an ExecutionLongID that includes both the RequesterNode ID and the supplied ExecutionID. Compute nodes can generate an InternalReference that is unique within the compute node scope to be used for local file naming and docker labeling. Or it can use the ExecutionLongID for that, but it needs to make sure it is friendly with its filesystem naming

Another point is that the Requester is aware of nodes who accepted AskForBid by checking if the ComputeReference is not empty, which you might have to change as well. There shouldn't be a reason for Compute nodes to expose the internal reference to the Requester node anymore:
https://github.com/bacalhau-project/bacalhau/blob/534960f7fcbef5a71d44dc23ee9883b7dcc638f7/pkg/model/execution_state.go#L130

wdbaruni: > I am not sure that's required at this point since the ID assigned by the requester node is generated from a UUID and should already be globally unique, especially considering we use the same apraoch for JobID's today.
I understand that UUIDs are globally unique, but since this is a decentralized network, we can't trust Requester nodes wont somehow still submit an execution with a conflicting ID with another requester, which will put the compute node in a funny state. The compute node is not relying on JobIDs today to identify or group executions. Adding the requester node ID as a namespace for executions on the compute side sounds reasonable to me.

I wouldn't mind doing this as a separate change to get this PR unblocked. Also a reminder about what we should use for local directory naming and docker container labeling

pkg/requester/scheduler.go
```diff
@@ -391,6 +391,7 @@ func (s *BaseScheduler) handleAskForBidResponse(ctx context.Context,
 			ExpectedState: model.ExecutionStateAskForBid,
 		},
 		NewValues: model.ExecutionState{
+			ReceivedBid:      true,
```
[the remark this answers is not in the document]
frrist: I am attempting to set this whenever we update an execution state with an ExecutionID/ComputeReference, but am realizing now that compute nodes can return a response with an empty ExecutionID for the case when they reject an `AskForBidRequest`

Will address in a subsequent commit.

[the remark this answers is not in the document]
frrist: corrected in c81b1c397bb6f6a30a27a44e000d77f1c7ce40fe

pkg/compute/endpoint.go
frrist: Sounds good, will land the compute node change in a follow on. Tracked via: https://github.com/bacalhau-project/bacalhau/issues/2297

wdbaruni (approved):

frrist:
Closing in favor of https://github.com/bacalhau-project/bacalhau/pull/2346

PR #2285, opened 2023-04-04
generic cli storage inputs

wdbaruni:
A generic way to define storage inputs using URIs instead of separate flags per input types. This change enables and exposes S3 inputs to CLI users.

Users can also pass additional options depending on the storage type, such as endpoint and region for S3 storage.

## Examples:
```
# Mount IPFS CID to /inputs directory
-i src=ipfs://QmeZRGhe4PmjctYVSVHuEiA9oSXnqmYa4kQubSHgWbjv72

# Mount IPFS CID using simple form to /inputs directory 
-i ipfs://QmeZRGhe4PmjctYVSVHuEiA9oSXnqmYa4kQubSHgWbjv72

# Mount S3 object to a specific path
-i src=s3://bucket/key,dst=/my/input/path

# Mount S3 object with specific endpoint and region
-i src=s3://bucket/key,dst=/my/input/path,opt=endpoint=http://s3.example.com,opt=region=us-east-1

# Mount S3 object using long flag names
--input source=s3://bucket/key,destination=/my/input/path

```

## Breaking Changes
1. Removed `--input-urls, -u` to mount URLs
2. Removed `--input-volumes, -v` to mount CIDs
3. Defining `ipfs://` schema is mandatory when defining input CIDs
4. Mount path is explicitly defined using `,dst=<path>` instead of using colon as delimiter

### DIFF
--- a/cmd/bacalhau/opts/storage.go
+++ b/cmd/bacalhau/opts/storage.go
@@ -3,6 +3,7 @@ package opts
 import (
 	"encoding/csv"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/bacalhau-project/bacalhau/pkg/job"
@@ -29,8 +30,22 @@ func (o *StorageOpt) Set(value string) error {
 		key, val, ok := strings.Cut(field, "=")
 
 		if !ok {
+			// parsing simple format of source:destination
 			if i == 0 {
-				sourceURI = field
+				parsedURI, err := url.Parse(field)
+				if err != nil {
+					return err
+				}
+				// find the last colon, excluding the schema part
+				schema := parsedURI.Scheme
+				trimmedURI := strings.TrimPrefix(field, schema+"://")
+				index := strings.LastIndex(trimmedURI, ":")
+				if index == -1 {
+					sourceURI = field
+				} else {
+					sourceURI = schema + "://" + trimmedURI[:index]
+					destination = trimmedURI[index+1:]
+				}
 				continue
 			} else {
 				return fmt.Errorf("invalid storage option: %s. Must be a key=value pair", field)
--- a/cmd/bacalhau/opts/storage.go
+++ b/cmd/bacalhau/opts/storage.go
@@ -1,0 +1,81 @@
+package opts
+
+import (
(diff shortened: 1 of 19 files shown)


cmd/bacalhau/opts/storage.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (o *StorageOpt) String() string {
+	storages := make([]string, len(o.values))
+	for _, storage := range o.values {
+		repr := fmt.Sprintf("%s %s %s", storage.StorageSource, storage.Name, storage.Path)
+		storages = append(storages, repr)
+	}
+	return strings.Join(storages, ", ")
+}
```
frrist, line 89: I am not sure this is correct. Allocating a slice `storages` size of `len(o.value)` via `make` and appending to it in the loop will add additional allocations and result in the first `len(o.value)` indices containing empty strings. 
Here is an example: https://go.dev/play/p/voW7jj6cQe0

You can address this with either of the following changes:
```go
func (o *StorageOpt) String() string {
	storages := make([]string, len(o.values))
	for i, storage := range o.values {
		repr := fmt.Sprintf("%s %s %s", storage.StorageSource, storage.Name, storage.Path)
		storages[i] = repr
	}
	return strings.Join(storages, ", ")
}
```
 or
 ```go
 func (o *StorageOpt) String() string {
	storages := make([]string, 0, len(o.values))
	for _, storage := range o.values {
		repr := fmt.Sprintf("%s %s %s", storage.StorageSource, storage.Name, storage.Path)
		storages = append(storages, repr)
	}
	return strings.Join(storages, ", ")
}
```

File: cmd/bacalhau/opts/storage.go [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+
+func (o *StorageOpt) Values() []model.StorageSpec {
+	return o.values
+}
+
+// compile-time check to ensure type implements the flag.Value interface
+var _ flag.Value = &StorageOpt{}
```
frrist, line 96: nit: move this near the struct definition to inform readers the interface(s) this struct satisfies.

PR #2302, opened 2023-04-05
Fix container entry points

### LINKED ISSUE #656: Both of the default entrypoint and the commands in a container are blatted
From a user perspective, if you define either an entrypoint or a command in a container, the defaults are overwritten with an empty value if you pass no arguments to bacalhau. This is unexpected, from a user perspective. Default container entrypoints and commands should be used.

From a code perspective, we only use an entrypoint, and overwrite that with the arguments passed to bacalhau (i.e. empty when you don't pass any). We don't use or pass the command at all.

E.g. entrypoint https://github.com/filecoin-project/bacalhau/blob/main/pkg/executor/docker/executor.go#L249

Here's a reference to the command part of the container config (which is not set): https://github.com/moby/moby/blob/master/api/types/container/config.go#L82

### LINKED ISSUE #2528: Adding any parameters overrides entrypoint
We should have a better solution for allowing people to add arguments to a command line. E.g.

`bacalhau docker run mycontainer:latest` <- runs the default container with the default entrypoint

But - `bacalhau docker run mycontainer:latest foo`

Expected:
uses the default entrypoint with these as an argument (as docker does)

Actual:
`bacalhau docker run mycontainer:latest foo` <- overrides the default entrypoint.

olgibbons:
Fixes #656. Fixes #2528.


pkg/model/job.go
```diff
@@ -215,6 +215,8 @@ type JobSpecDocker struct {
 	Image string `json:"Image,omitempty"`
 	// optionally override the default entrypoint
 	Entrypoint []string `json:"Entrypoint,omitempty"`
+	// Parameters holds the default arguments that are passed to the ENTRYPOINT when the container is run.
```
lukemarsden, line 218: I think this comment is wrong? It's the "extra arguments that are passed on the commandline when the container is run", no?

pkg/executor/docker/executor.go
```diff
@@ -204,6 +204,7 @@ func (e *Executor) Run(
 		Tty:        false,
 		Env:        useEnv,
 		Entrypoint: job.Spec.Docker.Entrypoint,
```
lukemarsden, line 206: Maybe we just need to only populate this if it's not empty

cmd/bacalhau/docker_run.go
```diff
@@ -87,6 +88,8 @@ type DockerRunOptions struct {
 }
 
 func NewDockerRunOptions() *DockerRunOptions {
+	//Olgibbons!! delete:
+	fmt.Println("NewDockerRunOptions has been called")
```
simonwo, line 92: Delete?

cmd/bacalhau/docker_run.go
```diff
@@ -262,13 +265,20 @@ Git-LFS repo (e.g. '-i gitlfs://huggingface.co/bigscience/test-bloomd.git' URI m
...
+
 	dockerRunCmd.PersistentFlags().AddFlagSet(NewRunTimeSettingsFlags(&ODR.RunTimeSettings))
 	dockerRunCmd.PersistentFlags().AddFlagSet(NewIPFSDownloadFlags(&ODR.DownloadFlags))
 
 	return dockerRunCmd
 }
 
 func dockerRun(cmd *cobra.Command, cmdArgs []string, ODR *DockerRunOptions) error {
+	//Olgibbons: delete
+	fmt.Println("dockerRun has been called")
```
simonwo, line 281: Delete?

cmd/bacalhau/docker_run.go
```diff
@@ -320,9 +330,12 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, ODR *DockerRunOptions) erro
...
 // CreateJob creates a job object from the given command line arguments and options.
 func CreateJob(ctx context.Context, cmdArgs []string, odr *DockerRunOptions) (*model.Job, error) { //nolint:funlen,gocyclo
+	//olgibbons: delete:
+	fmt.Println("CreateJobFunc has been called")
 	odr.Image = cmdArgs[0]
-	odr.Entrypoint = cmdArgs[1:]
-
+	odr.Cmd = cmdArgs[1:]
+	fmt.Println("cmdArgs[0] is", odr.Image)
+	fmt.Println("odr.Cmd is", odr.Cmd)
```
simonwo, line 338: Remove printlns?

cmd/bacalhau/docker_run.go
```diff
@@ -448,6 +462,7 @@ func CreateJob(ctx context.Context, cmdArgs []string, odr *DockerRunOptions) (*m
 		odr.NodeSelector,
 		odr.WorkingDirectory,
 	)
+	fmt.Println("odr.Entrypoint is", odr.Entrypoint)
```
simonwo, line 465: Remove println?

pkg/docker/docker.go
```diff
@@ -211,6 +211,21 @@ func (c *Client) RemoveContainer(ctx context.Context, id string) error {
...
+				OS:           info.Os,
+				OSVersion:    info.OsVersion,
+			},
+		}, nil
+	} else if !dockerclient.IsErrNotFound(err) {
+		// The only error we wanted to see was a not found error which means we don't have
+		// the image being requested.
+		return nil, err
+	}
+
```
simonwo, line 228: You can probably remove this – it will clash with what Ross has done

cmd/bacalhau/docker_run.go
```diff
@@ -73,7 +73,8 @@ type DockerRunOptions struct {
 	NodeSelector     string   // Selector (label query) to filter nodes on which this job can be executed
 
 	Image      string   // Image to execute
-	Entrypoint []string // Entrypoint to the docker image
+	Entrypoint string   // Entrypoint to the docker image
+	Cmd        []string //Command arguments for Entrypoint
```
simonwo, line 77: The field in the JobSpecDocker is called `Parameters` – should this be the same?

pkg/job/factory.go
```diff
@@ -99,15 +100,23 @@ func ConstructDockerJob( //nolint:funlen
...
 	j.APIVersion = a.String()
-
+	//If entrypoint is specified, construct a 1 element slice for it. Otherwise don't pass anything for that field
+	var entrypointSlice []string
+	if entrypoint == "" {
+		//tell docker to use default entrypoint, which can be defined in the Dockerfile
+		entrypointSlice = nil
+	} else {
+		entrypointSlice = []string{entrypoint}
+	}
```
simonwo, line 110: `entrypointSlice` will be nil by default, so you only need to handle the `else` case, e.g. `if entrypoint != ""`

pkg/test/executor/docker_entrypoint_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+//When the entrypoint flag is used on the CLI, eg "docker run --entrypoint /bin/echo image hello world" docker will ignore the CMD
+//Stored in the dockerfile. Any paramaters used after the image will be interpreted as the new CMD.
+//If no entrypoint is specified in the dockerfile and the CMD appended to the CLI does not contain an executable that can
+//be found in Alpine's $PATH, the docker daemon will throw an error.
+//Please note that if a dockerfile specifies neither the CMD or ENTRYPOINT then docker will use the base image's (if specified).
+//Alpine's base image has a CMD of '/bin/sh' and no entrypoint.
```
simonwo, line 33: This should be true of all images, not just alpine. So perhaps replace references to "Alpine" with "the chosen image"?

pkg/test/executor/docker_entrypoint_test.go
```diff
@@ -0,0 +1,281 @@
...
+		buildOptions := types.ImageBuildOptions{
+			Dockerfile: filepath.Join(tempDir, name),
+			Tags:       []string{tag + ":latest"},
+		}
+
+		response, err := cli.ImageBuild(ctx, buildContext, buildOptions)
+		//olgibbons delete after testing:
+		if err != nil {
+			suite.T().Fatal("Error building image: ", err)
+		}
```
simonwo, line 135: This can just be a `require.NoError`?

pkg/test/executor/docker_entrypoint_test.go
```diff
@@ -0,0 +1,281 @@
...
+	ctx := context.Background()
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	require.NoError(suite.T(), err)
+
+	imageTags := []string{
+		suite.testName + "-image-true-true",
+		suite.testName + "-image-false-true",
+		suite.testName + "-image-true-false",
+		suite.testName + "-image-false-false",
+	}
```
simonwo, line 154: This is a little bit brittle because it relies on:
* The logic of `createUniqueName`
* The `dockerfilePermutations`
I.e. if you change either of those things, this will be broken.

Could you keep a list of all the images created instead? Or could you use something like `suite.T().Cleanup`?

pkg/test/executor/docker_entrypoint_test.go
```diff
@@ -0,0 +1,281 @@
...
+		{stderr, "cdrom\nfloppy\nusb\n", image, defaultUserOverwrites.entrypoint, defaultUserOverwrites.cmd},
+		{stderr, "", image, nil, nil},
+	}
+	for _, test := range newTests {
+		testScenario := createTestScenario(test.expectedStderr, test.expectedStdout, test.image, test.entrypoint, test.parameters)
+		RunTestCase(suite.T(), testScenario)
+	}
+}
+
+//in cmd/docker_run_test.go: make sure flag is hooked up correctly
```
simonwo, line 281: Rogue comment?

pkg/test/scenario/scenario.go
```diff
@@ -68,6 +68,9 @@ type Scenario struct {
 	// whether it was successful or not. If empty, the job will not be waited
 	// for once it has been submitted.
 	JobCheckers []job.CheckStatesFunction
+
+	//olgibbons: delete this if it doesn't work:
+	CLIParameters []string
```
simonwo, line 73: Do you still need this? You don't seem to use it in the tests?

pkg/test/scenario/test_scenarios.go
```diff
@@ -266,5 +296,8 @@ func GetAllScenarios() map[string]Scenario {
 		"wasm_csv_transform": WasmCsvTransform,
 		"wasm_exit_code":     WasmExitCode,
 		"wasm_dynamic_link":  WasmDynamicLink,
+		//olgibbons: do this properly:
+		"DockerfileNoEntryPointNoCommand":  DockerFileNoEntryPointNoCommand,
+		"DockerfileNoEntryPointYesCommand": DockerFileNoEntryPointYesCommand,
```
simonwo, line 301: Do you still need these?

testdata/docker-entrypoints/cmd-echo/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM ubuntu:20.04
+#ENTRYPOINT ["/bin/sh", "-c"]
+CMD ["echo", "helloo from dockerfile"]
```
simonwo, line 3: Do you still need all this stuff in testdata?

pkg/test/executor/test_runner.go
```diff
@@ -70,9 +71,19 @@ func RunTestCase(
...
+		if strings.Contains(err.Error(), `"media": executable file not found in $PATH`) {
+			t.Log("Expected error occured for no entrypoint", err)
+			//skip rest of test if correct error occurs
+			return
+		} else {
+			require.NoError(t, err)
+		}
+	} else {
+		require.Empty(t, runnerOutput.ErrorMsg)
+	}
```
simonwo, line 84: Is this still necessary? It would be preferable to not encode test-specific logic in this generic code.

simonwo (approved):

PR #2307, opened 2023-04-06
Allow compute nodes to receive async job authorizations

### LINKED ISSUE #2290: Allow compute node to submit job moderation requests
At the moment the compute node can only check that a request has been moderated, and can't submit it's own, because it doesn't have a callback API for job approvals.

This is of interest for orgs that are part of a bigger Bacalhau cluster but maybe don't have all jobs going through their own requester node.

Add an API endpoint similar to the requester for job approval callbacks and ensure the compute node includes it in it's bid requests.

simonwo:
The compute node has always been able to call out to an external web hook as part of the bidding process but until now it has not been able to handle a waiting response and was required to respond to bids immediately. The compute node therefore had no need for a web callback once the job was moderated.

Now, the bid response is an asynchronous operation and the compute node can receive its own web callback. Compute nodes acknowledge that they have received a bid request but will respond to it in an async manner.

This allows compute nodes to exist in a separate trust domain from the requester node, because they can now have their own dashboard which will moderate jobs which can be separate from anything else.

Resolves #2290.

### DIFF
--- a/pkg/test/compute/data_locality_bid_test.go
+++ b/pkg/test/compute/data_locality_bid_test.go
@@ -1,0 +1,34 @@
+//go:build integration || !unit
+
+package compute
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type DataLocalityBidSuite struct {
+	AskForBidSuite
+}
+
+func TestDataLocalityBidSuite(t *testing.T) {
+	suite.Run(t, new(AskForBidSuite))
+}
+
+func (s *DataLocalityBidSuite) SetupTest() {
+	s.config.JobSelectionPolicy.RejectStatelessJobs = true
+	s.AskForBidSuite.SetupTest()
+}
+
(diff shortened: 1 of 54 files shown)


simonwo:
hello

pkg/test/compute/data_locality_bid_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+type DataLocalityBidSuite struct {
+	AskForBidSuite
+}
+
+func TestDataLocalityBidSuite(t *testing.T) {
+	suite.Run(t, new(AskForBidSuite))
+}
```
frrist, line 17: @simonwo ran into some issues here while working on https://github.com/bacalhau-project/bacalhau/pull/2712 Is this logic right? It looks like we never instantiate the `DataLocalityBidSuite` meaning its tests don't get run?

PR #2352, opened 2023-04-13
Allow network traffic to RPC Endpoints (Polygon, Ethereum, Binance, and Filecoin)

kelindi:
Enable network traffic to RPC endpoints on Polygon, Ethereum, Binance, and Filecoin networks to allow containers to access on-chain information for use cases like chain analysis and oracles.


simonwo:
OK, [the consensus](https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1681378745046769) seems to be that access to RPC endpoints does not allow general coin mining and can't be used in a malicious way. The use cases opened up by this are also limited whilst we don't have private computation on the public network, as any keys used to e.g. sign transactions would need to be in plaintext. So the attack vector on this looks small enough to warrant experimentation with this.

simonwo:
This change will be included in the next production release.

PR #2374, opened 2023-04-18
Allows bprotocol to process errors with responses 

### LINKED ISSUE #2371: Proper error handling for bprotocol requests
Today the requester node will fail to decode a bprotocol response and an error is returned instead of expected response type. 

This is an example of AskForBid request where the compute node didn't have S3 storage engine installed and returned an error when attempting to calculate the disk space requirement. While we should handle uninstalled engines gracefully, returning errors is expected in many other places in the code.

```
# compute error
error getting job disk space requirements: model.StorageSourceType is not installed: S3

# requester side
compute.AskForBidRequest: failed to decode response from peer QmXaXu9N5GNetatsvwnTfQqNtSeKAD6uCmarbh3LMRYAcF: stream reset
```

rossjones:
Currently the compute node end of the bprotocol is unable to return
errors to the requester node, resulting in the requester node generating
errors about decoding the request object.

What we want to happen is to allow the Compute Proxy (and the Callback
Proxy) to receive errors from the other process, so that we can log
useful error messages in the requester.

Should resolve #2371

### DIFF
--- a/pkg/transport/bprotocol/compute_handler.go
+++ b/pkg/transport/bprotocol/compute_handler.go
@@ -69,13 +69,21 @@ func handleStream[Request, Response any](ctx context.Context, stream network.Str
 	defer closer.CloseWithLogOnError("stream", stream)
 
 	response, err := f(ctx, *request)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


frrist (approved):
Good stuff!

Issue #2351, opened 2023-04-13
EngineSpec Refactor


frrist:
The gist of and motivation for this change is summarized in this [notion document]( https://www.notion.so/pl-strflt/Job-Schema-521ba6cdc06b4bdb940dbb151c576882)

The general idea is to replace the Specs Engine enum with the following:
```go
type Engine struct {
        // Type is a string representation of the Engines type
	Type string
        // Spec is the set of arguments(?) the engine will use during operation.
	Spec map[string]interface{}
}
```

Using a plain string for `Type` instead of dedicated enum type (as we do now) allows for the addition of new engines and submitting jobs at runtime without having to do a new release or approve the engine. For example, a compute provider wants to accept `spark_3.3` jobs. It should be able to do so just by publishing the engine in it's `NodeInfo` message. Users should be able to submit `spark_3.3` jobs without having to update their client to wait for the requester to update their server.

Using a `map[string]interface{}` for `Params` instead of a structure containing dedicated types (as we do now, e.g. `EngineSpecDocker`) enables the addition of new engines as their arguments will be dynamic to the engine `Type`. In future version we may consider making this a `[]byte`, but for now a `map` is more pragmatic.

### Acceptance Criteria
- [ ] Changes to `EngineSpecDocker` #2381 
  - [ ] move from the `modules` package to the `executor` package.
  - [ ] provide conversion between the `Engine.Param` and concrete `EngineSpecDocker` type
- [ ] Changes to `EngineSpecWasm` #2381
  - [ ] move from the `modules` package to the `executor` package.
  - [ ] rename `EntryPoint` to `Entrypoint`
  - [ ]  provide conversion between the `Engine.Param` and concrete `EngineSpecWasm` type
- [ ] Delete `JobSpecLanguage`. https://github.com/bacalhau-project/bacalhau/pull/2373
  - [ ] Delete its executor and related bits

frrist:
In addition, we must determine whether we should permit the creation of ["invalid jobs"](https://github.com/bacalhau-project/bacalhau/blob/89c97a4a3087b4ac2db4c988a8bd48c7c626e3cb/pkg/model/engine.go#L21) through the Command Line Interface (CLI). At present, during job creation, we validate its engine against a list of supported engines. In principle, this change would eliminate the notion of "valid jobs," as it grants users the flexibility to specify any engine name of their choice. Instead of rejecting jobs with unrecognized engines, we could issue a warning to users about the utilization of an "unknown" engine.

[...]

PR #2381, opened 2023-04-18
Job EngineSpec refactor

frrist:
### What
This PR modifies the `Spec` as outlined in #2351. The gist is we are removing `JobSpecWasm` and `JobSpecDocker` from the `Spec` and replacing the `Spec.Engine` field with `EngineSpec` which contains:
```go
type EngineSpec struct {
        // Type is a string representation of the Engines type
	Type Engine
        // Spec is the set of arguments(?) the engine will use during operation.
	Spec map[string]interface{}
}
``` 
An `EngineSpec` can represent both `JobSpecWasm` and `JobSpecDocker` generically. The `JobSpecWasm` and `JobSpecDocker` still exist, one implementation resides in the `pkg/model` package and another in the `executor` package. Both implementations contain methods to marshal between `EngineSpec` and `JobSpec*`, and are duplicated across package to permit inspection by the Requester node and Compute node (and also due to dependency ordering).

### Notes
- Marshaling between `EngineSpec` and `JobSpec*` is complicated for several reasons:
  - We lose type safety when crossing API boundaries between Requester and Compute nodes, and because the value of the `EngineSpec.Spec` map contains an interface _anything_ can be there. I have called this out explicitly [here (pkg/executor/wasm/spec/engine_spec_wasmm.go:111)](https://github.com/bacalhau-project/bacalhau/pull/2381/files#r1170769219) (As an example, when passed over an API a `StorageSpec` is converted to a `map[string]interface` which makes reflection-based decoding non-trivial)
  - I have explicitly defined constants as the keys of the `EngineSpec.Spec` map for the cases when it contains a `JobSpecWasm` or `JobSpecDocker` engine. The responsibility of correctly defining these values in a Job defined as josn/yaml provided by the user is implicitly left to their discretion, and this change actually breaks that workflow, more in the below "Decision" section below.
  - In https://github.com/bacalhau-project/bacalhau/tree/frrist/spec/engine-bytes I sketched a change that would make the `EngineSpec.Spec` field a JSON encoded byte array containing the Engine data. The change to bytes was motivated by a commented from @simonwo here: https://github.com/bacalhau-project/bacalhau/issues/2351#issuecomment-1508425576. I like this change a lot better as it preserves type safety and moves us closer to a self describing `EngineSpec` but more thought is needed regarding the details of implementation, which I have postponed in favor of completing this work as defined in #2351. 

- Since mutation of Jobs is occasionally performed by the Requester node I have provided methods like `MutateEngineSpec` to facilitate this. 
  - I voiced that we should move away from this mutable-job-pattern in https://github.com/bacalhau-project/bacalhau/issues/2351#issuecomment-1507833676 and have since written a [brief memo](https://www.notion.so/Separate-Protocols-and-APIs-2b8552ea1fe6428b832d111494b59f37?pvs=4) proposing ways to move away from this in the future.

### Decision

**Do we want to continue to allow users to specify jobs using the previous `Spec` format and adapt it to the new one, or do we want to make this a breaking change and require users to specify Jobs using the new `Spec` format?** 

This is the current reason of CI failures.

Since this PR is modifying the `Spec` it is both a protocol change and an API change:
- The protocol change primarily affects Compute and Requester nodes as they now need to deserialize `EngineSpec`s to an explicit type (`JobSpec*`) in order to operate over it- this was mostly a mechanical change. 
- The API change however requires a decision as it affects the way users issues jobs when defining them as either json or yaml as the format has now changed. 

(The [brief memo](https://www.notion.so/Separate-Protocols-and-APIs-2b8552ea1fe6428b832d111494b59f37?pvs=4) I mentioned above was further motivated by this mixture of Protocol and API, and the need to make such decisions as this.)


pkg/test/scenario/example_noop_test.go
```diff
@@ -22,10 +23,7 @@ var noopScenario Scenario = Scenario{
 		},
 	},
 	Spec: model.Spec{
-		Engine: model.EngineNoop,
-		Wasm: model.JobSpecWasm{
-			EntryPoint: "_start",
-		},
+		EngineSpec: model.EngineSpec{Type: model.EngineNoop},
```
frrist, line 26: I wasn't sure to handle this specific conversion to the new `EngineSpec`. But my sense was that this test expects a noop engine and I have attempted to keep the same behavior.

pkg/model/job.go
```diff
@@ -213,43 +217,14 @@ func (s *Spec) AllStorageSpecs() []*StorageSpec {
...
-	Parameters []string `json:"Parameters,omitempty"`
-
-	// The variables available in the environment of the running program.
-	EnvironmentVariables map[string]string `json:"EnvironmentVariables,omitempty"`
-
-	// TODO #880: Other WASM modules whose exports will be available as imports
-	// to the EntryModule.
-	ImportModules []StorageSpec `json:"ImportModules,omitempty"`
-}
-
```
frrist, line 252: These now live in `model/engine_spec_docker.go`, `model/engine_spec_wasm.go`, `executor/docker/spec` and `executor/was/spec` along with their helper methods. They are duplicated across these package to avoid cyclical dependencies and to satisfy the implicit server<-> client relationship of requester and compute nodes.

cmd/bacalhau/docker_run.go
```diff
@@ -241,18 +241,13 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, ODR *DockerRunOptions) erro
 	}
 	err = jobutils.VerifyJob(ctx, j)
 	if err != nil {
-		if _, ok := err.(*bacerrors.ImageNotFound); ok {
```
frrist, line 244: Nothing actually returns this error type so I removed it.

cmd/bacalhau/wasm_run.go
```diff
@@ -49,6 +53,15 @@ type WasmRunOptions struct {
...
 	Publisher       opts.PublisherOpt
 	Inputs          opts.StorageOpt
+
+	// Engine Params
+
+	Entrypoint           string
+	ImportModules        []model.StorageSpec
+	EnvironmentVariables map[string]string
+	// EntryModule are passed as an argument over the CLI.
+	// Parameters are passed as an argument over CLI.
```
frrist, line 63: These contain user input from the CLI. We write to these instead of the `EngineSpec` directly.

cmd/bacalhau/wasm_run.go
```diff
@@ -182,45 +197,6 @@ func runWasm(
...
-		time.Sleep(1 * time.Second)
-
-		storage := inline.NewStorage()
-		inlineData, err := storage.Upload(ctx, info.Name())
-		if err != nil {
-			return err
-		}
-		ODR.Job.Spec.Wasm.EntryModule = inlineData
-	}
-
```
frrist, line 223: move into the `parseWasmEntryModule` method.

cmd/bacalhau/wasm_run.go
```diff
@@ -229,43 +205,95 @@ func runWasm(
...
+	}
+
+	cmd.Printf("Uploading %q to server to execute command in context, press Ctrl+C to cancel\n", moduleArg)
+	time.Sleep(1 * time.Second)
+
+	return inline.NewStorage().Upload(ctx, info.Name())
+}
+
 func newValidateWasmCmd() *cobra.Command {
-	wasmJob := defaultWasmJobSpec()
```
frrist, line 244: Previously we constructed the "default wasm job", populated its `Entrypoint` from the CLI and then inspect the entry point in the verification code. This changes replaces the wasm job with a string as the entrypoint is all we are verifying.

dashboard/api/pkg/server/client.go
```diff
@@ -112,6 +116,10 @@ func runStableDiffusion(prompt string, testing bool) (string, error) {
 	} else {
 		s = realSpec
 	}
-	s.Docker.Entrypoint = append(s.Docker.Entrypoint, prompt)
+	// TODO this is really gross, we can probably do better.
+	if entryPoint, ok := s.EngineSpec.Spec[model.DockerEngineEntrypointKey].([]string); ok {
+		entryPoint = append(entryPoint, prompt)
+		s.EngineSpec.Spec[model.DockerEngineEntrypointKey] = append(s.EngineSpec.Spec[model.DockerEngineEntrypointKey].([]string), entryPoint...)
+	}
```
frrist, line 123: The alternative is to decode the `EngineSpec.Spec` map into its corresponding structure for field inspection. It's less efficient (hardly), but possibly more readable. thoughts?

pkg/bidstrategy/type.go
```diff
@@ -26,6 +26,7 @@ func NewShouldBidResponse() BidStrategyResponse {
 }
 
 type BidStrategy interface {
+	// TODO: remove the error from this interface method return, the response should contain the error as the interface def doesn't allow a nil value return for the response and most imples wrap the error in the response.
```
frrist, line 29: Drive-by TODO, unrelated to this change, open to opinions.

frrist:
Closing in favor of https://github.com/bacalhau-project/bacalhau/pull/2492.
Thank you for the early review @rossjones, I believe many of your comments remain valid for this implementation and I will be addressing them in #2493 in subsequent commits.

cmd/bacalhau/create.go
```diff
@@ -224,13 +224,8 @@ func create(cmd *cobra.Command, cmdArgs []string, OC *CreateOptions) error { //n
 
 	err = jobutils.VerifyJob(ctx, j)
 	if err != nil {
-		if _, ok := err.(*bacerrors.ImageNotFound); ok {
```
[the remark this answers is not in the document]
frrist: The idea of implementing bacerrors certainly has merit, but I have reservations regarding the creation and upkeep of our own error-type system. There are numerous [alternatives](https://awesome-go.com/error-handling/) available that could negate the necessity for us to develop an in-house solution. One notable alternative is https://github.com/cockroachdb/errors - designed to function as a direct replacement for both github.com/pkg/errors and Go's standard errors package. I think it's worth considering an existing solution before embarking on the development of our own error-handling system; this feels well timed considering many of the baccerrors are currently unused and could use a refactor anyways. What do you think?

cmd/bacalhau/wasm_run.go
```diff
@@ -229,43 +202,90 @@ func runWasm(
...
+		return model.StorageSpec{}, fmt.Errorf("%q should point to a single file", moduleArg)
+	}
+
+	err = os.Chdir(filepath.Dir(moduleArg))
+	if err != nil {
+		return model.StorageSpec{}, err
+	}
+
+	cmd.Printf("Uploading %q to server to execute command in context, press Ctrl+C to cancel\n", moduleArg)
+	time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
frrist: Yea, that was my guess as well. Users are required to have lightening fast reflexes 😉

PR #2384, opened 2023-04-19
Changes to repo provider logging

rossjones:
Changes to the repo provider to log output instead of writing to stdout in a few places. Also removes a bit of unused code and changes the visibility of a couple of functions that don't need to be public.


js-ts (approved):

js-ts:
LGTM

PR #2395, opened 2023-04-21
fix(bidder): ensure jobs pass semantic bidding before calculating resource usage

### LINKED ISSUE #2393: Bug(Compute): AskForBid calculates resource usage before executing bidding strategy
When a Compute node executes an AskForBid request it first calculates the resource usage via its [`usageCalculator`](https://github.com/bacalhau-project/bacalhau/blob/2f61dd2359f7293a3e812778fb6bc1f49bf1be4b/pkg/compute/endpoint.go#L56). If an input is an IPFS hash, and  one of the configured storage providers is IPFS this causes the compute node to resolve the hash (via `object stat` api):
https://github.com/bacalhau-project/bacalhau/blob/2f61dd2359f7293a3e812778fb6bc1f49bf1be4b/pkg/storage/ipfs/storage.go#L59-L65

If the content is not local this means the attached IPFS storage provide will l attempt to resolve the content on the IPFS network. There is no guarantee the content exists, or is reachable, and this operation will hang for a default value of 2 mins.

This clashes with the `InputLocalityStrategy` bidder (enabled by default), which instructs compute node to reject all bids requesting it to operator over content it does not contain locally. 

Compute nodes should instead swiftly reject asks that contain content that is remote to the compute node.

frrist:
- fixes #2393


pkg/compute/bidder.go
```diff
@@ -5,69 +5,104 @@ import (
...
-			ExecutionMetadata: NewExecutionMetadata(execution),
-			Err:               err.Error(),
-		})
+			ExecutionMetadata: ExecutionMetadata{
+				ExecutionID: request.ExecutionID,
+				JobID:       request.JobID,
+			},
+			Accepted: response.ShouldBid,
+			Reason:   response.Reason,
+		}
```
simonwo, line 85: We're now returning bid results from two places – one from here and one from ReturnBidResult, and they have subtly different content now. Is this a problem?

pkg/compute/bidder.go
```diff
@@ -5,69 +5,104 @@ import (
...
 
-	b.ReturnBidResult(ctx, execution, response)
+		// if we are bidding or waiting create an execution
+		if response.ShouldWait || response.ShouldBid {
+			execution := store.NewExecution(request.ExecutionID, request.Job, request.SourcePeerID, *resourceUsage)
+			if err := b.store.CreateExecution(ctx, *execution); err != nil {
+				log.Ctx(ctx).Error().Err(err).Msg("Unable to create execution state")
+				return
+			}
```
simonwo, line 98: We previously were creating an Execution in all cases, and now we only create one if we intend to bid. I can see some advantages from an auditing and retries perspective in the compute node remembering what it has declined to bid on. Should we be keeping the previous behaviour, but just not populating the resource usage?

pkg/bidstrategy/type.go
```diff
@@ -25,6 +25,14 @@ func NewShouldBidResponse() BidStrategyResponse {
...
 }
 
+type SemanticBidStrategy interface {
+	ShouldBid(ctx context.Context, request BidStrategyRequest) (BidStrategyResponse, error)
+}
+
+type ResourceBidStrategy interface {
+	ShouldBidBasedOnUsage(ctx context.Context, request BidStrategyRequest, usage model.ResourceUsageData) (BidStrategyResponse, error)
+}
+
```
simonwo, line 35: Double plus good change. Can we redefine the below `BidStrategy` to just be a combination of these two interfaces?

simonwo (commented):
👍🏻

pkg/compute/bidder.go
frrist: It shouldn't be. The only place `ReturnBidResult` gets called is in https://github.com/bacalhau-project/bacalhau/blob/e866a9156c1384c29f13cee84cc4421a42eb345f/pkg/compute/publicapi/endpoint_approve.go#L22-L45

And we can be certain the `execution` exists in the store at that point if the bid was approved or is waiting. Perhaps we could name this method differently, or ideally remove it from the `Bidder` since the method has little to do with bidding in the first place.

frrist: On second thought, this is a problem for our testing via mocked callbacks since it is unexpected to receive multiple `OnBidComplete` callbacks. Im going to give this a second look.

frrist: Alright [sorted the testing](https://github.com/bacalhau-project/bacalhau/pull/2395/files#diff-e40c658cdd9c7b652d7597a1e363a9606dd596daef8e54d3b6491eef166837c3R61-R67). Tests calling `AskForBid` must wait for the execution to be created if the bid is accepted or waited on as I have done in the linked test package.

pkg/compute/bidder.go
frrist: True. I made this change as creating an execution for each bid, regardless of it being accepted or rejected, probably won't scale as the network grows. I also wanted to avoid adding "incomplete executions" (i.e. without resource usage date) in the execution store. But this decision is ultimately up to the compute node implementer. However, it seems unreasonable (imo) from a protocol level to expect a compute nodes to keep execution states for jobs they are not bidding on/rejecting. (e.g. A Requester shouldn't expect compute nodes to maintain state for bids they are rejecting.) 

With respect to retries, compute nodes could use their execution store to track jobs they have accepted or rejected, and again this is up to their implementation, but it would probably be more efficient for them to use a separate store/cache specifically for this purpose. (assuming we want to swiftly reject previously reject jobs?). I suppose there could be a case for keeping this state when a compute node initially rejects a job, but then later via a retry accepts it if more resources became available. 

With respect to auditing, I would expect an audit of bid accept/reject to happen at the Requester level given it's the "thingy" that ultimately cares about what has been accepted/rejected. If we want to make the case that compute nodes can be audited in this way we could use a filesystem journal managed by each compute nodes (again up to them).

If you feel strongly about keeping this state in the compute nodes I am happy to revert this change.

pkg/bidstrategy/type.go
frrist: Yup, done in latest commit.

pkg/compute/bidder.go
```diff
@@ -104,21 +124,28 @@ func (b Bidder) ReturnBidResult(ctx context.Context, execution store.Execution,
...
+	// Check semantic bidding strategies before having to calculate resource usage
+	bidStrategyResponse, err := b.semanticStrategy.ShouldBid(ctx, bidStrategyRequest)
 	if err != nil {
-		return nil, fmt.Errorf("error asking bidding strategy if we should bid: %w", err)
+		return nil, nil, fmt.Errorf("error asking bidding strategy if we should bid: %w", err)
 	}
 
+	// if the request is semantically biddable check resource-based bidding.
+	var resourceUsage model.ResourceUsageData
 	if bidStrategyResponse.ShouldBid {
```
simonwo, line 138: If the semantic strategy responds with `ShouldBid: false, ShouldWait: true`, which is what will happen with using an async HTTP call, this will cause the resourceUsage config to not be evaluated, so the execution will be created with a nil resource usage. (The system effectively ignores the value of ShouldBid if ShouldWait is true, at the moment. So a tri-state value might have been a better choice.)

We then never go back to compute it or even run the resource-based strategies, because the async return call goes straight to ReturnBidResult. So sounds like any semantic strategy that returns ShouldWait will cause resource-based selection to be skipped.

pkg/bidstrategy/semantic/image_platform.go
```diff
@@ -1,25 +1,28 @@
-package docker
+package semantic
```
simonwo, line 1: I don't think this file should live in a general part of the code / I think that this being a Docker-specific strategy, it should live within the Docker executor package somewhere. My logic being that if/when we move to the world of executors being provided through plugins, then this strategy would need to be provided by the plugin and not by the core.

pkg/compute/bidder.go
frrist: ohh yeah, great catch! I've correct the code in the latest commits and added tests for this.

pkg/bidstrategy/semantic/image_platform.go
frrist: Yeah that's fair. Although if there is ever more than one docker executor (i.e. and alternative implementation to our own) it would also need to implement something like this. My thinking here was that we could provide a library/package of "standard" bid strategies that other implementations could implement, mainly motivated by our mono repo approach today.

In ee288ce09c525fa8a5b7ea9bb53b60b9eff89856 I have moved this back into the docker executor.

simonwo (approved):

Issue #2254, opened 2023-03-30
Cache docker manifests


rossjones:
The compute node will often ask docker registries for information about specific images, in order to determine what platforms that image is available for.  We rely on the docker client to do this for us, using information we provide. Unfortunately, some registries, notably the default docker registry, implement rate limiting reducing the number of calls that can be made in any 24 hour period. This has been a recent problem as we make more requests, we hit the rate limit more often.

We should implement a cache, storing the results of reading image manifests. This will dramatically reduce the number of times we get near to our rate limit. For each `image:tag` (or `user/image:tag` or `server/user/image:tag`) we would expect one single JSON file, but we would have to implement rules to vary how long the cached item should live in the cache. These are:

* 1 hour for untagged images, or images using the latest tag.  This will minimise the impact on our hourly rate limit, and still provide flexibility for users that have a workflow using `latest`
* 1 day if the image has a tag. Although tags can change over time, they do so only rarely.
* Permanently if the image is specified using a digest as that is immutable. We may want to implement an expiry time if it is not used for a noticeable period of time. 

Optional: Although these lifetimes are reasonable, we may want to provide a mechanism to clear the cache (either entirely, for a specific name (server/user/image:tag), or for old/unused manifests).  

Whilst it seems sensible to implement a cache per compute node, it is entirely possible to run more than one compute node on a machine, and therefore a shared cache seems more efficient.  

Implementation tbd, but either a filesystem based cache, or a shared sqlite DB, may work depending on preference for pain-points.  

- File-based cache can be atomic within the same filesystem, but there is limited metadata available, and there's a fair chance that `atime` may be disabled making last access time more difficult to ascertain.
- A shared sqlite database may appear more flexible in recording metadata, and multiple readers is fine, but a writer much obtain an exclusive lock on the entire database and must wait for shared locked to be complete.

frrist:
I want to flag that the description has the potential to clash with https://github.com/bacalhau-project/bacalhau/issues/2255 (recent comments in #2255 came from conversation in https://github.com/bacalhau-project/bacalhau/issues/2351). 

### latest and versioned tags
[...]

PR #2410, opened 2023-04-25
Resolve docker digests from supplied image ids

rossjones:
Provides a utility to translate a docker image identifier (repo/org/name:tag) to a docker image digest (repo/org/name@hash). As this requires an API call to the docker instance running locally (which may then call to a remote repository) thie map from the original identifier to the digest version is maintained in a small local cache.  

Currently the caching rules for how long the map from identifier->digest should live follows those described in https://github.com/bacalhau-project/bacalhau/issues/2254, but we may want to reduce the the no-tag/latest cache to avoid caching no-tag images too long if the user is in a dev-test-push cycle of getting the image up to date. 

### TODO

* Get agreement that this should be used in requester as a job spec transformer to update the docker image (iff docker is available on the requester)


pkg/docker/image_resolver.go
```diff
@@ -0,0 +1,42 @@
...
+
+// ResolveImageID will take the provided image identifier and a resolver,
+// and attempt to provide a version of the image id containing the digest
+// instead.
+func ResolveImageID(ctx context.Context, i ImageID, resolver imageResolverFunc) (ImageID, error) {
+	if i.HasDigest() {
+		return i, nil
+	}
+
+	// TODO: Look up i.String() in cache to see if we already have a digest for it
```
frrist, line 22: We'll need to do this carefully, right? If the `ImageID` contains the `latest` tag we'll want to avoid caching it, correct? (Mostly asking to clarify my understanding of the code)

pkg/docker/image_id.go
```diff
@@ -0,0 +1,90 @@
...
+	repository string
+
+	// e.g. ubuntu or organisation/ubuntu or organisation/user/ubuntu
+	name string
+
+	// e.g. latest or kinetic or @sha256:digest-string
+	tag ImageTag
+}
+
+func NewImageID(imageID string) ImageID {
```
frrist, line 52: Not implying I think this is wrong or anything, but would it be possible to re-use the existing docker code for parsing this information? e.g. https://github.com/moby/moby/blob/master/vendor/github.com/docker/distribution/reference/reference.go#L65

rossjones: Oh yeah, that'd make life a bit easier. We already have it as an indirect dependency of docker/docker so won't introduce any more dependencies.  I'll take a look at doing that.

pkg/docker/image_resolver.go
rossjones: It's really just a marker for me to remind me where to do the lookup when I get to https://github.com/bacalhau-project/bacalhau/issues/2254  - this suggests caching the digest/manifest for ~1 hour for latest/no-tag but that might affect users who are in a dev/push/test cycle with untagged images, so may need more thought.

pkg/docker/image_resolver.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/bacalhau-project/bacalhau/pkg/config"
+	"github.com/rs/zerolog/log"
+)
+
+type imageResolverFunc func(context.Context, string, config.DockerCredentials) (*ImageManifest, error)
+
+// ResolveImageID will take the provided image identifier and a resolver,
+// and attempt to provide a version of the image id containing the digest
+// instead.
+func ResolveImageID(ctx context.Context, img ImageID, resolver imageResolverFunc) (ImageID, error) {
```
simonwo, line 17: Where is this used??

pkg/docker/docker.go
```diff
@@ -241,6 +242,65 @@ func (c *Client) SupportedPlatforms(ctx context.Context) ([]v1.Platform, error)
...
+//	  }
+//
+// This is the image that will finally be installed.
+func (c *Client) ImageDistribution(ctx context.Context, image string, dockerCreds config.DockerCredentials) (*ImageManifest, error) {
+	authToken := getAuthToken(ctx, image, dockerCreds)
+
+	dist, err := c.DistributionInspect(ctx, image, authToken)
+	if err != nil {
+		return nil, err
+	}
```
simonwo, line 295: There is some local changes @olgibbons has which also adds checking the local image cache at this step – probably worth doing if the image is not in our (memory) cache.

pkg/docker/image_resolver.go
rossjones: Nowhere yet.

pkg/docker/image_id.go
rossjones: Followed this advice, found a bug in my parsing so definitely a good move.  The interface is kind of ugly so I haven't let it leak outside the ImageID module as we really just want to be able to pick apart sections of the image id, and it seems unnecessary to require knowledge of how to cast the docker module type to various subtypes.

frrist (approved):

pkg/requester/jobtransform/docker_image_digest.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+
+	"github.com/bacalhau-project/bacalhau/pkg/docker"
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+	"github.com/rs/zerolog/log"
+)
+
+func DockerImageDigest() Transformer {
+	client, err := docker.NewDockerClient()
+	if err != nil || !client.IsInstalled(context.Background()) {
```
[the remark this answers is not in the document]
rossjones: Oops, missed that.

pkg/cache/cache.go
```diff
@@ -21,8 +28,8 @@ type CacheItem[T any] struct {
 	expiresAt int64
 }
 
-func NewCache[T any](options CacheOptions) (*Cache[T], error) {
-	c := &Cache[T]{
+func NewBasicCache[T any](options CacheOptions) (*BasicCache[T], error) {
```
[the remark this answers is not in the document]
rossjones: I've made these changes (and a couple of others) in https://github.com/bacalhau-project/bacalhau/pull/2448 (after cherry-picking from this PR) so that I can keep this PR focused on non-cache changes, and have sent a review request if you have time.

pkg/requester/jobtransform/docker_image_digest.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func DockerImageDigest() Transformer {
+	return func(ctx context.Context, j *model.Job) (modified bool, err error) {
+		client, err := docker.NewDockerClient()
+		if err != nil || !client.IsInstalled(ctx) {
+			// Return a noop if docker is not installed as it means we
+			// won't be able to find digests for images in the requester
+			return false, nil
+		}
```
simonwo, line 18: Is it worth doing this stuff up front and forming a closure over the client? The `IsInstalled` check makes a req against the Docker daemon – is it bad to do this in the job submission hot path?

pkg/requester/jobtransform/docker_image_digest.go
```diff
@@ -0,0 +1,46 @@
...
+		image, err := docker.NewImageID(j.Spec.Docker.Image)
+		if err != nil {
+			return false, nil
+		}
+
+		resolver := docker.NewImageResolver(image)
+		err = resolver.Resolve(ctx, client.ImageDistribution, docker.DockerTagCache)
+		if err != nil {
+			log.Ctx(ctx).Debug().
+				Str("Image", image.String()).
```
simonwo, line 33: You can use `Stringer` here to log an item with a String method, the advantage being that is the String method is expensive it won't be called unless the log message will actually be sent. May not be relevant here depending on the String method, but I thought it's worth mentioning for the future.

pkg/docker/image_id.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+
+	if named, ok := repo.(reference.Named); ok {
+		id.repository = reference.Domain(named)
+		id.name = reference.Path(named)
+	}
+
+	if digested, ok := repo.(reference.Digested); ok {
+		digest := fmt.Sprintf("sha256:%s", digested.Digest().Encoded())
```
simonwo, line 82: It's not always guaranteed to be `sha256` right? Can we call `Digest().Algorithm()` for this?

pkg/docker/docker.go
```diff
@@ -241,6 +241,93 @@ func (c *Client) SupportedPlatforms(ctx context.Context) ([]v1.Platform, error)
...
+	// return those details, but only if the caller did not set forceRemote
+	if !forceRemote {
+		info, _, err := c.ImageInspectWithRaw(ctx, image)
+		if err == nil {
+			repos := info.RepoDigests
+			if len(repos) >= 1 {
+				// We only want the digest part of the name, otherwise we would have
+				// to go through supporting two different values in the returned
+				// ImageManifest (fully qualified IDs and also just digests)
+				digestParts := strings.Split(repos[0], ":")
```
simonwo, line 302: Would it be safer to use `digest.Digest(repos[0]).Encoded()` here?

Or even, shouldn't the `digest` field be a full `Digest` including the algorithm?

pkg/docker/docker.go
```diff
@@ -241,6 +241,93 @@ func (c *Client) SupportedPlatforms(ctx context.Context) ([]v1.Platform, error)
...
+//	  }
+//
+// This is the image that will finally be installed.
+func (c *Client) ImageDistribution(
+	ctx context.Context, image string,
+	forceRemote bool, creds config.DockerCredentials,
+) (*ImageManifest, error) {
+	// Check the local repository first to see if we already have this image and if so,
+	// return those details, but only if the caller did not set forceRemote
+	if !forceRemote {
```
simonwo, line 294: I can only see one place this function is used, and we don't set `forceRemote`. So is this bool necessary? Does this mean calls to this method always rely on a cached value if one exists, regardless of the cached time?

Perhaps I could do with a comment explaining the overall flow here – i.e. when does the image resolver and it's cache come into play, and how does that cache interact with this (implicit) one?

Do we also need to update our use of `ImagePlatforms` used by the bidding strategy to use the caches? https://github.com/bacalhau-project/bacalhau/blob/8d752fcef94c2048dbdf4d3bcfa77cfa0d445a09/pkg/docker/docker.go#L213

pkg/requester/jobtransform/docker_image_digest.go
rossjones: It was originally like that, but there was no context available and it got called out in earlier review.  Happy to move it back into the outer function so we can close over it instead (and return a nop function earlier) - but we will have to add a pointless context as there isn't one available in the caller and I don't want to pointlessly add contexts everywhere.

pkg/requester/jobtransform/docker_image_digest.go
rossjones: TIL, thanks!

pkg/docker/image_id.go
rossjones: Yup, you're right, it does support others but 256 is the canonical one, have changed.

pkg/docker/docker.go
rossjones: forceRemote was just a useful tool for manual testing of remote requests, will remove and update with a comment explaining how the flow works in the resolver. 

>Do we also need to update our use of ImagePlatforms used by the bidding strategy to use the caches? 
No, this is done in the compute node isn't it? The code in the PR is in the requester.

pkg/docker/types.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/bacalhau-project/bacalhau/pkg/config"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type imageResolverFunc func(context.Context, string, config.DockerCredentials) (*ImageManifest, error)
+
+type ImageManifest struct {
+	// We only ever expect the digest to be the `algorithm:hash`
+	digest    string
```
simonwo, line 14: Final thought, are there any type strength advantages to using `Digest` object here?

pkg/util/env_var.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import "os"
+
+type EnvParserFunc[T any] func(string) (T, error)
+
+func GetEnvAs[T any](envVar string, deflt T, parser EnvParserFunc[T]) T {
```
simonwo, line 7: Much needed 👍🏻

simonwo (approved):

pkg/docker/docker.go
rossjones: Have switched to using digest.Part, but each repo here is the full image name with a digest - so we still don't want the image name.

PR #2415, opened 2023-04-27
Implement a simple cache

rossjones:
Implements a simple, cost-based  cache that allows us to store items in-memory and mark them as expired after a configured period of time.   When a cache is created, a single goroutine is run to manage the eviction of expired items who will have had their expiry time specified on insertion into the cache.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -93,6 +93,11 @@ func (c *Cache[T]) Delete(key string) {
 	c.items.Delete(key)
 }
 
+func (c *Cache[T]) Close() {
+	close(c.closer)
+	delete(caches, c.name)
+}
+
 func (c *Cache[T]) cleanup(frequency time.Duration) {
 	ticker := time.NewTicker(frequency)
 	defer ticker.Stop()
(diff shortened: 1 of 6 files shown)


pkg/cache/cache.go
```diff
@@ -0,0 +1,130 @@
+package cache
+
+import (
+	"time"
```
frrist, line 4: The test for this package can be made less racy by using a mockable clock such as https://github.com/benbjohnson/clock embedded in the cache. It satisfies the interface of the functions in the standard library time package (meaning supports Ticker and such).

pkg/cache/types.go
```diff
@@ -0,0 +1,38 @@
+package cache
+
+import (
+	"errors"
+	"time"
+)
+
+var errWrongCacheType error = errors.New("requested cache type did not match previous cache with that name")
+var errTooCostly error = errors.New("item too costly for cache")
+var errTooFull error = errors.New("cache is full")
```
frrist, line 10: If we are going to return these error lets export them so that assertions can be made outside of this package.

pkg/cache/cache_test.go
```diff
@@ -0,0 +1,84 @@
+//go:build unit || !integration
+
+package cache
```
frrist, line 3: nit: slight preference for tests live in their own package (`cach_test`) when access to the internals of the "thing" being tested are not required.

PR #2418, opened 2023-04-27
fix: network config validation

frrist:
- NetworkConfig with HTTP networking enabled but empty Domain set is invalid.
- Motivated by conversation here: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1682548637776789

For example Users requesting networking who do not specify a domain will now see the following error message:

```bash
$ bacalhau docker run ghcr.io/kelindi/fvm-pricing-oracle:sha-4250033 --network=http

Error verifying job: invalid networking configuration, at least one domain is required when networking type HTTP is enabled
```

### DIFF
--- a/pkg/model/network.go
+++ b/pkg/model/network.go
@@ -113,6 +113,9 @@ func (n NetworkConfig) IsValid() (err error) {
 // expected by our Docker HTTP gateway, which complains and/or fails to start if
 // these requirements are not met.
 func (n NetworkConfig) DomainSet() []string {
+	if n.Domains == nil {
+		return []string{}
+	}
 	domains := slices.Clone(n.Domains)
 	slices.SortFunc(domains, func(a, b string) bool {
 		// If the domains "match", the match may be the result of a wildcard. We
--- a/pkg/executor/docker/network.go
+++ b/pkg/executor/docker/network.go
@@ -7,14 +7,15 @@ import (
 	"net"
 	"time"
 
-	"github.com/bacalhau-project/bacalhau/pkg/config"
-	"github.com/bacalhau-project/bacalhau/pkg/logger"
-	"github.com/bacalhau-project/bacalhau/pkg/model"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"go.uber.org/multierr"
+
+	"github.com/bacalhau-project/bacalhau/pkg/config"
+	"github.com/bacalhau-project/bacalhau/pkg/logger"
+	"github.com/bacalhau-project/bacalhau/pkg/model"
 )
 
 const (
@@ -121,6 +122,10 @@ func (e *Executor) createHTTPGateway(
 		return nil, nil, errors.Wrap(err, "error getting network subnet")
 	}
 	subnet := internalNetwork.IPAM.Config[0].Subnet
+
+	if len(job.Spec.Network.DomainSet()) == 0 {
+		return nil, nil, fmt.Errorf("invalid networking configuration, at least one domain is required when %s networking is enabled", model.NetworkHTTP)
+	}
 
 	// Create the gateway container initially attached to the *host* network
 	domainList, derr := json.Marshal(job.Spec.Network.DomainSet())


pkg/model/network.go
```diff
@@ -87,9 +86,15 @@ func (n NetworkConfig) Disabled() bool {
...
 // otherwise.
-func (n NetworkConfig) IsValid() (err error) {
+func (n NetworkConfig) IsValid() error {
 	if n.Type < NetworkNone || n.Type > NetworkHTTP {
-		err = multierr.Append(err, fmt.Errorf("invalid networking type %q", n.Type))
+		return fmt.Errorf("invalid networking type %q", n.Type)
+	}
+
+	// If the network is enabled, but no domains are specified, the NetworkConfig is considered invalid
+	if !n.Disabled() && len(n.DomainSet()) == 0 {
```
simonwo, line 95: We only need domains for HTTP jobs currently – full networking does not enforce an allowlist.

pkg/model/network.go
```diff
@@ -87,9 +86,15 @@ func (n NetworkConfig) Disabled() bool {
...
+		return fmt.Errorf("invalid networking configuration, at least one domain is required when networking is enabled\n" +
+			"For a list of allowed dowmains see https://docs.bacalhau.org/next-steps/networking/#support-for-networked-jobs-on-the-public-network")
```
simonwo, line 97: Hm… our allowlist is definitely only intended for our own nodes to match our own risk profile. For other public nodes or private clusters the list could/should be different. Does it always make sense to redirect the user to this?

pkg/model/network.go
frrist: 👍 addressed in latest commit.

pkg/model/network.go
frrist: ahh right, good point, I have removed the bit that directs users to our website. Perhaps it would make sense to move some job validation to the `requester` package so that more specific errors can be returned? I imagine it may also be the case that some Requester implementations allow HTTP networking without a domain list?

simonwo: In fact... the fact there is a domain allowlist is very much intended only as a starting point (which is why it's implemented as a shell script using the exec hook rather than in the core).

So... maybe the corrext fix is (unfortunately) not to reject network config that has zero domains as invalid, but to work out why the gateway container is failing with zero domains, and make it work under that configuration? (It definitely used to!) Or, instead, enforce the "no zero domains" requirement in our shellscript instead – which would still give an informative error to the user? That way, users implementing alternative selection strategies can support have zero declared domains if they want?

frrist: The comment here: https://github.com/bacalhau-project/bacalhau/blob/902f334d39aa90e9810b7bac255aec33e2fa6c97/pkg/model/network.go#L108-L114 loosely implies domains are required by the container. 

Based on the errors in the log messages below (I pulled this from the [slack thread](https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1682553700232479?thread_ts=1682548637.776789&cid=C02RLM3JHUY) triggering this issues) it would suggest that one of the `jq` commands in our shell scripts is encountering a null value 

```json
{
  "level": "debug",
  "NodeID": "QmUDAXvv",
  "trace_id": "e8d773ed94ecabd7738c8c12c9819775",
  "span_id": "72b9ff77c57eb749",
  "jobid": "ce6553ed-c80c-49e7-af38-0340c94f14e5",
  "Job": "ce6553ed-c80c-49e7-af38-0340c94f14e5",
  "ExecutionID": "e-94612ed1-c733-4cd2-851c-0df0b3f73e93",
  "Source": "stderr",
  "time": "2023-04-26T23:55:35.173076195Z",
  "caller": "pkg/logger/logger.go:245",
  "message": "jq: error (at <stdin>:1): Cannot iterate over null (null)"
}
```

Looking more closing at our code and script to run the gateway container:
1. We marshal an empty string slice here (because `DomainSet()` returns an empty slice): https://github.com/bacalhau-project/bacalhau/blob/902f334d39aa90e9810b7bac255aec33e2fa6c97/pkg/executor/docker/network.go#L126
    - which when [printed returns `null`](https://go.dev/play/p/Nq7pZ6Pu9oy)
2. We set `BACALHAU_HTTP_DOMAINS` to `null` here: https://github.com/bacalhau-project/bacalhau/blob/902f334d39aa90e9810b7bac255aec33e2fa6c97/pkg/executor/docker/network.go#L136
3. We run this command (which I assume produces the above error message): https://github.com/bacalhau-project/bacalhau/blob/902f334d39aa90e9810b7bac255aec33e2fa6c97/pkg/executor/docker/gateway/gateway.sh#L17

Putting this together locally reproduces the same error:
```bash
$ export BACALHAU_HTTP_DOMAINS="null"
$ echo "$BACALHAU_HTTP_DOMAINS" | jq -r '.[]'
jq: error (at <stdin>:1): Cannot iterate over null (null)
```

So similar to what you have suggested we could enforce a network domain list in the executor around here: https://github.com/bacalhau-project/bacalhau/blob/902f334d39aa90e9810b7bac255aec33e2fa6c97/pkg/executor/docker/network.go#L126

The downside to this is that we allow, what we consider to be an invalid spec, to make it all the way to an execution.

frrist: I have adjusted the fix in b305e33d1befd92b0a53d79a924250398d200e46

simonwo: Great analysis. I expect that step 1 was where I fell down (I wrote nearly all of this) – I would have expected an empty slice to be marshalled to an empty array rather than null. Fix looks good, thanks!

simonwo (approved):

pkg/model/network.go
frrist: > I would have expected an empty slice to be marshalled to an empty array rather than null

That would be the case if the slice was empty ([heres a quick example](https://go.dev/play/p/zncYOIRcFiU)), but since the `Domains` field of the `NetworkConfig` is never initialized the slice is nil rather than empty. I have pushed [97203fa](https://github.com/bacalhau-project/bacalhau/pull/2418/commits/97203fa2fed807e8edbb12d31a5d6a866a7b034c) to ensure future versions that call this method don't encounter the same problem

PR #2442, opened 2023-05-03
Enable automatic merge for all Dependabot requests

simonwo:
- We have been trialling this for just patch updates – it works fine, but most people are doing minor updates and not patches. 
- We would rather have many requests spaced out rather than big batches of updates at once, because that way we don't rebase and run tests for all of the PRs once one is merged.


rossjones (approved):
LGTM. The only thing I can imagine being a particular pain is those jumps up in major versions that sometimes happen, but presumably they won't pass tests and be auto-merged if they're problematic.

PR #2444, opened 2023-05-03
update slack link

aronchick:


frrist:
The `#bacalhau` channel doesn't currently exist in the bacalhau slack, should we create that channel, or remove it from the README?

PR #2448, opened 2023-05-04
Restructure cache package and the basic cache

rossjones:
Restructures the cache package to separate the different types of
caches, and also the configuration of those caches.  Implements options
as recommended for the BasicCache (basic/cache.go).

Removes use of the the mockable clock as it is racy around ticker
resets when using the mock clock. In its place it now enables the user
to provide a function to handle evictions and so we can override the
default (time check) function and specify the criteria for removal (as
well as providing a synchronisation mechanism for fixing the racy 
expiry test).


frrist (approved):
LGTM :shipit:

PR #2449, opened 2023-05-04
Storage cleanup of inputs

rossjones:
Cleanup downloads after the executor's Run method has finished.

Currently each executor is responsible for performing the download of content and then mapping it to their equivalent of docker's volumes. Indirectly this has led to multiple ways of implementing it within the executors, and triggering the cleanup task is not straight-forward.  

This PR ensures it defers a function call so that the downloaded inputs are cleaned up in line with the implementation within the storage provider (e.g. some are noops). This implementation also now duplicates almost identifical code but this is resolved in #2466 which is a longer term solution to the problem of cleaning up resources.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -325,7 +325,7 @@ require (
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/rs/cors v1.7.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
-	github.com/samber/lo v1.36.0 // indirect
+	github.com/samber/lo v1.36.0
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/spf13/afero v1.9.4 // indirect
 	github.com/spf13/cast v1.5.0 // indirect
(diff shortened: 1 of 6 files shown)


simonwo:
Did I see a similar PR where we took the storage preparation out of the executor and into an "environment"? Have we moved away from that idea? Personally I agree that it would be nice if executors didn't have to do this.

We have `ParallelPrepareStorage` which is an executor-generic method – should we also have `ParallelCleanStorage` which can do this generic cleaning? (Unless I'm blind the code seems to be the same across executors.) And we could then do

```golang
defer storage.ParallelCleanStorage(e.storageProvider, inputVolumes)
```

rossjones:
> Did I see a similar PR where we took the storage preparation out of the executor and into an "environment"? Have we moved away from that idea? Personally I agree that it would be nice if executors didn't have to do this.

After submitting this PR I realised I may as well do a larger refactor to remove the deep knowledge of storage from the executor and so started the environment PR in https://github.com/bacalhau-project/bacalhau/pull/2466 - I think 2466 is my preferred approach, but still needs further work to disconnect entirely from the executor.

simonwo (commented):
Fine by me – I'd prefer a generic function in the `storage` package but not a blocker. Do we need tests?

rossjones:
> Fine by me – I'd prefer a generic function in the `storage` package but not a blocker. Do we need tests?

We do, particularly because I suspect some of the cleanups don't clean up the folder they create, only the files.

simonwo (approved):

PR #2466, opened 2023-05-12
WIP: Run executors in an 'environment'

rossjones:
Currently executors have reponsibility for executing 'code', but also has responsibility for preparing the environment where it runs. This leads to a situation where each of the executors performs it's own retrieval from storage and it's own mapping onto whatever mechanism it uses to make the data available.  This makes it difficult to have a single location where we can ensure we clean up that environment as StorageProviders are attached directly to the executor.

This commit introduces an `executor.Environment` (cyclic dependencies made `compute.Environment` untenable) which can prepare the storage before providing enough information to the executor to map the data into it's compute-space. This new environment can also ensure that any resources created by the execution can be cleaned up once the process has completed.

Current blockers:

* DiskUsageCalculator depends on the executor being able to get the volume size - this should be the environment's responsibility 
* Scenario tests are ending up with nil StorageProviders and it's not straight-forward to unpick the dep injection alongside the factories to figure out why
*

### DIFF
--- a/pkg/compute/executor.go
+++ b/pkg/compute/executor.go
@@ -10,6 +10,7 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/executor"
 	"github.com/bacalhau-project/bacalhau/pkg/model"
 	"github.com/bacalhau-project/bacalhau/pkg/publisher"
+	"github.com/bacalhau-project/bacalhau/pkg/storage"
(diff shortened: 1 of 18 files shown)


rossjones:
This is going to be too problematic until we refactor storage as well, so this may as well wait for the pluggable work.

PR #2484, opened 2023-05-19
Allow users to specify preferred listen address

rossjones:
When listening for IPFS swarm, or peerings, we listen on 0.0.0.0, which will leave us trying to listen on any/all interfaces available. In various places though we try and choose a single address to use but it isn't always the one we want.

By specifying BACALHAU_PREFERRED_ADDRESS it will pick that address over any others for the IPFS swarm and the peering information.

### DIFF
--- a/pkg/requester/selection/all.go
+++ b/pkg/requester/selection/all.go
@@ -89,16 +89,12 @@ func (s *allNodeSelector) SelectNodesForRetry(ctx context.Context, job *model.Jo
 
 	failedNodes := make([]string, 0, len(jobState.Executions))
 	for nodeID, failureCount := range failureCounts {
-		if failureCount > 0 && failureCount < maxRetriesPerNode {
+		if failureCount > 0 {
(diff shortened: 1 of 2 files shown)


simonwo:
I don't really understand how this interacts with the `--host` flag you mention?

simonwo:
*Should* the flag be doing it? Should we be just using the flag instead of adding a new env var?

rossjones:
> _Should_ the flag be doing it? Should we be just using the flag instead of adding a new env var?

The flag would have to also work for the calculation of the peering, but we might want to have the peering to happen on an internal IP, the API and IPFS swarm on different addresses.  Using --host would only solve the case where a user wants everything on a single public address. 

I added https://github.com/bacalhau-project/bacalhau/issues/2486 as the start of a discussion about having better options around the network config, and hopefully a less scattergun approach to libp2p so we're not doing all of udp, udp+quic1, udp+quic, tcp etc.. particularly given the comment referenced in that issue about concurrency issues.

PR #2485, opened 2023-05-19
2482 output peering information to run

aronchick:


pkg/config/config.go
```diff
@@ -17,24 +17,79 @@ import (
...
+	// } else if _, err := os.Stat("/var/run"); err == nil {
+	// 	writePath = "/var/run" // Older Linux
+	// } else if _, err := os.Stat("/private/var/run"); err == nil {
+	// 	writePath = "/private/var/run" // MacOS
+	// } else {
+	// 	// otherwise write to temp dir, which should be available on all systems
+	// 	writePath = os.TempDir()
+	// }
+
+	RunInfoFilePath = filepath.Join(writePath, RunInfoFilename)
```
rossjones, line 57:
```suggestion
	// TODO: Use the below when we can figure out how to test write permissions in a cross platform way.
	// See this - https://stackoverflow.com/questions/20026320/how-to-tell-if-folder-exists-and-is-writable //nolint:lll
	// if _, err := os.Stat("/run"); err == nil {
	// 	writePath = "/run" // Linux
	// } else if _, err := os.Stat("/var/run"); err == nil {
	// 	writePath = "/var/run" // Older Linux
	// } else if _, err := os.Stat("/private/var/run"); err == nil {
	// 	writePath = "/private/var/run" // MacOS
	// } else {
	// 	// otherwise write to temp dir, which should be available on all systems
	// 	writePath = os.TempDir()
	// }

	RunInfoFilePath = DevstackEnvFile() 
	if RunInfoFilePath == "" {
	    filepath.Join(writePath, RunInfoFilename)
	 }
```

Doing this would mean we can, in dev, override the name of the file being written so that we can get the CLI to auto-load it (https://github.com/bacalhau-project/bacalhau/blob/main/main.go#L26-L30). In prod it would carry on as previously.

pkg/config/config.go
```diff
@@ -17,24 +17,79 @@ import (
 )
 
 const (
-	maxUInt16 uint16 = 0xFFFF
-	minUInt16 uint16 = 0x0000
+	maxUInt16              uint16 = 0xFFFF
+	minUInt16              uint16 = 0x0000
+	RunInfoFilename               = "bacalhauServe.run"
```
rossjones, line 22:
```suggestion
	DefaultRunInfoFilename               = "bacalhauServe.run"
```

pkg/config/config.go
```diff
@@ -12,29 +12,88 @@ import (
...
-	return os.Getenv("DEVSTACK_PRINT_INFO") != ""
+// WriteRunInfoFile writes the ShellVariables to a file that can be imported. It writes bacalhauServe.run to TempDir().
+func WriteRunInfoFile(ctx context.Context, summaryShellVariablesString string) error {
+	writePath := ""
+	if writeable, _ := filefs.IsWritable("/run"); writeable {
+		writePath = "/run" // Linux
+	} else if writeable, _ := filefs.IsWritable("/var/run"); writeable {
+		writePath = "/var/run" // Older Linux
+	} else if writeable, _ := filefs.IsWritable("/private/var/run"); writeable {
+		writePath = "/private/var/run" // MacOS
```
rossjones, line 49: These folders all tend to be owned by `root` so we'll only be able to write to them if bacalhau is run as root.   On Linux we _can_ write to /var/run/user/[uid] if we know the current user's UID.  Otherwise if it is a file intended to be owned by a specific (non-root) user, then it should probably be in `~/username/.bacalhau`

pkg/config/config.go
```diff
@@ -12,29 +12,88 @@ import (
...
+		log.Warn().Msg("Could not write to /run, /var/run, or /private/var/run, writing to temp dir instead." +
+			"This file contains sensitive information, so please ensure it is limited in visibility.")
+		writePath = os.TempDir()
+	}
+
+	if DevstackEnvFile() != "" {
+		RunInfoFilePath = DevstackEnvFile()
+	} else {
+		RunInfoFilePath = filepath.Join(writePath, DefaultRunInfoFilename)
+	}
```
rossjones, line 61:
```suggestion
	RunInfoFilePath = DevstackEnvFile()
	if RunInfoFilePath == "" {
		RunInfoFilePath = filepath.Join(writePath, DefaultRunInfoFilename)
	}
```

pkg/config/config.go
aronchick: agreed - this is what "IsWriteable" checks. Here's a question - do we expect bacalhau to be run as non-root? I'd imagine so - i'm just not sure writing a global service to a user's directory is the best.

pkg/config/config.go
aronchick: agreed

pkg/config/config.go
rossjones: We _should_ be able to run as non-root, but I'm not sure it's the case today. Either way ~/.bacalhau is probably not a useful place for info to be shared. If we default to /tmp/ when we can't write to any of the run directories, is that useful for something else trying to find it?

aronchick: my concern with /tmp is just it's not a very secure place (and information here will allow anyone to join a bacalhau network)

wdbaruni (approved):

PR #2487, opened 2023-05-22
Print errors if no nodes complete the job successfully

simonwo:
Currently we are printing unhelpful "not enough nodes to run job" error messages. This hides the real errors from users.

Now, print a summary of the errors per node if the job fails, and make sure to always display an error if the job is cancelled or rejected.

```
$ bacalhau docker run some-rando-image ls  
Using default tag: latest. Please specify a tag/digest for better reproducibility.
Job successfully submitted. Job ID: e2d7a10c-012c-4abb-a083-f1780890cf98

        Communicating with the network  ................  done ✅  0.1s
           Creating job for submission  ................  done ✅  0.0s
           Job waiting to be scheduled  ................  done ✅  0.0s
                       Job in progress  ................  err  ❌  1.3s

Error running job: not enough nodes to run job. requested: 1, available: 0
Job Results By Node:
• Node QmdZQ7Zb, QmYgxZiy, QmXaXu9N, QmUDAXvv, QmVAb7r2: 
        Could not inspect image "some-rando-image" - could be due to repo/image not existing, or registry needing authorization: Error response from daemon: errors:
        denied: requested access to the resource is denied
        unauthorized: authentication required

        bacalhau describe e2d7a10c-012c-4abb-a083-f1780890cf98
```

### DIFF
--- a/cmd/bacalhau/utils.go
+++ b/cmd/bacalhau/utils.go
@@ -29,6 +29,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"golang.org/x/exp/slices"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


rossjones (approved):
LGTM, minor comment about extra-large top level error message.

simonwo:
Oops, trigger happy on the auto merge.

Thankfully, I think I’m happy with the spewing large errors for now? In the case you mentioned you need nearly the whole error to understand what the issue is. 

In this case we could convert this into an easier to understand error in the executor if we wanted to? So it might be appropriate to err on the side of more verbose errors for now and target the common ones with specific error checks in the executor.

PR #2494, opened 2023-05-24
Record Canary Jobs in Postgres Database

OwainJones:
We want to be recording canary jobs in the database and dashboard, with the option to filter them out further down the line, so removed canary filters, but ensured that their conditions would appear in the database so that they can be indentified at a later date. Only affects new canary jobs.


simonwo (approved):
LGTM. We will have to solve the problem with seeing thousands of canary jobs in the dashboard, but we need to solve that anyway for amplify jobs so I don’t see that as a reason to block this.

PR #2499, opened 2023-05-26
enable mounting writable local paths

wdbaruni:
Allow Bacalhau jobs to have write access to allow-listed paths. A use case where this is helpful is when jobs need to checkpoint their progress so that future invocations/retries can resume instead of starting over, such as when processing local log files or local sensor data.

### DIFF
--- a/pkg/job/parser.go
+++ b/pkg/job/parser.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/bacalhau-project/bacalhau/pkg/clone"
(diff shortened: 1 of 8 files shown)


frrist:
Any chance we could base the changes landing here onto https://github.com/bacalhau-project/bacalhau/pull/2492? #2492 is rewriting the entire storage spec to be specific to the implementation by defined independent structures  for each type of storage rather than continuing to append to the existing storage spec

wdbaruni:
@frrist I see your point. Though this PR introduces are other changes and exposes a new feature, and not only adding a field to the spec. I prefer that https://github.com/bacalhau-project/bacalhau/pull/2492 only has changes related to re-factoring the spec and not bundle this change with it. I can hold off from merging this PR and rebasing it after https://github.com/bacalhau-project/bacalhau/pull/2492 is merged if you think that will happen soon. Keep in mind there is a [demo](https://github.com/bacalhau-project/sample-event-processor) that is blocked on this PR

frrist:
In that case merge it and I'll sort the conflicts in my branch when it's rebased onto master.

wdbaruni:
Thanks. I've just added you as reviewer whenever you have a chance :)

pkg/job/parser.go
```diff
@@ -64,6 +65,24 @@ func ParseStorageString(sourceURI, destinationPath string, options map[string]st
 			StorageSource: model.StorageSourceLocalDirectory,
 			SourcePath:    filepath.Join(parsedURI.Host, parsedURI.Path),
 		}
+		for key, value := range options {
+			switch key {
+			case "ro", "read-only", "read_only", "readonly":
+				readonly, parseErr := strconv.ParseBool(value)
+				if parseErr != nil {
+					return model.StorageSpec{}, fmt.Errorf("failed to parse read-only option: %s", parseErr)
```
frrist, line 73: nit: user `%w` instead of `%s` to wrap the error. https://pkg.go.dev/fmt#Errorf (also below this line)

pkg/model/storage_spec.go
```diff
@@ -28,6 +28,9 @@ type StorageSpec struct {
 	// The path of the host data if we are using local directory paths
 	SourcePath string `json:"SourcePath,omitempty"`
 
+	// Allow write access for locally mounted inputs
+	ReadWrite bool `json:"ReadWrite,omitempty"`
```
frrist, line 32: Which `StorageSourceType` does this field apply to? Could you leave a comment stating this? It will make my rebase of the `StorageSpec` changes easier.

pkg/storage/local_directory/storage_test.go
```diff
@@ -201,20 +220,42 @@ func (s *LocalDirectorySuite) TestPrepareStorage() {
...
+		name      string
+		readWrite bool
+	}{
+		{name: "readonly", readWrite: false},
+		{name: "readWrite", readWrite: true},
+	} {
+		s.Run(tc.name, func() {
+			path := folderPath
+			if tc.readWrite {
+				path += ":rw"
```
frrist, line 236: nit: can the path be a field of the test case?

frrist (approved):
LGTM

PR #2516, opened 2023-06-01
Improve config of docker manifest caching

rossjones:
Adds configuration options for the docker manifest cache, allowing users to specify the maximum size (in items), the frequency of the staleness check and the default time-to-live for each entry.

Also removes the now unused tag cache pending clarity on whether we can implement this in a way that doesn't make the DX worse.


pkg/cache/basic/cache.go
```diff
@@ -8,11 +8,16 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/util/generic"
 )
 
+const (
+	DefaultTTLSeconds = int64(3600)
```
frrist, line 12: Let's move this to the default config structures.

frrist (approved):

pkg/cache/basic/cache.go
rossjones: We need this here in case somebody calls NewCache() without specify the option that sets TTL - so this is just to appease the linter (although it doesn't complain about 1000 a few lines above). I think it'd make sense if ttl was a required parameter though.

PR #2518, opened 2023-06-01
Add a node selection algorithm to target all nodes

simonwo:
Currently our algorithm for selecting and retrying nodes will select a number of nodes from an appropriate set until a user-defined number of successful executions is reached. This is useful for targeting data processing jobs against a wide and disparate network of different nodes, as is the common use case for the public Bacalhau network.

We want to meet another use case, which is for a job to be able to target all nodes that match a given criteria, regardless of how many nodes that is. This is useful in private networks for running jobs against a homogeneous set of nodes (i.e. all nodes with some attached sensor or that are storing the same types of logs).

To achieve this, we split out all of the node targeting logic from the scheduler and create two new modules – one for the "target any nodes" case and one for the "target all nodes" case. The scheduler is then abstracted from the exact logic by which nodes are picked, and just manages the executions and job state accordingly.

This introduces a new field into the job spec called "TargetAll": a boolean that switches between the node selection modes (something more complex may be appropriate in the future that also includes the fields currently in the Deal struct which are only appropriate for the "any" case). The default value in old job specs is "any", matching the old default behaviour. There is currently no way to trigger "all" jobs from the CLI and this will be added in a later commit.


pkg/requester/types.go
```diff
@@ -47,6 +47,23 @@ type NodeRanker interface {
...
+type NodeSelector interface {
+	// SelectNodes returns the nodes that should be used to execute the passed job.
+	SelectNodes(context.Context, *model.Job) ([]model.NodeInfo, error)
+	// SelectNodesForRetry returns the nodes that should be used to retry the
+	// passed failed executions in the context of the passed job. If no nodes
+	// are returned, the executions do not need to be retried just yet.
+	SelectNodesForRetry(context.Context, *model.Job, *model.JobState) ([]model.NodeInfo, error)
+	// SelectBids returns the pending bids on the passed job that should be
+	// accepted or rejected.
+	SelectBids(context.Context, *model.Job, *model.JobState) (accept, reject []model.ExecutionState)
```
frrist, line 60: nit: by virtue of the interface name it is implied these methods select some set of nodes, these methods could be shortened to:
`Nodes`
`NodesForRetry`
`Bids`

pkg/requester/selection/all.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, execution := range jobState.Executions {
+		numFails, seen := failureCounts[execution.NodeID]
+		if !seen {
+			numFails = 0
+		}
+
+		if numFails < 0 || !execution.State.IsDiscarded() {
+			// The job is in progress or completed: don't retry this node again.
+			failureCounts[execution.NodeID] = -1
+		} else if execution.State.IsDiscarded() {
```
frrist, line 85:
```go
func (s ExecutionStateType) IsDiscarded() bool {
	return s == ExecutionStateAskForBidRejected || s == ExecutionStateBidRejected || s == ExecutionStateResultRejected ||
		s == ExecutionStateCanceled || s == ExecutionStateFailed
}
```
`IsDiscarded` could mean many things. Would it be better to be explicit about the states we're expecting, or not expecting? Although I assume `*BidRejected*` states will never make it in here; would we still retry if the result was rejected?

pkg/requester/selection/all.go
```diff
@@ -0,0 +1,139 @@
...
+	// failedNodes counts the number of failures that the node has seen so far.
+	// A negative value means the job is successful or in progress.
+	failureCounts := make(map[string]int, len(jobState.Executions))
+	for _, execution := range jobState.Executions {
+		numFails, seen := failureCounts[execution.NodeID]
+		if !seen {
+			numFails = 0
+		}
+
+		if numFails < 0 || !execution.State.IsDiscarded() {
```
frrist, line 82: Would `IsActive` be more appropriate than `!IsDiscarded`? `IsActive` also includes the `ExecutionStateCompleted` state.

pkg/requester/selection/all.go
```diff
@@ -0,0 +1,139 @@
...
+			failedNodes = append(failedNodes, nodeID)
+		}
+	}
+
+	log.Ctx(ctx).Debug().Int("Retryable", len(failedNodes)).Msg("Found nodes to retry")
+	if len(failedNodes) == 0 {
+		return nil, nil
+	}
+
+	foundNodes, err := s.nodeDiscoverer.FindNodes(ctx, *job)
```
frrist, line 103: The `jobState` contains an `ExecutoState` which contains the nodeID for the node running the execution, do we need to re-discover the nodes here? Or could we reuse the value from the execution state?

pkg/requester/selection/any.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	enoughExecutions := jobState.NonDiscardedCount() >= minExecutions
+	if enoughExecutions {
+		// There are enough running executions, so no action to take right now.
+		return []model.NodeInfo{}, nil
+	}
+
+	desiredNodeCount := minExecutions - jobState.NonDiscardedCount()
+	return s.Select(ctx, job, desiredNodeCount, desiredNodeCount)
```
frrist, line 103: Similar to my comment in the "all selector", do we need to re-discover nodes here? (`Select` will rerun node discovery)

pkg/requester/selection/any.go
```diff
@@ -0,0 +1,129 @@
...
+	return executionsWaiting[:requiredNewExecutions], executionsWaiting[requiredNewExecutions:]
+}
+
+// We can mark a job as partially complete if we have a single complete
+// execution, and fully complete once we are over the user's required confidence.
+func (*anyNodeSelector) CanCompleteJob(ctx context.Context, job *model.Job, jobState *model.JobState) (bool, model.JobStateType) {
+	completedCount := jobState.CompletedCount()
+	if completedCount < 1 {
+		return false, jobState.State
+	} else if completedCount < job.Spec.Deal.GetConfidence() {
```
frrist, line 124: Say we have a confidence of 4, 2 successful executions, and 2 failed executions (`ExecutionStateFailed`). I am not sure it's correct to say the execution was partially completed since `JobStateCompletedPartially` means some nodes have failed to publish their results when that may not be the case here - executions can fail for other reason. Do you think there is an opportunity to be more granular here with the returned `JobStateType`?

This comment also applies to the implementation in `allNodeSelector`.

I may also be way off in the weeds here as it could be the case that this method is never called in the circumstances I have described above.

frrist (commented):
Good stuff! I would like to get some clarification on my comments before approving for merge.

pkg/model/job_state.go
```diff
@@ -2,6 +2,8 @@ package model
 
 import (
 	"time"
+
+	"github.com/samber/lo"
```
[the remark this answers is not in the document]
rossjones: Was already an indirect dependency so not really a new addition.

pkg/requester/selection/all.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	log.Ctx(ctx).Debug().Int("Discovered", len(nodeIDs)).Msg("Found nodes for job")
+
+	rankedNodes, err := s.nodeRanker.RankNodes(ctx, *job, nodeIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	// filter nodes with rank below 0
+	filteredNodes := lo.Filter(rankedNodes, func(node requester.NodeRank, _ int) bool { return node.Rank >= 0 })
```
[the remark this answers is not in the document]
simonwo: Good catch – the comment is wrong, it should read non-negative (to include zero). The logic is <0 doesn't meet requirements, >0 does, with higher numbers being "preferred".

Actually, this is a good place for a more semantic way of talking about ranks than using magic numbers. I will push a new commit to factor out the magic numbers into some constants and provide a method to talk about which nodes "meet requirements".

pkg/requester/selection/all.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, err
+	}
+	log.Ctx(ctx).Debug().Int("Discovered", len(foundNodes)).Msg("Found nodes for job")
+
+	retryNodes := make([]model.NodeInfo, 0, len(failedNodes))
+	for _, foundNode := range foundNodes {
+		if slices.Contains(failedNodes, string(foundNode.PeerInfo.ID)) {
+			retryNodes = append(retryNodes, foundNode)
+		}
+	}
```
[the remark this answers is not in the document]
simonwo: The issue is that we're comparing on the ID field of the struct, and those methods only compare via struct equality, i.e. there is no `IntersectBy`.

pkg/requester/selection/all.go
simonwo: I've replaced both calls to `IsDiscarded` with one to `IsActive` and a catch-all `else`, which is a bit more sane anyway. But your point still stands in that there are some subtleties here:

* Should the job be retried if the compute node rejects the bid? I say yes - it might be because the node was overloaded in that moment but could run the job later. (There is a back-off algorithm that we have in `lilypad` that would be useful to add to our retry handling here.)
* We shouldn't ever reject a bid in this algorithm because we want all nodes to run the job, so yes, `ExecutionStateBidRejected` is redundant for jobs of this type.
* Should the job be retried if the result is rejected? That would mean the verification stage has failed. I don't think there's any point retrying the job here? The node has executed the code and come back with some unsuitable result – not clear why executiing it again would produce something suitable.
* Should the job be retried if the execution was cancelled? Presumably not?
* Should the job be retried if the execution failed? Presumably yes.

pkg/requester/selection/all.go
simonwo: Having this here is for two reasons:

1. Our interface expects a `model.NodeInfo` and this is the only way to get one. However… this interface is only consumed in one place and it only ever uses the node ID. So if we wanted to, we could refactor this to only talk about node IDs. Worth it?
2. This call means that if we need to retry a node but the node has disappeared from the network, we have a chance to throw an error (below). Do we need to do this? Atm without this the job would just sit in `CompletedPartially` if the node never comes back. Is that desirable? This is a more broad question, because atm we'll also end up in that state if the node fails more than the retry limit – it won't be retried and we won't throw an error.

pkg/requester/selection/all.go
simonwo: I'm refactoring this.

pkg/requester/selection/any.go
simonwo: In this case it's useful, because between the job being orchestrated and this moment, new nodes that might be able to run the job may have joined the network. So it makes sense to re-orcehstrate this execution with as current a view as possible of what nodes are available.

pkg/requester/types.go
simonwo: Agreed... I think the `Selector` part of the name is the part that I am not happy with. It now seems to be doing more than just "selecting". It will:

* Decide whether there are enough nodes to run the job, and fail it if not
* Select the nodes that should run the job
* Decide whether we are ready to execute the job (i.e. did we get enough bids)
* Select the bids that should be accepted
* Decide whether the job is complete enough to report success
* Select which executions should be retried, and the nodes to retry them on

So selection is a big part of it, but now it is also encoding the "logic of the deal", i.e. ensuring that the "deal" that the user requested is fulfilled or failing if not. I almost want to call it a very whimsical `Dealmaker` but I don't think that is an improvement in clarity!

simonwo:
@rossjones @frrist on naming:

I think the Selector part of the name is the part that I am not happy with. It now seems to be doing more than just "selecting". It will:

- Decide whether there are enough nodes to run the job, and fail it if not
- Select the nodes that should run the job
- Decide whether we are ready to execute the job (i.e. did we get enough bids)
- Select the bids that should be accepted
- Decide whether the job is complete enough to report success
- Select which executions should be retried, and the nodes to retry them on

So selection is a big part of it, but now it is also encoding the "logic of the deal", i.e. ensuring that the "deal" that the user requested is fulfilled or failing if not.

There is something here relating to what @rossjones said above: 
> it seems like we could allow a Deal implementation to provide a NodeSelector?

Perhaps we are just seeing that we need a name that reflects this? I almost want to call it a very whimsical `Dealmaker` but I don't think that is an improvement in clarity!

pkg/model/job.go
```diff
@@ -91,6 +91,9 @@ type JobWithInfo struct {
 // The deal the client has made with the bacalhau network.
 // This is updateable by the client who submitted the job
 type Deal struct {
+	// Whether the job should be run on any matching node (false) or all
+	// matching nodes (true). If true, other fields in this struct are ignored.
+	TargetAll bool `json:"TargetAll,omitempty"`
```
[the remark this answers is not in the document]
simonwo: So I didn't do the full blown refactor for some reasons:

1. This is a backwards compatible change that comes "for free" – all existing job specs automatically get interpreted correctly. So I didn't want to introduce a compat breakage unless strictly necessary.
2. I can't actually think of any more implementations! It does seem like there _might_ be more, but I also can't see us implementing any others soon. (I think `TargetMin` is actually a better name for our `any` strategy, as it is targeting a minimum number of nodes set by the `Concurrency`). 
3. Is this all about to change anyway when we move to a more modular API? My understanding of that world is that we might have some sort of `TargetMinJob` or `TargetAllJob` which would be submitted to a specific module. So doing the work to make the spec dynamically extendable might be a waste of time if we are about to refactor things?

WDYT?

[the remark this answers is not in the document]
simonwo: So I didn't do the full blown refactor for some reasons:

- This is a backwards compatible change that comes "for free" – all existing job specs automatically get interpreted correctly. So I didn't want to introduce a compat breakage unless strictly necessary.
- I can't actually think of any more implementations! It does seem like there might be more, but I also can't see us implementing any others soon. (I think `TargetMin` is actually a better name for our `any` strategy, as it is targeting a minimum number of nodes set by the `Concurrency`).
- Is this all about to change anyway when we move to a more modular API? My understanding of that world is that we might have some sort of `TargetMinJob` or `TargetAllJob` which would be submitted to a specific module. So doing the work to make the spec dynamically extendable might be a waste of time if we are about to refactor things? cc @wdbaruni 

WDYT?

pkg/requester/selection/switch.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/bacalhau-project/bacalhau/pkg/requester"
+)
+
+type switcher struct {
+	selectors map[bool]requester.NodeSelector
+}
+
+// Returns a node selector that switches between node selectors dependent on the
+// targeting type defined in the job.
+func NewNodeSelectorSwitch(forAny, forAll requester.NodeSelector) requester.NodeSelector {
```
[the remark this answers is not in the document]
simonwo: Do love this idea, and doing it in the same vein as the executors where the `Deal` is actually just a CID to a schema that describes the deal, pluggable deals basically. And if we do the work to put WASM in the executor spec (e.g. to allow validation in the CLI of pluggable executors) then this would be another place we could use that feature.

pkg/requester/selection/any.go
simonwo: I think the decision to be made here is what to do if some/any nodes fail to execute (after retries). Do we:

1. Mark the whole job as `Failed`? After all, the user said run it on all nodes and we couldn't. This is what we do today when the user says "run it on X nodes" and we only manage <X.
2. Mark the whole job as "partially completed" (using some state? could update `CompletedPartially` to be broader than just "results failed to publish" or introduce something new)

rossjones:
> Or is `Deal` the wrong word?

I think `Deal` is the wrong word, but that may be because I think of the verb before I think of the noun.
It feels more like an `Agreement` or, dare I suggest, a `Contract`, but neither of these feel like the correct answer either because we're really asking for the world to look like whatever is in this object. Agreement and Contract (and Deal) weight far too far to being about the agreement aspect rather than it being a requested blueprint for how the computation is laid out.  Maybe we have to stick with `Deal` :(

pkg/model/job.go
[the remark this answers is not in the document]
frrist: > This is a backwards compatible change that comes "for free" – all existing job specs automatically get interpreted correctly. So I didn't want to introduce a compat breakage unless strictly necessary.

It's free in the sense this change is compatible with the dashboard and previous Bacalhau versions. But I do want to point out that we call [this version of the spec](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/model/apiversion.go#L18) the same version as [v1beta1](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/model/v1beta1/apiversion.go#L21) despite being different in content, and at some point, we will need to pay that debt down, and I view this change as adding to that debt. 
I prefer Ross's suggestion to make the `Deal` dynamic to it's content.

> I can't actually think of any more implementations!

As we discussed out of band, we'll want to remove `Confidence` from `Deal` at some point as it has more to do with verification than the actual `Deal`. That _will_ be a breaking change as it removes a field from the structure.
Another change could be the porposal [here (wasm)](https://github.com/bacalhau-project/bacalhau/pull/2518#discussion_r1213557715), although I acknowledge that is probably a far-off thing.

> Is this all about to change anyway when we move to a more modular API?

My sense is no, but I will wait for @wdbaruni to clarify. 
Generally, I would expect any API that accepts a `Deal` to deserialize it according to its expected usage - APIs would accept an opaque `Deal` to satisfy some interface and interpret its content internally.

pkg/requester/selection/all.go
frrist: Looking at the `FindNode` implementations, the operations look relatively cheap, so I'd be fine just leaving this as it is.

pkg/requester/selection/any.go
frrist: I'd prefer option one as it keeps behavior inline with the existing expectations.

frrist (approved):
My comments aside, the change looks good. We can implement the feedback in a follow on if/when it becomes necessary.

PR #2546, opened 2023-06-05
Fix up the runfile writing to write to config dir as fallback

rossjones:
When the run file is not able to be written to /run (or similar) it currently writes to the current users home directory - which is probably correct, but we also have the bacalhau config directory (~/.bacalhau most of the time) which we will use instead.

Also checks early for the devstack_env_file variable to short circuit the rest of the checks in dev mode


rossjones:
Not sure this is mergeable with recent config changes.  Doesn't seem to have been as much of an issue as expected.

Issue #2535, opened 2023-06-02
Remove v1alpha and v1beta model types


frrist:
### Background
v1alpha and v1beta models are legacy model specifications. Their primary purpose is to interact with the dashboard: The backing database used by the dashboard contains v1alpha, v1beta, and "current" model specifications. v1alpha and v1beta exist to facilitate inspecting their respective types from the database using go code in the Bacalhau project. A secondary usage is allowing users to specify jobs to Bacalhau using legacy job specification files, although there is low confidence of this actually working in practice as it's currently untested.

Additionally, the Dashboard API still operates using the v1beta schema, and we implicitly rely on JSON marshallers to handle the transformation of "current" models to v1beta for read/write operations on the database. Changes to the "current" model specifications must result in the "current" model schema being a superset of the fields of the v1beta schema. As an example, the `Publisher` field is marked as [deprecated in the current job spec](https://github.com/bacalhau-project/bacalhau/blob/6d02cb64959a541c445c9fddc9b4eb14978190e5/pkg/model/job.go#L157 ) but remains in the spec (and is occasionally populate) to maintain the implicit JSON marshaling outlined above.

### Action
It will likely be the case that this work must be completed in tandem with the work described in [#2536](https://github.com/bacalhau-project/bacalhau/issues/2536)

Issue #2536, opened 2023-06-02
Update Dashbaord to latest model implementation


frrist:
### Background
As outlined in [#2535](https://github.com/bacalhau-project/bacalhau/issues/2535), the Dashboard still relies on the v1beta model specification.

### Action
Update the dashboard type system to the latest version of the model specification

simonwo:
What is the plan for the existing data in the dashboard database that is currently in the legacy specifications?

frrist:
There is undoubtedly value in keeping it. We could export it to CSV and persist it somewhere (estuary, s3, r2, w/e) for historical reference. But the effort to keep the dashboard working as it does today (its queries and ingestion) with the legacy data and new formats we are moving towards will burden development as the spec continues to evolve.

How does this sound:
1. Export all existing data to CSV, and pin it on a storage backend.
2. Hibernate the dashboard, and possibly truncate it's database (would save $$$)
3. If we want to query the historical data use bacalhau + duckdb.
4. Later, awaken the dashboard, and at that point decide if we want to import historical data.

frrist:
To mediate the loss of metrics we'll face from dashboard hibernation we can move almost all of its (metric related) functionality to Grafana. This will require implementing some new metrics in the Requester node, but that's something we intended to do anyways - so a win-win in my book.

PR #2550, opened 2023-06-06
Deprecate v1alpha1 and v1beta1 models, correct API version, & hiberante dashboard

frrist:
**Note: this change targets `dev1`; subsequent breaking changes will continue to target `dev1`.**
### What
- Remove all references to `v1alpha1` and `v1beta1` job specifications.
- Assign a new (and correct) version to the current model specification `v1beta2`.
- Hibernate the dashboard (due to breakages from `v1alpha1` and `v1beta1` removal)
  - https://github.com/bacalhau-project/bacalhau/issues/2536 tracks re-adding support for dashboard

Closes https://github.com/bacalhau-project/bacalhau/issues/2535 & #2551


rossjones:
-18,683 ... nice. I love a good tidy up. But I have concerns around hibernating the dashboard on a branch that is intended to be long lived.  Unless we've already agreed this with product peeps, we're essentially committing to refactoring the dashboard into a workable state before we ever merge in dev1. Is that intentional?

pkg/model/apiversion.go
```diff
@@ -9,8 +9,7 @@ type APIVersion int
 
 const (
 	apiVersionUnknown APIVersion = iota // must be first
-	V1alpha1
-	V1beta1
+	V0
```
frrist, line 12: @wdbaruni Are we okay calling this version of the spec `V0`?

wdbaruni: better to use `V1gamma`. V0 indicates it is the very first version but in this case it isn't

frrist:
@rossjones inlight of @wdbaruni's design on a module API and plans for backwards compatibility this will now go into `main`

frrist:
For some reason, CI is failing to install the precommit hook; once I get this sorted, I'll get this merged.

frrist:
@rossjones and @simonwo this is a relatively large change, so when you have time I'd like to get a review and ✅ from both of you before merging.

rossjones (approved):
I'm not sure where we got to with the discussion about insulated jobs, but removing the old models looks good!

Issue #2538, opened 2023-06-03
Remove Filecoin and Lotus Storage types


frrist:
### Background
The Filecoin storage type's maintenance burden is not worth the value. Especially considering the Estuary and IPFS storage specifications can be used to achieve a similar result.

### Action
Remove the Filecoin Storage spec and related test fixtures. The Filecoin Storage Spec mey be re-added later via a pluggable storage spec.

Issue #1705, opened 2023-01-19
TestLotusNodeSuite/TestLotusNode is flaky


simonwo:
Test often times out after 10 mins, see e.g. https://app.circleci.com/pipelines/github/filecoin-project/bacalhau/3064/workflows/ba830836-1de4-41a4-a622-d45b8157691e/jobs/30586

wjam:
I think one of the causes of this issue is that each block must finish mining within 4 seconds, for the `localnet`, to be considered valid but it's possible for the CI instance to be too overloaded to mine fast enough.

Another issue I've spotted is that it takes roughly 4 minutes to perform this call https://github.com/filecoin-project/bacalhau/blob/430bbb06bebc6614bcdd3ad956157c23dc6b709b/pkg/publisher/filecoin_lotus/publisher.go#L295

philwinder:
I've seen this too: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1674649121250959

wjam:
The test has been marked as skip as it's been failing too much recently

https://github.com/filecoin-project/bacalhau/pull/1898

frrist:
We can close this when [#2538](https://github.com/bacalhau-project/bacalhau/issues/2538) is addressed as we are removing this storage option.

Issue #2555, opened 2023-06-06
Fix NewIPFSPublishers


frrist:
The method is named incorrectly, and it makes all publishers, such as S3 and Noop, available 
Its registering an IPFS publisher as an estuary publisher, resulting in nodes advertising estuary is installed when its not (its just IPFS

simonwo:
The reason for this was/is that:
* Estuary was/is our preferred publishing platform, because data there gets written into Filecoin = good
* We want as many jobs to use this as possible, so the default on the CLI is today still Estuary
* But we don't want jobs to fail if Estuary isn't installed, and we just want a fallback to IPFS – hence we make sure that the Estuary publisher is always installed and just uses IPFS if necessary.

Not ideal, but that's the reasoning.

Is there a case for even having the Estuary type separate from IPFS? Could we just make the IPFS publisher a fanout publisher that sends to both IPFS and Estuary, if installed?

aronchick:
Still needed?

Issue #2543, opened 2023-06-03
Implement Storage Specification with IPLD Schema


frrist:
For now a place holder. Relate context in #2541.

frrist:
were not doing this with IPLD, closing.

PR #2556, opened 2023-06-06
remove lotus filecoin storage and simplify publishers

frrist:
- closes https://github.com/bacalhau-project/bacalhau/issues/2538 and closes https://github.com/bacalhau-project/bacalhau/issues/1705
- Wrote https://github.com/bacalhau-project/bacalhau/issues/2555 while implementing. This is a fix for a follow on.


pkg/model/publisher.go
```diff
@@ -11,7 +11,6 @@ const (
 	publisherUnknown Publisher = iota // must be first
 	PublisherNoop
 	PublisherIpfs
-	PublisherFilecoin
```
simonwo, line 14: Removing this means that people who have in the past submitted Estuary or S3 jobs now have data that is interpreted incorrectly, e.g. previously Estuary was 4 but now S3 has that value, so places where the numbers are used (existing data) will be wrong. Can we be sure that noone is using these numbers? If not, can we just use `PublisherEstuary = 4` and make sure that 3 isn't used?

pkg/model/schemas/bacalhau.ipldsch
```diff
@@ -36,7 +36,6 @@ type Publisher enum {
     | Noop ("1")
     | IPFS ("2")
     | Estuary ("3")
-    | FilecoinLotus ("4")
 } representation int
```
simonwo, line 39: Lol this was wrong? Can we correct the value for Estuary here?

pkg/model/publisher.go
frrist: In a very similar vein to https://github.com/bacalhau-project/bacalhau/pull/2552#discussion_r1224572846: These will go away when https://github.com/bacalhau-project/bacalhau/issues/2543 lands. 

The pattern of using enums/constants for these values comes at the cost of them being hard to change later, the work to complete #2543 will replace these with CID's backed by a schema similar to how https://github.com/bacalhau-project/bacalhau/issues/2542 will replace engine enums with CIDs.

wdbaruni: This shouldn't be a problem as we always marshal and unmarshal publisher types to their string value before sending them around

wdbaruni (approved):

PR #2558, opened 2023-06-07
Pin the curl docker image used in tests

rossjones:
We currently don't specify a tag, and we recently got a broken image. This pins us to 8.1.0 which we know works.


.circleci/config.yml
```diff
@@ -74,15 +74,15 @@ commands:
 
       - restore_cache:
           name: Restoring Go cache
-          key: go-mod-{{ arch }}-{{ checksum "go.sum" }}
+          key: go-mod-v1-{{ arch }}-{{ checksum "go.sum" }}
```
frrist, line 77: What do these changes do?

frrist (approved):

.circleci/config.yml
rossjones: Because you can't delete caches in circle you have to update it - and in theory the checksum of go.sum should be enough to differentiate, but I was just forcing the issue in case there was something wrong with go.sum

PR #2559, opened 2023-06-07
Exploratory Sqlite/GORM based object store

rossjones:
An exploration on how we can provide an object store using sqlite/gorm as the persistence/query layers.  Even without taking advantage of the relational nature of sqlite we encounter problems that make it difficult to meet the brief for our object store.

* If we attempt to build a kv store in sqlite, then we're not really taking advantage of sqlite's best features.

* It's necessary to build DTOs (Data Transfer Objects) for every entity we want to store in the database to provide the timestamps. The alternative is to mark up our current structures with orm specific tags and embedded an ORM provided structure.  This is likely to get unwieldy very quickly.

* It's hard to provide a deep module with a narrow interface, instead the use of the ORM leads towards a very wide interface that is reasonably shallow with us barely obscuring the use of sqlite/gorm.

* Working with GORM is error prone as it requires pointers everywhere when `interface {}` is used in every function, so passing `Product{}` instead of `&Product{}` into a function will compile and panic at runtime.

It might be possible to build a Repository interface on top of the sqlite/gorm so that it could be bound to each type to improve the interface, but by the time we implement this on top of an ORM on top of Sqlite it feels like we've strayed quite a way from the simplicity of an object stores typical Put/Get interface.


pkg/objectstore/options.go
```diff
@@ -0,0 +1,9 @@
+package objectstore
+
+type Option func(*DBConfig)
+
+func WithLocation(loc string) Option {
```
frrist, line 5: ❤️

pkg/objectstore/objectstore.go
```diff
@@ -0,0 +1,100 @@
...
+	registeredTypes []string
+}
+
+type DBConfig struct {
+	Location string
+}
+
+func NewObjectStore(options ...Option) *DB {
+	config := DBConfig{
+		Location: ":memory:",
```
frrist, line 26: You maybe want to modify this to `:memory:?busy_timeout=5000`:
> busy_timeout=5000 sets a busy handler that sleeps for a specified amount of time when a table is locked.
The handler will sleep multiple times until at least "ms" milliseconds of sleeping have accumulated.
After at least "ms" milliseconds of sleeping, the handler returns 0 which causes sqlite3_step() to return SQLITE_BUSY
We do this to prevent the error "database is locked" from being returned on parallel reads and writes.

pkg/objectstore/objectstore.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	for _, opt := range options {
+		opt(&config)
+	}
+
+	return &DB{
+		location:        config.Location,
+		registeredTypes: make([]string, 8), //nolint:gomnd
```
frrist, line 35: Should this be `make([]string, 0, 8)`? Else the append statement in `Connect` is going to append new values to a slice of 8 empty strings.

frrist (commented):
just some drive by comments

rossjones:
I don't think we're going to follow this path due to some of the complexity of managing type wrappers and trying to decompose object graphs into their component entities.

PR #2560, opened 2023-06-07
refactor CLI to align with design guidelines of cobra

frrist:
A refactor of the CLI to align with the principles described here: https://github.com/spf13/cobra/blob/main/user_guide.md#organizing-subcommands


cmd/util/handler/logging.go
```diff
@@ -7,70 +7,31 @@ import (
...
 
-	return logsCmd
-}
+const (
+	JSONFormat string = "json"
+	YAMLFormat string = "yaml"
+)
 
-func logs(cmd *cobra.Command, cmdArgs []string, options LogCommandOptions) error {
+func Logs(cmd *cobra.Command, jobID string, follow, history bool) error {
```
simonwo, line 26: Even though this is the word "logs", `util` is not the right place for this – this is a stand-alone command for collecting Bacalhau logs, not utils.

pkg/job/factory.go
```diff
@@ -1,22 +1,191 @@
...
+	spec, err := MakeSpec(append(opts, WithWasmEngine(entryModule, entrypoint, parameters, envvar, importModules))...)
+	if err != nil {
+		return model.Spec{}, err
+	}
+	return spec, nil
+}
+
+// TODO(forrest): find a home
+const DefaultTimeout = 30 * time.Minute
+
```
simonwo, line 154: TODO?

cmd/util/handler/logging.go
frrist: While it is a command, it's also used to print the execution of a job (via `PrintJobExecution`) if the user passes the `--follow` flag to `create`, `docker run`, or `wasm run`: https://github.com/bacalhau-project/bacalhau/pull/2560/files#diff-34e9172ae6c1602e3ad866de40c1ab94e28de0040e0ec71fbc9effcab8ff47b1R93 (printer.go line 93)

pkg/job/factory.go
```diff
@@ -1,22 +1,191 @@
...
+	return *spec, nil
+}
+
+/*
 // these are util methods for the CLI
 // to pass in the collection of CLI args as strings
 // and have a Job struct returned
 func ConstructDockerJob( //nolint:funlen
 	ctx context.Context,
```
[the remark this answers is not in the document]
frrist: Whoops, I've replaced this method with the builder pattern in this file. Will remove the comment.

cmd/util/handler/logging.go
rossjones: Those flags are implemented by passing control to the top level `logs` command :D

rossjones:
Without the deletions, could this be merged into `main` rather than another branch?  With it moving a lot of files about I'm going to be nervous about making changes to any of these 90+ files for fear of changes getting lost in a few months time when all the branches get merged into each other.

frrist:
@rossjones (and all) I have rebased this onto main as a follow on to https://github.com/bacalhau-project/bacalhau/pull/2552

cmd/cli/cancel/cancel.go
```diff
@@ -55,9 +58,13 @@ func newCancelCmd() *cobra.Command {
...
-		PreRun:  applyPorcelainLogLevel,
+		PreRun:  handler.ApplyPorcelainLogLevel,
 		RunE: func(cmd *cobra.Command, cmdArgs []string) error {
-			return cancel(cmd, cmdArgs, cancelOptions)
+			err, exitcode := cancel(cmd, cmdArgs, cancelOptions)
+			if err != nil {
+				handler.Fatal(cmd, err, exitcode)
+			}
+			return nil
 		},
```
wdbaruni, line 68: since we are not returning errors, should we use `Run` instead of `RunE`?

cmd/cli/docker/docker_run.go
```diff
@@ -0,0 +1,279 @@
...
+
+const (
+	DefaultDockerRunWaitSeconds = 600
+)
+
+func NewDockerRunOptions() *DockerRunOptions {
+	return &DockerRunOptions{
+		SpecSettings: flags.SpecFlagSettings{
+			Verifier: model.VerifierNoop.String(),
+			// TODO most users would probably prefer IPFS
```
wdbaruni, line 72: since we are retuning default values, that are also expected to be common between docker and wasm (mostly), does it make sense to move this logic to `flags` package and have `flags.NewSpecFlagSettings()` return default settings? The same goes for the other settings

cmd/cli/docker/docker_run.go
```diff
@@ -0,0 +1,279 @@
...
+func newDockerRunCmd() *cobra.Command { //nolint:funlen
+	opts := NewDockerRunOptions()
+
+	dockerRunCmd := &cobra.Command{
+		Use:     "run [flags] IMAGE[:TAG|@DIGEST] [COMMAND] [ARG...]",
+		Short:   "Run a docker job on the network",
+		Long:    runLong,
+		Example: runExample,
+		Args:    cobra.MinimumNArgs(1),
+		//PreRun:  bacalhau.ApplyPorcelainLogLevel,
```
wdbaruni, line 145: is this intentional?

cmd/cli/list/list.go
```diff
@@ -263,3 +270,35 @@ func summarizeJob(j *model.JobWithInfo, OL *ListOptions) (table.Row, error) {
...
+	}
+
+	if len(st) < DefaultShortenStringLength {
+		return st
+	}
+
+	return st[:20] + "..."
+}
+
+func ShortID(outputWide bool, id string) string {
```
wdbaruni, line 296: do you need to expose this method outside of list package?

cmd/cli/root.go
```diff
@@ -8,24 +8,36 @@ import (
...
+	"github.com/bacalhau-project/bacalhau/cmd/cli/devstack"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/docker"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/get"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/id"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/list"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/logs"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/serve"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/simulate"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/validate"
+	version2 "github.com/bacalhau-project/bacalhau/cmd/cli/version"
```
wdbaruni, line 28: why `version2`?

cmd/cli/serve/serve.go
```diff
@@ -427,10 +432,10 @@ func serve(cmd *cobra.Command, OS *ServeOptions) error {
...
 
 		shellVariablesString := fmt.Sprintf(`
-export BACALHAU_IPFS_SWARM_ADDRESSES=%s
-export BACALHAU_API_HOST=%s
-export BACALHAU_API_PORT=%d
-export BACALHAU_PEER_CONNECT=%s`, ipfsSwarmAddress, OS.HostAddress, apiPort, peerAddress)
+			export BACALHAU_IPFS_SWARM_ADDRESSES=%s
+			export BACALHAU_API_HOST=%s
+			export BACALHAU_API_PORT=%d
+			export BACALHAU_PEER_CONNECT=%s`, ipfsSwarmAddress, OS.HostAddress, handler.GetAPIPort(ctx), peerAddress)
```
wdbaruni, line 438: this will add empty spaces to wherever we are printing these variables

cmd/cli/wasm/wasm_run.go
```diff
@@ -0,0 +1,402 @@
...
+		`The name of the WASM function in the entry module to call. This should be a zero-parameter zero-result function that
+		will execute the job.`,
+	)
+	// removed this for commonality with docker
+	/*
+		wasmRunCmd.PersistentFlags().VarP(
+			flags.EnvVarMapFlag(&ODR.Job.Spec.Wasm.EnvironmentVariables), "env", "e",
+			`The environment variables to supply to the job (e.g. --env FOO=bar --env BAR=baz)`,
+		)
+	*/
```
wdbaruni, line 178: are you planning to remove this?

cmd/util/handler/logging.go
wdbaruni: this is confusing to me as well. Does it make sense to have this part of `logs` command, and have `create`, `docker run` and `wasm run` depend on `logs`? Otherwise, a comment on why this method exists here would be helpful

cmd/cli/cancel/cancel.go
frrist: hum yeah good point, since we are calling `Fatal` from the root if there is an error it would probably be best to modify all command functions (`cancel`, `create`, `logs`, etc.) only return an error and handle the exit codes in a single place - the root command. wdyt?

cmd/cli/cancel/cancel.go
```diff
@@ -89,48 +96,45 @@ func cancel(cmd *cobra.Command, cmdArgs []string, options *CancelOptions) error
...
-			return err
+			return fmt.Errorf("unknown error reading from file: %s", err), handler.ExitError
 		}
 		requestedJobID = string(byteResult)
 	}
 
 	// Let the user know we are initiating the request
 	spinner.NextStep(connectingMessage)
-	apiClient := GetAPIClient()
+	apiClient := handler.GetAPIClient(ctx)
```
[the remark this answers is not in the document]
frrist: Open to suggestions on better names for it, it's become a a bit of a dumping ground for things I don't know where to place. Alternatively, we could remove the `handler` package and place all of these files in `util/`. wdyt?

cmd/cli/list/list.go
frrist: Yes its used in list_test.go which is a separate package named `list_test`.

cmd/cli/root.go
frrist: my IDE did it when moving files around, will remove the alias so this is imported as `version`.

cmd/cli/serve/serve.go
frrist: addressed.

cmd/cli/wasm/wasm_run.go
frrist: yes, removed the comment.

cmd/cli/docker/docker_run.go
frrist: Good idea, will do this for other common flags as well.

cmd/cli/cancel/cancel.go
wdbaruni: Makes sense. What happens if we return an error here? Why do need to handle the errors in root?

cmd/cli/cancel/cancel.go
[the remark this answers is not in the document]
wdbaruni: util or helper can be good indicator for what the package does. We can break things out of it as it makes sense in the future similar to what you did with printer pkg

cmd/util/handler/logging.go
frrist: What would make this less confusing? This is the dependency graph:
```
 +-----+    +-----+    +-----+
 |docker|    |wasm|    |create|
 +-----+    +-----+    +-----+
      \       |       /
       \      |      /
              ↓
        +-----------------+
        |PrintJobExecution|
        +-----------------+
              ↓
        +-----+        +-----+
        | log |  <---- | Log |
        +-----+        +-----+
```

wdbaruni (approved):

cmd/cli/cancel/cancel.go
frrist: So I thought it made sense too, but I was wrong, went with your suggestion in https://github.com/bacalhau-project/bacalhau/pull/2560/commits/20e66c7409e9230033a7c9b785aa5c5d0c79f5cc

cmd/cli/cancel/cancel.go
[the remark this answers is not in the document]
frrist: moved everything to `util` the parent package of `handler`.

cmd/util/exitcodes.go
```diff
@@ -0,0 +1,21 @@
+package util
+
+const (
+	// ExitSuccess Successful termination
+	ExitSuccess = 0
+
+	// ExitError Catchall for general errors
+	ExitError = 1
+
+	// ExitBuiltinMisuse Misuse of shell builtins (according to Bash documentation)
```
rossjones, line 10: This is a bash-ism, there are some POSIX defined error codes I believe but it is otherwise surprisingly non-standard.  The best we can do for now is probably using the Linux error codes that should be listed in the man page for errno `man errno` (they are in osx)

cmd/util/fatal.go
```diff
@@ -0,0 +1,21 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+var Fatal = fatalError
+
+func fatalError(cmd *cobra.Command, err error, code int) {
+	if msg := err.Error(); msg != "" {
+		// add newline if needed
```
rossjones, line 14: We can probably skip this comment.

cmd/util/parse/parse.go
```diff
@@ -0,0 +1,91 @@
...
+	if len(unSafeAnnotations) > 0 {
+		log.Ctx(ctx).Error().Msgf("The following labels are unsafe. Labels must fit the regex '/%s/' (and all emjois): %+v",
+			RegexString,
+			strings.Join(unSafeAnnotations, ", "))
+		/*
+			return nil, fmt.Errorf("the following labels are unsafe. Labels must fit the regex '/%s/' (and all emjois): %+v",
+				RegexString,
+				strings.Join(unSafeAnnotations, ", "))
+
+		*/
```
rossjones, line 38: Can remove this

cmd/util/exitcodes.go
frrist: These are no longer used, so will just remove the file.

cmd/util/api.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"strconv"
+
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/viper"
+
+	"github.com/bacalhau-project/bacalhau/pkg/requester/publicapi"
+)
+
+// TODO you must vlidate your assumptions of how this works before continuing.
```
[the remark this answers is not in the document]
frrist: extraneous comment, removed.

cmd/cli/list/list.go
```diff
@@ -263,3 +269,35 @@ func summarizeJob(j *model.JobWithInfo, OL *ListOptions) (table.Row, error) {
 
 	return row, nil
 }
+
+func shortenTime(outputWide bool, t time.Time) string { //nolint:unused // Useful function, holding here
+	if outputWide {
+		return t.Format("06-01-02-15:04:05")
```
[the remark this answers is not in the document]
frrist: Not sure, the additions here were from me moving things around, going to leave this unchanged for now and we can modify it in a follow on if needed.

cmd/cli/create/create.go
```diff
@@ -111,81 +107,73 @@ func create(cmd *cobra.Command, cmdArgs []string, OC *CreateOptions) error { //n
...
+			return fmt.Errorf("%s: %w", userstrings.JobSpecBad, err)
 		}
 
-		job, taskErr := model.NewJobWithSaneProductionDefaults()
-		if taskErr != nil {
-			panic(taskErr)
+		job, err := model.NewJobWithSaneProductionDefaults()
+		if err != nil {
+			// TODO this is a bit extream, maybe just ensure the above call doesn't return an error? the mergo package is a bit pointless there.
+			panic(err)
```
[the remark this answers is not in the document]
frrist: I am going to leave the logic here as is and the TODO in place, we can remove this at a later point if needed.

cmd/cli/list/list.go
frrist: it may be set to `true` by a user flag, so will leave in place for now.

PR #2576, opened 2023-06-16
Job Info Event consumer

wdbaruni:
### introduce completed jobs gossipsub topic

a new gossipsub topic that requesters publish info about jobs when they are successfully completed, cancelled or failed

### job info consumer
    
A separate application that consumes job info events and feed them into postgres database. It doesn't expose any interface to query or visualize the data as that is currently possible through grafana and looker.
    
This is a replacement to our current dashboard. It follows a very simple schema that should withstand changes in our job and status specs. Separate postgres views are created to provide more valuable summary of the data which can be updated with future changes in job specs


apps/job-info-consumer/consumer/pkg/postgres.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/golang-migrate/migrate/v4/source/iofs"
+	"github.com/rs/zerolog/log"
+
+	_ "github.com/lib/pq"
+)
+
+type PostgresDatastoreParams struct {
+	Host        string
+	Port        int
+	Database    string
```
rossjones, line 22: Should we have an explicit sslmode parameter too? We use `sslmode=disable` below but I'm sure we'd need this on AWS.

apps/job-info-consumer/consumer/cmd/utils.go
```diff
@@ -0,0 +1,77 @@
...
+func getDefaultOptionBool(envName string, defaultValue bool) bool {
+	envValue := os.Getenv(envName)
+	if envValue != "" {
+		i, err := strconv.ParseBool(envValue)
+		if err == nil {
+			return i
+		}
+	}
+	return defaultValue
+}
```
rossjones, line 45: For these you can use 

```
# util.GetEnvAs
# e.g. 
v := util.GetEnvAs[bool]("TEST_BOOL", true, strconv.ParseBool)
v := util.GetEnvAs[int]("TEST_INT", 10, strconv.Atoi)
```

apps/job-info-consumer/README.md
```diff
@@ -0,0 +1,41 @@
+## dev
+
```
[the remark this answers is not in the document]
wdbaruni: Done

apps/job-info-consumer/consumer/cmd/utils.go
wdbaruni: Didn't know about this util. Thanks for the pointer Ross :)

apps/job-info-consumer/Makefile
```diff
@@ -0,0 +1,51 @@
+.PHONY: all
```
frrist, line 1: It would be nice to have directives in the top level Makefile of the project to build (and test!) this package when the command `make all` is issued.

apps/job-info-consumer/consumer/pkg/consumer.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (c *Consumer) Start(ctx context.Context) error {
+	var err error
+	ctx, cancel := context.WithCancel(ctx)
+	defer func() {
+		if err != nil {
+			cancel()
+		}
+	}()
```
frrist, line 44: Not sure this is right since `err` is being reassigned below, we'll probably want a named return instead.

pkg/model/job_event.go
```diff
@@ -69,20 +66,16 @@ const (
...
 
+	// a job has been completed
+	JobEventCompleted
+
 	jobEventDone // must be last
 )
 
 // IsTerminal returns true if the given event type signals the end of the
 // lifecycle of a job. After this, all nodes can safely ignore the job.
 func (je JobEventType) IsTerminal() bool {
```
frrist, line 77: is `JobEventInvalidRequest` also terminal?

apps/job-info-consumer/consumer/pkg/postgres.go
```diff
@@ -0,0 +1,126 @@
...
+	files, err := iofs.New(fs, "migrations")
+	if err != nil {
+		return nil, err
+	}
+
+	drive, err := postgres.WithInstance(d.db, &postgres.Config{})
+	if err != nil {
+		return nil, err
+	}
+	migrations, err := migrate.NewWithInstance("iofs", files, d.dbname, drive)
```
frrist, line 100: Should also return this error.

pkg/model/job_event.go
wdbaruni: actually we are not using `JobEventInvalidRequest` any more. Will create a new PR to remove it along with other dead event types

PR #2581, opened 2023-06-19
Show node ranking failures in the CLI

### LINKED ISSUE #2398: Have better errors around "no nodes with <StorageType> are available on the network"
image

simonwo:
Previously when so many nodes fail ranking that not enough are available to run the job, the CLI would just print an unhelpful "not enough nodes to run job" message. This makes it hard to diagnose why the job isn't suitable.

This commit adds a reason to the node rankings, and in the event of total job failure displays the reasons for each node in the CLI.

E.g.
```
% go run . create job-local.yaml
Error executing job: failed to submit job: not enough nodes to run job. requested: 1, available: 0. 
        Node QmW7rsRr: job requires more resources {CPU: 0, Memory: 0 B, Disk: 0 B, GPU: 2} than are available per job {CPU: 6400m, Memory: 12.8 GB, Disk: 97.5 GB, GPU: 0}
```

```
% go run . create job-local.yaml
Error executing job: failed to submit job: not enough nodes to run job. requested: 1, available: 0. 
        Node QmZdjfNa: does not support model.StorageSourceType LocalDirectory, only [IPFS URLDownload FilecoinUnsealed Inline]
```

Resolves #2398.

### DIFF
--- a/pkg/requester/ranking/features.go
+++ b/pkg/requester/ranking/features.go
@@ -2,6 +2,7 @@ package ranking
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/bacalhau-project/bacalhau/pkg/model"
 	"github.com/bacalhau-project/bacalhau/pkg/requester"
@@ -56,12 +57,12 @@ func NewStoragesNodeRanker() *featureNodeRanker[model.StorageSourceType] {
 // - Rank 10: Node is supporting the type(s) the job is requiring.
 // - Rank 0: We don't have information on what the node supports.
 // - Rank -1: Node is not supporting a type the job is requiring.
-func (s *featureNodeRanker[Key]) rankNode(ctx context.Context, node model.NodeInfo, requiredKeys []Key) int {
+func (s *featureNodeRanker[Key]) rankNode(ctx context.Context, node model.NodeInfo, requiredKeys []Key) (rank int, reason string) {
 	if node.ComputeNodeInfo == nil {
 		// Node supported types are not set, or the node was discovered not
 		// through nodeInfoPublisher (e.g. identity protocol). We will give the
 		// node the benefit of the doubt and ask it to bid.
-		return requester.RankPossible
+		return requester.RankPossible, "supported types are not known"
 	}
 
 	providedKeys := s.getNodeProvidedKeys(*node.ComputeNodeInfo)
@@ -74,15 +75,14 @@ func (s *featureNodeRanker[Key]) rankNode(ctx context.Context, node model.NodeIn
 			}
 		}
 
-		log.Ctx(ctx).Trace().Stringer("Requirement", requiredKey).Bool("Supported", found).Send()
 		if !found {
 			// Target wasn't found – we can end early as we won't use this node.
-			return requester.RankUnsuitable
+			return requester.RankUnsuitable, fmt.Sprintf("does not support %T %s, only %s", requiredKey, requiredKey, providedKeys)
 		}
 	}
 
 	// Node provides all the specified required types.
-	return requester.RankPreferred
+	return requester.RankPreferred, "provides all the specified required types"
 }
 
 func (s *featureNodeRanker[Key]) RankNodes(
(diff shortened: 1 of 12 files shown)


pkg/requester/ranking/features.go
```diff
@@ -56,12 +56,12 @@ func NewStoragesNodeRanker() *featureNodeRanker[model.StorageSourceType] {
 // - Rank 10: Node is supporting the type(s) the job is requiring.
 // - Rank 0: We don't have information on what the node supports.
 // - Rank -1: Node is not supporting a type the job is requiring.
-func (s *featureNodeRanker[Key]) rankNode(ctx context.Context, node model.NodeInfo, requiredKeys []Key) int {
+func (s *featureNodeRanker[Key]) rankNode(ctx context.Context, node model.NodeInfo, requiredKeys []Key) (int, string) {
```
rossjones, line 59: Only a minor point, but I think this _might_ be one of those cases where named return values provide some useful info.  If this were 

```go 
func (s *featureNodeRanker[Key]) rankNode(
    ctx context.Context, 
    node model.NodeInfo, 
    requiredKeys []Key) (rank int, reason string) {
```
even if they weren't used they would still be useful as documentation.

pkg/requester/ranking/previous_executions.go
```diff
@@ -47,19 +47,25 @@ func (s *PreviousExecutionsNodeRanker) RankNodes(ctx context.Context, job model.
...
 		if previousExecutions, ok := previousExecutors[node.PeerInfo.ID.String()]; ok {
 			if previousExecutions > 1 {
 				rank = requester.RankUnsuitable
+				reason = "job already executed on this node"
 			} else if _, filterOut := toFilterOut[node.PeerInfo.ID.String()]; filterOut {
 				rank = requester.RankUnsuitable
+				reason = "job rejected or produced invalid result"
 			} else {
 				rank = requester.RankPossible
+				reason = "job already executed on this node"
```
rossjones, line 60: This is slightly confusing as it matches the reason on line 54, but with a different rank.  Is it possible to clarify the else block is for `previousExecutions == 1` and with a slightly different reason?

rossjones (commented):
LGTM, 2 rather minor points.

pkg/requester/ranking/utils.go
```diff
@@ -0,0 +1,19 @@
...
+	"context"
+
+	"github.com/bacalhau-project/bacalhau/pkg/requester"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+)
+
+const logLevel = zerolog.TraceLevel
+
+func logRank(ctx context.Context, rank requester.NodeRank) {
```
frrist, line 13: nit: I'm not keen on have a separate function specifically for logging. Could we instead implement [`LogObjectMarshaler`](https://github.com/rs/zerolog/blob/9070d49a1a4154c819e9c5ae6c923f3cc916d959/event.go#L48) on the `requester.NodeRank` type and then log it without this method like:
```go
log.Ctx(ctx).Trace().Object(rank).Msg("Ranked node")
```

We could also use `LogArrayMarshaler` if we are logging an array of these structures.

pkg/requester/ranking/labels.go
```diff
@@ -51,26 +51,30 @@ func (s *LabelsNodeRanker) RankNodes(ctx context.Context, job model.Job, nodes [
 		}
 	}
 	for i, node := range nodes {
-		rank := 0
+		rank := requester.RankPossible
+		reason := "selectors not set or unknown"
 		if !mustSelector.Empty() {
 			if mustSelector.Matches(labels.Set(node.Labels)) {
-				rank = 10
+				rank = requester.RankPreferred
```
frrist, line 58: ❤️ the constant for this!

frrist (approved):
Looks good!

Question about the example:
> Error executing job: failed to submit job: not enough nodes to run job. requested: 1, available: 0. 
        Node QmZdjfNa: does not support model.StorageSourceType LocalDirectory, only [IPFS URLDownload FilecoinUnsealed Inline]
        
What does this print when there's more than 1 node requested? E.g. if 3 nodes are requested and none match - is there a message for each?

pkg/requester/ranking/features.go
```diff
@@ -74,15 +74,14 @@ func (s *featureNodeRanker[Key]) rankNode(ctx context.Context, node model.NodeIn
...
 		if !found {
 			// Target wasn't found – we can end early as we won't use this node.
-			return requester.RankUnsuitable
+			return requester.RankUnsuitable, fmt.Sprintf("does not support %T %s, only %s", requiredKey, requiredKey, providedKeys)
 		}
 	}
 
 	// Node provides all the specified required types.
-	return requester.RankPreferred
+	return requester.RankPreferred, "provides all the specified required types"
```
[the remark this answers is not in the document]
simonwo: I figured it was at worst just more documentation and I would rather return a meaningful string rather than an empty one. In the future it may be helpful if diagnosing why some nodes are preferred over others.

pkg/requester/ranking/utils.go
simonwo: Cool, did'nt know you could do that!

Issue #2542, opened 2023-06-03
Implement Pluggable Enginge Specification


frrist:
### Background
initial Proposal: https://www.notion.so/expanso/Pluggable-Engines-9d16239f569341b2afabb160c3ae4e59
Bacalhau supports several Executors: Docker, Language, Python, and WASM. Per https://github.com/bacalhau-project/bacalhau/issues/2537 Python and Language executors will be removed which leaves us with Docker and WASM - the focus of this work.

Docker Executor:
https://github.com/bacalhau-project/bacalhau/blob/5651cd8b45ffad1c1352e8a1a7b91cb9e53128ae/pkg/model/job.go#L221-L231

Wasm Executor:
https://github.com/bacalhau-project/bacalhau/blob/5651cd8b45ffad1c1352e8a1a7b91cb9e53128ae/pkg/model/job.go#L250-L269

### Action
- Define a pluggable generic engine specification:
```go
type EngineSpec struct {
  Type string
  Params map[string]interface{}
}
```
- Replace the engine type enum constants (`int`'s) with a `string`. 
- Implement the `JobSpecDocker` and `JobSpecWASM` engine specs using the new `EngineSpec` type.

PR #2594, opened 2023-06-27
Implement Pluggable EngineSpec

frrist:
## What
This PR goes part way to implement a pluggable engine specification as outlined in #2542. Below are some decisions we need to make regarding compatibility with previous versions of bacalhau.

- [x] Per https://github.com/bacalhau-project/bacalhau/discussions/2605 we will be supporting backwards compatibility so **keeping** these types, but in a package for previous versions
**Keep or Kill `Spec.Engine`, `Spec.JobSpecDocker`, and `Spec.JobSpecWasm`?**: This decision will affect engine pluggability moving forward.
  - Keep:
    - We remain backwards compatible with previous versions of bacalhau
    - We take on the extra work of handling both fields
    - Sub Decisions:
      - Do compute nodes advertise the `Engine` enum for engines they support, or do they advertise a string representing their supported engine?
        - Continue using Enum: Engines aren't pluggable.
        - Use a string for Engines: Engines are pluggable.
      - Do we maintain two versions of Spec files here: https://github.com/bacalhau-project/bacalhau/pull/2594/commits/f72f900f96254c484931381d8eb892520d6549d2, one for the old version and one for the new?
  - Kill:
    - We break backwards compatibility with previous versions of bacalhau
    - We have less code to maintain
    - Engines are truly pluggable
  - Thoughts: I am leaning towards kill, at some point we'll probably want to break backwards compatibility, and I would prefer to do it while its a simple change to make rather than later when more complexity is introduced.
- [ ] Find a more robust solution for decoding `EngineSpec.Params` from `map[string]interface{}` to its concrete type (e.g. `WasmEngine`) than https://github.com/bacalhau-project/bacalhau/pull/2594/commits/d6145ddb38dc152a8cf481b015f5264fe3428a50
  - The problem here is that `StorageSpecType` is an `int` internally but is passed as a `string` when specifying a job and the reflection based decoding used by `mapstrucutre.Decode` cannot handle conversion of a `string` to an `int` without special handling.

### TODO
- [ ] Address the `TODO(forrest)` comments after some review.


pkg/executor/docker/bidstrategy/semantic/image_platform.go
```diff
@@ -32,26 +33,38 @@ func (s *ImagePlatformBidStrategy) ShouldBid(
...
+	// TODO(forrest): [correctness] keep same (probably incorrect) behavior as before.
+	if request.Job.Spec.EngineSpec.Type != model.EngineTypeDocker {
+		return bidstrategy.NewShouldBidResponse(), nil
+	}
+
+	// TODO(forrest): [correctness] keep same (probably incorrect) behavior as before.
+	dockerEngine, engineErr := model.DockerEngineFromEngineSpec(request.Job.Spec.EngineSpec)
+	if engineErr != nil {
+		return bidstrategy.NewShouldBidResponse(), engineErr
+	}
```
frrist, line 49: This all looks like a bug as we are returning an affirmative bidding response for image based bidding the job **is not docker**. Thoughts?

pkg/executor/wasm/spec.go
```diff
@@ -0,0 +1,107 @@
...
+		from reflect.Type,
+		to reflect.Type,
+		data interface{}) (interface{}, error) {
+		// TODO(forrest): [hack] this is unsafe, but I am unsure how else to handle this case.
+		if to.String() != "model.StorageSourceType" {
+			return data, nil
+		}
+		return model.ParseStorageSourceType(data.(string))
+	}
+}
```
frrist, line 107: This is the "hack" I am referring to in the PR description, it works for now but is fragile to changes. Suggestions on alternatives welcome!

pkg/executor/wasm/validator.go
```diff
@@ -3,29 +3,12 @@ package wasm
...
 	"github.com/rs/zerolog/log"
 	"github.com/tetratelabs/wazero"
 	"github.com/tetratelabs/wazero/api"
 	"golang.org/x/exp/maps"
 )
 
-// ValidateModuleAgainstJob will return an error if the passed job does not
-// represent a valid WASM executor job or the passed module is not able to be
-// run to fulfill the job.
-func ValidateModuleAgainstJob(
```
frrist, line 16: unused.

pkg/model/ipld.go
```diff
@@ -93,3 +94,16 @@ type IPLDMap[K comparable, V any] struct {
...
 }
+
+func (m IPLDMap[K, V]) ToStringSlice() []string {
+	var result []string
+	for _, key := range m.Keys {
+		value, exists := m.Values[key]
+		if exists {
+			result = append(result, fmt.Sprintf("%v", key), fmt.Sprintf("%v", value))
+		} else {
+			panic(fmt.Sprintf("invalid IPLD map: key '%v' missing value", key))
```
frrist, line 105: alternatively this could return an error, panicking here as I assume it's not possible to construct an IPLD map containing a key without a value, I am probably wrong on this.

pkg/model/job.go
```diff
@@ -275,8 +278,13 @@ func (s *Spec) GetTimeout() time.Duration {
...
 func (s *Spec) AllStorageSpecs() []*StorageSpec {
-	storages := []*StorageSpec{
-		&s.Wasm.EntryModule,
+	storages := []*StorageSpec{}
+	if s.EngineSpec.Type == EngineTypeWasm {
+		wasmEngine, err := WasmEngineFromEngineSpec(s.EngineSpec)
+		if err != nil {
+			panic(err)
+		}
+		storages = append(storages, &wasmEngine.EntryModule)
```
frrist, line 287: bubbling up the error from this method will be a moderately size code change, if we feel strongly that panicking is incorrect here I will make said change.

pkg/requester/jobtransform/docker_image_digest.go
```diff
@@ -1,51 +0,0 @@
...
-
-import (
-	"context"
-
-	"github.com/bacalhau-project/bacalhau/pkg/docker"
-	"github.com/bacalhau-project/bacalhau/pkg/model"
-	"github.com/rs/zerolog/log"
-)
-
-func DockerImageDigest() Transformer {
```
frrist, line 11: unused.

pkg/model/wasm_task.go
```diff
@@ -19,25 +19,24 @@ var _ JobType = (*WasmInputs)(nil)
...
 	for _, resource := range wasm.Modules {
+		// TODO(forrest): [correctness] this is correct but easily bug prone: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable
 		resource := resource
-		spec.Wasm.ImportModules = append(spec.Wasm.ImportModules, parseStorageSource("", &resource))
+		importModules = append(importModules, parseStorageSource("", &resource))
 	}
 
+	// TODO(forrest): [logic change] I am not sure this is the correct way to set an entryModule, previously we never set it, attempting to now.
+	spec.EngineSpec = NewWasmEngineSpec(parseStorageSource("", entryModule), wasm.Entrypoint, wasm.Parameters, wasm.Env.Values, importModules)
+	spec.EngineDeprecated = EngineWasm
```
frrist, line 38: This needs review:
previously we never assigned the wasm engine an entry module and instead put it in the Inputs, I have keep the behavior the same regarding inputs, but have changed the logic slightly by also including it in the wasm engine, I am not sure what the original intention of this method is.

pkg/test/devstack/submit_test.go
```diff
@@ -59,7 +59,7 @@ func (suite *DevstackSubmitSuite) TestEmptySpec() {
 	require.Error(suite.T(), missingSpecError)
 
 	j = &model.Job{}
-	j.Spec = model.Spec{Engine: model.EngineDocker}
+	j.Spec = model.Spec{EngineDeprecated: model.EngineDocker, EngineSpec: model.NewDockerEngineSpec("", nil, nil, "")}
```
frrist, line 62: It would probably be better to make this a noop engine as I believe the intention is to assert that the spec is invalid due to an empty deal structure, thoughts?

pkg/test/scenario/example_noop_test.go
```diff
@@ -22,10 +23,9 @@ var noopScenario Scenario = Scenario{
 		},
 	},
 	Spec: model.Spec{
-		Engine: model.EngineNoop,
-		Wasm: model.JobSpecWasm{
-			EntryPoint: "_start",
-		},
+		// TODO(forrest): [correctness] this doesn't make sense, why wasm?
+		EngineDeprecated: model.EngineNoop,
+		EngineSpec:       model.NewWasmEngineSpec(model.StorageSpec{}, "_start", nil, nil, nil),
```
frrist, line 28: Not sure what the correct change is here, I'd assume replacing wasm with noop would be fine?

frrist (commented):
Left some comments to guide review.

cmd/cli/create/create.go
```diff
@@ -213,7 +213,12 @@ func create(cmd *cobra.Command, cmdArgs []string, OC *CreateOptions) error { //n
 	err = jobutils.VerifyJob(ctx, j)
 	if err != nil {
 		if _, ok := err.(*bacerrors.ImageNotFound); ok {
-			return fmt.Errorf("docker image '%s' not found in the registry, or needs authorization", j.Spec.Docker.Image)
+			// TODO(forrest): [fixme] the below error ought to be impossible.
+			dockerEngine, err := model.DockerEngineFromEngineSpec(j.Spec.EngineSpec)
```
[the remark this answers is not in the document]
rossjones: At present we get `Could not inspect image "ss" - could be due to repo/image not existing, or registry needing authorization` which is misleading because it's serving a dual purpose - we should be able to tell which it is.  This is probably the place we should be using ImageNotFound so that we can present a nicer and more accurate error message.

pkg/model/ipld.go
rossjones: Even if you create an IPLDMap with a nillable V then you should still be able to retrieve it and sprintf it as %v - seems unlikely to have a key with no value.  But might be best to return an error, panicking is probably a bit excessive :D

pkg/model/job.go
rossjones: Not a fan of panics, and AllStorageSpecs() isn't called from too many places, _but_ the problem is that places like `NewStoragesNodeRanker` will need to handle the error, and don't currently return an error, which means its use in the ranker chain will need to change to accommodate it.

pkg/model/task.go
```diff
@@ -129,6 +129,7 @@ func parseStorageSource(path string, resource *Resource) StorageSpec {
 
 func parseInputs(mounts IPLDMap[string, Resource]) ([]StorageSpec, error) {
 	inputs := []StorageSpec{}
+	// TODO(forrest): [correctness] this is correct but easily bug prone: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable
```
rossjones, line 132: I think this whole function could be 

```go
return lo.MapToSlice(mounts.Values, func(path string, resource Resource) StorageSpec {
    return parseStorageSource(path, &resource)
})
```
if parseStorageSource didn't require a pointer parameter, it could be 

```go
return lo.MapToSlice(mounts.Values, parseStorageSource)
```

rossjones (commented):
Couple of comments, but nothing earth-shattering.

cmd/cli/describe/describe_test.go
```diff
@@ -51,8 +51,9 @@ func (s *DescribeSuite) TestDescribeJob() {
 
 				for i := 0; i < tc.numberOfAcceptNodes; i++ {
 					for k := 0; k < n.numOfJobs; k++ {
+						// TODO(forrest): [correctness] were making a noop job and assigning a docker entry point, which ought to be ignored?
 						j := testutils.MakeNoopJob()
-						j.Spec.Docker.Entrypoint = []string{"Entrypoint-Unique-Array", uuid.NewString()}
+						j.Spec.EngineSpec = model.NewDockerEngineSpec("", []string{"Entrypoint-Unique-Array", uuid.NewString()}, nil, "")
```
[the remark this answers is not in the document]
simonwo: > I think the intent here was to execute a job that did nothing and then assert its description contained some metedata about the job.

Yes, I think the test is just that the `describe` API does actually return details for the job (in the past this has regressed). If there is some other field that we can fill in with testable data then sure, we can replace this. I think we should add some tests that `describe` can handle pluggable engine specs though.

> I think the intent here was to execute a job that did nothing and then assert its description contained some metedata about the job.

pkg/executor/wasm/spec.go
```diff
@@ -0,0 +1,107 @@
...
+	Entrypoint string `json:"EntryPoint,omitempty"`
+
+	// The arguments supplied to the program (i.e. as ARGV).
+	Parameters []string `json:"Parameters,omitempty"`
+
+	// The variables available in the environment of the running program.
+	EnvironmentVariables map[string]string `json:"EnvironmentVariables,omitempty"`
+
+	// TODO #880: Other WASM modules whose exports will be available as imports
+	// to the EntryModule.
```
simonwo, line 57: You can remove the todo here now.

pkg/model/job.go
```diff
@@ -220,11 +221,17 @@ type PublisherSpec struct {
...
+}
+
 // Spec is a complete specification of a job that can be run on some
 // execution provider.
 type Spec struct {
-	// e.g. docker or language
-	Engine Engine `json:"Engine,omitempty"`
+	// deprecated: user EngineSpec instead.
+	EngineDeprecated Engine `json:"Engine,omitempty"`
+	EngineSpec       EngineSpec
```
simonwo, line 234: Is there any point deprecating this if the Docker and Wasm fields are completely gone? As that change will already have broken every previous job.

pkg/executor/wasm/spec.go
```diff
@@ -0,0 +1,107 @@
...
+			EngineKeyEntrypoint:           entrypoint,
+			EngineKeyParameters:           parameters,
+			EngineKeyEnvironmentVariables: environmentVariables,
+			EngineKeyImportModules:        importModules,
+		},
+	}
+}
+
+// Describes a raw WASM job
+type Engine struct {
```
simonwo, line 40: Why do we have this in both `model` and here?

pkg/model/wasm_engine.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return out, nil
+}
+
+func StorageSpecUnmarshalledHookFunc() mapstructure.DecodeHookFuncType {
+	return func(
+		from reflect.Type,
+		to reflect.Type,
+		data interface{}) (interface{}, error) {
+		// TODO(forrest): [hack] this is unsafe, but I am unsure how else to handle this case.
```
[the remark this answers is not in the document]
frrist: Another option (this just occurred to me re your raw json comment) would be to first marshal the `map[string]interface{}` to json bytes and then unmarshal that into its corresponding structure. e.g.
```go
func AsEngine(e model.EngineSpec) (Engine, error) {
	if e.Type != EngineType {
		return Engine{}, fmt.Errorf("expected type %s got %s", EngineType, e.Type)
	}
	if e.Params == nil {
		return Engine{}, fmt.Errorf("engine params uninitialized")
	}
	pb, err := json.Marshal(e.Params)
	if err != nil {
		return Engine{}, err
	}
	var out Engine
	if err := json.Unmarshal(pb, &out); err != nil {
		return Engine{}, err
	}
	return out, nil
}
```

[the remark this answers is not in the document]
rossjones: Oh, that'd be neat if it works as it should.

pkg/executor/docker/spec.go
```diff
@@ -0,0 +1,67 @@
...
+type Engine struct {
+	// this should be pullable by docker
+	Image string `json:"Image,omitempty"`
+	// optionally override the default entrypoint
+	Entrypoint []string `json:"Entrypoint,omitempty"`
+	// a map of env to run the container with
+	EnvironmentVariables []string `json:"EnvironmentVariables,omitempty"`
+	// working directory inside the container
+	WorkingDirectory string `json:"WorkingDirectory,omitempty"`
+}
```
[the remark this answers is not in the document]
frrist: How about `EngineSpec`? Imports will read like `docker.EngineSpec`.

pkg/model/wasm_engine.go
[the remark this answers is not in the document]
wdbaruni: I like the json marshall/unmarshall approach compared to using mapstructure. It is not very efficient, but we can accept that for now. 

A more efficient way would be to do explicit and per field decoding such as:
```
func WasmEngineFromEngineSpec(e EngineSpec) (WasmEngine, error) {
	if e.Type != EngineTypeWasm {
		return WasmEngine{}, fmt.Errorf("expected type %s got %s", EngineTypeWasm, e.Type)
	}
	if e.Params == nil {
		return WasmEngine{}, fmt.Errorf("engine params uninitialized")
	}
	var out WasmEngine
	out.Entrypoint = decoder.ToString(e.Params, "entrypoint")
	out.EntryModule = decoder.ToType[StorageSpec](e.Params, "entrymodule", funcThatConvertsToStorageSpec)
	out.Parameters = decoder.ToArray[String](e.Params, "parameters")
	out.EnvironmentVariables = decoder.ToMap[String,String](e.Params, "environmentvariables")
	out.ImportModules = decoder.ToArrayType[StorageSpec(e.Params, "importmodules", funcThatConvertsToStorageSpec)
	return out, nil
}
```

pkg/executor/wasm/spec.go
frrist: Eventually these specifications will live inside plugins - I'll need to figure our a way of exposing their "schemas" to client and requester nodes, but punting that task for now. 
I'll dedup these specs to a single location (probably `pkg/model`) before merging.

pkg/model/job.go
frrist: I will remove these fields from this version of the job spec and also preserve the previous job spec schema to maintain backwards compatibility as described in https://github.com/bacalhau-project/bacalhau/discussions/2605. Doing that now in the following commits.

pkg/executor/docker/bidstrategy/semantic/image_platform.go
simonwo: An affirmative response is more like "don't discount based on this". So we have to have an affirmative response from everyone, or just a single negative response.

pkg/model/wasm_engine.go
```diff
@@ -0,0 +1,95 @@
...
+	EngineKeyEnvironmentVariablesWasm = "environmentvariables"
+	EngineKeyImportModulesWasm        = "importmodules"
+)
+
+// NewWasmEngineSpec returns an EngineSpec of type EngineTypeWasm with the provided arguments as EngineSpec.Params.
+func NewWasmEngineSpec(
+	entryModule StorageSpec,
+	entrypoint string,
+	parameters []string,
+	environmentVariables map[string]string,
```
rossjones, line 22: The environmentVariables here are a map of string->string, but the environmentVariables in the docker engine are []string.  Should we fix them so they are both consistent?

frrist:
Closing in favor of https://github.com/bacalhau-project/bacalhau/pull/2706

PR #2597, opened 2023-06-27
Compute Object Store

rossjones:
Currently execution data is held in memory by the individual compute nodes.  This means that should a compute node be restarted, all knowledge of the work it was doing is lost.  As we move toward different types of jobs, where we expect some to always be running we need to persist this state so that compute nodes can resume work from where they left off.

To provide this resilience, this PR introduces a local persistent store for compute nodes using boltdb key-value database (using the etcd variant). This new execution store is interchangeable with the current inmemory store, implementing the same interface and passing the same tests.  The inmemory store is currently the default but the kvstore can be enabled using environment variables:

```shell
# Enable the boltdb storage, alternative option is inmemory (or "" for the default)
export BACALHAU_COMPUTE_STORE_TYPE=boltdb
# Specifies the file to be used by the compute node's database
export BACALHAU_COMPUTE_STORE_PATH=/tmp/database.db
```

The database created is a single file, stored in `BACALHAU_COMPUTE_STORE_PATH` or the config directory if this is not specified.  When not specified, the filename structure is `NODEID-compute.db` and is persistent. Any databases created during tests are, and should be, cleaned up during teardown.

As part of this work the execution store model has changed to allow `GetExecutionCount` to receive a `store.ExecutionState` for which the count is required.  For this store it is held in-memory and loaded at runtime from the current state in the database at launch time.  This background task might provide the initial basis for start-up compaction.


pkg/config/config.go
```diff
@@ -295,3 +296,35 @@ func GetDockerCredentials() DockerCredentials {
...
+	StoreType ExecutionStoreType
+	Location  string
+}
+
+func GetComputeStorageConfig(nodeID string) ComputeStorageConfig {
+	c := ComputeStorageConfig{}
+
+	computeStore := strings.ToLower(os.Getenv("BACALHAU_COMPUTE_STORE"))
+	if computeStore == "kvstore" {
+		c.StoreType = ExecutionStoreKVStore
```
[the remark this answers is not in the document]
rossjones: Following on from other comments, I've just directly built an execution store around boltdb directly as it is much easier to manage transactions without the added complication of too many layers of abstraction.  So I'm removing store.kvstore in favour of store.boltdb and then I guess the types of stores are more about the explicit storage mechanism.

pkg/config/config.go
```diff
@@ -295,3 +296,35 @@ func GetDockerCredentials() DockerCredentials {
...
+		c.StoreType = ExecutionStoreInMemory
+	}
+
+	f := fmt.Sprintf("%s-compute.db", nodeID)
+	path := os.Getenv("BACALHAU_COMPUTE_STORE_PATH")
+	if path == "" {
+		path = GetConfigPath()
+	}
+	c.Location = filepath.Join(path, f)
+
```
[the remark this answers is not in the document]
rossjones: I'm making that change, although I'm expecting it'll be little used to specify the environment variable as there's no obvious way to narrow it down to an individual compute node (in cases where more than one is on the same machine).

pkg/node/compute.go
```diff
@@ -306,8 +310,27 @@ func createExecutionStore(host host.Host) (store.ExecutionStore, error) {
...
+			localstore.WithLocation(storageConfig.Location),
+			localstore.WithPrefixes(kvstore.ExecutionPrefixes...),
+		)
+		if err != nil {
+			return nil, err
+		}
+		store = kvstore.NewStore(ctx, database)
+
+		// Ensure the database underlying the store is closed at shutdown
+		cm.RegisterCallbackWithContext(database.Close)
```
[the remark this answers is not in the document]
rossjones: I've added a Close(context) method to `ExecutionStore` as there wasn't one, and so now no matter what execution store we create, Close() will be called in the cleanupFunc (which is currently empty).

pkg/compute/store/boltdb/store.go
```diff
@@ -0,0 +1,439 @@
...
+	if err != nil {
+		errorList = errors.Join(errorList, store.NewErrExecutionNotFound(executionID))
+	}
+
+	// Delete single execution entry
+	executionsBucket := tx.Bucket([]byte(BucketExecutionsName))
+	err = executionsBucket.Delete([]byte(executionID))
+	if err != nil {
+		errorList = errors.Join(errorList, err)
+	}
```
wdbaruni, line 394: If I understand correctly, an error returned within a transaction means the whole transaction will be rolled-back. Is there a value of appending errors instead of just returning the first encountered error?
Also we are not consistent as we fail early if the job index couldn't be deleted

pkg/compute/store/boltdb/store.go
```diff
@@ -0,0 +1,439 @@
...
+func (s *Store) Close(ctx context.Context) error {
+	return s.database.Close()
+}
+
+func (s *Store) GetExecutionCount(ctx context.Context) (uint, error) {
+	log.Ctx(ctx).Trace().
+		Msg("boltdb.GetExecutionCount")
+
+	// TODO(ross) So, counting based on execution state .... seems like
+	// maintaining a prefix for a counter is excessive
```
wdbaruni, line 426: I just to make sure you saw my comment in your previous revision about calculating counts

pkg/compute/store/kvstore/envelopes.go
```diff
@@ -0,0 +1,38 @@
+package kvstore
+
```
wdbaruni, line 2: I think you forgot to delete this package

pkg/node/compute.go
```diff
@@ -63,7 +67,7 @@ func NewComputeNode(
 	// create the execution store
 	if config.ExecutionStore == nil {
 		var err error
-		executionStore, err = createExecutionStore(host)
+		executionStore, err = createExecutionStore(ctx, host, cleanupManager)
```
wdbaruni, line 70: I don't think you are using cleanupManager inside `creaetExecutionStore`

pkg/objectstore/errors.go
```diff
@@ -0,0 +1,31 @@
+package objectstore
+
```
wdbaruni, line 2: I think you forgot to delete this package as well

pkg/node/compute.go
```diff
@@ -41,6 +44,7 @@ type Compute struct {
 	Executors           executor.ExecutorProvider
 	LogServer           *logstream.LogStreamServer
 	Bidder              compute.Bidder
+	Database            *localstore.LocalStore
```
wdbaruni, line 47: this is not being used, and you are no longer using localstore anywhere

pkg/compute/store/boltdb/store.go
```diff
@@ -0,0 +1,439 @@
...
+)
+
+const (
+	newExecutionComment = "Execution created"
+
+	DefaultSliceRetrievalCapacity = 10
+
+	BucketExecutionsName = "executions"
+	BucketHistoryName    = "history"
+	BucketJobIndexName   = "execution-index"
```
wdbaruni, line 23: I suggest to prefix the bucket names with `executions` as the compute node may use the same bolt database in the future to store info other than executiosn:

```
	BucketExecutionsName = "executions"
	BucketHistoryName    = "executions_history"
	BucketJobIndexName   = "executions_jobid_index"
```

or something similar

pkg/compute/store/kvstore/store.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if err != nil || len(identifiers) == 0 {
+		return nil, store.NewErrExecutionsNotFoundForJob(jobID)
+	}
+
+	// TODO: We can optimize this by fetching en masse rather than one read per
+	// execution (if we extend the client impl)
+	executions := make([]store.Execution, len(identifiers))
+	for i, key := range identifiers {
+		execEnv, err := s.executions.Get(key)
```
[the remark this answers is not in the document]
rossjones: I've implemented the 'indexing' in-line with the transaction for the compute node persistence.  I think when we find more use-cases and can refactor out the commonalities, we can follow the approach to running in a transaction that we've implemented here (a private version of the called method that _must_ have a trxn and is expected to process all related data/tasks within that transaction).

pkg/compute/store/kvstore/store.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/bacalhau-project/bacalhau/pkg/objectstore/localstore"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	newExecutionComment = "Execution created"
+
+	PrefixJobs       = "jobs"
+	PrefixExecutions = "executions"
+	PrefixHistory    = "history"
```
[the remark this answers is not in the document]
rossjones: These were specific to the execution store, and are now renamed to Bucket*. The idea was for each specific use of the objectstore to use it's own prefixes, but I agree they should have been qualified/namespaced.

pkg/compute/store/kvstore/store.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (s *Store) CreateExecution(ctx context.Context, execution store.Execution) error {
+	log.Ctx(ctx).Debug().
+		Str("ExecutionID", execution.ID).
+		Msg("kvstore.CreateExecution")
+
+	_, err := s.executions.Get(execution.ID)
+	if err == nil {
+		return store.NewErrExecutionAlreadyExists(execution.ID)
+	}
```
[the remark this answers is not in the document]
rossjones: I did this (moved `kvstore` to `boltdb`) so now those packages (`inmemory` etc) really describe the storage mechanism explicitly, which I think makes more sense to me (for now).  The other change is really that we followed the pattern where each method in the interface is implemented like 

```
func (s *Store) CreateExecution(ctx context.Context, execution store.Execution) error {
... some basic validation/logging ...
	return s.database.Update(func(tx *bolt.Tx) (err error) {
	
        }
...
}

func (s *Store) createExecution(tx *boltdb.Tx, execution store.Execution) error {
    // check existing 
    // validation 
    //  persist
    // update relevant 'index' buckets
}
```

pkg/compute/store/kvstore/envelopes.go
rossjones: Was leaving it until I was sure I didn't need it :D

pkg/compute/store/boltdb/store.go
```diff
@@ -0,0 +1,439 @@
...
+		return
+	})
+
+	return execution, err
+}
+
+func (s *Store) getExecution(tx *bolt.Tx, executionID string) (store.Execution, error) {
+	var execution store.Execution
+
+	executionsBucket := tx.Bucket([]byte(BucketExecutionsName))
```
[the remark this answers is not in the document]
rossjones: It absolutely would, but (and this surprised me quite a bit), `[]byte()` isn't const (`(value of type []byte) is not constant`) and we can't have constant array in Go. We could make them vars I suppose

pkg/compute/store/boltdb/store.go
rossjones: Yes, working on that today.

pkg/compute/store/boltdb/store.go
rossjones: You're right about the transaction rollback, I think I was starting to experiment with the idea of returning errors separately and running with the partial changes, but that's definitely risky and will leave us in a bit of a mess. Will fix to return early as we do elsewhere.

rossjones: Fixed in 8532dc1d

pkg/compute/store/boltdb/store.go
```diff
@@ -0,0 +1,439 @@
...
+
+	// History entry keys within the execution bucket will be iterated over
+	// in lexographical order so we want to pad the numbers with leading 0s
+	// although this adds a hard limit of 999 history items for an execution
+	seqNum, err := executionHistoryBucket.NextSequence()
+	if err != nil {
+		return err
+	}
+	seq := fmt.Sprintf("%03d", seqNum)
+	return executionHistoryBucket.Put([]byte(seq), historyEntryData)
```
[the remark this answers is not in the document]
rossjones: No, unfortunately not.  They're retrieved in lexicographic order, so if we were to write 12 entries, we'd get them back in this order.
```
[1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9]
```

If we pad them to 3 spaces, we get them the following order.
```
[001, 002, 003, 004, 005 ... 012]
```

There is a side-effect that now we've got a limit on the number of history entries (per execution) but we might have a bug if we get to 1000 state changes.

pkg/compute/store/boltdb/store.go
```diff
@@ -0,0 +1,439 @@
...
+// each job is represented by a new bucket, named after the job ID.  Within
+// that job, each execution is represented by a key which is the id of that
+// execution, the value being nil.
+//
+// execution-index
+//     |--> job-id
+//               |-> <execution-id> -> {}
+//
+
+type Store struct {
```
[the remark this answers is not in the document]
rossjones: For my benefit as much as anybody elses :D

pkg/compute/store/boltdb/counter.go
```diff
@@ -0,0 +1,50 @@
+package boltdb
+
+import (
+	"github.com/bacalhau-project/bacalhau/pkg/compute/store"
+	"go.uber.org/atomic"
+)
+
+type StateCounter struct {
+	data map[store.ExecutionState]*atomic.Uint64
+}
```
wdbaruni, line 10: I am assuming it is safe to use a map instead of syncMap since the counter is only mutated within blot transactions, right?

pkg/compute/store/boltdb/store.go
```diff
@@ -404,35 +444,64 @@ func (s *Store) deleteExecution(tx *bolt.Tx, executionID string) error {
...
-	// 	if execution.State == store.ExecutionStateCompleted {
-	// 		counter++
-	// 	}
-	// }
-	// return counter, nil
-	return 0, nil
+	// We have to wait here to ensure the counter has been populated,
+	// so we will wait until we know it has finished
+	s.starting.Wait()
+	return uint(s.stateCounter.Get(state)), nil
```
wdbaruni, line 468: does it make sense to change `GetExecutionCount` to return uint64 instead of down casting?

pkg/compute/store/boltdb/store.go
```diff
@@ -404,35 +444,64 @@ func (s *Store) deleteExecution(tx *bolt.Tx, executionID string) error {
...
+		return err
+	})
+
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("failed to generate state counter for execution store")
+	}
+
+	// As more items may have been added whilst we were running, merge the
+	// statecounter we have generated into the existing one for the store
+	s.stateCounter.Include(acc)
```
wdbaruni, line 501: Is that possible given we are populating within a transaction? Does boltdb allow concurrent transactions on the same bucket?

wdbaruni (commented):
Looks great. Just minor comments, specially around using expected state while updating the counter

pkg/compute/store/boltdb/store.go
rossjones: One transaction per goroutine, but otherwise the only other restriction seems to be that there's only a single writer txn at a time.

pkg/compute/store/boltdb/counter.go
rossjones: More specifically, because it's only mutated within the single allowed writer transaction.

wdbaruni (approved):

PR #2599, opened 2023-06-27
Add `bacalhau nodes` to output details of known nodes.

simonwo:
This commit adds a new `nodes` command to Bacalhau that will output the list of known nodes from the requester. By default the output will include the labels and capacity of each node but can be customised to include features and versions as well.

CLI commands now use a common method for outputting in YAML, JSON, CSV or a text-based table, rather than all supporting a smattering of different things in this area.


rossjones (approved):
LGTM, really like the extensible table with `--show`

pkg/model/nodeinfo.go
```diff
@@ -6,10 +6,11 @@ import (
...
 )
 
+//go:generate stringer -type=NodeType -trimprefix=NodeType -output=nodeinfo_string.go
 type NodeType int
 
 const (
-	nodeTypeUnknown NodeType = iota
+	NodeTypeRequester NodeType = iota
 	NodeTypeCompute
 )
```
frrist, line 15: This will mean previous versions of the code treat unknown nodes as Requester nodes. Could we have instead made this a new value instead of modifying the existing?

cmd/util/output/output.go
```diff
@@ -0,0 +1,122 @@
...
+	switch options.Format {
+	case JSONFormat:
+		return json.NewEncoder(cmd.OutOrStdout()).Encode(items)
+	case YAMLFormat:
+		return yaml.NewEncoder(cmd.OutOrStdout()).Encode(items)
+	case CSVFormat:
+		fallthrough
+	case TableFormat:
+		outputTable[T](cmd, columns, options, items)
+		return nil
```
frrist, line 62: nit: I find `case CSVFormat, TableFormat:` clearer as a reader and typically use `fallthrough` when taking some action before falling through to the next case. Not a big deal though.

frrist (commented):
Good stuff! 

I missed the review window, but left a couple comments. I do have some concerns regarding the changes to nodeinfo.go as it relates to backwards compatibility - not that it will break anything, but rather that all previously unknown nodes will be treated as requester nodes, maybe this was always the case though?

Perhaps in a follow on we could make `list` the parent command of `nodes` and `jobs`, where `jobs` takes the behavior of `list`. I am guessing there will be other things we want to list in the future (maybe executions?) e.g
`bacalhau list nodes`: lists information about nodes on the network.
`bacalhau list jobs`: list information about jobs on the network.

cmd/util/output/output.go
simonwo: Ah, I didn't know you could include multiple values like this. Thanks!

pkg/model/nodeinfo.go
simonwo: My careful reading of the code concluded that this was always true (i.e. the only time we used unknown is when we definitely knew it was a requester anyway) so it's not introducing any ambiguity here. If we add a new node type in the future, previous versions will continue to "not know" what type it is as long as we use an appropriate value.

PR #2609, opened 2023-06-28
use math.Max/Min instead of system.Max/Min

wdbaruni:


frrist (approved):
I am assuming this is mostly a mechanical change? If so LGTM

I assume we will drop the `lib/math` package in favor of https://tip.golang.org/ref/spec#Min_and_max when it's released sometime in August?

PR #2615, opened 2023-06-30
Ensure prefixes in S3 are unique

rossjones:
This is to make sure that when multiple tests are being run in CI that they don't conflict (e.g. windows tests might delete before linux test does a read), as they are not run in sequence.

### DIFF
--- a/pkg/publisher/s3/publisher_test.go
+++ b/pkg/publisher/s3/publisher_test.go
@@ -8,10 +8,12 @@ import (
 	"compress/gzip"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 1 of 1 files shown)


simonwo (approved):

pkg/publisher/s3/publisher_test.go
```diff
@@ -24,13 +26,16 @@ import (
...
 const region = "eu-west-1"
 const endpoint = "https://s3-eu-west-1.amazonaws.com"
 
 var jobID = uuid.NewString()
 var executionID = uuid.NewString()
 
+// Ensure unique prefix
+var timestamp = time.Now().UTC().Format("20060102T150405") // yyyyMMddThhmmss
+var prefix = fmt.Sprintf("integration-tests-publisher-%s/%s/", timestamp, executionID)
+
```
wdbaruni, line 38: Looks great. The only concern I have is we might end up with a noisy bucket if some tests fail to delete their test directories. We are using the bucket for some static datasets used by other test scenarios.

My recommendation is to change the prefix to `fmt.Sprintf("integration-tests-publisher/%s-%s/", timestamp, executionID)`

Which will give a better console experience and limit the noise to one directory/prefix. An alternative is using a dedicate bucket just for this test scenario, but that is on overkill, and changing the prefix will give us the same benefit.

PR #2616, opened 2023-07-03
implement evaluation worker

wdbaruni:
Worker is a long-running process that dequeues evaluations, invokes the scheduler to process the evaluation, and acknowledges or rejects the evaluation back to the broker. The worker is single-threaded and processes one evaluation at a time. An orchestrator can have multiple workers working in parallel.

Future PRs will include actual implementation of schedulers, and linking things together


pkg/orchestrator/interfaces.go
```diff
@@ -58,3 +59,23 @@ type EvaluationBroker interface {
...
+	// Process handles a new evaluation. It applies the necessary logic to determine
+	// task placements based on the provided evaluation.
+	Process(*model.Evaluation) error
+}
+
+// SchedulerProvider returns a scheduler instance that is capable of handling
+// jobs requiring scheduling.
+type SchedulerProvider interface {
+	// Scheduler returns a scheduler for the given job type
+	Scheduler(jobType string) Scheduler
```
rossjones, line 77: This should probably return an error?  Although this is tested as `unknown returns nil` it isn't immediately obvious from looking at this that `Scheduler` is an interface (and therefore nillable), and perhaps having an error makes it more explicit that it can fail. 

[aside] I really wish Go mandated that interfaces _had_ to be returned as `*Scheduler`, I bet it'd safe lots of nil panics.

pkg/orchestrator/worker.go
```diff
@@ -0,0 +1,197 @@
...
+	// Check if worker is shutting down while dequeueing
+	if w.isShuttingDown() {
+		log.Warn().Msgf("Worker is shutting down, not scheduling evaluation %s", evaluation.ID)
+		return
+	}
+
+	// Schedule the evaluation
+	scheduler := w.schedulerProvider.Scheduler(evaluation.Type)
+	if scheduler == nil {
+		log.Error().Msgf("Failed to retrieve scheduler for evaluation %s", evaluation.ID)
```
rossjones, line 149: Is it worth including the evaluation.Type in here?

rossjones (approved):
LGTM, minor comments about returning an error from a func and added to a log message.

pkg/orchestrator/interfaces.go
wdbaruni: I was actually thinking about this and whether an error should be returned or not. My thinking was:

- If there can be different issues when returning a scheduler, then retuning an error with a descriptive message on what went wrong is helpful
- In this case, it is either a scheduler exists or does not exist. Thats why I thought of just returning a nullable result. However, this is based on `MappedSchedulerProvider` implementation 
- Another option could be `Scheduler(jobType string) (Scheduler, bool)`, where true means a scheduler was found

Overall, I need to fix this as it troubled me and you also raised a concern. I am just looking for the right _Go_ approach

pkg/lib/backoff/types.go
```diff
@@ -0,0 +1,11 @@
+//go:generate mockgen --source types.go --destination mock_backoff.go --package backoff
```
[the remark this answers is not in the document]
frrist: fwiw testify also has a mock package https://github.com/stretchr/testify#mock-package and a similar gen tool: https://github.com/vektra/mockery. Unsure if we value sticking with the same packages for testing/assertion/mocking.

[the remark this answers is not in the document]
wdbaruni: Yeah I spent over half a day comparing and implementing the test suite in both frameworks. The deal breaker was with error messages as testify/mock didn't give a helpful error message on which expected call wasn't triggered. It just tells you the method name, but not the arguments, which is not helpful when debugging

pkg/orchestrator/worker.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// Start triggers the worker to start processing evaluations.
+// The worker can only start once, and subsequent calls to Start will be ignored.
+func (w *Worker) Start() {
+	w.startOnce.Do(func() {
+		w.setStatus(WorkerStatusStarting)
+		go w.run()
+	})
+}
```
frrist, line 72: Alternatively, context could be passed to `run()` via `Start()` as its typically advised to [avoid embedding a context in structure](https://go.dev/blog/context-and-structs)

pkg/orchestrator/worker.go
```diff
@@ -0,0 +1,197 @@
...
+
+	var dequeueFailures int
+	for !w.isShuttingDown() {
+		// Dequeue an evaluation and apply backoff if dequeueing fails
+		evaluationReceipt, err := w.dequeueEvaluation()
+		if err != nil {
+			dequeueFailures++
+			w.dequeueFailureBackoff.Backoff(w.shuttingDownCtx, dequeueFailures)
+			continue
+		}
```
frrist, line 100: It might be worth exploring [sync.Cond](https://pkg.go.dev/sync#NewCond) in a follow for synchronizing dequeue events from the evaluation broker.

frrist (commented):
left some drive-by comments, feel free to ignore.

pkg/orchestrator/worker.go
wdbaruni: That makes sense. Pushed a new commit

pkg/orchestrator/worker.go
wdbaruni: Looks useful. Thanks for sharing

pkg/orchestrator/interfaces.go
wdbaruni: I went with returning an error

PR #2623, opened 2023-07-04
Refactor: Implement API for v1beta2 models and latest models

frrist:
### What
This PR introduces the following changes:
- `pkg/model/v1beta2` types, which are a copy of the current `pkg/model` types.
- modified the `requester/publicapi/*` to operate over `v1beta2` types
- implements the `RequesterAPIClientWrapper` wrapping the requester client to convert between `v1beta2` types and `model` types. We are using this everywhere right now, and it can serve as a stub for a v2 API until we decide on an implementation for a new API.

Motivated by 1. here: https://github.com/bacalhau-project/bacalhau/discussions/2605
I tried to make this as mechanical as I could. Fairly happy with the client wrapper but was unable to apply the same pattern to the server due to its internal fields expecting the latest version of model structures. I chose to make the type conversions in the server code directly rather than duplicate more types.

### Open Questions
The CLI currently uses the `RequesterAPIClientWrapper` and `model` types. Should it instead interact with the unwrapped client type and thus the `v1beta2` model types? This would mean either calling the converter methods in all CLI methods, or duplicating the CLI such that there is a v1 and v2. This will be a large code change so I'd like some feedback before undertaking.

### Alternatives
https://github.com/bacalhau-project/bacalhau/pull/2625


pkg/model/convert.go
```diff
@@ -0,0 +1,526 @@
+package model
```
frrist, line 1: Here is where model conversion occurs. The file contains methods for going from old types to new ones and vice versa. I have explicitly removed struct key names in the conversation method to ensure builds fail if a field is added or removed from an existing type as a reminder to update conversion methods.

pkg/requester/publicapi/client.go
```diff
@@ -7,14 +7,16 @@ import (
 	"strings"
```
frrist, line 7: The requester API client only takes as input or outputs the `v1beta2` types.

pkg/requester/publicapi/wrapper_client.go
```diff
@@ -0,0 +1,155 @@
...
+	"context"
+
+	"github.com/gorilla/websocket"
+
+	"github.com/bacalhau-project/bacalhau/pkg/bidstrategy"
+	"github.com/bacalhau-project/bacalhau/pkg/job"
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+)
+
+type RequesterAPIClientWrapper struct {
```
frrist, line 13: The `RequesterAPIClientWrapper` wraps a `RequesterAPIClient`. `RequesterAPIClientWrapper` accepts `model` types as inputs, converts them to `v1beta2` types for its wrapped client, and converts `v1beta2` type returned from the wrapped client to `model` types.

pkg/requester/publicapi/wrapper_utils.go
```diff
@@ -0,0 +1,39 @@
+package publicapi
```
frrist, line 1: Wrappers used to construct `RequesterAPIServer` params

PR #2625, opened 2023-07-04
feat: modify job submit endpoint to support versions

frrist:
This PR serves as an alternative to https://github.com/bacalhau-project/bacalhau/pull/2623. Instead of duplicating the API methods on the requester it instead modifies the `submit` endpoint to accept a dynamic `JobCreatePayload` type with a `Spec` encoded as bytes (json). The spec is then decoded based on its API version, and if contains a later version (`v1beta2`) converted to the latest API version.
Pros:
- Simpler to implement and smaller change
- Continues to support job creation via old and new job spec file

Cons:
- Commands that output job data (like `describe`) will change as we mutate models.


pkg/model/job.go
```diff
@@ -343,7 +343,7 @@ type JobCreatePayload struct {
 	APIVersion string `json:"APIVersion,omitempty" example:"V1beta1" validate:"required"`
 
 	// The specification of this job.
-	Spec *Spec `json:"Spec,omitempty" validate:"required"`
+	Spec []byte `json:"Spec,omitempty" validate:"required"`
```
frrist, line 346: we determine how to decode this value based on `APIVersion`

pkg/requester/publicapi/client.go
```diff
@@ -260,16 +262,19 @@ func (apiClient *RequesterAPIClient) Submit(
...
 
+	specbytes, err := json.Marshal(j.Spec)
+	if err != nil {
+		return nil, err
+	}
 	data := model.JobCreatePayload{
 		ClientID:   system.GetClientID(),
 		APIVersion: j.APIVersion,
-		Spec:       &j.Spec,
+		Spec:       specbytes,
```
frrist, line 272: Marshal the spec to JSON bytes before submitting.

pkg/requester/publicapi/endpoints_submit.go
```diff
@@ -46,12 +49,58 @@ func (s *RequesterAPIServer) submit(res http.ResponseWriter, req *http.Request)
...
+	default:
+		publicapi.HTTPError(ctx, res, fmt.Errorf("unhandled api version: %s", apiVersion), http.StatusInternalServerError)
+		return
+	}
+
+	// the job has been migrated to the latest version
+	submittedJob := model.Job{
+		APIVersion: model.V1beta3.String(),
+		Spec:       spec,
+	}
```
frrist, line 93: decode the spec based on API version and convert it to the latest type. Allowing client to submit either the "current" (`v1beta2`) version of the spec, or a previous (`v1beta2`)

PR #2627, opened 2023-07-05
implement MetricRecorder

wdbaruni:
MetricRecorder is a helper for recording metrics.Specifically, it provides methods to record latency and faults.
Example usage:
```
	recorder := NewMetricRecorder()
	defer recorder.RecordFault(ctx, faultCounter) // Records fault if success was not reported
	defer recorder.RecordTotalLatency(ctx, totalLatency) // Records total latency since start

	// Alternatively, you can use the combined method to record fault and total latency
	// defer recorder.RecordFaultAndLatency(ctx, faultCounter, totalLatency)

	msg := queue.Receive()
	recorder.RecordLatency(ctx, dequeueLatency) // Records dequeue latency since start

	process(msg)
	recorder.RecordLatency(ctx, processLatency) // Records process latency since queue.Receive()

	recorder.Success()
```


pkg/telemetry/metric_recorder.go
```diff
@@ -0,0 +1,86 @@
...
+	success       bool
+}
+
+func NewMetricRecorder() *MetricRecorder {
+	now := time.Now()
+	return &MetricRecorder{
+		start:         now,
+		lastOperation: now,
+	}
+}
```
frrist, line 41: If I understand correctly, the intent of the `telemetry` package is to be shared by many different parts of the system, (imo) having methods specific to Faults and Latency might not make sense for all use cases. As an alternative we could implement separate recorders for counter and histogram metrics, e.g. `CounterRecorder` and a `DurationRecorder` respectively (naming is hard, but something like that), aiming to make them more generic,. Example usage:
```go
func (w *Worker) processEvaluation(ctx context.Context, evaluation *model.Evaluation) (ack bool) {
	stopwatch := telemetry.RecordDuration()
	defer stopwatch.Stop(ctx, WorkerProcessEvaluationLatancy, EvalTypeAttribute(evaluation.Type))
	
	// the name of the variable indicates what is being recorded rather than its constructor.
	faults := telemetry.RecordCount()

	// Check if worker is shutting down while dequeueing
	if w.isShuttingDown(ctx) {
		log.Warn().Msgf("Worker is shutting down, not scheduling evaluation %s", evaluation.ID)
		return
	}

	// Schedule the evaluation
	scheduler, err := w.schedulerProvider.Scheduler(evaluation.Type)
	if err != nil {
		log.Error().Err(err).Msgf("Failed to retrieve scheduler for evaluation %s", evaluation.ID)
		faults.Inc(ctx, WorkerProcessFaults, EvalTypeAttribute(evaluation.Type))
		return
	}
	if err = scheduler.Process(evaluation); err != nil {
		log.Error().Err(err).Msgf("Failed to process evaluation %s", evaluation.ID)
		faults.Inc(ctx, WorkerProcessFaults, EvalTypeAttribute(evaluation.Type))
		return
	}
	ack = true
	return
}
```
I am unsure of the `Success` pattern for recording a metric, for the specific case of recording faults, its nice UX, but the pattern may not apply well to other usages of a Counter. It also doesn't really make sense for there to be a `Success` method on the recorder for the case of measuring a duration, while it's not called for the case of latency metrics, I don't see them used yet so maybe I am misunderstanding, it's still present on the type and a little confusing as to developer using the package.

 For example, we could also use `telemetry` for something like the below:
```go
func FindHowManyPeersImConnectedToOrWhatever() {
	stopwatch := telemetry.RecordDuration()
	defer stopwatch.Stop(ctx, PeerFindLatency, EvalTypeAttribute(evaluation.Type))
	connections := telemetry.RecordCount()
	badPeers := telemetry.RecordCount()

	for _, p := range peers {
		func() {
			stopwatch := telemetry.RecordDuration()
			// ignore the cardinality issue of recording a peerID as a metric tags, this is just an example.
			defer stopwatch.Stop(ctx, PeerConnectDuration, EvalTypeAttribute(peer.ID))
			if err := p.Conntect(); err != nil {
				badPeers.Inc(ctx, BadPeers, EvalTypeAttribute(peer.ID))
			}
			conntections.Inc(ctx, ConnectedPeers)
			// we could also have a Dec() method on a counter for the case when it's used as a Gauge.
		}()
	}
}
```
The `RecordLatency` and `RecordFault` method could still exist as helpers, but would wrap a more generic counter/duration recorder instead.

frrist (approved):
If this is blocking other in-flight work my comments are non-blocking, otherwise I'd like to explore the alternative before landing.

pkg/telemetry/metric_recorder.go
wdbaruni: Thanks for the feedback. From my experience, most operations would benefit from a metric that records the total latency, metrics that record breakdown of the latency, and a metric that indicate whether the operation failed or not. The `MetricRecorder` is trying to simplify the capture of those common metric patterns. Now there can be counters other the failure of the operation, and you can publish those directory without a recorder as it wouldn't add value.

Also I am not sure about Prometheus, but CloudWatch doesn't play well with stale metrics, or metrics that you publish rarely when a failure or edge case is triggered. The recommendation is to always publish these metrics with 0 zero values if no failures, and with 1 values if there are failures. The `MetricRecord` helps with as it always publishes failure metrics in the `defer` block with zero values if `Success()` was called, or `1` if failure reported before

Having a single recorder makes more sense to me as both latency and failure metrics are commonly published together, and there wouldn't be harm to create a recorder and use it only to publish latency or failure

However, things are a different than what I am used to with CloudWatch and Java, and the patterns I wanted to implement weren't as easy. My plans is to keep iterating on this as I add metrics to the scheduler and new components. It will be easier to evaluate different patterns once we have enough metrics

frrist: > From my experience, most operations would benefit from a metric that records the total latency, metrics that record breakdown of the latency, and a metric that indicate whether the operation failed or not>

Agreed - these are best practices for [Online-serving systems](https://prometheus.io/docs/practices/instrumentation/#online-serving-systems). 
I am also considering the other case's we'll need metrics for, such as [Caches](https://prometheus.io/docs/practices/instrumentation/#caches), [Threadpools](https://prometheus.io/docs/practices/instrumentation/#threadpools), and [Offline Processing](https://prometheus.io/docs/practices/instrumentation/#offline-processing). I assume these will come at a later point.

PR #2629, opened 2023-07-06
Secrets

rossjones:
This PR attempts to provide a basis for wider work around secrets, but providing an initial implementation that encrypts/decrypts environment variables for jobs.  While this doesn't cover all use-cases it will enable the safe transfer of API keys, and any symmetric key values that might be used for decrypting data. 

A new key pair is generated for each node, both requester and compute, from which the public key is distributed as part of the node's NodeInfo and transmitted over the P2P network.  When on the network the public keys are encoded as PKCS #1 using ASN.1 DER. Currently the libp2p code also generates a keypair but this is only used for supplying a secure connection and a verified ID for that node and wouldn't be shareable between nodes (should we want nodes grouped by the keypair they use for secrets). 

## TODO 

* [ ] Expose requester node's public key via API 
* [ ] Have CLI use public key to encrypt environment 
* [ ] Requester node to recrypt the environment once the target compute is known 
* [ ] Compute node should decrypt environment variables before making available to executor


rossjones:
Closing in favour of https://github.com/bacalhau-project/bacalhau/pull/2692 which is much simpler, touches less code, and is easier to reason about.

PR #2637, opened 2023-07-06
Implement pluggable executors 

frrist:
### What 
This PR implements logic to support pluggable executors using the [go-plugin](https://github.com/hashicorp/go-plugin) library. Each commit _should_ be reviewable on its own.
- Pluggable Executors use gRPC for communication, gRPC supports streaming responses allowing plugged executors to stream their logs (logs of the job running within the executor) regardless if they are running in process or as an external process (true plugin).
- StorageProviders are now an attribute of the compute node rather than the executors themselves, which simplifies configuration.
- Executors (wasm and docker) are now responsible for canceling their executions, previously, this was handled by the compute node.
- Result output limits are now provided as params to an executors `Run` command.   
- Pluggable executors are disabled by default, but may be enabled on the dev stack by passing the `--pluggable-executors=true` flag.
- The Makefile has been modified with directives to build and install executor plugins (disabled on Windows).

### How do I run devstack with this?
1. Build bacalhau and the plugins: `make build`
  - Binaries are located in `pkg/executor/plugins/executors/<docker|wasm>` 
2. Install the plugins `make install-plugins`
  - Copies binaries from the aforementioned path to `BACALHAU_PATH/plugins/` e.g `~/.bacalhau/plugins`
3. Start the devstack with plugins enabled: `./bin/darwin_arm64/bacalhau devstack --pluggable-executors=true`

### What's next?
Follow on change https://github.com/bacalhau-project/bacalhau/pull/2594 which modifies the engine spec to support arbitrary engine names


pkg/executor/plugins/docker/plugin/plugin.go
```diff
@@ -0,0 +1,81 @@
+package main
```
frrist, line 1: This is the pluggable executor, right now it is wrapping the docker executor. Later implementation will need to satisfy the `PluggableExecutor` interface (as this is doing) and implement a similar `main` function to be pluggable to a compute node.

pkg/executor/plugins/docker/executor.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/hashicorp/go-plugin"
+
+	"github.com/bacalhau-project/bacalhau/pkg/executor/plugins"
+)
+
+func NewDockerPlugin() (plugins.PluggableExecutor, error) {
```
frrist, line 12: Instantiates the docker executor plugin (assumes its binary has been build). The `Cmd` feild below informs the client of the shelling point to the plugin. The returned value of this method is a client that speaks RPC to the executor.

pkg/executor/plugins/rpc.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"context"
+	"io"
+	"net/rpc"
+
+	"github.com/bacalhau-project/bacalhau/pkg/bidstrategy"
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+)
+
+type ServerRPC struct {
```
frrist, line 12: Server that translates RPC calls to arguments passed to the executor and translates responses from the executor to RPC.

pkg/executor/plugins/rpc.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return err
+	}
+	reply.Result = result
+	return nil
+}
+
+var _ PluggableExecutor = (*ClientRPC)(nil)
+
+type ClientRPC struct {
```
frrist, line 83: The compute node talks to this client which forwards requests to `ServerRPC` as RPC calls.

wdbaruni:
Thanks Forrest for writing this up.

> The implementation of the pluggable docker executor does not support "storage" options as there isn't a simple way to plumb node configuration to a pluggable executor
1. what kind of configurations are you referring to?
1. it seems we are doing something wrong where we are asking the executor to download and prepare the job inputs from storage sources. A separate component should prepare the storage inputs, download them from remote sources, and always ask the executor to run with local inputs.

> Decide if we like the utility provided by go-plugins enough to continue using it, or if we would prefer to implement our own plugin system with greater flexibility around the RPC protocol we use
1. If I understand correctly, streaming responses are possible with go-plugins + gRPC. What are the problems with gRPC that is pushing us to evaluating implementing a new plugin framework?
2. Is streaming responses only needed for logs? or do we foresee other use cases for it? I am really questioning if each executor should implement a way to stream their logs, or maybe we need to revisit that. An alternative is to have each executor log to a local file, and have our logging APIs stream from the local file instead of asking the executor. This should simplify replay of logs and starting from certain offsets

> Explore separating compute nodes out into a standalone binary - this would simplify its configuration, and configuration of its executors.
I would avoid doing that at the moment. In the future I would expect the compute nodes to also expose HTTP APIs not only for debugging and for operators, but also for users to submit jobs directly to the compute node, which can either be handled by the compute node or forwarded to a requester. 
In summary, we don't know yet the common parts that requester and the compute will share, and we should delay the decision to split the binaries until we have more clarity

> Explore standardizing APIs for communicating with various processes:
That makes sense. We should move away from bprotocol between requester and compute nodes. RPC can be an option, but HTTP might make more sense if we plan to expose some compute APIs to the end users

frrist:
> it seems we are doing something wrong where we are asking the executor to download and prepare the job inputs from storage sources.

I think you are right, the set of support storages are an attribute of both the compute node and its executors, this isn't necessary. We can remove the storage dependency from the executors and handle it at the compute node level, which will greatly simplify the configuration of executors I was referring to. Stated simply - a compute node will be configured with a set of supported storage types and make them available as inputs to an executor as a local source (as you stated).

> If I understand correctly, streaming responses are possible with go-plugins + gRPC. What are the problems with gRPC that is pushing us to evaluating implementing a new plugin framework? 

Correct. And there isn't a problem with gRPC - I just wanted to get buy on gRPC as the communication protocol between a compute node and its executor(s) before implementing as it's slightly more complex than the `net/rpc` implementation currently used in this PR. 

The other option would be to implement our own plugin system yielding greater flexibility of the protocol used by a compute node to communicate with its executor(s). However, we'd have to pay the cost of implementing our own plugin system to support this, and I suspect that isn't something we are interested in right now, but I wanted to leave it as an option before implementing - we can always revisit at a later point. 

> Is streaming responses only needed for logs?

At present, yes.

> do we foresee other use cases for it?

Possibly. At present executors are passed a `ResultsDirectory` when they execute a job. Once the execution is complete it's `RunCommandOutput` is written to this directory. After a Requester verifies the execution, the contents of `ResultsDirectory` is published (I think I have this flow right). Perhaps in the future we may wish to bypass this verify step and publish the job output as it's executed. I could imagine a workflow like this requiring the ability to stream data from an executor directly to a an output (or to an intermediary that passes along the stream), e.g. an S3 bucket or HTTP endpoint. I would guess it's probably worth leaving this option open. Adopting gRPC, as stated above, leaves this option open for future implementations. 

> In summary, we don't know yet the common parts that requester and the compute will share, and we should delay the decision to split the binaries until we have more clarity
> That makes sense. We should move away from bprotocol between requester and compute nodes. RPC can be an option, but HTTP might make more sense if we plan to expose some compute APIs to the end users

👍 I won't change anything here in this PR.

________________________

In summary:
- we'll move forward with go-plugins and gRPC as a means of implementing pluggable executors in this PR.
- we'll make storages an attribute of only the compute node and modify it to prepare local storage before passing it to an executor.
- we'll leave streaming logs as a part of the executor for now and revisit it's move the the compute node (as you described) in a follow on.

rossjones (approved):
Couple of hopefully useful comments, but I think it's looking good, and is a step forward.  I did want to ask how we control how many instances of a the plugin are active (if we wanted to perhaps scale up the number of executors).

When I try and run it though, I get the following (where I suspect it is using the Name instead of Command somewhere). 

```
bacalhau on  frrist/engine/pluggable-executor [$] is 🐹 v1.20.6 
❯ ./bin/darwin_arm64/bacalhau devstack --pluggable-executors
stat /Users/ross/.bacalhau/plugins/Docker: no such file or directory
```
Think it might be a mix of case of executable name and built writing into the source tree. Renaming and putting it where expected make it work.

pkg/executor/util/pluggable_executor.go
```diff
@@ -0,0 +1,130 @@
...
+	if ok {
+		return fmt.Errorf("duplicate registration of exector %s", config.Name)
+	}
+
+	if pluginBin, err := os.Stat(filepath.Join(config.Path, config.Name)); err != nil {
+		return err
+	} else if pluginBin.IsDir() {
+		return fmt.Errorf("plugin location is directory, expected binary")
+	}
+	// TODO check if binary is executable
```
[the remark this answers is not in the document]
frrist: It's a "thingy" that can provide plugins (implements the provider interface), register plugins, as well as start and stop plugins. I think Manager is fitting 👍

pkg/executor/util/pluggable_executor.go
```diff
@@ -0,0 +1,130 @@
...
+	"path/filepath"
+
+	"github.com/hashicorp/go-plugin"
+
+	"github.com/bacalhau-project/bacalhau/pkg/executor"
+	"github.com/bacalhau-project/bacalhau/pkg/executor/plugins/grpc"
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+)
+
+func NewPluggableExecutor() *PluggableExecutor {
```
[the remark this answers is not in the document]
frrist: It's a "thingy" that can provide plugins (implements the provider interface), register plugins, as well as start and stop plugins. I think Manager is fitting 👍

pkg/executor/util/pluggable_executor.go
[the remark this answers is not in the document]
frrist: eghick yeah windows will make this hard, didn't think about that. Alright if we leave this TODO here for now and assume good faith for the time being?

pkg/executor/plugins/executors/docker/main.go
```diff
@@ -0,0 +1,54 @@
...
+
+func main() { // Create an hclog.Logger
+	logger := hclog.New(&hclog.LoggerOptions{
+		Name:   "docker-plugin",
+		Output: os.Stderr,
+		Level:  hclog.Trace,
+	})
+	cm := system.NewCleanupManager()
+	dockerExecutor, err := docker.NewExecutor(
+		context.TODO(),
```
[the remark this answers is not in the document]
frrist: Whoops, yeah should only be creating this once - fixed in 028b0bc279ab90ddb573f42822d340a9371a8326

cmd/cli/devstack/devstack.go
```diff
@@ -120,6 +121,10 @@ func NewCmd() *cobra.Command {
 		&ODs.AllowListedLocalPaths, "allow-listed-local-paths", ODs.AllowListedLocalPaths,
 		"Local paths that are allowed to be mounted into jobs",
 	)
+	devstackCmd.PersistentFlags().BoolVar(
```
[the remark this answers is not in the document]
frrist: There are still some changes I need to make around redirecting executor logs to a file before this is production ready. If it's alright I'd like to enable them in a follow on?

pkg/executor/plugins/grpc/proto/executor.proto
```diff
@@ -0,0 +1,125 @@
...
+  bool Installed = 1;
+}
+
+message ShouldBidRequest {
+  bytes BidRequest = 1;
+}
+
+message ShouldBidBasedOnUsageRequest {
+  bytes BidRequest = 1;
+  bytes Usage = 2;
```
wdbaruni, line 35: Similar to a comment I left above, we should define the actual `BidRequest` and `Usage` fields in the proto schema instead of just bytes.

Also a similar comment I left for `JobStore` PR that @rossjones is working on. We should wrap requests (and possibly responses) in an envelope to help maintain backward compatibility with future changes. Meaning all requests that we send over the write should be wrapped in an envelope that might look like:

```
syntax = "proto3";

message Envelope {
    string version = 1;
    string timestamp = 2;
    Metadata metadata = 3;
    bytes payload = 4; // or maybe use `any`
    string checksum = 5;

    message Metadata {
        string serialization_format = 1;
        // ... any other metadata fields you need.
    }
}

```

**Envelope Schema**

- version: (String) Represents the version of the schema used for the payload. When a backward-incompatible change to the object schema occurs, this version number can be incremented. Code that reads this data can then decide on different processing paths based on the version.

- timestamp: (String) Represents the time at which the data was stored or transmitted. It's typically in ISO 8601 format, for example "2023-07-30T13:45:00Z". This can aid in conflict resolution, debugging, and tracing.

- metadata: (Object) This is a map that stores additional contextual information about the object or the transmission. It can include fields such as:
  - serialization_format: (String) This specifies the serialization format used for the payload. It allows for the payload to be properly deserialized, even if the serialization method changes in the future.
  - Additional fields can be added as required to provide further contextual information about the payload or the transmission.
- payload: (Byte Array) This contains the serialized form of the actual object being stored or transmitted. The method of serialization is indicated in the serialization_format field in the metadata object.

- checksum: (String) This contains a cryptographic hash of the payload. It's used to verify the integrity of the payload data, ensuring it hasn't been tampered with during storage or transmission. When the data is read, the hash of the payload can be computed and compared to this value to verify they match.

This schema is designed to be flexible and can be extended or adjusted based on specific needs and constraints. The primary goal is to ensure backward compatibility, maintain data integrity, and provide flexibility for future changes in serialization methods and object schema.

pkg/executor/plugins/grpc/server.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	result, err := s.Impl.Run(ctx, args)
+	if err != nil {
+		return nil, err
+	}
+	b, err := json.Marshal(result)
+	if err != nil {
+		return nil, err
+	}
+	return &proto.RunCommandResponse{Params: b}, nil
```
wdbaruni, line 32: this is a blocking call and the rpc request will return after the execution is completed, which can be an hour today, and days in the future. Wouldn't that cause issues? I understand this is the case today with our current executors, but those are in-process executors and are running on a separate goroutine

pkg/executor/util/utils.go
```diff
@@ -146,3 +145,27 @@ func NewNoopExecutors(config noop_executor.ExecutorConfig) executor.ExecutorProv
...
+	for _, cfg := range pluginOptions.Plugins {
+		if err := pe.RegisterPlugin(cfg); err != nil {
+			return nil, err
+		}
+	}
+	if err := pe.Start(ctx); err != nil {
+		return nil, err
+	}
+
+	cm.RegisterCallbackWithContext(pe.Stop)
```
wdbaruni, line 168: I am hoping to minimize our usage of cleanup manager inside different places of the code. I am hoping that we limit its usage to `/pkg/node` package and have better control on the sequence on how we close our components

pkg/executor/plugins/grpc/proto/executor.proto
frrist: Adding an Envelope sounds good, I can make that change in this PR. However, as I mentioned [above](https://github.com/bacalhau-project/bacalhau/pull/2637#discussion_r1279665465) I am hesitant to implement protobuf messages for all of our models at this point in implementation.

Would serializing the go structure as json bytes in the envelop payload field be a reasonable compromise?

pkg/executor/util/utils.go
frrist: I am following the same pattern use by the `New*Provider` methods in the package, is that alright for now or should I do something different?

I'd like to remove usage of the CleanupManager completely, and I think we can achieve that using something like FX https://uber-go.github.io/fx/lifecycle.html#lifecycle-hooks

pkg/executor/plugins/grpc/client.go
```diff
@@ -0,0 +1,129 @@
...
+	resp, err := c.client.IsInstalled(ctx, &proto.IsInstalledRequest{})
+	if err != nil {
+		return false, err
+	}
+	return resp.Installed, nil
+}
+
+func (c *GRPCClient) ShouldBid(ctx context.Context, request bidstrategy.BidStrategyRequest) (bidstrategy.BidStrategyResponse, error) {
+	b, err := json.Marshal(request)
+	if err != nil {
```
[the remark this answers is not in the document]
frrist: Added TODO as code comment linking to issue: https://github.com/bacalhau-project/bacalhau/issues/2700

pkg/executor/plugins/grpc/server.go
frrist: Added TODO as code comment linking to issue: https://github.com/bacalhau-project/bacalhau/issues/2702

pkg/executor/util/pluggable_executor.go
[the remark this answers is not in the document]
frrist: done in 55da1450ac8fd6afb38860d6cbc7388e1319e48d

pkg/executor/plugins/executors/wasm/Makefile
```diff
@@ -0,0 +1,5 @@
+wasm: main.go
+	go build -o ./bin/wasm main.go
```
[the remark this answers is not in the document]
frrist: done in 55da1450ac8fd6afb38860d6cbc7388e1319e48d

pkg/executor/plugins/executors/docker/Makefile
```diff
@@ -0,0 +1,5 @@
+docker: main.go
+	go build -o ./bin/docker main.go
```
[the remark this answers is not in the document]
frrist: done in 55da1450ac8fd6afb38860d6cbc7388e1319e48d

pkg/executor/plugins/grpc/proto/executor.proto
frrist: Filed issue for this work in https://github.com/bacalhau-project/bacalhau/issues/2701

wdbaruni (approved):

PR #2642, opened 2023-07-10
chore: update Kubo to v0.21

Jorropo:
v0.19.1 is the latest bug free enough version, anything bellow that is considered completely borked.

This new release bring also a new intresting feature, github.com/ipfs/go-ipfs-http-client has been moved to github.com/ipfs/kubo/client/rpc, this is because we were constantly failing at keeping them in sync.

Now that they are in one repo it's way easier to keep them in sync (the compiler creates compilation failures and you don't need to solve a cycle of PRs where neither is green because they both need each other for CI to pass).


rossjones (approved):

pkg/ipfs/node.go
```diff
@@ -332,7 +332,6 @@ func serveAPI(cm *system.CleanupManager, node *core.IpfsNode, repoPath string) (
 	// Options determine which functionality the API should include:
 	var opts = []corehttp.ServeOption{
 		corehttp.VersionOption(),
-		corehttp.GatewayOption(false),
```
Jorropo, line 335: Support for writable gateway has been removed from Kubo.

rossjones:
Thanks for the PR @Jorropo, very much appreciated.

PR #2643, opened 2023-07-10
implement batch job scheduler

wdbaruni:
First implementation of a job scheduler thats gets handled an evaluation, and generates a plan of updates, including new executions and stopping existing ones if no longer healthy or if over-subscribed.

Some work still remains for follow up PRs, including:
- Implementation of a `Planner` to update the `JobStore` and notify compute nodes with new executions
- Implementation of `SystemBatch` jobs that targets all matching nodes
- Implementation of `NodeManager` that keeps track of node status, including nodes that are disconnected but still have active executions. 
- Cover more edge cases 
- Deprecate `MinBids`. It is adding more complexity than value based on the direction we are heading


pkg/orchestrator/scheduler/batch_job.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	return b.planner.Process(ctx, plan)
+}
+
+// failUnhealthyExecs marks executions that are running on nodes that are not healthy as failed
+func (b *BatchJobScheduler) failUnhealthyExecs(
+	ctx context.Context, existingExecutions []*model.ExecutionState, nodeInfos map[string]*model.NodeInfo, plan *models.Plan) error {
+	for _, execution := range existingExecutions {
+		if _, ok := nodeInfos[execution.NodeID]; !ok {
```
rossjones, line 128: For this case, where the node has dropped off the network, we're marking the execution as stopped even though the compute node may not find out about it and we won't be able to connect to the compute node.  When the compute node restarts/rejoins, it'll have this execution in store.  For batch jobs like this, when the compute node restarts, I was considering marking anything active as stopped/cancelled on a reboot and trying to report the new state back to the requester node. Does it make sense to do that?

pkg/orchestrator/scheduler/batch_job.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// existingNodeInfos returns a map of nodeID to NodeInfo for all the nodes that have executions for this job
+func (b *BatchJobScheduler) existingNodeInfos(ctx context.Context, existingExecutions []*model.ExecutionState) (map[string]*model.NodeInfo, error) {
+	out := make(map[string]*model.NodeInfo)
+	checked := make(map[string]struct{})
+
+	// TODO: implement a better way to retrieve node info instead of listing all nodes
+	nodesMap := make(map[string]*model.NodeInfo)
+	discoveredNodes, err := b.nodeDiscoverer.ListNodes(ctx)
```
rossjones, line 142: As I start looking at a persistent job store, would the job store be the correct place to implement something like `NodesForJob(jobID string) map[executionID]NodeID` to improve this lookup, or is there an issue with using job id in this context?

pkg/orchestrator/scheduler/batch_job.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+type BatchJobSchedulerParams struct {
+	JobStore       jobstore.Store
+	Planner        orchestrator.Planner
+	NodeDiscoverer requester.NodeDiscoverer
+	NodeRanker     requester.NodeRanker
+}
+
+func NewBatchJobScheduler(params *BatchJobSchedulerParams) *BatchJobScheduler {
```
rossjones, line 35: Are we planning to add anything to the Job model to identify what type of job it is (batch, system, cron etc)?

rossjones (approved):
Looks good, couple of questions and a suggestion around UpdateExecutionRequest.

pkg/orchestrator/scheduler/batch_job.go
wdbaruni: That might make sense. I'll see what APIs we need after NodeManager's implementation

pkg/orchestrator/scheduler/batch_job.go
wdbaruni: Yes, eventually. There are two options to consider https://www.notion.so/expanso/Modular-Architecture-In-Depth-2bd19c8acbe046c298319045edde18af

Still though, my next plan is to introduce a new internal job type that is different than we expose to end users through the CLI or APIs.

pkg/orchestrator/scheduler/batch_job.go
```diff
@@ -0,0 +1,202 @@
...
+	if desiredExecutionCount > 0 {
+		selectedNodes, err := b.selectNodes(ctx, &job, desiredExecutionCount)
+		if err != nil {
+			return err
+		}
+		for i := 0; i < desiredExecutionCount; i++ {
+			executionID := model.ExecutionID{
+				JobID:       job.Metadata.ID,
+				NodeID:      selectedNodes[i].PeerInfo.ID.String(),
+				ExecutionID: "e-" + uuid.NewString(),
```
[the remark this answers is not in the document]
wdbaruni: This is great. I'll create helper functions to generate the different IDs for the different objects with some object identifying prefix

pkg/models/plan.go
```diff
@@ -0,0 +1,65 @@
...
+	StoppedExecutions []*jobstore.UpdateExecutionRequest
+}
+
+// NewPlan creates a new Plan instance.
+func NewPlan(eval *Evaluation) *Plan {
+	return &Plan{
+		EvalID:            eval.ID,
+		Priority:          eval.Priority,
+		NewExecutions:     []*model.ExecutionState{},
+		StoppedExecutions: []*jobstore.UpdateExecutionRequest{},
```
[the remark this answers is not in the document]
wdbaruni: That makes sense. I am still not sure if creating `jobstore` requests is the right thing in the plan or not. Will keep this in mind as I introduce the other schedulers

PR #2647, opened 2023-07-10
remove MinBids feature

wdbaruni:
Removing the feature to wait for `MinBids` before starting accepting or rejecting compute nodes bids. It is adding a significant complexity. Also its value when operating on public networks, which lilypad project has moved to a layer above than bacalhau to achieve this.


rossjones (approved):
👍🏻

PR #2650, opened 2023-07-11
Persistent Jobstore

### LINKED ISSUE #2645: Add bolt-db backed job store
Implement a bolt-db backed jobstore to expand the range of options and improve reliability of the requester node. This should follow the same pattern as the execution store, along with clear in-code documentation on the schema.

rossjones:
Currently we have only an in-memory implementation of a jobstore, and this isn't very resistant to failures.  This PR introduces a boltdb backed jobstore to ensure jobs and their state and remember across restarts of the requester node.

The jobs in the database are structured as follows, with uppercase names reflecting dynamic values, and `key:val` showing key-value mappings. In cases where the the path does not end in a `key:val` entry, these are the index buckets described below. 

### Buckets

```
Jobs 
    |---> bucket <JobID>
                |----> key:Spec value:{model.Job}
                |----> key:State value:{model.JobState}
                |----> bucket job_history
                                       |----> key:seq value:{model.JobHistory}
                |----> bucket execution_history
                                       |----> key:seq value:{model.JobHistory}
                |----> bucket executions
                                       |----> key:execID value:{model.ExecutionState}
                |----> bucket evaluations
                                       |----> key:evalID value:{models.Evaluation}
```                  

### Indexes

Some of the entries in the database, such as the list of Job IDs for a given client-id are actually indexes.  These are structured as buckets, with the root being "idx_NAME". So for client ids we have "idx_clients" and then the next bucket showing the "CLIENT-ID" and then a list of buckets showing the jobs from the client-id.  They could have been structured as buckets labelled "clientid;jobid" and found with partial matches but this requires we iterate over all of the buckets.  This way we can do a direct lookup on the clientid, resulting in iterating many fewer buckets.

Current indexes are 

* TagsIndex =  tag -> Job id
* ProgressIndex = job-id -> {}
* ClientsIndex   =  client-id -> Job id
* EvaluationsIndex = evaluation-id -> Job id

### Bucket paths

This PR also introduces bucket paths which are used for finding buckets from strings like "root.bucket.with.subbucket". This performs the same number of lookups as doing this work each time, but also provides the option to create any buckets in the path that don't yet exist.

### Watching events 

The jobstore now contains a method (Watch) which will provide the caller with an event channel filtered to their requirements.  By specifying the event types (Create, Update, Delete) and the target objects (JobWatcher or ExecutionWatcher) the returned channel will receive events that only match those criteria. The `WatchEvent` received on the channel also contains a timestamp, and a `[]byte` format json-encoded of the related object type.  

|Watcher type|Event type|Returned object|
|--|--|--|
| JobWatcher | Create | Job |
| JobWatcher | Update | JobState |
| JobWatcher | Delete | Job |
| ExecutionWatcher | Create | Execution |
| ExecutionWatcher | Update | Execution |
| ExecutionWatcher | Delete | - |
|EvaluationWatcher|Create| Evaluation |
|EvaluationWatcher|Update| - |
|EvaluationWatcher|Delete|Evaluation|

Resolves #2645


pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	bktJobs := tx.Bucket(BucketJobs)
+	bktState, err := NewBucketPath(BucketPathState).Get(tx, false)
+	if err != nil {
+		return nil, err
+	}
+
```
wdbaruni, line 434: sometimes you are instantiating bucket directly and other times you are using bucket path, which can get a little confusing. What is the reason for not instantiating the state bucket directly here? The bucket should already exist as it was created in the constructor

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+		JobID:      job.Metadata.ID,
+		State:      model.JobStateNew,
+		Version:    1,
+		CreateTime: b.clock.Now().UTC(),
+		UpdateTime: b.clock.Now().UTC(),
+	}
+	data, err := json.Marshal(jobState)
+	if err != nil {
+		return err
+	}
```
wdbaruni, line 557: as discussed earlier, adding an envelope around each payload before it is written to the data store can improve our versioning, backward compatibility and resiliency to future changes, such as changing the serialization from json to a binary format

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+			return err
+		}
+	}
+
+	if err = b.inProgressIndex.Add(tx, BucketJobsInProgress, jobIDKey); err != nil {
+		return err
+	}
+
+	if job.Metadata.ClientID == "" {
+		fmt.Println("TEST missing clientID in job")
```
wdbaruni, line 586: extra print statement

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+
+	// Write sentinels keys for specific tags
+	for _, tag := range job.Spec.Annotations {
+		tagBytes := []byte(strings.ToLower(tag))
+		if err = b.tagsIndex.Add(tx, tagBytes, jobIDKey); err != nil {
+			return err
+		}
+	}
+
+	b.triggerEvent(jobstore.JobWatcher, jobstore.CreateEvent, jobData)
```
wdbaruni, line 602: the events are triggered within the transaction, which is great as it guarantees ordering of events published to match the order of mutations and updates to the underlying store. However, a slow event consumer can block the transaction, and you might see a deadlock if the event consumer tries to read data from the datastore while processing the event.

I think the transactions should just publish the events to a channel and have another go routine read from the channel and notify the watchers. Does that make sense? Though I don't want to add complexity to this PR and we can do it in a follow up PR if you prefer

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+	// Delete sentinels keys for specific tags
+	for _, tag := range job.Spec.Annotations {
+		tagBytes := []byte(strings.ToLower(tag))
+		if err = b.tagsIndex.Remove(tx, tagBytes, jobIDKey); err != nil {
+			return err
+		}
+	}
+
+	encodedJob, _ := json.Marshal(job)
+	b.triggerEvent(jobstore.JobWatcher, jobstore.DeleteEvent, encodedJob)
```
wdbaruni, line 657: wouldn't it be better to pass the job object instead of its bytes in the event?

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+		return b.updateJobState(tx, request)
+	})
+}
+
+func (b *BoltJobStore) updateJobState(tx *bolt.Tx, request jobstore.UpdateJobStateRequest) error {
+	job, err := b.getJob(tx, request.JobID)
+	if err != nil {
+		return jobstore.NewErrJobNotFound(request.JobID)
+	}
+
```
wdbaruni, line 675: do you need to pull the job object?

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+	return b.database.Update(func(tx *bolt.Tx) (err error) {
+		return b.updateExecution(tx, request)
+	})
+}
+
+func (b *BoltJobStore) updateExecution(tx *bolt.Tx, request jobstore.UpdateExecutionRequest) error {
+	if !b.jobExists(tx, request.ExecutionID.JobID) {
+		return jobstore.NewErrJobNotFound(request.ExecutionID.JobID)
+	}
+
```
wdbaruni, line 823: when updating an execution, we should just care if the execution exists or not, but not the job. I understand it is a bug where an execution remains but its job was deleted or never created, but this check won't help much

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+	newExecution := request.NewValues
+	if newExecution.CreateTime.IsZero() {
+		newExecution.CreateTime = b.clock.Now().UTC()
+	}
+	if newExecution.UpdateTime.IsZero() {
+		newExecution.UpdateTime = existingExecution.CreateTime
+	}
+	if newExecution.Version == 0 {
+		newExecution.Version = existingExecution.Version + 1
+	}
```
wdbaruni, line 847: we should use createTime of the existing execution, and now() for the update time as the default values.

pkg/jobstore/boltdb/store.go
```diff
@@ -0,0 +1,1013 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	encodedExec, _ := json.Marshal(existingExecution)
+	b.triggerEvent(jobstore.ExecutionWatcher, jobstore.UpdateEvent, encodedExec)
+
+	return b.appendExecutionHistory(tx, newExecution, existingExecution.State, request.Comment)
+}
```
wdbaruni, line 873: we should trigger event after appending execution history and making sure the transaction won't fail

pkg/jobstore/inmemory/inmemory.go
```diff
@@ -41,18 +47,38 @@ func NewJobStore() *JobStore {
 	return res
 }
 
+func (d *InMemoryJobStore) Watch(c context.Context, t jobstore.StoreWatcherType, e jobstore.StoreEventType) chan jobstore.WatchEvent {
+	w := jobstore.NewWatcher(t, e)
+	d.watchers = append(d.watchers, *w)
+	return w.Channel()
```
wdbaruni, line 53: lock block

pkg/jobstore/inmemory/inmemory.go
```diff
@@ -198,19 +214,33 @@ func (d *JobStore) CreateJob(_ context.Context, job model.Job) error {
...
+
+	d.triggerEvent(jobstore.JobWatcher, jobstore.CreateEvent, job)
+
+	return nil
+}
+
+// DeleteJob removes a job from storage
+func (d *InMemoryJobStore) DeleteJob(ctx context.Context, jobID string) error {
+	d.triggerEvent(jobstore.JobWatcher, jobstore.DeleteEvent, d.jobs[jobID])
+
```
wdbaruni, line 232: we should trigger the even the deletion

pkg/jobstore/watcher.go
```diff
@@ -0,0 +1,114 @@
...
+	Object    []byte
+	Timestamp int64
+}
+
+func NewWatchEvent(kind StoreWatcherType, event StoreEventType, object []byte) WatchEvent {
+	return WatchEvent{
+		Kind:      kind,
+		Event:     event,
+		Object:    append([]byte(nil), object...),
+		Timestamp: time.Now().Unix(),
```
wdbaruni, line 62: Timestamp is the time the event was published and not necessary when the change has happened. It can be confusing and I am concerned people will use it wrongly. My recommendation is to remove the field and publish the actual event time, which I understand can be challenging

pkg/jobstore/watcher.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// Watcher is used by the jobstore to keep a record of parties interested in events happening
+// in the jobstore.  This allows for watching of job and execution types (or both), and for
+// create, update and delete events (or any combination).
+type Watcher struct {
+	types       StoreWatcherType // a bitmask of types being watched
+	events      StoreEventType   // a bitmask of events being watched
```
wdbaruni, line 71: a watcher is only watching a single object type and single event type, but the field names and comments say differently

pkg/jobstore/watcher.go
```diff
@@ -0,0 +1,114 @@
...
+	if len(w.channel) == w.channelSize && !allowBlock {
+		return false
+	}
+
+	w.channel <- WatchEvent{
+		Kind:   kind,
+		Event:  event,
+		Object: object,
+	}
+	return true
```
wdbaruni, line 109: I see. So the watcher is just a buffer that is non-blocking and enqueues events for others to consume. The only issue here is if a consumer is interested in all evaluations that were created and updated, they will have to create two watchers, but the order of events can interleave since each watcher is maintaining their own ordered events. A consumer might poll an updated evaluation event before getting created evaluation event from the other watcher

pkg/requester/endpoint.go
```diff
@@ -121,6 +121,13 @@ func (e *BaseEndpoint) SubmitJob(ctx context.Context, data model.JobCreatePayloa
...
 		ModifyTime:  job.Metadata.CreatedAt.UnixNano(),
 	}
+
+	err = e.store.CreateEvaluation(ctx, *eval)
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msgf("failed to save evaluation for job %s", jobID)
+		return job, err
+	}
+
 	err = e.evaluationBroker.Enqueue(eval)
```
wdbaruni, line 131: since the requester is creating an evaluation in the datastore, it doesn't need to enqueue the evaluation to the broker since there should be a watcher that polls from the data store and enqueues to the broker. This goes for all other evaluations here

Though it would be nice to create the job and evaluation in the datastore in a single transaction. I understand that is not easy to do now, and we can delay it to a later stage. A TODO would be nice for now :)

wdbaruni (commented):
Looks great. Nothing major

pkg/jobstore/watcher.go
rossjones: Actually, if you want multiple events for a type, you can just use the following to get create and update events, and then they'll arrive in order.

```
WatchEvent(jobstore.EvaluationWather, jobstore.CreateEvent | jobstore.UpdateEvent)
```

cmd/cli/serve/serve.go
```diff
@@ -301,10 +303,18 @@ func serve(cmd *cobra.Command, OS *ServeOptions) error {
...
-	if err != nil {
-		return fmt.Errorf("error creating in memory datastore: %w", err)
+	var datastore jobstore.Store
+	jobStoreConf := config.GetJobStoreConfig(libp2pHost.ID().String())
+	if jobStoreConf.StoreType == config.JobStoreBoltDB {
+		datastore, err = boltjobstore.NewBoltJobStore(jobStoreConf.Location)
+		if err != nil {
+			return fmt.Errorf("error creating datastore: %w", err)
+		}
+		cm.RegisterCallbackWithContext(datastore.Close)
```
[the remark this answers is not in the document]
rossjones: Yes, that makes a lot of sense.  There's some other stuff that'll need to move as well, in particular the job info publisher as that relies on the jobstore, but that isn't too problematic.   
Fixed in 5bbd6177

pkg/jobstore/boltdb/database.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return bucket, nil
+}
+
+// BucketSequenceString returns the next sequence in the provided
+// bucket, formatted as a 3 character padded string to ensure that
+// bolt's lexicographic ordering will return them in the correct
+// order
+func BucketSequenceString(_ *bolt.Tx, bucket *bolt.Bucket) string {
```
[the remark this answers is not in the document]
rossjones: When I looked back at where I got the usage pattern from, it uses 16 leading chars. which is a lot of history for an individual job's history, so will bump it to that.

pkg/jobstore/boltdb/store.go
rossjones: Have changed to be a bit more consistent.

pkg/jobstore/boltdb/store.go
rossjones: I might attempt this in another PR if that's okay.  

If we take this approach, won't the new go routine reading from the channel block if any of the called watchers block when handling the notification?   Or do we need a new goroutine per notification and hope that the recipient returns at some point?

pkg/jobstore/boltdb/store.go
rossjones: Fixed in the InMemory store as well.

pkg/jobstore/inmemory/inmemory.go
rossjones: I've noticed there's actually an OnCommit() we can use for the transaction to do the notify once we know it'll be successful.  Which feels cleaner than the current way where there is implied knowledge that returning an err means a fail means no notification.

pkg/jobstore/watcher.go
rossjones: Watchers can mask the watcher type and the events (see https://github.com/bacalhau-project/bacalhau/pull/2650#discussion_r1275916813) 

So `store.JobWatcher | store.EvaluationWatcher` is valid, as is just `store.JobWatcher`.

pkg/jobstore/boltdb/store.go
rossjones: Resolved with OnCommit() handler

pkg/jobstore/boltdb/store.go
rossjones: Agreed

pkg/jobstore/boltdb/store.go
```diff
@@ -281,46 +235,64 @@ func (b *BoltJobStore) getJobs(tx *bolt.Tx, query jobstore.JobQuery) ([]model.Jo
...
+		tagKeys := maps.Keys(tagSet)
+		tagKeysLen := len(tagKeys)
+
+		var jobIDs []string
+		if clientKeysLen > 0 && tagKeysLen > 0 {
+			jobIDs = lo.Intersect(clientKeys, tagKeys)
+		} else if clientKeysLen > 0 && tagKeysLen == 0 { // being explicit
+			jobIDs = clientKeys
+		} else if tagKeysLen > 0 && clientKeysLen == 0 { // being explicit
+			jobIDs = tagKeys
```
wdbaruni, line 290: we shouldn't check if the keys are empty, but if the query specified clientID or IncludeTags. If the user specified a tag that didn't have any matching jobs, then we return all jobs which is not the desired behaviour

pkg/jobstore/boltdb/store.go
```diff
@@ -421,28 +407,25 @@ func (b *BoltJobStore) getInProgressJobs(tx *bolt.Tx) ([]model.JobWithInfo, erro
...
-	bktJobs := tx.Bucket(BucketJobs)
-	bktState, err := NewBucketPath(BucketPathState).Get(tx, false)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, key := range keys {
+	for _, jobIDKey := range keys {
 		var job model.Job
 		var jobState model.JobState
```
wdbaruni: I don't think we are populating the job state with executions here

pkg/jobstore/envelope.go
```diff
@@ -0,0 +1,63 @@
...
+
+import "encoding/json"
+
+// Envelope provides a wrapper around types that can be stored in a jobstore.
+// It takes responsibility for ser/de for the wrapped type, ensuring that
+type Envelope[T any] struct {
+	Body      T
+	marshal   MarshalFunc
+	unmarshal UnmarshalFunc
+}
```
wdbaruni, line 11: Nice. We need to include other metadata to help deserialize the object and maintain backward compatibility. At a minimum we need a `version` field, which will be the version of the envelope. If we decide to include more fields or blocks in the future (e.g. encoding type, object type, object version), we can bump the version and add those later

pkg/jobstore/boltdb/index.go
```diff
@@ -18,50 +36,35 @@ func NewIndex(bucketPath string) *Index {
...
 		return err
 	}
 
-	bktLabel, err := bkt.CreateBucketIfNotExists(label)
-	if err != nil {
-		return err
-	}
-
-	_, err = bktLabel.CreateBucketIfNotExists(identifier)
+	_, err = bkt.CreateBucketIfNotExists(identifier)
```
wdbaruni: I thought I left this comment in my previous review, but I can't find it! In the index, we are creating a bucket for each index value/identifier instead of having a single bucket and each identifier is a key (with empty value) in that bucket.

I don't know how boltdb works internally, but my assumption is that buckets are more expensive to create and query than key/value entries in existing buckets. If this assumption is true, then we we need to revisit how indexes are implemented, and minimize our use of dynamic buckets in other parts of the jobstore.

I understand I suggested exploring nested buckets with a parent `jobs` bucket and have executions, spec and state as child buckets, but please push back if that is not an efficient way to implement boltdb schemas

pkg/jobstore/boltdb/store.go
rossjones: Fixed in e7bdaa55

pkg/jobstore/envelope.go
rossjones: I'll copy the pattern suggestedi n the plugin PR. I wonder whether we should have the Envelope only work for structs that implement a VersionedType struct which could supply the version number as part of it's interface?

pkg/jobstore/boltdb/index.go
rossjones: Yes, you're absolutely right, I thought Buckets were really cheap to create, but turns out each one is a newly create btree and so there's a lot of looking up of the location of other btrees that seems unnecessary.  Luckily fixing this was pretty straight-forward in `index.go` so I've switched it to writing 0 byte values.  The only downside of this approach I've seen is that there's a temptation to have a single bucket and prefix the keys, but then we trade a lot of space for the single occassional lookup, so I've left it with a btree per index for now. Impl in fe8f5b7fe8f5b772d01951392e63b8abbcc5df826d1d2a8

pkg/jobstore/boltdb/store.go
wdbaruni: Actually the expected behaviour is:
- `ReturnAll` should ignore the `clientID` filter and will include jobs for different clients
-`IncludeTags` if specified, will filter out any jobs that don't include the specified tags
- `ExcludeTags` if specified, will filter out any jobs that has the specified tags

The filer names are confusing and we need to revisit them. Since you are away, I can update this code to get your PR merged.

pkg/jobstore/envelope.go
wdbaruni: Yeah having a unified struct or at least a pattern for jobstore and grpc/http requests would be nice. I won't be blocking the PR for this change and we can do it when you are back and before our next release

wdbaruni (approved):
Approving the changes. We can work on improving the envelope on a follow up PR

PR #2675, opened 2023-07-17
refactor: executor interface and storage dependency

frrist:
### What
The goal of this PR is to simplify the implementation #2637 by removing the requirement of configuring storage providers for executors. It includes several refactors to the bid strategies, executors interface, and corresponding implementation.
- Executors no longer contain storage providers. Instead the compute node responsible for orchestrating executors maintains the set of available storage providers. The compute node makes storage available to executors as local volumes. 
- ParallelPrepareStorage now returns a slice of prepared storage (instead of a map). This allows the returned `PreparedStorage` type to be serialized and passes across RPC boundaries.
- `NewModuleLoader` will not fetch remote wasm modules, and instead operates over a slice of `PreparedStorage`.
  - This change was required to remove the Storage Provider type from the wasm executor which previously fetched remote content from `Job.Input` during execution.
- Modifies the `Executor` interface by removing methods related to providing storage and reducing the complexity of parameters passes to the `Run` method.

### DIFF
--- a/pkg/compute/executor.go
+++ b/pkg/compute/executor.go
@@ -5,18 +5,23 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/rs/zerolog/log"
+
 	"github.com/bacalhau-project/bacalhau/pkg/compute/store"
 	"github.com/bacalhau-project/bacalhau/pkg/executor"
+	"github.com/bacalhau-project/bacalhau/pkg/executor/wasm"
 	"github.com/bacalhau-project/bacalhau/pkg/model"
 	"github.com/bacalhau-project/bacalhau/pkg/publisher"
+	"github.com/bacalhau-project/bacalhau/pkg/storage"
+	"github.com/bacalhau-project/bacalhau/pkg/system"
 	"github.com/bacalhau-project/bacalhau/pkg/util/generic"
-	"github.com/rs/zerolog/log"
 )
 
 type BaseExecutorParams struct {
 	ID                     string
 	Callback               Callback
 	Store                  store.ExecutionStore
+	Storages               storage.StorageProvider
 	Executors              executor.ExecutorProvider
 	ResultsPath            ResultsPath
 	Publishers             publisher.PublisherProvider
@@ -30,6 +35,7 @@ type BaseExecutor struct {
 	callback         Callback
 	store            store.ExecutionStore
 	cancellers       generic.SyncMap[string, context.CancelFunc]
+	Storages         storage.StorageProvider
 	executors        executor.ExecutorProvider
 	publishers       publisher.PublisherProvider
 	resultsPath      ResultsPath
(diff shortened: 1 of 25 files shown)


pkg/compute/executor.go
```diff
@@ -41,13 +47,75 @@ func NewBaseExecutor(params BaseExecutorParams) *BaseExecutor {
...
+			return nil
+		})
+		engineArgs = &wasm.Arguments{
+			EntryPoint:           execution.Job.Spec.Wasm.EntryPoint,
+			Parameters:           execution.Job.Spec.Wasm.Parameters,
+			EnvironmentVariables: execution.Job.Spec.Wasm.EnvironmentVariables,
+			EntryModule:          entryModuleVolumes[0],
+			ImportModules:        importModuleVolumes,
+		}
+	}
```
frrist, line 102: I hate special casing the wasm engine, but for the time being this is workable since wasm is a "builtin" executor. This change fetches wasm modules on the compute node (from either IPFS or a URL) and makes them available locally before running the wasm executor. Previously the wasm executor was making network calls to compile the wasm binary before executing it in process.

pkg/executor/docker/executor.go
```diff
@@ -44,38 +45,31 @@ type Executor struct {
 	// used to allow multiple docker executors to run against the same docker server
 	ID string
 	// the storage providers we can implement for a job
-	StorageProvider storage.StorageProvider
```
frrist, line 47: Docker executor no longer contains a storage provider and thus no longer needs to (or can) implement the `HasStorageLocally` and `GetVolumeSize` methods.

pkg/executor/docker/executor.go
```diff
@@ -200,39 +167,43 @@ func (e *Executor) Run(
...
+			jsonJobSpec, err := model.JSONMarshalWithMax(job.Spec)
+			if err != nil {
+				return executor.FailResult(err)
+			}
+			log.Ctx(ctx).Debug().Msgf("Job Spec JSON: %s", jsonJobSpec)
+
+		useEnv := append(job.Spec.Docker.EnvironmentVariables,
+			fmt.Sprintf("BACALHAU_JOB_SPEC=%s", string(jsonJobSpec)),
+		)
+	*/
```
frrist, line 192: Need to address this before merging, I don't think it's necessary and I can't find any usage of the `BACALHAU_JOB_SPEC` environment variable. I believe there may have been some docker images that required access to the full job spec for some reason - I have linked to the commit that introduced this change in the code comment for the reviewer.

pkg/storage/parallel.go
```diff
@@ -4,25 +4,31 @@ import (
...
 // storage specs in parallel, and returns a map of specs to their download
 // volume counterparts.
 func ParallelPrepareStorage(
 	ctx context.Context,
 	provider StorageProvider,
-	specs []model.StorageSpec,
-) (map[*model.StorageSpec]StorageVolume, error) {
-	volumes := generic.SyncMap[*model.StorageSpec, StorageVolume]{}
+	specs ...model.StorageSpec,
+) ([]PreparedStorage, error) {
```
frrist, line 25: Returns a slice now as the returned map keyed with pointers cannot cross an RPC boundary.

pkg/executor/wasm/executor.go
```diff
@@ -91,31 +67,31 @@ func (*Executor) GetResourceBidStrategy(context.Context) (bidstrategy.ResourceBi
...
-		stat, err = os.Stat(volume.Source)
+		stat, err = os.Stat(v.Volume.Source)
 		if err != nil {
 			return nil, err
 		}
 
 		var inputFs fs.FS
 		if stat.IsDir() {
-			inputFs = os.DirFS(volume.Source)
+			inputFs = os.DirFS(v.Volume.Source)
```
rossjones, line 89: I think @simonwo is working on the dir/file breakdown, so just adding this comment to let him know when he's back from vacation.

pkg/executor/wasm/executor.go
```diff
@@ -174,78 +174,59 @@ func (e *Executor) Run(ctx context.Context, executionID string, job model.Job, j
...
 	// filesystem which contains our input and output.
 
-	args := append([]string{job.Spec.Wasm.EntryModule.Name}, job.Spec.Wasm.Parameters...)
-
 	config := wazero.NewModuleConfig().
 		WithStartFunctions().
 		WithStdout(stdout).
 		WithStderr(stderr).
-		WithArgs(args...).
+		WithArgs(append([]string{wasmArgs.EntryModule.Spec.Name}, wasmArgs.Parameters...)...).
```
rossjones, line 208: Not a big deal, but I think I preferred the old way because this is quite dense in the middle of this builder.

pkg/executor/wasm/loader.go
```diff
@@ -60,46 +55,38 @@ func (loader *ModuleLoader) Load(ctx context.Context, path string) (wazero.Compi
...
-		programPath := volume.Source
-
-		var info fs.FileInfo
-		info, err = os.Stat(programPath)
+	// We expect the input to be a single WASM file. It is common however for
+	// IPFS implementations to wrap files into a directory. So we make a special
+	// case here – if the input is a single file in a directory, we will assume
+	// this is the program file and load it.
+	if info.IsDir() {
+		files, err := os.ReadDir(programPath)
```
rossjones, line 74: We should probably correct the method name and method command if we're only returning a single module (previously this method returned several).

pkg/storage/parallel.go
```diff
@@ -4,25 +4,31 @@ import (
...
 	waitgroup := multierrgroup.Group{}
 
-	for _, inputStorageSpec := range specs {
+	out := make([]PreparedStorage, len(specs))
+	for i, inputStorageSpec := range specs {
 		spec := inputStorageSpec // https://golang.org/doc/faq#closures_and_goroutines
+		i := i
```
rossjones, line 31: (nit) Instead of shadowing the same name (I first read it as 1 and got confused), should we give it a different name and capture that in the function instead?

rossjones (approved):
Couple of comments that may/may-not be useful, and one opportunity to move a bit more code out of the executor (creating output dir) but LGTM, definitely a positive change.

pkg/executor/docker/executor.go
```diff
@@ -175,7 +142,7 @@ func (e *Executor) Run(
 			return executor.FailResult(err)
 		}
 
-		srcd := filepath.Join(jobResultsDir, output.Name)
+		srcd := filepath.Join(args.ResultsDir, output.Name)
```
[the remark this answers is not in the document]
frrist: Agreed! Would it be alright if land this as a follow on? 

I think we'll need to do some (re)thinking regarding a potentially different `Output` type, but am unsure what it should resemble. As an example, the wasm executor requires an `Output` to have a `Path`, but doesn't use it - it just joins the results directory with the output name. The docker executor does similar, but uses the `Path` as a target for it's mounted filesystem. 

I think a more ideal case would be passing to each executor a set of inputs and set of outputs as interfaces to a filesystem, but we also need a way to serialize this information across the RPC used by a compute node to orchestrate and executor.

pkg/executor/types.go
```diff
@@ -16,30 +18,40 @@ type ExecutorProvider = model.Provider[model.Engine, Executor]
...
-	Run(
-		ctx context.Context,
-		executionID string,
-		job model.Job,
-		resultsDir string,
-	) (*model.RunCommandResult, error)
+	Run(ctx context.Context, Args *RunCommandRequest) (*model.RunCommandResult, error)
+}
+
+type RunCommandRequest struct {
```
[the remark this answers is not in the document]
frrist: Would that mean EnvironmentVariables becoming its own field on the JobSpec? Generally seems fine to me.

pkg/executor/wasm/loader.go
frrist: Good catch, will also make it private as it's only used internally.

pkg/executor/types.go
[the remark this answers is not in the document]
wdbaruni: I also like the idea of moving Envs outside of the engine spec to the `Job.Spec` level. My only concern would be if an executor and a publisher within the same spec want to use different values for the same env variable. We don't support Envs in publishers today, so it may not be an immediate concern

pkg/executor/docker/network.go
```diff
@@ -87,16 +88,17 @@ func (e *Executor) setupNetworkForJob(
...
-		err = fmt.Errorf("unsupported network type %q", job.Spec.Network.Type.String())
+		err = fmt.Errorf("unsupported network type %q", network.Type.String())
 	}
 	return
 }
 
 //nolint:funlen,gocyclo
 func (e *Executor) createHTTPGateway(
 	ctx context.Context,
+	job string,
```
wdbaruni, line 99: nit: helpful to rename this to jobID

wdbaruni (approved):
Just minor comments that shouldn't block the PR

PR #2680, opened 2023-07-20
bug: Does this manifest a bug in CI?

frrist:


frrist:
Current theory is operations that add time to node construction for devstack interfere with https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/config_defaults.go#L50

wdbaruni:
I am not able to reproduce the issue locally, but it does make sense to only start the node info publisher once the node is ready to handle requests

wdbaruni (approved):

PR #2687, opened 2023-07-25
Fix all outstanding/broken lints

rossjones:
`make lint` failing on main with quite a few complaints.  This PR fixes those that are obvious, and nolints those that are not.

Also removes the complaint about the spelling of cancelled/canceled by way of the lint config.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -62,6 +62,7 @@ linters-settings:
     locale: US
     ignore-words:
     - favour
+    - cancelled
   nolintlint:
     allow-leading-space: true # don't require machine-readable nolint directives (i.e. with no leading space)
     allow-unused: true # report any unused nolint directives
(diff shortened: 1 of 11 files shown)


.golangci.yml
```diff
@@ -162,5 +162,6 @@ run:
 
 issues:
   exclude:
-    # Naming of `error` variables as `err` is standard practise for Go
-    - 'shadow: declaration of "err" shadows declaration at'
+  # Naming of `error` variables as `err` is standard practise for Go
+  - 'shadow: declaration of "err" shadows declaration at'
+  - '`cancelled` is a misspelling of `canceled`'
```
wdbaruni, line 167: THANK YOU! I also see `favour` as an ignored misspelled word. Should we add cancelled/canceled in that section instead of here?

wdbaruni (approved):

PR #2690, opened 2023-07-25
introduce ops job type, a batch job that targets all matching nodes

wdbaruni:
Other key changes with this new job type:
- Requester can now ask a compute node to run a job immediately when there is available capacity without having to first ask the requester for approval (ie BidAccepted or BidRejected). `OpsJob` uses this pattern as there is no reason to approve or reject these types of jobs. `BatchJob` still follows the existing approval approach. The code is still ugly because we are responding to bid requests async, which I am planning to revisit when Simon is back. Also will revisit changing the method names
- The requester doesn't retry when a node fails an `OpsJob` as there is no value in attempting to retry on another node in the network since we've selected all matching nodes


pkg/orchestrator/scheduler/ops_job.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			b.handleFailure(nonTerminalExecs, allFailed, plan, err)
+			return b.planner.Process(ctx, plan)
+		}
+	}
+
+	// mark job as completed if there are no more active or new executions
+	if len(nonTerminalExecs) == 0 && len(newExecs) == 0 {
+		if len(allFailed) > 0 {
+			b.handleFailure(nonTerminalExecs, allFailed, plan, errors.New(""))
```
frrist, line 90: should `errors.New("")` contain a helpful message?

pkg/orchestrator/scheduler/types.go
```diff
@@ -104,6 +106,109 @@ func (set execSet) filterFailed() execSet {
...
+	for _, exec := range set {
+		if _, ok := nodeInfos[exec.NodeID]; !ok {
+			lost[exec.ComputeReference] = exec
+			log.Debug().Msgf("Execution %s is running on node %s which is not healthy", exec.ComputeReference, exec.NodeID)
+		} else {
+			healthy[exec.ComputeReference] = exec
+		}
+	}
+	return healthy, lost
+}
```
frrist, line 140: Potential candidates for named returns

pkg/orchestrator/scheduler/types.go
```diff
@@ -104,6 +106,109 @@ func (set execSet) filterFailed() execSet {
...
+// filterByApprovalStatus partitions executions based on their approval status.
+func (set execSet) filterByApprovalStatus(desiredCount int) executionsByApprovalStatus {
+	nonTermExecs := set.filterNonTerminal()
+	running := nonTermExecs.filterRunning()
+	toApprove := make(execSet)
+	toReject := make(execSet)
+	pending := make(execSet)
+
+	//TODO: we are approving the oldest executions first, we should probably
+	// approve the ones with highest rank first
```
frrist, line 164: What is the rank in this context?

frrist (approved):

pkg/orchestrator/scheduler/ops_job.go
wdbaruni: `handleFailure` will use the failure message from the latest execution, and fall back to the passed error if no failed execution was found, which won't be the case here

pkg/orchestrator/scheduler/types.go
wdbaruni: Nice suggestion. Done in [a274f77](https://github.com/bacalhau-project/bacalhau/pull/2690/commits/a274f777a5a5c220fdce71f78d0936fba186aed1)

pkg/orchestrator/scheduler/types.go
wdbaruni: Nodes are ranked when creating executions. We should persist the rank that was used when selecting nodes in the execution itself, and then discard executions with lower rank if we need to discard some

PR #2692, opened 2023-07-26
WIP: Encrypted environment variables

rossjones:
Adds a `--encrypt` option to `docker run` to encrypt the values for any specified environment variables. This is intended to give us the option to test the functionality before we commit to encrypting all values by default.  It also only currently targets docker because of the different ways that docker and wasm store their environment variables (map vs []string). 

PR is just being kept around so we can cherry-pick any useful commits when we implement this feature.


rossjones:
Closing PR but not removing branch.

PR #2696, opened 2023-07-28
Promotes EnvironmentVariables into the spec.

rossjones:
Currently the environment variables are held (differently) within the engine specs, but they're really a property of the environment where the executor will be run - and so it feels like it might make more sense for them to be part of the JobSpec. They're also held there as a map[string]string rather than []string.

### DIFF
--- a/cmd/cli/docker/docker_run.go
+++ b/cmd/cli/docker/docker_run.go
@@ -189,8 +189,13 @@ func CreateJob(ctx context.Context, cmdArgs []string, opts *DockerRunOptions) (*
 		return nil, err
 	}
 
+	env, err := util.ParseArrayAsMap(opts.SpecSettings.EnvVar, "=")
+	if err != nil {
+		return nil, err
+	}
+
 	spec, err := jobutils.MakeDockerSpec(
-		image, opts.WorkingDirectory, opts.Entrypoint, opts.SpecSettings.EnvVar, parameters,
+		image, opts.WorkingDirectory, opts.Entrypoint, parameters,
 		jobutils.WithPublisher(opts.SpecSettings.Publisher.Value()),
 		jobutils.WithResources(
 			opts.ResourceSettings.CPU,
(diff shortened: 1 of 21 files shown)


simonwo (commented):
My yardstick for what should be in the spec and what should not is "is this something we expect all executors to support"? The concept of environment variables is very much a low-level POSIX-style thing that happens to be supported by our two executors. I can think of higher level executors that would not be expected to process environment variables, such as a "DuckDB query plan" executor or a "Spark queue" executor. Configuration of those things should be done via properly typed parameters.

Is there an actual problem being caused? If not, I vote we make a decision on this later when we have more executors. Note that this is the conclusion we reached last time we discussed this.

rossjones:
Just to prefix: I marked it for discussion because it's come up a few times, and it's easier to discuss over code.

> My yardstick for what should be in the spec and what should not is "is this something we expect all executors to support"? The concept of environment variables is very much a low-level POSIX-style thing that happens to be supported by our two executors. 

I'm not sure the concept of environment vars is really that tied to POSIX any more in most people's minds, although it's a very weak argument, I'd stuggle to find an execution environment that doesn't support them or can't ignore them if necessary.  I think being able to ignore them is key as I'm sure there's an edge use-case where `Inputs` is unnecessary and we'll have the executor ignore it.   

I have a slightly different view of what the contents of the engine parameters contains, which is "what does this engine require to build the executor for this compute" where-as the job spec is, in part, "what does the executor need to configure that environment runtime (input/output/config)". At present it feels like the env vars in the engine params should be configuring the engine, rather than the runtime env.   Maybe we should document some definitive/clarifying detail on what each is supposed to contain?  Apologies if it already exists and I missed it.

> I can think of higher level executors that would not be expected to process environment variables, such as a "DuckDB query plan" executor or a "Spark queue" executor. Configuration of those things should be done via properly typed parameters.

Both of the provided examples allow the code they're running to interact with environment variables, to specify the S3 location they want to read/write from for example. They're not to configure spark/duck themselves, and so feel like they should be job-specific rather than executor specific. 

> Is there an actual problem being caused? If not, I vote we make a decision on this later when we have more executors. Note that this is the conclusion we reached last time we discussed this.

At the moment, just that it's a little clumsy and requires the CLI to understand how those vars will be provided to the runtime on the server (wasm as map[str]str and docker as []str). I expect this will get very repetetive when we add more plugins.

rossjones:
These moved into the task in the latest jobspec (https://docs.bacalhau.org/references/job-specification/task) which feels better than the engine spec, so I am content :D

PR #2705, opened 2023-08-01
Allow maximum job execution time to be configured

### LINKED ISSUE #2561: Configurable compute timeouts
Having a maximum timeout makes sense on a public network where non-paying users should not be able to hog capacity. For private clusters, the behaviour should be configurable – users may wish to run jobs that span many days.

At the moment we enforce a default compute timeout of 60 minutes:
https://github.com/bacalhau-project/bacalhau/blob/516a1570aa40376ae822474c184f58cdf7fb45d6/pkg/node/config_defaults.go#L20
Any job that is submitted that exceeds this timeout is [rejected by compute nodes at the bidding stage](https://github.com/bacalhau-project/bacalhau/blob/516a1570aa40376ae822474c184f58cdf7fb45d6/pkg/bidstrategy/semantic/timeout_strategy.go#L43).

Requester nodes have separate config that control timeouts:
https://github.com/bacalhau-project/bacalhau/blob/516a1570aa40376ae822474c184f58cdf7fb45d6/pkg/node/config_defaults.go#L27-L28
If a job is submitted with a timeout less than or equal to the `MinJobExecutionTimeout` (or without a Timeout field at all), the job's timeout is set to the `DefaultJobExecutionTimeout`.

We also by default set the timeout on a job to a seperate 30 minutes:
https://github.com/bacalhau-project/bacalhau/blob/516a1570aa40376ae822474c184f58cdf7fb45d6/cmd/bacalhau/utils.go#L44

We should:
* by default, set the timeout to "infinite" or thereabouts – we should only apply limits if operators tell us to
* allow compute and requester node operators to set the maximum timeout they will accept
* configure the public Bacalhau network to still have a 60 minute timeout

## Setting an "infinite" timeout

We will need to:
* Remove the default timeout on the CLI  – jobs by default should use the maximum timeout. We can probably just set the `DefaultTimeout` in the CLI to zero.
* Set the default `MaxJobExecutionTimeout` and `DefaultJobExecutionTimeout` to the maximum value of a `time.Duration` parameter (e.g. `math.MaxInt64`) – this is not technically infinite but covers many centuries, so is fine.
* Leave the `MinJobExecutionTimeout` at zero – so that jobs without a timeout specified will always use the max. 

## Set timeouts

Add a flag to the `devstack` and `serve` commands that will allow the timeout to be set. The same flag should be used to set both the compute node's `MaxJobExecutionTimeout` and requester node's `DefaultJobExecutionTimeout`. 

We already have a configuration parameter for timeouts, so you only need to ensure they are set:
https://github.com/bacalhau-project/bacalhau/blob/516a1570aa40376ae822474c184f58cdf7fb45d6/cmd/bacalhau/serve.go#L183

If the flags are not used, the default values set above should be used.

## Configure public Bacalhau

Ensure the public Bacalhau network still sets a timeout of 60 mins. Ensure the `bacalhau serve` command as used by the Bacalhau public nodes includes your new flag with the correct value.

https://github.com/bacalhau-project/bacalhau/blob/516a1570aa40376ae822474c184f58cdf7fb45d6/ops/terraform/remote_files/scripts/start-bacalhau.sh#L58-L68

simonwo:
Previously Bacalhau had some hard-coded values for maximum job time. Having a maximum timeout makes sense on a public network where non-paying users should not be able to hog capacity. For private clusters, the behaviour should be configurable – users may wish to run jobs that span many days.

In this commit we provide a new config flag `--max-timeout` respected by `bacalhau serve` that allows these maximum job times to be set by a node operator. It also changes the default when the flag is unused to jobs having no maximum execution time.

The behaviour of the Bacalhau demo network is unchanged.

The timeout value has been changed from a floating point number of seconds to an integer number of seconds. There are no instances of anyone submitting jobs with fractional seconds on the public network, so this is a backwards compatible change.

Resolves #2561.

### DIFF
--- a/cmd/cli/serve/timeout_test.go
+++ b/cmd/cli/serve/timeout_test.go
@@ -1,0 +1,74 @@
+//go:build unit || !integration
+
+package serve_test
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/bacalhau-project/bacalhau/pkg/docker"
+	"github.com/bacalhau-project/bacalhau/pkg/job"
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+	"github.com/bacalhau-project/bacalhau/pkg/requester/publicapi"
+)
+
+var (
+	noTimeout          = model.NoJobTimeout
+	nonZeroTimeout     = 30 * time.Second
+	halfNonZeroTimeout = nonZeroTimeout / 2
+)
+
+func (s *ServeSuite) TestNoTimeoutSetOrApplied() {
+	docker.MustHaveDocker(s.T())
+
+	cases := []struct {
+		configuredMax    *time.Duration
+		timeoutSpecified *time.Duration
+		timeoutApplied   time.Duration
+		stateExpected    model.JobStateType
+	}{
+		{configuredMax: nil, timeoutSpecified: nil, timeoutApplied: model.NoJobTimeout, stateExpected: model.JobStateCompleted},
+		{configuredMax: nil, timeoutSpecified: &nonZeroTimeout, timeoutApplied: nonZeroTimeout, stateExpected: model.JobStateCompleted},
+		{configuredMax: &nonZeroTimeout, timeoutSpecified: nil, timeoutApplied: nonZeroTimeout, stateExpected: model.JobStateCompleted},
+		{configuredMax: &nonZeroTimeout, timeoutSpecified: &halfNonZeroTimeout, timeoutApplied: halfNonZeroTimeout, stateExpected: model.JobStateCompleted},
+		{configuredMax: &nonZeroTimeout, timeoutSpecified: &noTimeout, timeoutApplied: noTimeout, stateExpected: model.JobStateError},
+	}
+
+	for _, tc := range cases {
+		name := fmt.Sprintf(
+			"job in %s has timeout %s after configuring %s and specifying %s",
+			tc.stateExpected,
+			tc.timeoutApplied,
+			tc.configuredMax,
+			tc.timeoutSpecified,
+		)
+
+		s.Run(name, func() {
+			args := []string{}
+			if tc.configuredMax != nil {
+				args = append(args, "--max-timeout", tc.configuredMax.String())
+			}
+
+			port, err := s.serve(args...)
+			s.Require().NoError(err)
+
+			client := publicapi.NewRequesterAPIClient("localhost", port)
+			s.Require().NoError(publicapi.WaitForHealthy(s.ctx, client))
+
+			testJob := model.NewJob()
+			specOpts := []job.SpecOpt{}
+			if tc.timeoutSpecified != nil {
+				specOpts = append(specOpts, job.WithTimeout(int64(tc.timeoutSpecified.Seconds())))
+			}
+			testJob.Spec, err = job.MakeSpec(specOpts...)
+			s.Require().NoError(err)
+
+			returnedJob, err := client.Submit(s.ctx, testJob)
+			s.Require().NoError(err)
+
+			jobState, err := client.GetJobState(s.ctx, returnedJob.ID())
+			s.Require().NoError(err)
+			s.Require().Equal(jobState.State, model.JobStateError)
+		})
+	}
+}
--- a/cmd/util/flags/spec.go
+++ b/cmd/util/flags/spec.go
@@ -4,7 +4,6 @@ import (
 	"github.com/spf13/pflag"
 
 	"github.com/bacalhau-project/bacalhau/cmd/util/opts"
-	jobutils "github.com/bacalhau-project/bacalhau/pkg/job"
 	"github.com/bacalhau-project/bacalhau/pkg/model"
 )
 
@@ -27,7 +26,7 @@ func NewSpecFlagDefaultSettings() *SpecFlagSettings {
 		Inputs:        opts.StorageOpt{},
 		OutputVolumes: []string{"outputs:/outputs"},
 		EnvVar:        []string{},
-		Timeout:       jobutils.DefaultTimeout.Seconds(),
+		Timeout:       int64(model.DefaultJobTimeout.Seconds()),
 		Labels:        []string{},
 		Selector:      "",
 		DoNotTrack:    false,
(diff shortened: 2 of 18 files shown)


pkg/compute/executor_buffer.go
```diff
@@ -141,13 +142,13 @@ func (s *ExecutorBuffer) doRun(ctx context.Context, task *bufferTask) {
 
 	select {
 	case <-ctx.Done():
-		s.callback.OnComputeFailure(ctx, ComputeError{
+		log.Ctx(ctx).Info().Str("ID", task.execution.ID).Dur("Timeout", timeout).Msg("Execution timed out")
+		s.callback.OnCancelComplete(ctx, CancelResult{
```
frrist, line 146: I'm not entirely convinced that a cancellation of context necessarily means the job was canceled, so I'm hesitant to suggest we modify this logic right away. E.g. there could be a situation where the compute node is shut down. In this case, the context will be canceled, not because of a timeout, but due to a failure in the compute node's operations. It's like if I, acting as a compute provider, decide to shut down my node gracefully, causing your job to stop running. You'd naturally assume it failed, yea?

Also, if a user expects their job to be finished in 5 minutes and it doesn't, I'd be inclined to label that job as 'Failed.' To me, a canceled job is something initiated directly by the user through the API (Our [code comments](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/model/execution_state.go#L32) kind of suggest this). It's not something that happens indirectly through a timeout specified in the job spec. Wdyt?

frrist (commented):
Would like to discuss canceled v failed jobs a bit more before merging.

cmd/util/flags/spec.go
```diff
@@ -73,11 +72,11 @@ func SpecFlags(settings *SpecFlagSettings) *pflag.FlagSet {
 		settings.EnvVar,
 		`The environment variables to supply to the job (e.g. --env FOO=bar --env BAR=baz)`,
 	)
-	flags.Float64Var(
+	flags.Int64Var(
 		&settings.Timeout,
 		"timeout",
 		settings.Timeout,
-		`Job execution timeout in seconds (e.g. 300 for 5 minutes and 0.1 for 100ms)`,
+		`Job execution timeout in seconds (e.g. 300 for 5 minutes)`,
```
rossjones, line 79: Might this flag, like `max-timeout` be nicer as a duration?

rossjones (commented):
LGTM, with a minor comment, but in the time since I added it I'm not sure yet more conversion between duration and seconds would improve anything.

cmd/cli/serve/timeout_test.go
```diff
@@ -0,0 +1,72 @@
+//go:build unit || !integration
```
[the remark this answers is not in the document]
simonwo: At one point we were defining integration tests as "anything you need a connection to an external service for"... the point of this being all of the unit tests you can run offline but the integration tests you cannot, and the _value_ of that being that when Dave was on a trans-Atlantic flight and wanted to work on the code it was clear what tests needed to pass :)

I'm not sure if that's a requirement any longer though. So do we have a definition of the two that brings us some value? I know there are common definitions that are mainly just semantic, but ideally I'd like to achieve something by splitting them out.

cmd/cli/serve/timeout_test.go
```diff
@@ -0,0 +1,72 @@
...
+				args = append(args, "--max-timeout", tc.configuredMax.String())
+			}
+
+			port, err := s.serve(args...)
+			s.Require().NoError(err)
+
+			client := publicapi.NewRequesterAPIClient("localhost", port)
+			s.Require().NoError(publicapi.WaitForHealthy(s.ctx, client))
+
+			job, err := model.NewJobWithSaneProductionDefaults()
```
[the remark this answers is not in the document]
simonwo: Me too! At the very least, I'd like more tests to start using WASM jobs because they run on more platforms in CI than Docker.

How would you feel if I:

1. Changed the `NewJobWithSaneProductionDefaults` to a real fully valid WASM example?
2. Removed the use of the function entirely?

pkg/model/job.go
```diff
@@ -232,9 +233,17 @@ type Spec struct {
...
 }
 
+var (
+	// The job should automatically receive the longest possible timeout.
+	MaxAllowedJobTimeout time.Duration = time.Duration(0)
+
+	// The job should not be subject to timeouts.
+	NoJobTimeout time.Duration = time.Duration(math.MaxInt64).Truncate(time.Second)
+)
+
```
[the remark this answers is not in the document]
simonwo: 1. Sure thing.
2. Yes, perhaps this is poorly named. The zero value for the timeout has been used for a while as "pick whatever the largest timeout is that I can have" - basically, the user did not supply a timeout. Would you be more happy with:
    - A better name, e.g. `AutomaticTimeout` or sometihng.
    - Turning the timeout field into a pointer and if it's missing apply the default logic. I think this would be backwards compatible in JSON world.

pkg/compute/executor_buffer.go
simonwo: There's no modification of the logic here – the existing code already assumes that a context cancellation is a timeout (see the error message). The context is only ever the background atm: 
https://github.com/bacalhau-project/bacalhau/blob/07d2c37a40630c70d286d99511c2ceaee1eabcd4/pkg/compute/executor_buffer.go#L187

This would be OK if the delegate executor was guaranteed to stop the job on node shutdown. But I can't see any logic that does this. So this does look like a bug – I can't see any logic anywhere for ending jobs when the node shuts down? However, I think that's a bigger bug than I am trying to handle here. So I'd prefer to leave this as is for now and investigate that issue separately.

cmd/util/flags/spec.go
simonwo: From what I can see specifying no unit is (rightly) an error for duration values. Today we accept values with no unit as seconds, so this would be backwards-incompatible. However it should be pretty obvious and definitely a better UX, so I'd be happy to change this if we think the impact is low?

rossjones: I'm not sure I've a good grip on what level of breakage we're comfortable with atm, so perhaps safer to leave it like this.

cmd/cli/serve/timeout_test.go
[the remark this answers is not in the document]
simonwo: I have gone for the secret third option which is just changing this case. I figure it will be less burdensome for your work if we do a change like that after it's merged.

rossjones (approved):

simonwo:
Happy to continue discussions above and open a further PR if we decide any more changes are needed 👍🏻

NiklasTR:
great to see this feature added - thank you @simonwo & team!

PR #2712, opened 2023-08-04
wip: first pass at config with cbora+viper

frrist:
### What
(this is a bit rough at the moment, so apologies for messy code and panics)
- I've defined a value `DefaultProductionBacalhauConfig` which contains default values for running bacalhau server in a production environment, I plan to define other values for the rest of the environments (testing, staging, etc.)
- `DefaultProductionBacalhauConfig` is used to generate a config file: `sample_config/config.toml` (its toml for now, but easy to make yaml, json, etc. via Viper - it handles all that for us)
- When the serve command starts it defaults to the values in `DefaultProductionBacalhauConfig`, if a configuration file is found it will use those values instead, if env vars then those, and if flags are passed then those. flags > env vars > config > defaults.
- I have added a helper method `registerFlags` that accepts a list of flags, registers them for a command, and binds viper to their values. Based on the viper and cobra documentation binding viper to flags is idiomatic usage. 
  - Unsure if I like `registerFlags` pattern as its a bit obtuse and has the draw back of not supporting custom flag value types. i.e. it doesn't support things like `DataLocalityFlag`, `IncludedTagFlag`, `NewURLStorageSpecArrayFlag`, etc.  
  - I wanted to have clean way to register and bind a whole bunch of flags, as typing out the viper bind code for each flag seems like a lot of work. If we want custom flag types (and I think we do) I'll modify this to register each flag on it's own - feedback here would be very appreciated.


rossjones:
Thank you for starting this! I much much prefer this to the current implementation, I have a couple of questions though:

1. What would env vars look like? Are they auto-determined from the var name?
2. We would/should need to provide a complete config file so we can comment the relevant parts. Should we ship the defaults as a config file? 
3. Should they live in the pkg/config or in cmd/cli?  If we're going to provide each layer everything they need to run, then cmd/cli makes sense, but it feels like it might be more accessible in the package?
4. We currently have a lot of helper functions in various places, making use of config differently (some only using env vars for instance).  Should we attach those to the relevant config structs for some consistency?

cmd/cli/serve/flags.go
```diff
@@ -0,0 +1,262 @@
...
+				fset.String(def.Name, v, def.Description)
+			case []string:
+				fset.StringSlice(def.Name, v, def.Description)
+			case map[string]string:
+				fset.StringToString(def.Name, v, def.Description)
+			}
+			viper.SetDefault(def.Path, def.Default)
+			if err := viper.BindPFlag(def.Path, fset.Lookup(def.Name)); err != nil {
+				return fmt.Errorf("")
+			}
```
frrist, line 38: Binding the flag value to a string that also adds structure to the config.toml, i.e. `Node.Libp2p.SwarmPort` has a value of `1234`, `Node.Libp2p.PeerConnect` has value `none`. Viper will produce the below toml when writing a config:
```toml
[node.libp2p]
peerconnect = 'none'
swarmport = 1235
```
If the flag's value was bound to the flag name (e.g. `swarm-port`) we'd get a flat config, which I thought was a bit gross and hard to edit. These paths are also how we access values from viper, e.g. `viper.GetInt("Node.Libp2p.SwarmPort")`. This pattern allows us to have simple flags names that don't need to reflect config layout, and also allows users to generate a config file on demand based on their current bacalhau configuration from passed flags or env vars.

cmd/cli/serve/serve.go
```diff
@@ -60,6 +60,84 @@ var (
 `))
 )
 
+func GetServerOptions() (*ServeOptions, error) {
```
frrist, line 63: This method generates a `ServerOptions` based on the current configuration, viper handles returning the correct values from the default, file, env vars, and/or flags.

pkg/model/job_selection.go
```diff
@@ -46,3 +46,13 @@ func NewDefaultJobSelectionPolicy() JobSelectionPolicy {
...
+
+func (e JobSelectionDataLocality) MarshalText() ([]byte, error) {
+	return []byte(e.String()), nil
+}
+
+func (e *JobSelectionDataLocality) UnmarshalText(text []byte) (err error) {
+	name := string(text)
+	*e, err = ParseJobSelectionDataLocality(name)
+	return
+}
```
frrist, line 58: This is dead code: I was attempting to allow `BacalhauConfig` to contain the actual values we use in the code, but for the values to be read and written from the config as human readable values (e.g. `JobSelectionDataLocality` in the config could sting ("local"|"anywhere", but read as a type (1, 2), I couldn't get it to work)

rossjones: That would be neat if we can get an extra level of type safety, from what I can tell with a cursory glance it seems like a lot of boiler plate to do this with viper/mapstructure/decodehooks - hopefully you can find a way.

cmd/cli/serve/config.go
```diff
@@ -0,0 +1,120 @@
...
+				JobGPU:      "",
+				TotalCPU:    "",
+				TotalMemory: "",
+				TotalGPU:    "",
+			},
+		},
+		Requester: RequesterConfig{
+			ExternalVerifierHook: "",
+			JobSelectionPolicy: JobSelectionPolicyConfig{
+				Locality:            model.Anywhere.String(),
```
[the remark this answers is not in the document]
simonwo: Ah, you discuss this below. Yes, I agree that the extra type safety would be worth spending time for.

pkg/system/config.go
```diff
@@ -39,22 +40,19 @@ func InitConfig() error {
...
 	viper.SetDefault("user-id-key", userIDKey) // rsa key for identifying user
 
 	// Settings and initialisation for viper:
-	viper.SetConfigFile(configFile) // provided or created config file
-	viper.SetConfigType("yaml")     // config is always a yaml file
-	viper.AutomaticEnv()            // try to read config from env if possible
-	viper.SetEnvPrefix("bacalhau")  // BACALHAU_<key> is encoding for env vars
+	viper.SetConfigName("config")
+	viper.SetConfigType("toml")
+	//viper.SetConfigType("yaml")
```
simonwo, line 52: Why TOML when our other stuff uses YAML?

sample_configs/config.toml
```diff
@@ -0,0 +1,40 @@
+api-host = 'bootstrap.production.bacalhau.org'
```
simonwo, line 1: Is it possible to generate this file with comments for each key? In sample config files of yore, everything was commented so that you could open the file and read the documentation inline for the config value you are changing.

Lets also add a CI step to make sure this is up to date like we do for Swagger?

simonwo (commented):
Agreed re: using the proper types if we can, but otherwise looks great to me.

pkg/model/job_selection.go
frrist: Brilliant, just needed to register the text decoder, custom types work!
```golang
	if err := viper.Unmarshal(&cfg, viper.DecodeHook(mapstructure.TextUnmarshallerHookFunc())); err != nil {
		return err
	}
```

pkg/system/config.go
frrist: For development purposes I made it TOML; personally, I find it easier to read. But Viper will handle this for us: it supports JSON, TOML, YAML, HCL, and others. I'll be sure to make the final change default to YAML.

sample_configs/config.toml
frrist: I'm not away of any feature in viper that generates config comments, that might have to come later.

cmd/cli/serve/config.go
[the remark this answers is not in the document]
frrist: got it working: https://github.com/bacalhau-project/bacalhau/pull/2712#discussion_r1284812756

frrist:
@rossjones 

> What would env vars look like? Are they auto-determined from the var name?

Viper uses keys to set and get values, all keys are defined [here](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/generated_constants.go). Code gen for building that file is in [here](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/gen_paths/generate.go), and may be called from [here](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/config.go#L10) using `go generate`. The [code gen](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/gen_paths/generate.go#L63) reads our config structure to generate keys for it. Whenever you update the config structure you re-run the codegen. I have defined a helper method that generates the environment variable names from the keys [here](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/helper.go#L8). [Here](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/devstack/devstack.go#L382-L402) is some example usage of it in the devstack code. By using these keys, instead of the flag names we can provide a clear config file as they dictate the config field names, rather than the flag names themselves.

Some example env vars right now are `BACALHAU_NODE_API_HOST=0.0.0.0 BACALHAU_NODE_API_PORT=63454`

> We would/should need to provide a complete config file so we can comment the relevant parts. Should we ship the defaults as a config file?

Upon start up bacalhau [generates a default config](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/main.go#L46) file [from our default configuration](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/main.go#L41) by setting default values for viper, then instructing it to write a config. The [default config](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/production.go#L12) lives in code, and the file we generate is produced from it. By using viper to write the config we have the flexibility to: change our config format (json, toml, yaml) at will, and use custom types in our config structure (like model.Engine, model.JobSelectionDataLocality, etc.) - their formats are human readable in the config and we don't need to parse them when accessing with viper (just type case). 

We can provide a config file, and if its present bacalhau will read that instead of generating one. At some point I would like to generate comments on the config based on the comments we put on our config structure.

> Should they live in the pkg/config or in cmd/cli? If we're going to provide each layer everything they need to run, then cmd/cli makes sense, but it feels like it might be more accessible in the package.

I think they should live in `pkg/config` (right now its called `pkg/config_v2` as I remove the old one). If a flag has a corresponding value in the config structure it can import `pkg/config` to bind to it. I have messily done this in the current code - working to make it better.

> We currently have a lot of helper functions in various places, making use of config differently (some only using env vars for instance). Should we attach those to the relevant config structs for some consistency?

Yes, but I am not sure we can get it all in one PR, we've been (ab)using our global config values for a while and its hard to untangle without touching everything 🍝 . For example all nodes in a devstack (from what I can tell) share the same config. I am wondering if the devstack should have its own config file.

rossjones:
> Viper uses keys to set and get values, all keys are defined [here](https://github.com/bacalhau-project/bacalhau/blob/9dc532e4b449f18165397d4a44693da053c6d4ca/pkg/config_v2/generated_constants.go).

The only thing I think here is that having `Node` in the value is okay, but having it in the const name is a bit superfluous.  May be better to just have it be ..

```go
const APIHost = "Node.API.Host"
const APIPort = "Node.API.Port"
const AllowListedLocalPaths = "Node.AllowListedLocalPaths"
```
 
> Some example env vars right now are BACALHAU_NODE_API_HOST=0.0.0.0 BACALHAU_NODE_API_PORT=63454

Again, I think the NODE here might be unnecessary if every env var starts `BACALHAU_NODE`.

Everything else is great, and I think a huge improvement over how we do it now.

frrist:
replacing with https://github.com/bacalhau-project/bacalhau/pull/2740

PR #2714, opened 2023-08-06
fix: refactor all tests to use suite.Require()

frrist:
🤞 this wasn't hurting us anywhere. I'll remove the code I used to make this refactor before merging, but left it in-place for now to demonstrate the mechanical nature of this change. We could also consider turning it into a linter rule.


rossjones (approved):
LGTM.

refactor/main.go
```diff
@@ -0,0 +1,87 @@
+package main
```
simonwo, line 1: FYI you might like to have a read about [`sed`](https://man7.org/linux/man-pages/man1/sed.1.html) which is a unix command that does this sort of thing. 😇

    find -name '*.go' | xargs sed -i 's/find-regex/replacement/'

simonwo (approved):
Looks good to me!

PR #2719, opened 2023-08-07
New model types

wdbaruni:
Introduce new model types and job specs that unblock more job types, extensible engines and storage sources, and updatable jobs.

#### Job 

`Job` is an internal and common representation for the different job types submitted by end users. Users won't submit `Job` directly, but instead will submit different request and spec depending on the job type (not introduced yet), and then those requests will be translated to a `Job` spec, which allows introducing new job types with no or minimal change to existing schedulers or executors. Changes to existing `Job` include:

- `Name` field defined by the user, which will allow updating existing jobs if a job with the same name already exist for the same user
- `Spec` has been replaced by a list of `Tasks` to enable running parallel or sequential local tasks per job in the same compute node. (e.g. parallel task to publish results as they are made available by the main task, or a task at the end to publish results which can replace our current publishers). Won't introduce this capability with this PR.
- `JobState` type has been replaced by a single `State` field in the `Job` type itself.

#### Task

Represents the previous `Spec` type, which the following changes:
- `Engine`, `Publisher`, `Inputs` and `Outputs` are all represented by a the same flexible `SpecConfig` type, which has a `Type` string field, and a `Params` map of strings to interfaces. This works nicely with existing `Publishers`, new pluggable executors, and will unlock the same flexibility for storage specs
- Previously `Engine`, `Publisher` and `StorageSource` names were _indirectly_ case insensitive as we ignored the case when deserializing those names into their enums. As I've switched to strings instead of enums in the new models, extra care was required to maintain this behaviour. Switching to snake case for future names might simplify future work (e.g. `urlDownload` -> `url_download`)

### Main Changes
This turned out to be a very large PR, but most of the changes are because of using the new types and fixing tests. These are the main changes to help with the review:
1. `ExcludedTag` and `IncludedTag` are now represented as strings instead of a dedicated type. Also `inmemory` job store now gives higher precedence to `ExcludedTag` so that if a tag shows in both `IncludedTag` and `ExcludedTag`, it will be excluded. This is a similar behaviour to boltdb implementation, and can change results of `bacalhau list` for some users
2. `JobSelectionPolicy` struct has been moved from `model` to `node` package as it represents configuration for _some_ of the bidding strategies. We might remove this struct with proper configuration setup
3. Moved marshallers and providers from `model` package to their packages under `lib`. Providers now accept a string a key, and does case insensitive matching as explained above
4. Compute `store.ExecutionState`, which is almost a duplicate of the job store's `models.ExecutionState`, has been merged and replaced with `LocalexecutionState` that wraps `models.ExecutionState` along with additional fields to track the local compute's state


pkg/models/constants.go
```diff
@@ -0,0 +1,13 @@
...
+const (
+	// DefaultNamespace is the default namespace.
+	DefaultNamespace = "default"
+)
+
+const (
+	JobTypeService = "service"
+	JobTypeBatch   = "batch"
+	JobTypeOps     = "ops" // TODO: revisit the job naming
+	JobTypeDaemon  = "daemon"
```
rossjones, line 12: A one-line comment here will help with remembering what the different types do so our editors can show us without having to look it up in notion.

pkg/models/execution.go
```diff
@@ -0,0 +1,126 @@
+//go:generate stringer -type=ExecutionStateType --trimprefix=ExecutionState --output execution_state_string.go
+//go:generate stringer -type=ExecutionDesiredStateType --trimprefix=ExecutionDesiredState --output execution_desired_state_string.go
+package models
+
+// ExecutionStateType The state of an execution. An execution represents a single attempt to execute a job on a node.
+// A compute node can have multiple executions for the same job due to retries, but there can only be a single active execution
+// per node at any given time.
+type ExecutionStateType int
+
+// TODO: change states to reflect non-bidding scheduling
```
rossjones, line 10: When this happens will we replace the bid based states with New/Pending/Running? 
I'm keen for us to try and differentiate between "Preparing the environment for the run" and "The docker image is actually starting".

pkg/models/execution.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Name is a logical name of the execution.
+	Name string
+
+	// NodeID is the node this is being placed on
+	NodeID string
+
+	// Job is the parent job of the task being allocated.
+	// This is copied at execution time to avoid issues if the job
+	// definition is updated.
```
rossjones, line 60: Should it be a non-pointer type just to make sure the referenced var isn't modified?

pkg/models/execution.go
```diff
@@ -0,0 +1,126 @@
...
+	Job   *Job
+
+	// AllocatedResources is the total resources allocated for the execution tasks.
+	AllocatedResources *AllocatedResources
+
+	// DesiredState of the execution on the compute node
+	DesiredState ExecutionDesiredStateType
+
+	// DesiredStateDescription is meant to provide more human useful information
+	DesiredStateDescription string
```
rossjones, line 71: These two seem like they're co-dependent but might in fact be changed independently?  Should we wrap them up together?

pkg/models/execution.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// ComputeState observed state of the execution on the compute node
+	ComputeState ExecutionStateType
+
+	// ComputeStateDescription is meant to provide more human useful information
+	ComputeStateDescription string
```
rossjones, line 77: Same with these two.  If they get out of sync it'll be super-confusing.

pkg/models/execution.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// the published results for this execution
+	PublishedResult *SpecConfig
+
+	// PreviousExecution is the execution that this execution is replacing
```
rossjones, line 82:
```suggestion
	// PreviousExecution is the execution that this execution is replacing, if any
```

If there is none, is this "" ?

pkg/models/execution.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (a *Execution) JobNamespacedID() NamespacedID {
+	return NewNamespacedID(a.JobID, a.Namespace)
+}
+
+// Normalize Allocation to ensure fields are initialized to the expectations
+// of this version of Nomad. Should be called when restoring persisted
+// Allocations or receiving Allocations from Nomad agents potentially on an
+// older version of Nomad.
```
rossjones, line 109: We probably should explain how this is used in bacalhau.

pkg/models/job.go
```diff
@@ -0,0 +1,187 @@
...
+	Tasks []*Task
+
+	// State is the current state of the job.
+	State JobStateType
+
+	// StateDescription is meant to provide more human useful information
+	StateDescription string
+
+	// Version is a monotonically increasing version number that is incremented
+	// on each job update.
```
rossjones, line 75:
```suggestion
	// Version is a per-job monotonically increasing version number that is incremented
	// on each job update.
```

After suggesting that, I'm wondering whether now the version applies to the Name/ID combo for this job, or just the ID.

pkg/models/job.go
```diff
@@ -0,0 +1,187 @@
...
+	} else if validate.ContainsSpaces(j.ID) {
+		mErr.Errors = append(mErr.Errors, errors.New("job ID contains a space"))
+	} else if validate.ContainsNull(j.ID) {
+		mErr.Errors = append(mErr.Errors, errors.New("job ID contains a null character"))
+	}
+	if validate.IsBlank(j.Name) {
+		mErr.Errors = append(mErr.Errors, errors.New("missing job name"))
+	} else if validate.ContainsNull(j.Name) {
+		mErr.Errors = append(mErr.Errors, errors.New("job Name contains a null character"))
+	}
```
rossjones, line 153: Perhaps `Name` should be optional, might make it safer to use without the risk of even accidentally overwriting previous jobs.

rossjones (commented):
In general I like it, it feels cleaner than the current models, and I like that we're expecting more of them (validation etc).

The things that make me nervous are:

* Introducing more nils, I think I'd have a preference for defaulting to the starnewt ( `*new([]string)` ) rather than nil. I think it'll trip us up later on
* Job names having meaning make me think that it'll be a cause of much support if it is mandatory with people accidentally overwriting their original jobs from repeating commands.
* A few places we have structs that have a couple of dependent fields (often type and description) that may end up out of sync and a source of confusion.

Overall though I think this is better than what we have and a cleaner based to build on.

pkg/models/constants.go
wdbaruni: Done in 0fc4b259977f5fc4019018abdef1346ee35b35eb

pkg/models/execution.go
wdbaruni: Done in b37de3db8db7907ebbf04b788c46b5e5796ecb72

pkg/models/job.go
```diff
@@ -0,0 +1,187 @@
...
+	// Meta is used to associate arbitrary metadata with this job.
+	Meta map[string]string
+
+	Tasks []*Task
+
+	// State is the current state of the job.
+	State JobStateType
+
+	// StateDescription is meant to provide more human useful information
+	StateDescription string
```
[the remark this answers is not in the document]
wdbaruni: Good point with making sure the state and its description don't go out of sync. Done in b37de3db8db7907ebbf04b788c46b5e5796ecb72

pkg/models/job.go
```diff
@@ -0,0 +1,187 @@
...
+// called when registering a Job.
+func (j *Job) Normalize() {
+	if j == nil {
+		return
+	}
+
+	// Ensure that an empty and nil map are treated the same to avoid scheduling
+	// problems since we use reflect DeepEquals.
+	if len(j.Meta) == 0 {
+		j.Meta = nil
```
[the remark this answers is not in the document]
wdbaruni: make sense. setting empty types to nils _might_ play nicer with serialization and deserialization, but I agree it will be better to use empty types overall and worry about serde where and when it happens. Fixed in 88149bf98be8945ad8f9df44f765bb2e9cbcbb0d

pkg/models/execution.go
wdbaruni: I am still planning to keep the two phase of job scheduling where the scheduler can ask multiple nodes to run a job, but only select few at the end. I am planning to change the language and order of events so that we end up with a [Sparrow](https://cs.stanford.edu/~matei/papers/2013/sosp_sparrow.pdf) implementation for job scheduling. It will look as follows:
1. The scheduler will submit a job to multiple nodes. If they can run the job, they will enqueue it, otherwise they will fail the request. This is similar to `AskForBid`, but I am hoping to make the response synchronous to simplify things. Will need to sync up with @simonwo about this. The other difference is the compute nodes will enqueue the job right away and won't wait for approval
2. When the compute node is ready to execute the job, it will first ask the scheduler for approval in case another compute node already started the execution. If this is the first node that is ready to execute, then the scheduler will approve the request, send rejection to all other compute nodes, and will reject any future request to approve the execution in case the compute node didn't hear about the rejection yet
3.  This two phase job scheduling will only be used for `Batch` and `Service` jobs, but not for `Daemon` or `Ops` jobs

pkg/models/job.go
wdbaruni: should be only for the id

pkg/models/job.go
wdbaruni: left a comment on how nomad solves this. Will see if we can follow a similar approach

pkg/models/execution.go
wdbaruni: I left a TODO to evaluate this. I'll have more clarity once this is integrated with the job store, scheduler and executors

pkg/models/execution.go
simonwo: The reason we have async bid approval is so that the external approval hooks (HTTP and exec) can take a long time to approve things (e.g. many seconds to do a network request) if they want to. I think that is generally a positive quality of life improvement that we should not remove.

simonwo:
My comments:

> Name field defined by the user, which will allow updating existing jobs if a job with the same name already exist for the same user

Can this be something that only long running jobs et. al. have? And that way it's not possible to accidentally overwrite `batch` jobs because they don't have a name, only an ID.

> Spec has been replaced by a list of Tasks to enable running parallel or sequential local tasks per job in the same compute node. (e.g. parallel task to publish results as they are made available by the main task, or a task at the end to publish results which can replace our current publishers). Won't introduce this capability with this PR.

So is this us introducing DAGs then? Will Tasks have dependencies on each other?

> Changed Inputs into Artifacts which represent remote files that must be downloaded before the job starts, such as from S3 and IPFS, and Volumes which represent mounted volumes that can have both readonly or readwrite access.

Our existing Storage Spec system is flexible enough to represent the `LocalDirectory` storage type. This sounds like we are removing this flexibility. And I think we are about to hit similar features like 'connect me to this local database represented by this storage spec' etc. How will this type work in this structure? As a local directory mount is not "a remote file that must be downloaded". If the answer is "we'll represent them using some other method" then I think that's a -1 from me because of the loss of generality and increase in complexity here. 

There is also an implicit assumption here that "mounting" is the method by which each executor will bring data into the job – so that data must first be downloaded and then mounted. Today this does not have to be true. For example, if a storage spec specifies some remote location, the eecutor might be able to stream the data directly rather than having to have it downloaded and mounted. We currently have an issue with large datasets because we have to be able to hold the entire dataset on disk – it would be preferable to not do this. Generally I see this as a step backward on our journey to handling datasets of any meaningful size.

Why are we making this change now – is it actually necessary for this PR? I think it warrants more design. 

> Removed RunCommandResult as those can be represented by bacalhau logs. Feedback is needed on this change

Can you explain this a bit more? `bacalhau logs` only works for jobs that are currently executing – are we proposing to change that? RunCommandResult also contains system-generated errors that go into the error field and an exit code – where do we get that information now? It was certainly convenient to have short messages in the `describe` output – are we just removing that wholesale?

pkg/lib/validate/collection.go
```diff
@@ -0,0 +1,9 @@
+package validate
+
+func IsEmpty[T any](s []T) bool {
+	return len(s) == 0
+}
+
+func IsNotEmpty[T any](s []T) bool {
+	return !IsEmpty(s)
+}
```
frrist, line 9: if we stick with this pattern we'll need to:
- implement more methods for asserting the length of maps
- ensure subsequent changes to the model code use these helper methods in PR review (image an OS contribution making changes to this whose unaware of these helper methods) 


So, I'd prefer we make direct use of the builtin `len()` function, rather than adopt usage of these methods.
Comparing the current usage:
```go
if validate.IsEmpty(r.Values) {
	mErr.Errors = append(mErr.Errors, errors.New("selector values cannot be empty for In or NotIn operators"))
}
```
vs
```go
if len(r.Values) == 0 {
	mErr.Errors = append(mErr.Errors, errors.New("selector values cannot be empty for In or NotIn operators"))
}
```
I find the latter clearer; This just feels overly complicated. But I won't die on this hill, if you feel strongly about it then let's keep it.

frrist:
@simonwo 
> Our existing Storage Spec system is flexible enough to represent the LocalDirectory storage type. This sounds like we are removing this flexibility. And I think we are about to hit similar features like 'connect me to this local database represented by this storage spec' etc. How will this type work in this structure?

One way we might do this is via a `SpecConfig` of `Type` "Local" with params that include information about the path. Alternatively the `Volume` could contain such information in a similar fashion. Similarly, for the case of a database, the `Volume` could contain the information required to establish a connection to a database (I am ignoring the bits around a "database storage spec" containing sensitive information like passwords and API keys here).

> For example, if a storage spec specifies some remote location, the executor might be able to stream the data directly rather than having to have it downloaded and mounted.

FWIW I would expect this to be the Compute Nodes responsibility rather than the executors. An off-the-cuff example could be the Compute Node establishing a NFS/SSH-FS connection to a remote storage source (very similar to the database example above), and providing the NFS/SSH-FS connection to the executor. The details of the NFS/SSH-FS connection could come in the form of a `Volume` or `Artifact` - I'd expect the former. I don't think establishing such connections should be in the realm of responsibilities of an executor.

frrist:
My major concern so far is for the need of a `Copy` method on (almost) all of these structures. When do we expect such a method to be necessary? I assume we are worried about immutability, and want to prevent the (accidental) modification of the original Job structure by calling code? (I am picking on Job as an example here).

wdbaruni:
Thanks @simonwo for your comments:


> Can this be something that only long running jobs et. al. have? And that way it's not possible to accidentally overwrite batch jobs because they don't have a name, only an ID.

Yeah it does make sense to evaluate this option for batch jobs when we implement updates. This means users can't update a batch job once its submitted, and they can only cancel it and re-submit a new version. Name field will exist, but only as a reference and won't be used for de-duplicating jobs. 

> So is this us introducing DAGs then? Will Tasks have dependencies on each other?

We can decide to implement DAGs, or simple lifecycle such as:
- `pre-start`: tasks in this phase will run first, and will mainly include tasks that will prepare the environment for the main task, such as downloading data for the main tasks if `StorageSource` is not available or feasible to implement
- `start` or `main`: the main tasks for this job and will run after `pre-start` tasks are completed. There are can be multiple tasks running together, such as a main application and a side-car. For long running jobs, the main tasks are not expected to complete, and will be treated as a failure if a main task exits, event with code 0. Thats why it make sense to flag which tasks are the main ones, explicitly such as proposed here, or implicitly in a DAG which would be the leaf tasks
- `post-start`: makes sense for batch jobs and will run after the main task is completed.    

Changed Inputs into Artifacts which represent remote files that must be downloaded before the job starts, such as from S3 and IPFS, and Volumes which represent mounted volumes that can have both readonly or readwrite access.

> Our existing Storage Spec system is flexible enough to represent the LocalDirectory storage type. This sounds like we are removing this flexibility. 

Inputs are only restricted to content that must be downloaded to the local file system as the compute node downloads them before execution, and the executor is expecting them to be local before executing. This means you only have read access, and all the data must be fetched before executing and not while executing. LocalDirectory is a special case as the data is already local and we can provide read-write access. 
Inputs is great for the majority of use cases and are easy to implement new sources. In the future I am hoping to support Container Storage Interface (CSI) in _addition_ to Inputs to support mounting remote sources with read-write access support. Both are useful are useful and can co-exist. 

> There is also an implicit assumption here that "mounting" is the method by which each executor will bring data into the job – so that data must first be downloaded and then mounted. Today this does not have to be true. 

Actually this is not an assumption. The executor is no longer responsible or fetching inputs. The compute node is the one that is now downloading inputs into local directories, and passing those local directories to the pluggable executors. Each executor handles those local directories differently, (e.g. docker would mount them as volumes, and python might work on them directly), but each input will be translated eventually into a local directly to unify the executors' inputs and simplify their implementations. 

> Why are we making this change now – is it actually necessary for this PR? I think it warrants more design.

Mainly because I am working on the job specs and trying to look ahead. However, I agree a design is required as this is not a simple decision. I'll keep Inputs as they are, and we can evaluate introducing CSI Volumes in the future


> Can you explain this a bit more (removing RunCommandResult)? bacalhau logs only works for jobs that are currently executing – are we proposing to change that? 

I thought we are keeping logs for jobs after they were completed as well, but yeah it is not possible with today's implementation. Will keep RunCommandResult and re-evaluate its usefulness after we have better logs story.

rossjones:
> I thought we are keeping logs for jobs after they were completed as well,

Sort of. Currently they're held with the published outputs, but we did discuss publishing logs separately so that they can be fetched independently and also stored alongside other logs that the enterprise might be using. Haven't done that yet though.

cmd/cli/serve/serve.go
```diff
@@ -177,23 +177,37 @@ func GetPeers(OS *ServeOptions) ([]multiaddr.Multiaddr, error) {
...
+		CPU:    OS.LimitJobCPU,
+		Memory: OS.LimitJobMemory,
+		GPU:    OS.LimitJobGPU,
+	}
+
+	parsedTotalLimits, err := totalResourcLimitsConfig.ToResources()
+	if err != nil {
+		return node.ComputeConfig{}, err
+	}
+
```
wdbaruni, line 197: we used to silently ignore resource parsing errors

ops/terraform/remote_files/scripts/apply-http-allowlist.sh
```diff
@@ -8,12 +8,12 @@ set -euxo pipefail
 
 ALLOWLIST=./http-domain-allowlist.txt
 
-TYPE=$(echo "$BACALHAU_JOB_SELECTION_PROBE_DATA" | jq -r '.spec.Network.Type')
+TYPE=$(echo "$BACALHAU_JOB_SELECTION_PROBE_DATA" | jq -r '.Job.Tasks[0].Network.Type')
 test "$TYPE" = 'HTTP' || test "$TYPE" = 'None'
```
wdbaruni, line 12: This will break existing allow-listing user scripts that depend on parsing the previous job schema. Do we have enough users that use this feature for this to be a concern?

pkg/requester/jobtransform/wasm_stroage_spec.go
```diff
@@ -0,0 +1,27 @@
+package jobtransform
```
rossjones, line 1: Typo in filename (stroage->storage)

pkg/models/job.go
```diff
@@ -0,0 +1,184 @@
...
+func (j *Job) NamespacedID() NamespacedID {
+	return NamespacedID{
+		ID:        j.ID,
+		Namespace: j.Namespace,
+	}
+}
+
+// Normalize is used to canonicalize fields in the Job. This should be
+// called when registering a Job.
+func (j *Job) Normalize() {
```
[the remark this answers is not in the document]
wdbaruni: `Normalize` should be common among them, but `Validate` might introduce additional logic per job type

pkg/models/job.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		nj.Tasks = tasks
+	}
+
+	nj.Meta = maps.Clone(nj.Meta)
+	return nj
+}
+
+// Validate is used to check a job for reasonable configuration
+func (j *Job) Validate() error {
```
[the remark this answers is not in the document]
wdbaruni: answered above

pkg/models/spec_config.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (s *SpecConfig) Validate() error {
+	if s == nil {
+		return errors.New("nil spec config")
+	}
+	if validate.IsBlank(s.Type) {
+		return errors.New("missing spec type")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
wdbaruni: `Normalize` mainly converts the type to its standard format to maintain consistency, which is useful after deserializing types from data store or after they are sent over the wire. Examples empty maps are deserialized as nil maps, and `Normalize` will convert them to empty maps. Other examples include trimming spaces

pkg/models/job.go
```diff
@@ -0,0 +1,184 @@
...
+		tasks := make([]*Task, len(nj.Tasks))
+		for i, t := range nj.Tasks {
+			tasks[i] = t.Copy()
+		}
+		nj.Tasks = tasks
+	}
+
+	nj.Meta = maps.Clone(nj.Meta)
+	return nj
+}
```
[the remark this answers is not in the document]
wdbaruni: Today we are passing copies of objects all over the place instead of passing pointers, which I understand it is better for safety and data consistency, but not cheep specially for our large Job and Execution types. For example `boltdb` job store deserializes objects from disk and then return a copy of the deserialized object instead of the new object itself. I am hoping that we use more pointers where it makes more sense, and utilize `Copy()` when we want to return a new instance and avoid leaking the original instance. e.g. in-memory data stores can return a copy, and on-disk data stores can return the same deserialized instance

pkg/requester/jobtransform/wasm_stroage_spec.go
wdbaruni: fixed in [3037a12](https://github.com/bacalhau-project/bacalhau/pull/2719/commits/3037a12bda26e32657e585cd3add2a6e9afa81c4)

Issue #2657, opened 2023-07-12
Job archival and deletion for jobstore


rossjones:
Once the persistent jobstore is complete, we will have persistent job info and job states, and history but this will persist forever. Or until we run out of disk space. Whichever happens sooner.

We should provide an archival policy that states how long jobs are maintained in the jobstore once they are in a terminal state, before they are removed and perhaps archived in a different store. We should also implement that policy.  We might also take the opportunity to reap jobs that are in a broken state for too long (e.g. waiting in non-terminal states for days. ExecutionStateAskForBid for example).

aronchick:
Great point. Might also be worth having a (larger) feature that exports this information to an external store (flat file, redshift, whatever).

wdbaruni:
Tracked by https://github.com/bacalhau-project/bacalhau/issues/4174

PR #2723, opened 2023-08-08
WIP: Delete old/expired jobs from jobstore

rossjones:
When jobs enter a terminal state, they should have a limited lifetime after which they are removed from the jobstore. 

This PR implements the necessary logic to clean up those expired jobs when the jobstore is created, as the first step in providing hooks to a general housekeeping task that will be necessary in the requester node (for other cleanup tasks, like managing jobs in invalid states etc).  This PR does not add the housekeeping.

## Blocked by

* [ ] Need to configure the lifetimes per job type. Will add a config struct for jobstore until new configv2 is ready

### DIFF
--- a/pkg/jobstore/boltdb/bucketpath.go
+++ b/pkg/jobstore/boltdb/bucketpath.go
@@ -1,0 +1,71 @@
+package boltjobstore
+
+import (
+	"fmt"
+	"strings"
+
+	bolt "go.etcd.io/bbolt"
+)
+
+type BucketPath struct {
+	path string
+}
+
+// NewBucketPath creates a bucket path which can be used to describe the
+// nested relationship between buckets, rather than calling b.Bucket() on
+// each b found.  BucketPaths are typically described using strings like
+// "root.bucket.here".
+func NewBucketPath(sections ...string) *BucketPath {
+	return &BucketPath{
+		path: strings.Join(sections, BucketPathDelimiter),
+	}
+}
+
+// Get retrieves the Bucket, or an error, for the bucket found at this path
+func (bp *BucketPath) Get(tx *bolt.Tx, create bool) (*bolt.Bucket, error) {
+	path := strings.Split(bp.path, BucketPathDelimiter)
+
+	type BucketMaker interface {
+		Bucket([]byte) *bolt.Bucket
+		CreateBucketIfNotExists([]byte) (*bolt.Bucket, error)
+	}
+
+	getBucket := func(root BucketMaker, name string) (*bolt.Bucket, error) {
+		bucket := root.Bucket([]byte(name))
+		return bucket, nil
+	}
+	if create {
+		getBucket = func(root BucketMaker, name string) (*bolt.Bucket, error) {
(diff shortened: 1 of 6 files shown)


simonwo:
> they should have a limited lifetime after which they are removed from the jobstore.

Why? Are we worried about disk space? If I was a node operator I would expect a list of all executed jobs by default to be retained indefinitely.

rossjones:
> Why? Are we worried about disk space? 

Disk space, as well as performance, iterating through jobs is something we do reasonably often. We could move jobs (and related objects) into a 'dead-job' bucket and maintain another index, although we'd still need to compact (vacuum in postgres-land) the database from time to time.  

> I would expect a list of all executed jobs by default to be retained indefinitely, for debugging and auditing purposes

I'd certainly expect them to be available, but I'm not sure I'd have any expectations on where the audit was held.  I'd hope it to be somewhere immutable with limited accessibility. I think this is probably why I thought an archive/delete approach might be better in that  we can wrap all of the job objects into a single document and push it somewhere so that it's retrieval wasn't dependent on the current structure of the job (and related object). This was mentioned briefly in the original ticket -> https://github.com/bacalhau-project/bacalhau/issues/2657

simonwo:
> Disk space, as well as performance, iterating through jobs is something we do reasonably often.

It sounds to me like the access pattern is at fault here rather than the data size – it's not exactly big data. Shouldn't we be doing more intelligent querying that simply iterating all jobs? Are we using a proper query engine?

> I'd hope it to be somewhere immutable with limited accessibility. 

Another user-visible issue with this is that after a certain period of time (presumably quite invisible from the user) their job will no longer be available using `bacalhau describe` and retrieving their results will not be possible using `bacalhau get`. 

I don't disagree with the principle of things being performant of course, but I think we should at least know what the performance impact is and explore the options for sorting that out before we choose to degrade the UX.

rossjones:
> It sounds to me like the access pattern is at fault here rather than the data size – it's not exactly big data. Shouldn't we be doing more intelligent querying that simply iterating all jobs? Are we using a proper query engine?

We're using a kvstore, and using it inline with how you'd normally use those, alongside our own indexes for various lookups.  And yet, for cases where we might want to reconstitute the eval broker after a case, there's little alternative than iterating all of the jobs whose state is currently in-progress. 

What do you mean by a proper query engine?  Like a relational one?

> Another user-visible issue with this is that after a certain period of time (presumably quite invisible from the user) their job will no longer be available using bacalhau describe and retrieving their results will not be possible using bacalhau get.

Yes, that's true, and will likely cause confusion if we do nothing about it. That's definitely a place where usage data would be really useful, do users use the CLI to access outputs months after their job has completed?

> I don't disagree with the principle of things being performant of course, but I think we should at least know what the performance impact is and explore the options for sorting that out before we choose to degrade the UX.

Maybe best waiting then until we have some usage data so we can work out if it is likely to become an issue or not.  Happy to pause this for now in that case.

rossjones:
There's been no further discussion about cleaning up old jobs, so seems safe to close this.

PR #2730, opened 2023-08-09
Remove estuary storage/publisher

rossjones:
Removes support for estuary as a storage provider or a downloader but retains the ability to parse the name to avoid crashes from removing an enum.


frrist:
We will still technically support estuary as a publisher after this change, but as a URL publisher instead, correct? Just need a story for sharing API keys with compute nodes which I believe is the intention of https://github.com/bacalhau-project/bacalhau/pull/2692?

simonwo:
I think we should actually go one step further and remove `StorageSourceEstuary` from the publisher mapped provider, so that a job submission with Estuary will not get scheduled to any node. Yes, this will require people to change their job descriptions which is something we now need people to do anyway as Estuary is permanently gone, but it doesn't make past job specs invalid which would happen if we removed it wholesale.

rossjones:
> remove StorageSourceEstuary from the publisher mapped provider, so that a job submission with Estuary will not get scheduled to any node

Yes that makes a lot of sense, as long as we can still parse a jobspec, that that'll be enough. Does suggest we'll need to have a think about how we handle disappearing plugins in future.

simonwo (approved):

PR #2748, opened 2023-08-16
WIP: Requester node TLS

### LINKED ISSUE #2743: Requester node TLS support
Enable the requester node to serve endpoints over TLS. This should work for clients on both public and private networks. 

```[tasklist]
### Tasks
- [x] Requester node should support --tlscert and --tlskey to for TLS config
- [x] Requester node should support --autotls to use ACME for certs.
- [x] Update client to support TLS (with optional --cacert)
- [x] Client should support --insecure for self-signed server certs
- [ ] Document usage in server (and client)
- [ ] Fix developer experience
```

rossjones:
Adds TLS support to the bacalhau client and requester node so that all communication between the client and the server is encrypted.  This can be used in a private network as part of a zero-trust architecture, and as a trust/security measure for public networks.  

Resolves #2743 

## Client changes

By default, all communication with the requester API will be via HTTPS. For clusters where the server certificate was issued by a CA in the clients certstore, everything will work transparently.  Where the certificate was self-signed, or issued by an untrusted/unknown CA, then further steps will be required.

### Specifying the CA certificate 

By specifying `--cacert [filename.crt]` the client will use the provided CA certificate to verify the certificate when communicating with the requester node.  This certificate (often called ca.crt) must be provided by the certificate issuer (which may be the cluster owner in a self-signed scenario).

### Skipping certificate validation 

A new top level parameter was added to the CLI, `--insecure` will allow requests to be made to a HTTPS endpoint and the server certificate will not be verified.  

### Using legacy http connections

To connect to requester nodes that are running old HTTP endpoints, the client will need to specify the `--http` endpoint so that the clients created by the CLI will be able to downgrade the connections.

### Development mode 

By default, the devserver will run with no TLS settings and therefore will be in HTTP mode.  This means the CLI also needs to be in HTTP mode and so there are two ways to configure this.

```shell 
./bacalhau --http list 

export BACALHAU_HTTP=true 
./bacalhau list 
``` 

The alternative approach is to self-sign a certificate for your local machine, and you can achieve this (if you have openssl installed) with `make certs`.  This will generate the required materials in the ./certs directory but will not include them in any git transactions.

Once generated you can run a devstack with 

```shell
./bacalhau --tlscert ./certs/dev-server.crt --tlskey ./certs/dev-server.key devstack
```

Now the problem is that the CLI cannot verify the certificate (as it is extremely unlikely you will be developing on a public IP with a public DNS address - required for autocert).  You can use either of these approaches

```shell 
./bacalhau --insecure list 

./bacalhau --cacert ./certs/dev-ca.crt list 
```

or you may make use of the relevant environment variables (`BACALHAU_INSECURE, BACALHAU_CACERT`).

In future we may be able to provide options for a local CA and storing the cert in your operating system's certstore (to allow for verifiable certificates), but for now ubuntu users can do the following _every time they update the dev certificate_.

```
sudo apt-get install -y ca-certificates
cp ./certs/dev-ca.crt /usr/local/share/ca-certificates
sudo update-ca-certificates
```

## Server changes 

By default, if none of the following options are used, the server will continue to serve its API endpoints over HTTP. We _should_ default to HTTPS with a special opt-out, but we want to support some level of backwards compability. For the next major version, this should setting should be reversed and HSTS implemented.

### Autocert 

Using the `--autocert [hostname]` parameter to the CLI in the`serve` and `devstack` commands, a certificate will be obtained automatically from Lets Encrypt. The provided hostname can be a comma-separated list of hostnames, but they should all be publicly resolvable as Lets Encrypt will attempt to connect to the server to verify ownership (using the [ACME HTTP-01](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) challenge).  On the very first request this can take a short time whilst the first certificate is issued.  

As a result of the Lets Encrypt verification step, it is necessary for the server to be able to handle requests on port 80.  This typically requires elevated privileges, and rather than obtain these through a privileged account (such as root), we can instead use setcap to grant the executable the right to bind to ports <1024.  

```shell
sudo setcap CAP_NET_BIND_SERVICE+ep ./bacalhau 
```

A cache of ACME data will be held in the config directory, by default `~/.bacalhau/autocert-cache`, and this will be used to manage renewals as the certificates are typically short-lived. 

### Manually specified certificates

In cases where ACME/autocert is not suitable, or not desirable, it is possible to obtain certificates using the traditional route - create and submit a CSR (Certificate Signing Request) to a Certificate Authority. Or, if so desired a self-signed certificate can be created.

The server certificate and key are required, and the filenames passed to the `--tlscert` and `--tlskey` parameters respectively.  

### Development mode 

Unless testing TLS support, and until we can transparently solve issues with local CAs, it is recommended to not specify `--autocert`, `--tlskey` or `--tlscert` for the devstack and run with the default of http. 

## SDK 

TBD 

## TODO 

* [x] Developer experience 
* [ ] SDK to work in both modes (TLS on/off).
* [ ] Write-up tickets for investigating https://github.com/cert-manager


rossjones:
It's extremely unlikely that it will make sense to fix the merge conflicts as they're quite significant and wide ranging.  I'm going to keep the PR open though to crib for smaller partial PRs implementing the same thing.

rossjones:
Contents extracted into other PRs.

PR #2765, opened 2023-08-23
Use priority field for compute node execute buffer

rossjones:
Currently the executor_buffer which maintains the queue for the compute node running tasks where there are available resources to do so. This is currently a FIFO queue where jobs are processed in order, although it is able to skip entries in case a later job has fewer resource requirements which can be satisfied now.  

This PR changes the executor_buffer over to using a priority queue so that jobs are processed in order of priority (currently the job priority field). If dequeued tasks have resource requirements which can't currently be met, they will be requeued once the iteration has completed.

### DIFF
--- a/pkg/compute/executor_buffer.go
+++ b/pkg/compute/executor_buffer.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/bacalhau-project/bacalhau/pkg/compute/capacity"
 	"github.com/bacalhau-project/bacalhau/pkg/compute/store"
+	"github.com/bacalhau-project/bacalhau/pkg/lib/collections"
 	"github.com/bacalhau-project/bacalhau/pkg/logger"
 	"github.com/bacalhau-project/bacalhau/pkg/system"
 	sync "github.com/bacalhau-project/golang-mutex-tracer"
@@ -48,8 +49,8 @@ type ExecutorBuffer struct {
 	delegateService            Executor
 	callback                   Callback
 	running                    map[string]*bufferTask
-	enqueued                   map[string]*bufferTask
-	enqueuedList               []string
+	queuedTasks                *collections.PriorityQueue[*bufferTask]
+	enqueuedIDs                map[string]struct{}
 	defaultJobExecutionTimeout time.Duration
 	backoffDuration            time.Duration
 	backoffUntil               time.Time
@@ -64,8 +65,8 @@ func NewExecutorBuffer(params ExecutorBufferParams) *ExecutorBuffer {
 		delegateService:            params.DelegateExecutor,
 		callback:                   params.Callback,
 		running:                    make(map[string]*bufferTask),
-		enqueued:                   make(map[string]*bufferTask),
-		enqueuedList:               make([]string, 0),
+		queuedTasks:                collections.NewPriorityQueue[*bufferTask](),
+		enqueuedIDs:                make(map[string]struct{}),
 		defaultJobExecutionTimeout: params.DefaultJobExecutionTimeout,
 		backoffDuration:            params.BackoffDuration,
 	}
@@ -103,7 +104,7 @@ func (s *ExecutorBuffer) Run(ctx context.Context, localExecutionState store.Loca
 		err = fmt.Errorf("not enough capacity to run job")
 		return
 	}
-	if _, ok := s.enqueued[execution.ID]; ok {
+	if _, ok := s.enqueuedIDs[execution.ID]; ok {
 		err = fmt.Errorf("execution %s already enqueued", execution.ID)
 		return
 	}
@@ -116,8 +117,8 @@ func (s *ExecutorBuffer) Run(ctx context.Context, localExecutionState store.Loca
 		return
 	}
 
-	s.enqueued[execution.ID] = newBufferTask(localExecutionState)
-	s.enqueuedList = append(s.enqueuedList, execution.ID)
+	s.queuedTasks.Enqueue(newBufferTask(localExecutionState), execution.Job.Priority)
+	s.enqueuedIDs[execution.ID] = struct{}{}
 	s.deque()
 	return err
 }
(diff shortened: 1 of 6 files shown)


pkg/lib/collections/priority_queue.go
```diff
@@ -0,0 +1,205 @@
...
+
+	return pq.dequeue()
+}
+
+// dequeue is a lock-free version of Dequeue() for use by internal
+// methods that already have a lock
+func (pq *PriorityQueue[T]) dequeue() (T, int, error) {
+	if pq.IsEmpty() {
+		return *new(T), 0, ErrEmptyQueue
+	}
```
wdbaruni, line 76: the only valid error is when dequeueing an empty queue. Since we are using generics and strong typing, we shouldn't face casting error. I hate nils, but I also hate returning tuples and returning error when there can only be a single error possible as it can complicate caller handling of dequeues. 

This is a personal preference and what you are doing here is perfectly fine

pkg/lib/collections/priority_queue.go
```diff
@@ -0,0 +1,205 @@
...
+	pq.mu.Lock()
+	defer pq.mu.Unlock()
+
+	var result T
+	var priority int
+	var found bool
+
+	// Create a new Q to hold items that are not matches, this is suboptimal for time
+	// but not really an issue for space as we'll be using the same.
+	newQ := NewPriorityQueue[T]()
```
wdbaruni, line 106: it doesn't look like you need a temporary queue to hold the polled items as you are going to merge and sort them again when merging with the main queue. Unless merging sorted items is more efficient, but in this case the `newQ` should be in opposite order so you are merging the least priority item first. I might be wrong about this :S

pkg/lib/collections/priority_queue_test.go
```diff
@@ -0,0 +1,89 @@
...
+		{"A", 3},
+		{"B", 2},
+		{"C", 1},
+		{"C", 1},
+	}
+
+	pq := collections.NewPriorityQueue[string]()
+	for _, tc := range testcases {
+		pq.Enqueue(tc.v, tc.p)
+	}
```
wdbaruni, line 36: While testing I like to enqueue in an order that is different than the expected dequeue order to make sure we our queue is based on the priority and not on insertion order. So maybe:

```
inputs := []testcases{{"B", 2}, {"A", 3}, {"C", 1}, {"A", 3}, {"C", 1}, {"B", 2}}
expected := []testcases{{"A", 3},  {"A", 3}, {"B", 2}, {"B", 2} , {"C", 1},{"C", 1}}
```
```

File: pkg/lib/collections/priority_queue.go [hunk 4]
```diff
@@ -0,0 +1,205 @@
...
+
+	return item, heapItem.priority, nil
+}
+
+// DequeueWhere allows the caller to iterate through the queue, in priority order, and
+// attempt to match an item using the provided `MatchingFunction`.  This method has a high
+// time cost as dequeued but non-matching items must be held and requeued once the process
+// is complete.  Luckily, we use the same amount of space (bar a few bytes for the
+// extra PriorityQueue) for the dequeued items.
+func (pq *PriorityQueue[T]) DequeueWhere(matcher MatchingFunction[T]) (T, int, error) {
```
wdbaruni, line 96: Nice!

pkg/compute/executor_buffer.go
```diff
@@ -48,8 +50,8 @@ type ExecutorBuffer struct {
 	delegateService            Executor
 	callback                   Callback
 	running                    map[string]*bufferTask
-	enqueued                   map[string]*bufferTask
-	enqueuedList               []string
+	queuedTasks                *collections.PriorityQueue[*bufferTask]
+	enqueuedIDs                map[string]struct{}
```
wdbaruni, line 54: does it make sense to create a `HashedPriorityQueue` or `PriorityHashedQueue` under `collections` package that wraps a queue and a map for us and implements a queue with a `contains()` method that returns in `O(1)`? Might just simplify our testing if we extract this logic from the executor buffer

pkg/compute/executor_buffer.go
```diff
@@ -173,24 +175,37 @@ func (s *ExecutorBuffer) deque() {
...
+		if err != nil {
+			if errors.Is(err, collections.ErrNoMatch) {
+				break // we're all out of possible matches
+			}
+			log.Ctx(ctx).Error().Err(err).Msg("failed when searching for a task to run")
 		}
+
+		// Move the execution to the running list and remove from the list of enqueued IDs
+		// before we actually run the task
+		execID := task.localExecutionState.Execution.ID
```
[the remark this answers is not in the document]
rossjones: Can only be EmptyQueue or NoMatch, so in both cases we can't continue and should just return pending the next run.

pkg/lib/collections/priority_queue.go
rossjones: I agree, but there's some go-isms here that make it unpleasant.  One is, I think is that T can mask that it is a pointer and there's no clean way to specify (that I've found so far) as a generic type constraint so that returning nil will always be valid. If we accept *T we end up actually with **bufTask. 

What I'd much prefer is an Option type (Optional in Java8) but I gather they're frowned upon in Go.  Will see if we can wrap it into a useful type that we can return pointers to, perhaps something like 

```go
type QueueItem[T any] struct {
    val T 
    prio int 
}
```

and then we can more easily return a ptr to a concrete type.

rossjones: Implemented in e4535c4a and looks okay actually

pkg/lib/collections/priority_queue.go
```diff
@@ -0,0 +1,231 @@
...
+	// Re-add the items that were not matched back onto the Q
+	lo.ForEach(unmatched, func(item *QueueItem[T], _ int) {
+		pq.enqueue(item.Value, item.Priority)
+	})
+
+	// return the result we found, which might still be nil (not found)
+	return result
+}
+
+func (pq *PriorityQueue[T]) Merge(other *PriorityQueue[T]) {
```
wdbaruni, line 165: not used at the moment, but okay to leave it if you think useful

pkg/lib/collections/hashed_priority_queue.go
```diff
@@ -0,0 +1,84 @@
...
+	return ok
+}
+
+// Enqueue will add the item specified by `data` to the queue with the
+// the priority given by `priority`.
+func (q *HashedPriorityQueue[K, T]) Enqueue(data T, priority int) {
+	k := q.indexer(data)
+
+	q.identifiers[k] = struct{}{}
+	q.queue.Enqueue(data, priority)
```
wdbaruni, line 37: not thread safe. we should sync these operations

wdbaruni (approved):
looks great with couple of minor comments. feel free to merge after addressing them

pkg/lib/collections/priority_queue.go
rossjones: Meant to remove it actually, trivial enough to re-add if needed later.

PR #2767, opened 2023-08-24
Nodejs SDK

leostelon:


simonwo:
Hi Leo, just an update on this. We've made some changes to our v2 APIs which will necessitate some SDK changes, and we also want a Js SDK to work in the browser. So we're going to be combining what you've done here with some other work we are doing as part of building a Web UI. Thanks for your contributions so far, and we'll let you know when that work is merged!

leostelon:
That's awesome to know, happy to contribute😊

PR #2775, opened 2023-08-30
use echo web framework

wdbaruni:
This PR migrates from chi router to echo framework. 

The main benefit that I see with echo is the central error handling where the http handler signature `func(c echo.Context) error` returns an error instead of the standard `func(w http.ResponseWriter, r *http.Request)`. This allows injecting an error handler that captures all returned and unhandled errors, process and transform them before returning to the user. I am not utilizing the error handler in this PR because of how we are handling errors in the existing code. A similar behaviour is still possible in chi by passing an error handler and calling it inside the http handlers. We are already doing something similar, but it can be improved https://github.com/bacalhau-project/bacalhau/blob/chi/pkg/publicapi/endpoint/requester/endpoints_cancel.go#L37

The drawback is deviating from the standard library http handlers, which will make migrating away from echo to be more challenging, and that we will have to wrap 3rd party middleware that were implemented for the standard library and try to make them compatible with echo.

In my opinion, I don't see strong value with using echo and deviating away from the standard library handlers. I see go-chi more than enough for our needs, but I don't have strong opinion about this and I am okay with the switch given Forrest's positive experience with echo in a previous project.


pkg/publicapi/endpoint/compute/endpoint_approve.go
```diff
@@ -20,27 +20,25 @@ import (
...
 //	@Failure	500	{object}	string
 //	@Router		/api/v1/compute/approve [get]
-func (s *Endpoint) approve(res http.ResponseWriter, req *http.Request) {
-	request, err := signatures.UnmarshalSigned[bidstrategy.ModerateJobRequest](req.Context(), req.Body)
+func (s *Endpoint) approve(c echo.Context) error {
+	request, err := signatures.UnmarshalSigned[bidstrategy.ModerateJobRequest](c.Request().Context(), c.Request().Body)
 	if err != nil {
-		publicapi.HTTPError(req.Context(), res, err, http.StatusBadRequest)
-		return
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
```
frrist, line 26: Perhaps we are not at the point where we are ready to utilize this feature yet, but we could improve this through some middleware by doing:
```go
return echo.NewHTTPError(
    http.StatusBadRequest,
    "failed to unmarshal request").
    WithInternal(err)
```
This will allow use to make assertions on the error type in middleware. It will also allow us to include the specific error in our logs message via the `Err()` method. Also has some security benefit since the `Internal` field is tagged with `json:"-"`, which means it will be ignored when the HTTPError struct is serialized to JSON. Great for preventing potentially sensitive internal error details from being exposed to our end users.

pkg/publicapi/endpoint/compute/endpoint_approve.go
```diff
@@ -20,27 +20,25 @@ import (
...
+	executions, err := s.store.GetExecutions(c.Request().Context(), request.JobID)
 	if err != nil {
-		publicapi.HTTPError(req.Context(), res, err, http.StatusInternalServerError)
-		return
+		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 	}
 
 	for _, execution := range executions {
-		go s.bidder.ReturnBidResult(req.Context(), execution, &request.Response)
+		go s.bidder.ReturnBidResult(c.Request().Context(), execution, &request.Response)
```
frrist, line 40: I don't think we want to use the request context for this routine, as I believe the context will end before the bid result can be returned (might be a bug, although I realize the behavior wasn't added here). Since we have been using the in-memory job store that doesn't respect context cancelation (unlike the new Bolt store) it has probably gone unnoticed. But the change @rossjones is making in https://github.com/bacalhau-project/bacalhau/pull/2779 might cause this to surface at some point.

Do we even need the go-routine here, the BaseEndpoint may fail to update all execution states, in which case we incorrectly(?) signal to the user that the job has been approved when in fact it may be in a bad state. Perhaps this should be made synchronous, our modified use a waitgroup allowing parallelism, but avoiding a premature return or canceled request context.

pkg/publicapi/endpoint/compute/endpoint_approve.go
```diff
@@ -20,27 +20,25 @@ import (
...
-		return
+		return echo.NewHTTPError(http.StatusUnauthorized, errors.New("approval submitted by unknown client"))
 	}
 
-	executions, err := s.store.GetExecutions(req.Context(), request.JobID)
+	executions, err := s.store.GetExecutions(c.Request().Context(), request.JobID)
 	if err != nil {
-		publicapi.HTTPError(req.Context(), res, err, http.StatusInternalServerError)
-		return
+		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
```
frrist, line 36: Could this be a 404?

pkg/publicapi/endpoint/compute/endpoint_debug.go
```diff
@@ -16,16 +16,15 @@ import (
...
 //	@Router		/api/v1/compute/debug [get]
-func (s *Endpoint) debug(res http.ResponseWriter, req *http.Request) {
+func (s *Endpoint) debug(c echo.Context) error {
 	debugInfoMap := make(map[string]interface{})
 	for _, provider := range s.debugInfoProviders {
-		debugInfo, err := provider.GetDebugInfo(req.Context())
+		debugInfo, err := provider.GetDebugInfo(c.Request().Context())
 		if err != nil {
-			log.Ctx(req.Context()).Error().Msgf("could not get debug info from some providers: %s", err)
+			log.Ctx(c.Request().Context()).Error().Msgf("could not get debug info from some providers: %s", err)
```
frrist, line 24: Let's also include the error in the log message with `Err()`.

pkg/publicapi/endpoint/requester/endpoint_debug.go
```diff
@@ -16,16 +16,15 @@ import (
...
 //	@Router		/api/v1/requester/debug [get]
-func (s *Endpoint) debug(res http.ResponseWriter, req *http.Request) {
+func (s *Endpoint) debug(c echo.Context) error {
 	debugInfoMap := make(map[string]interface{})
 	for _, provider := range s.debugInfoProviders {
-		debugInfo, err := provider.GetDebugInfo(req.Context())
+		debugInfo, err := provider.GetDebugInfo(c.Request().Context())
 		if err != nil {
-			log.Ctx(req.Context()).Error().Msgf("could not get debug info from some providers: %s", err)
+			log.Ctx(c.Request().Context()).Error().Msgf("could not get debug info from some providers: %s", err)
```
frrist, line 24: `Err()` log the error

pkg/publicapi/endpoint/requester/endpoints_cancel.go
```diff
@@ -30,22 +30,22 @@ import (
...
-	ctx := req.Context()
-	jobCancelPayload, err := signatures.UnmarshalSigned[model.JobCancelPayload](ctx, req.Body)
+func (s *Endpoint) cancel(c echo.Context) error {
+	ctx := c.Request().Context()
+	jobCancelPayload, err := signatures.UnmarshalSigned[model.JobCancelPayload](ctx, c.Request().Body)
 	if err != nil {
-		publicapi.HTTPError(ctx, res, err, http.StatusBadRequest)
-		return
+		publicapi.HTTPError(c, err, http.StatusBadRequest)
+		return nil
```
frrist, line 38: Why do we return errors in someplaces and nil in other places? `publicapi.HTTPError` seems like a bad pattern.

pkg/publicapi/endpoint/requester/endpoints_result.go
```diff
@@ -25,32 +25,29 @@ import (
...
 	}
 
 	results := make([]model.PublishedResult, 0)
 	for _, execution := range executions {
 		if execution.ComputeState.StateType == models.ExecutionStateCompleted {
 			storageConfig, err := legacy.ToLegacyStorageSpec(execution.PublishedResult)
 			if err != nil {
-				http.Error(res, err.Error(), http.StatusInternalServerError)
-				return
+				return c.String(http.StatusInternalServerError, err.Error())
```
frrist, line 50: Why are we returning strings instead of json?

pkg/publicapi/endpoint/requester/endpoints_submit.go
```diff
@@ -27,34 +27,35 @@ import (
...
+	if otherJobID := c.Request().Header.Get("X-Bacalhau-Job-ID"); otherJobID != "" {
 		err := fmt.Errorf("rejecting job because HTTP header X-Bacalhau-Job-ID was set")
-		publicapi.HTTPError(ctx, res, err, http.StatusBadRequest)
-		return
+		publicapi.HTTPError(c, err, http.StatusBadRequest)
+		return nil
 	}
 
-	jobCreatePayload, err := signatures.UnmarshalSigned[model.JobCreatePayload](ctx, req.Body)
+	jobCreatePayload, err := signatures.UnmarshalSigned[model.JobCreatePayload](ctx, c.Request().Body)
```
frrist, line 39: I'd like to see the verification of these requests handled by middleware. Then, when we clean up our signature code we can modify the middleware implementation and leave methods like these unmodified.

frrist (commented):
Nice. Looks like we were able to remove a lot of our custom middlewares and adopt those already provided by echo 😀 (`+377 −691`).

Seems there are some test failures, I'll wait for those to be addresses before approving. Although you aren't targeting `main` with this, so feel free to merge without my approval.

pkg/publicapi/endpoint/compute/endpoint_approve.go
wdbaruni: yeah we would need to inject our own error handler for central handling of errors, which is what I like the most about echo. We need to rethink `bacerrors` package first

pkg/publicapi/endpoint/compute/endpoint_approve.go
wdbaruni: I don't think we need to worry much about `approve` APIs at the moment as we've removed that path in the requester's side. I created a backlog item to remove this https://expanso.atlassian.net/browse/ENG-86

pkg/publicapi/endpoint/requester/endpoints_cancel.go
wdbaruni: yeah it is because of `publicapi.HTTPError` implementation and how errors of type `bacerrors.ErrorResponse` are handled. Originally I only returned errors, and injected an error handler that handles `bacerrors.ErrorResponse` differently, but the change grew too much and was afraid I might've broken more stuff.

These are legacy APIs that I don't want to spend more time fixing or improving them

pkg/publicapi/endpoint/requester/endpoints_result.go
wdbaruni: good catch

pkg/publicapi/endpoint/requester/endpoints_submit.go
rossjones: Ideally we'd replace our request signing with something that doesn't need to modify the body of the request a likely cause of pain in building clients.  It would be great if https://www.ietf.org/archive/id/draft-ietf-httpbis-message-signatures-19.html were closer to completion, but even if it is too early to have an ecosystem around it (e.g. libraries for most languages) we can copy the approach (or chose a similar 'standard'). 

Adding this to middleware and then adding the relevant info to the echo Context will simplify a lot of the endpoints which will no longer need to worry about unravelling payloads.

wdbaruni:
> Comparatively, chi is designed to serve as a simple HTTP router for building a RESTful api service:

Yeah the difference is `chi` is a simple router leaving you the option to chose your own data binder, renderer, middlewares and so on, while `echo` is an opinionated framework (not necessary a bad thing) that comes with all those together within the same framework.

Lets close this topic and go with echo. Can you please finish reviewing #2772 as it contains more than just using chi, and then we can continue working on this PR to switch to echo

pkg/publicapi/endpoint/requester/endpoints_submit.go
wdbaruni: Yeah I am not planning to add the same signing methods to the new APIs as we will need to figure out a better way as you mentioned

wdbaruni:
bypassing PR approval as the comments have been addressed and Monday is a holiday in the US

pkg/publicapi/endpoint/compute/endpoint_approve.go
frrist: We could adopt one of the existing solutions from: https://awesome-go.com/error-handling/

frrist (commented):
LGTM

PR #2776, opened 2023-08-30
Introduces a RestartPolicy to reflect desired behaviour of tasks

rossjones:
Currently the compute makes decisions on how to restart executions when the compute node starts up, based on what it believes each job type should do.

This PR introduces a requester node determined RestartPolicy which informs the compute node how to restart executions without it needing to know anything about the job type.


pkg/models/restart_policy.go
```diff
@@ -0,0 +1,24 @@
+//go:generate stringer -type=RestartPolicyType --trimprefix=RestartPolicy --output restart_policy_string.go
+package models
+
+type RestartPolicyType int
+
```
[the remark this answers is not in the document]
rossjones: Took your advice and switched to a simple struct with an attempts count.  This number is decremented everytime we enter a Running state, but it did mean a slight change to the state machine allowing executions to exit terminal states (in this case cancellation) into Running states if they have any attempts left.

pkg/compute/executor.go
```diff
@@ -126,14 +126,15 @@ func PrepareRunArguments(
...
-		Outputs:      execution.Job.Task().ResultPaths,
-		Inputs:       inputVolumes,
-		ResultsDir:   resultsDir,
-		EngineParams: engineArgs,
+		JobID:         execution.Job.ID,
+		ExecutionID:   execution.ID,
+		Resources:     execution.TotalAllocatedResources(),
+		Network:       execution.Job.Task().Network,
+		RestartPolicy: execution.Job.Task().RestartPolicy,
+		Outputs:       execution.Job.Task().ResultPaths,
```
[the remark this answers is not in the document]
rossjones: Good point, think this was a pre-emptive change that I didn't end up needing.

pkg/compute/executor.go
```diff
@@ -167,10 +167,18 @@ func (e *BaseExecutor) Run(ctx context.Context, localExecutionState store.LocalE
...
 	}()
 
+	// We expect the current state to be BidAccepted but in some cases we will be restarting
+	// a failed job which will be in the Cancelled state (although this may later be upgraded
+	// to a FailedState).
+	expectedState := store.ExecutionStateBidAccepted
+	if localExecutionState.State == store.ExecutionStateCancelled {
+		expectedState = store.ExecutionStateCancelled
+	}
+
```
wdbaruni, line 177: I don't get this. If the execution is in cancelled state, then we shouldn't run or retry it

pkg/compute/startup.go
```diff
@@ -53,23 +51,21 @@ func (s *Startup) ensureLiveJobs(ctx context.Context) error {
...
-			{
-				// Service and System jobs are long running jobs and so we need to make sure it is running
-				err = s.runExecution(ctx, localExecution)
-				if err != nil {
-					errs = multierror.Append(errs, err)
-				}
+		// Fetch the restart policy for the current task.
+		policy := localExecution.Execution.Job.Task().RestartPolicy
+
+		if policy.Attempts > 0 {
```
wdbaruni, line 57: `models.Task` should not be updated by the system and should only be updated by the user. You shouldn't decrement `Attempts`, and instead increment something in the `localExecutionState` or in the `models.Execution`, and compare that with task attempts

pkg/compute/store/boltdb/store.go
```diff
@@ -413,6 +421,11 @@ func (s *Store) updateExecutionState(tx *bolt.Tx, request store.UpdateExecutionS
 	localExecutionState.Version += 1
 	localExecutionState.UpdateTime = time.Now()
 
+	// Each time we move into the running state, we want to use up one of the restart attempts
+	if request.NewState == store.ExecutionStateRunning {
+		localExecutionState.Execution.Job.Task().RestartPolicy.Attempts = retryAttempts - 1
+	}
+
```
wdbaruni, line 428: it sounds to me we are doing too much logic inside the data store. A data store should be as simple and stupid as possible, and let the business logic live somewhere else. It should just compare conditions, and validate the integrity of the end state, but not change the models on behalf of the caller. Or else you would need to do this in all data store implementations

wdbaruni (commented):
left some comments, but based on the discussion we just had and urgency to prepare for the next release, I believe this change can wait until we better plan for a restart policy. I don't see a problem with keeping your previous logic with restarting jobs based on their type as this PR is effectively doing the same

pkg/compute/executor.go
rossjones: This is why I'm hoping for a Failed state.  At the moment, it is us who cancelled the execution in startup.go

pkg/compute/startup.go
rossjones: Ah yeah, that's what I was trying to get at this morning with my questions around copying stuff from tasks into executions.

rossjones:
Closing pending clearer plan for restart policies.

PR #2777, opened 2023-08-30
Detect system GPUs using `nvidia-smi` binary

simonwo:
Previously we used to use the `nvidia-container-cli` binary but this is not present within Docker containers that have GPU access, whereas this new binary is. This allows Bacalhau running in a container to still query for system GPUs.


rossjones (approved):
LGTM, unfortunately don't have a GPU to try it with, but code looks safe.

PR #2778, opened 2023-08-30
Async Compute Executors 

frrist:
### What
This PR modifies the executor interface to include the following additions:
```go
Start(ctx context.Context, request *RunCommandRequest) error
Wait(ctx context.Context, executionID string) (<-chan *models.RunCommandResult, error)
```
It then updates the Noop, Wasm and Docker executors to satisfy the new interface requirements. Each of these executor changes needs some review as the changes are more than pure mechanical changes. A `handler` pattern has been implemented to manage each executors active executions.

This PR updates the `BaseExecutor` to also include a `Start` and `Wait` method, however these methods are not used outside of the BaseExecutor, and are instead called via the BaseExecutors Run method. This allows the ExecutorBuffer to continue using the BaseExecutor as it did before this change.

### Call outs
- Previously when a docker container was passed an invalid entrypoint/command it would return an error rather than start the execution. There were some tests that made assertions on this. Since starting the docker container is the same as running the job, and we want to run jobs async, it is no longer the case that these errors are returned immediately. They are instead shown when describing the failed executions state. I have modified the required tests to account for this change in behavior.
- The ExecutorBuffer triggers job timeouts via context cancellation by passing the Run method a context with a timeout. I am dubious this will continue to work as intended when operating against pluggable executors and suspect it will instead end the client connection to the plugin server. For the time being I have worked around this [here](https://github.com/bacalhau-project/bacalhau/blob/9834a589ee491f74f65b5e3b603a904348ab8453/pkg/compute/executor.go#L288) and added a TODO comment. IMO we should instead explicitly call the `Cancel` method on an executor to cancel its operation instead of relying on the passed context. As mentioned in the TODO comment there is more context on canceled job states [here](https://github.com/bacalhau-project/bacalhau/pull/2705#discussion_r1284228115)
- As Ross has already brought attention to [here](https://github.com/bacalhau-project/bacalhau/pull/2778/files#r1314919569) `PrepareRunArguments` now returns a function that may be called to trigger cleanup of input volumes. We should continue discussion in the linked thread regarding further changes to this design. I believe this is related to some work @simonwo is landing regarding restarts.
- WriteJobResults function signature has been modified to no longer return an error, instead the error is embedded (as it already was) in the returned `RunCommandResult` structure. This change was made to simplify logic around checking errors returned by the method.


pkg/compute/executor.go
```diff
@@ -86,141 +121,220 @@ func PrepareRunArguments(
...
+			Inputs:       inputVolumes,
+			ResultsDir:   resultsDir,
+			EngineParams: engineArgs,
+			OutputLimits: executor.OutputLimits{
+				MaxStdoutFileLength:   system.MaxStdoutFileLength,
+				MaxStdoutReturnLength: system.MaxStdoutReturnLength,
+				MaxStderrFileLength:   system.MaxStderrFileLength,
+				MaxStderrReturnLength: system.MaxStderrReturnLength,
+			},
+		}, func(ctx context.Context) error {
```
rossjones, line 167: Might it be worth moving this cleanup function into the `RunCommandRequest`?  At the moment it's an anonymous looking func and it isn't clear in the signature what it is/what it's for.  By adding it to the request we can turn it into a post-execute/cleanup step, it's clearer what it is.  Understand if you'd like to keep it outside the structure, but in that case it would be good to explain what it is in the func doc/comment.

pkg/executor/docker/executor.go
```diff
@@ -95,277 +110,259 @@ func (e *Executor) Run(
...
+		Str("jobID", request.JobID).
+		Msg("starting execution")
+
+	if handler, found := e.handlers.Get(request.ExecutionID); found {
+		if handler.active() {
+			// TODO we can make this method idempotent by not returning an error here if the execution is already active
+			return fmt.Errorf("execution (%s) already started", request.ExecutionID)
 		} else {
-			return executor.FailResult(fmt.Errorf("unknown storage volume type: %s", input.Volume.Type))
+			// TODO what should we do if an execution has already completed and we try to start it again? Just rerun it?
```
rossjones, line 141: Currently, it's not possible to go from any terminal state (cancelled, failed, complete) to a running state again.  

Except long-running jobs, if they end up in a failed state (due to a compute node restart) we should be able to restart the execution _without_ obtaining a new allocation from the requester node which may, or may not, be available.

pkg/compute/executor.go
frrist: Agreed, the current code structure is lacking comments describing it's usage, I'll add them before taking this out of draft.

> Might it be worth moving this cleanup function into the `RunCommandRequest`?

I think adding the cleanup to a structure related to the bits being cleaned up would be cleaner, so generally in favor of this.  However, I think it might also fit on the `models.Execution`, which would allow us to perform the cleanup when the execution reaches a terminal state. Since the state we are cleaning here is related to inputs we could probably (safely?) keep it around in the event that an execution is rescheduled or restarted. What do you think?

For now I will keep the cleanup logic as a clearly-commented anonymous function.

pkg/executor/docker/executor.go
frrist: It might be too early to answer this, but would we expect a restarted execution (recovered execution?) to go through a this same code path? Or would there be a different method on the executor that would "recover" and start a previously failed execution?

For now do you think it makes sense to continue erroring here?

pkg/executor/docker/executor.go
```diff
@@ -95,277 +110,259 @@ func (e *Executor) Run(
...
+	return handler.kill(ctx)
+}
 
-	if _, set := os.LookupEnv("SKIP_IMAGE_PULL"); !set {
-		dockerCreds := config.GetDockerCredentials()
-		if pullErr := e.client.PullImage(ctx, dockerArgs.Image, dockerCreds); pullErr != nil {
-			pullErr = errors.Wrapf(pullErr, docker.ImagePullError, dockerArgs.Image)
-			return executor.FailResult(pullErr)
-		}
+func (e *Executor) GetOutputStream(ctx context.Context, executionID string, withHistory bool, follow bool) (io.ReadCloser, error) {
```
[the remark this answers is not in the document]
frrist: I'd like to avoid forcing plugin implementations to meet the requirement of implementing `GetOutputStream` as doing so can be fairly complicated and burdensome. So agreed that we should re-think this methods presence on the interface. Ideally we solve this problem once for all current and future executor implementations. 

Accessing `stderr` and `stdout` of plugin process is [fairly straightforward](https://github.com/hashicorp/go-plugin/blob/main/client.go#L183-L193) via the client config on the go-plugin process. However, since an executor plugin is responsible for executing **many** executions, muxing the streams will still be necessary, as spinning off a plugin process for each execution doesn't seem practical. If we can enforce that every execution prefixes its output with its executionID then we should be able to filter the stream on the compute side to preserve methods such as `bacalhau log`

pkg/compute/executor.go
rossjones: I think I'd rather leave it where it is than add it to the Execution as it's a pretty inactive model at present and intended to be immutable. I don't think it's definitively wrong, it just feels weird returning a 3-tuple from a method instead of the usual 2-tuple.

pkg/executor/docker/executor.go
rossjones: I think this makes sense here, but I need to fix up the Startup so that for any long-running jobs it _only_ calls Run() (and not cancel/fail) and let's the executor deal with the idempotence.

pkg/executor/types.go
```diff
@@ -28,6 +28,8 @@ type Executor interface {
 	// Details in: https://github.com/bacalhau-project/bacalhau/issues/2702
 
 	Run(ctx context.Context, Args *RunCommandRequest) (*models.RunCommandResult, error)
+	Start(ctx context.Context, request *RunCommandRequest) error
```
rossjones, line 31: Possibly the stupidest nit ever, but could `Start` be above `Run`?

pkg/executor/wasm/handler.go
```diff
@@ -0,0 +1,186 @@
...
+	// TODO this is a little racey
+	h.running.Store(true)
+	close(h.activeCh)
+
+	// The function should exit which results in a sys.ExitError. So we capture
+	// the exit code for inclusion in the job output, and ignore the return code
+	// from the function (most WASI compilers will not give one). Some compilers
+	// though do not set an exit code, so we use a default of -1.
+	_, wasmErr := entryFunc.Call(wasmCtx)
+	exitCode := int64(-1)
```
rossjones, line 157: Default exit code should probably be 0 (success).  Using negative numbers for exit codes is a bit archaic and seems to suggest we're defaulting to fail (for which we should use 1 as a sort of generic fail).

pkg/test/executor/docker_entrypoint_test.go
```diff
@@ -158,21 +127,146 @@ func (suite *DockerEntrypointTestSuite) TearDownSuite() {
...
 }
 
-type tests []struct {
-	expectedStderr string
-	expectedStdout string
-	image          string
-	entrypoint     []string
-	parameters     []string
-	expectError    bool
+func (suite *DockerEntrypointTestSuite) TestTableDriven() {
```
rossjones, line 130: 😍

pkg/compute/executor.go
```diff
@@ -56,26 +59,58 @@ func NewBaseExecutor(params BaseExecutorParams) *BaseExecutor {
...
+
 func PrepareRunArguments(
 	ctx context.Context,
 	strgprovider storage.StorageProvider,
 	execution *models.Execution,
 	resultsDir string,
-	cleanup *system.CleanupManager,
-) (*executor.RunCommandRequest, error) {
-	inputVolumes, err := storage.ParallelPrepareStorage(ctx, strgprovider, execution.Job.Task().InputSources...)
+) (*executor.RunCommandRequest, func(context.Context) error, error) {
```
rossjones, line 105: Although `func(context.Context) error` is nice and explicit, do you think we should use a type to describe the Cleanup function as a little bit of extra inline documentation? Perhaps as 

```go
type CleanupFn(context.Context) error
```

pkg/executor/docker/executor.go
```diff
@@ -95,277 +110,259 @@ func (e *Executor) Run(
...
+func (e *Executor) Start(ctx context.Context, request *executor.RunCommandRequest) error {
+	log.Ctx(ctx).Info().
+		Str("executionID", request.ExecutionID).
+		Str("jobID", request.JobID).
+		Msg("starting execution")
+
+	if handler, found := e.handlers.Get(request.ExecutionID); found {
+		if handler.active() {
+			// TODO we can make this method idempotent by not returning an error here if the execution is already active
+			return fmt.Errorf("execution (%s) already started", request.ExecutionID)
```
rossjones, line 139: Instead of returning no error, how about returning ErrAlreadyStarted (or ErrAlreadyRunning) so that caller can make the decision as to whether to allow it (long running job) or whether it's a mistake (batch job).

pkg/executor/docker/executor.go
```diff
@@ -95,277 +110,259 @@ func (e *Executor) Run(
...
+	return nil
+}
 
-			Type: mount.TypeBind,
-			// this is an output volume so can be written to
-			ReadOnly: false,
+// Wait waits for the completion of a specific execution by its executionID.
+// A channel is returned that will emit the result of the execution.
+// TODO: The current implementation spawns a new Goroutine for each Wait call,
+// which could become resource-intensive with a large number of concurrent Wait calls.
```
rossjones, line 188: I don't think this is much of a problem, will we really create many of this per execution?  Even if for some reason we end up with dozens of goroutines it doesn't seem excessive. We have a broadcaster in pkg/util/generic/broadcaster.go which hands out channels, but is super-cautious to avoid writing to closed channels and as a result is probably not as performant as it could be. 

Is there a case that would generate many goroutines that I'm missing?

rossjones (commented):
I _think_ this looks okay (great even), but I think it should handle the idempotence problem with Run() before it lands. A few other comments/opinions too.

I will have a play with the branch manually while you consider :D

pkg/executor/types.go
frrist: 😂 you got it!

pkg/executor/wasm/handler.go
frrist: I'll default it to `0` here, and in the below code:
```go
if wasmErr != nil {
	h.logger.Warn().Int64("exit_code", exitCode).Err(wasmErr).Msg("execution ended")
}
```
set it to `1`

pkg/executor/docker/executor.go
frrist: Good idea, adding these errors now. 
It's probably worth pointing out that:
1. The BaseExecutor will likely catch this case before we hit it when updating its store since it would fail to transition an execution from the `ExecutionStateRunning` to the `ExecutionStateRunning` again. Similarity from a terminal state to the `ExecutionStateRunning`.
2. We'll need to be diligent in our review of executor plugins that follow these implementation ensuring they also use these error messages if we are going to be making assertions on them in calls higher up the stack.

rossjones: > transition an execution from the ExecutionStateRunning to the ExecutionStateRunning again.

Yes, I think we'll need to get agreement that this is a valid non-transition. And terminal to running is only valid for long-running jobs.

rossjones:
It'll also be necessary to update the protobufs with the new interface otherwise the plugins fail to load (as the `GRPCClient` that is dispensed can't be cast to an `executor.Executor` impl. This manifests currently as a crash at startup with running devstack with `--pluggable-executors`.

simonwo:
I don't think I get this just yet. I thought the reason we were making them async was so that we could pick them up and resume them if the compute node goes down. I can't see a way to do that just with `Start` and `Wait`, doesn't `Start` need to be idempotent or we need some way to detect executions already in progress so we can selectively start them?

frrist:
> It'll also be necessary to update the protobufs with the new interface otherwise the plugins fail to load (as the GRPCClient that is dispensed can't be cast to an executor.Executor impl. This manifests currently as a crash at startup with running devstack with --pluggable-executors.

@rossjones this has been done in: [2e1b46d](https://github.com/bacalhau-project/bacalhau/pull/2778/commits/2e1b46d4ad11daad61642e8facc0019f2a6ddcee)

pkg/executor/docker/executor.go
frrist: I can't think of any, was leaving it more as a hint for future implementers. I have removed the TODO and changed the comment to be more specific to the function of the method.

frrist:
@rossjones the `BaseExecutor.Run` method has been made idempotent for duplicate calls to `Start` here: https://github.com/bacalhau-project/bacalhau/pull/2778/commits/ed60b77ff773b55e276f603885d6b6ebeb821076

pkg/compute/executor.go
frrist: this has been addressed in a subsequent commit

pkg/docker/docker.go
```diff
@@ -195,13 +196,7 @@ func (c *Client) GetOutputStream(ctx context.Context, id string, since string, f
...
 func (c *Client) RemoveContainer(ctx context.Context, id string) error {
 	log.Ctx(ctx).Debug().Str("id", id).Msgf("Container Stop")
-	timeout := time.Millisecond * 100
-	if err := c.ContainerStop(ctx, id, timeout); err != nil {
-		if dockerclient.IsErrNotFound(err) {
-			return nil
-		}
-		return errors.WithStack(err)
-	}
+	// ContainerRemove kills and removes a container from the docker host.
```
rossjones, line 199: Is it safe to do this without calling stop first to give it a chance to shut down gracefully?  
From what I can tell Stop sends a SIGINT (or SIGTERM),  but kill will send SIGKILL (which is uncatchable).
Maybe in a future PR we should look at the docker lifecycle.

pkg/executor/docker/executor.go
```diff
@@ -43,15 +40,18 @@ const (
 
 type Executor struct {
 	// used to allow multiple docker executors to run against the same docker server
-	ID          string
+	ID string
+
+	handlers generic.SyncMap[string, *executionHandler]
```
rossjones, line 45: I'm guessing it's execution id -> handler, but is it worth a brief comment?  
In general though I like this pattern, it makes it much clearer that `one executor -> many active executions` and it wasn't super clear from our code before.

pkg/executor/docker/executor.go
```diff
@@ -90,282 +105,300 @@ func (e *Executor) ShouldBidBasedOnUsage(
...
+	log.Ctx(ctx).Info().
+		Str("executionID", request.ExecutionID).
+		Str("jobID", request.JobID).
+		Msg("starting execution")
+
+	if handler, found := e.handlers.Get(request.ExecutionID); found {
+		if handler.active() {
+			return fmt.Errorf("starting execution (%s): %w", request.ExecutionID, executor.AlreadyStartedErr)
+		} else {
+			return fmt.Errorf("starting execution (%s): %w", request.ExecutionID, executor.AlreadyCompleteErr)
```
rossjones, line 125: I'm sure I've seen a linter (or gopls) complain before that errors should be `ErrSomething` instead of `SomethingErr`.

rossjones (approved):
Looks good and I was able to test it manually in addition to CI.  A few comments in a few places, but nothing that should stop us merging.  I wasn't able to test restarting as I think there are some commits related to repo location that are not available on this branch yet, but will test/fix afterwards.

pkg/compute/executor.go
```diff
@@ -56,26 +59,71 @@ func NewBaseExecutor(params BaseExecutorParams) *BaseExecutor {
...
+	inputVolumes, err := storage.ParallelPrepareStorage(ctx, strgprovider, inputSources...)
+	if err != nil {
+		return nil, nil, err
+	}
+	return inputVolumes, func(ctx context.Context) error {
+		return storage.ParallelCleanStorage(ctx, strgprovider, inputVolumes)
+	}, nil
+}
+
+func prepareWasmVolumes(ctx context.Context, strgprovider storage.StorageProvider, wasmEngine wasmmodels.EngineSpec) (map[string][]storage.PreparedStorage, func(context.Context) error, error) {
```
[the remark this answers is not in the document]
frrist: There is an existing TODO for this where the method is called: https://github.com/bacalhau-project/bacalhau/blob/frrist/async-executor/pkg/compute/executor.go#L127

pkg/compute/executor.go
```diff
@@ -86,141 +134,241 @@ func PrepareRunArguments(
...
+		} else if errors.Is(err, executor.AlreadyCompleteErr) {
+			/*
+				if execution.Job.Type == models.JobTypeDaemon || execution.Job.Type == models.JobTypeService {
+					 // TODO: [decision]
+					 if the execution is already completed then it's in a terminal state. We need to decide if
+					 going from terminal to running is valid for job execution. Seems it could be valid for
+					 long-running jobs, but shouldn't be the case for "batch" and "ops" jobs.
+					 Gut check says we should create a new execution for the job and start that, reason being
+					 that we can preserve the history that a job failed and was restarted in the job history.
+				}
```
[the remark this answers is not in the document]
frrist: 👍 I have removed this error assertion and comment.

pkg/compute/executor.go
```diff
@@ -86,141 +134,241 @@ func PrepareRunArguments(
...
+			// that have timed out to be in state 'Canceled', rather than 'Failed'. IMO An execution that doesn't
+			// complete in the timeframe requested by a user should be 'Failed'. 'Canceled' probably ought to be
+			// reserved for actions initiated by a user. We are ignoring _only_ context.DeadlineExceeded
+			// errors and allowing context.Canceled errors be to returned so that when a compute node is shutdown
+			// any active executions will be labeled as 'Failed' instead of canceled.
+			// There is prior discussion regarding this point here:
+			// https://github.com/bacalhau-project/bacalhau/pull/2705#discussion_r1283543457
+			//
+			// Moving forward we must avoid canceling executions via the context.Context. When pluggable executors
+			// become the default since canceling the context will simply result in the RPC connection closing (I think)
```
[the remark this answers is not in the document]
frrist: We don't necessarily want to kill the plugin _process_ associated with this execution when the context is canceled. This specific error assertion is due to how we trigger timeout cancellations here: https://github.com/bacalhau-project/bacalhau/blob/frrist/async-executor/pkg/compute/executor_buffer.go#L139 
We only want to kill the plugin process when the bacalhau node is shutdown. And we ensure the plugin processes are ended when the node shuts down through the cleanup manager here: https://github.com/bacalhau-project/bacalhau/blob/frrist/async-executor/pkg/executor/util/utils.go#L168

pkg/compute/executor.go
```diff
@@ -266,32 +414,37 @@ func (e *BaseExecutor) publish(ctx context.Context, localExecutionState store.Lo
...
 	log.Ctx(ctx).Debug().Str("Execution", execution.ID).Msg("Canceling execution")
-	if cancel, found := e.cancellers.Get(execution.ID); found {
-		e.cancellers.Delete(execution.ID)
-		cancel()
+
+	// TODO is returning the error the correct behaviour here?
+	exe, err := e.executors.Get(ctx, execution.Job.Task().Engine.Type)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
frrist: I'll yeet this TODO then!

pkg/docker/docker.go
frrist: I removed the above lines as I didn't like the arbitrary timeout for stopping the container, I expected this method wrapper to _only_ remove the container and was surprised to see it doing more. If we want the extra stop call back, I can add it.

The docker `executionHander` uses its `kill` method to stop the container, and internally `kill` calls `ContainerStop`. We only use `RemoveContainer` in `destroy`, which calls `kill` before removing the container and associated objects. I have left a [TODO](https://github.com/bacalhau-project/bacalhau/blob/frrist/async-executor/pkg/executor/docker/handler.go#L151) in the aforementioned `kill` method regarding the need for better lifecycle management. In future changes we should allow a signal to be sent to the container when stopping it. The underlying docker client supports this behavior, we just need to wire it up.

pkg/executor/docker/executor.go
frrist: I have added a comment describing the maps purpose. You are correct, executionID -> handler.

pkg/executor/docker/executor.go
frrist: I promise I ran with the precommit hook on every commit, I event edited these lines and re-ran the commit hook - no complaints. This either indicates we are missing a linter rule, or its misconfiguration. Regardless, I have renamed these vars to begin with `Err*`.

pkg/executor/wasm/handler.go
```diff
@@ -0,0 +1,189 @@
...
+		exitCode = int64(errExit.ExitCode())
+		wasmErr = nil
+		h.logger.Info().Int64("exit_code", exitCode).Msg("execution ended")
+	}
+	if wasmErr != nil {
+		// in the event that an error is returned without an exist code we'll assume the operation
+		// failed and set the exit code to 1
+		exitCode = 1
+		h.logger.Warn().Int64("exit_code", exitCode).Err(wasmErr).Msg("execution ended")
+	}
```
simonwo, line 169: This a change from the original code which I don't think is necessary or explained anywhere. As explained in the comment above, the default exit code should be -1 unless one is supplied by the WASM runtime. There is no requirement to return any specific exit code in the case of an error.

Please revert this and we can discuss elsewhere.

pkg/executor/wasm/executor.go
```diff
@@ -61,6 +57,210 @@ func (*Executor) ShouldBidBasedOnUsage(
...
+//
+// Returns:
+// - A pointer to a RunCommandResult object, containing the result of the execution.
+// - An error if either starting or waiting for the execution fails, or if the context is cancelled.
+//
+// Steps:
+//  1. Starts the execution using the Start method. If it fails, returns an error.
+//  2. Waits for the execution to complete using the Wait method. If it fails, returns an error.
+//  3. Listens on the channel returned by Wait, and returns the result when available.
+//     If the context is cancelled, it returns a context error.
```
simonwo, line 245: Generally, comments that explain the internal workings of the method in detail are a code smell because they quickly become out of sync as the code changes and there is no enforcement that they stay correct. I'm guessing GPT generated this?

pkg/executor/wasm/executor.go
```diff
@@ -61,6 +57,210 @@ func (*Executor) ShouldBidBasedOnUsage(
...
 
+// Start initiates the execution of a WebAssembly (WASM) task using the provided RunCommandRequest.
+// It first checks if the execution with the given ExecutionID is already active or completed,
+// returning an error if so.
+//
+// This method performs several steps to initiate the execution:
+// 1. Decodes the engine parameters specific to the WASM engine.
+// 2. Configures memory limits for the WASM runtime, based on the resources specified in the request.
+// 3. Creates a virtual filesystem from storage to be used by the WASM execution.
+// 4. Sets up a new log manager specific to the execution for capturing logs.
```
simonwo, line 68: See other message about comments detailing implementation.

pkg/executor/results.go
```diff
@@ -140,7 +143,11 @@ func WriteJobResults(resultsDir string, stdout, stderr io.Reader,
 	}
 
 	result.ExitCode = exitcode
-	return result, err
+	return result
+}
+
+func NewFailedResult(reason string) *models.RunCommandResult {
+	return &models.RunCommandResult{ErrorMsg: reason, ExitCode: -1}
 }
```
simonwo, line 151: Do not set the exit code if the underlying runtime does not provide one. There is no requirement to provide an exit code in all circumstances.

pkg/executor/docker/executor_test.go
```diff
@@ -337,14 +381,17 @@ func (s *ExecutorTestSuite) TestDockerNetworkingAppendsHTTPHeader() {
...
 
-	result, err := s.runJob(task)
+	executionID := uuid.New().String()
+	result, err := s.runJob(task, executionID)
 	require.NoError(s.T(), err)
-	require.Equal(s.T(), "test", result.STDOUT, result.STDOUT)
+	require.Equal(s.T(), executionID, result.STDOUT, result.STDOUT)
 }
 
+/*
```
simonwo, line 390: Why is this commented out? The test is about jobs timing out – presumably we still expect that to work? If the method by which we do that has changed, then just update this to test the new method.

pkg/compute/executor.go
```diff
@@ -86,141 +134,229 @@ func PrepareRunArguments(
...
+			// failed to start? Perhaps it would be better to have metrics for exection failures here and job failures
+			// higher up the call stack?
+			jobsFailed.Add(ctx, 1)
+			return err
+		}
 	}
-	defer runCommandCleanup.Cleanup(ctx)
 
-	runCommandResult, err := jobExecutor.Run(ctx, runCommandArguments)
+	result, err := e.Wait(ctx, state)
```
simonwo, line 307: Given that we no longer call `jobExectuor.Run`, do we need executor plugins to implement a `Run` method anymore? Can we remove it from the interface?

simonwo (changes requested):

pkg/executor/wasm/handler.go
rossjones: Sorry, this is my bad. -1 seemed like an odd number to return for an error code, so I asked that it be changed to something more traditional.

pkg/executor/wasm/executor.go
frrist: Yes GPT wrote this comment as well as others. Would you prefer the `Steps:` section be removed?

simonwo: Yes please!

simonwo:
Unblocking so you can marge when you're ready in my PM.

pkg/compute/executor.go
frrist: It's still used extensively in our executor scenario test cases as the method offers some conveniences for running an execution and returning its output. Would be be okay to punt the removal of this from the interface into a follow on when we are more prepared for the full async nature of the new executor interface?

simonwo: Sure thing!

pkg/executor/docker/executor_test.go
frrist: fixed in latest commit.

PR #2779, opened 2023-08-31
Default to persistent stores

rossjones:
Now prod, test, dev, staging envs will use boltdb by default for persisting state,
with devstack continuing to use inmemory. This is easily changed for devstack
using the `COMPUTE_STORE_TYPE`, and `JOB_STORE_TYPE` environment
variables set to "boltdb".

In addition to making them default, it also changes the local of where the database 
files are stored. They continue to be part of the config repo, but are stored 
according to their node id and type. For each database file (executions.db and 
jobs.db) they can be found in a folder like `~/.bacalhau/{node-id}-{node-type}`.

### DIFF
--- a/pkg/config/configenv/test.go
+++ b/pkg/config/configenv/test.go
@@ -1,3 +1,4 @@
+//nolint:gomnd
 package configenv
 
 import (
--- a/pkg/config/configenv/test.go
+++ b/pkg/config/configenv/test.go
@@ -85,7 +85,7 @@ var TestingComputeConfig = types.ComputeConfig{
 		},
 	},
 	ExecutionStore: types.StorageConfig{
-		Type: types.InMemory,
+		Type: types.BoltDB,
 		Path: "",
 	},
 	JobTimeouts: types.JobTimeoutConfig{
@@ -120,7 +120,7 @@ var TestingRequesterConfig = types.RequesterConfig{
 		ProbeExec:           "",
 	},
 	JobStore: types.StorageConfig{
-		Type: types.InMemory,
+		Type: types.BoltDB,
 		Path: "",
 	},
 	HousekeepingBackgroundTaskInterval: types.Duration(30 * time.Second),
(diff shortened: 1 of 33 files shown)


pkg/repo/execution_store.go
```diff
@@ -14,14 +14,17 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/config/types"
 )
 
-// InitExecutionStore must be called after Init
+// InitExecutionStore must be called after Init and uses the configuration to create a
+// new ExecutionStore.  Where BoltDB is chosen, and no path is specified, then the database
+// will be created in the repo in a folder labeledafter the node ID.  For example:
+// `~/.bacalhau/Qmd1BEyR4RsLdYTEym1YxxaeXFdwWCMANYN7XCcpPYbTRs-compute/executions.db`
 func (fsr *FsRepo) InitExecutionStore(ctx context.Context, nodeID string) (store.ExecutionStore, error) {
```
frrist, line 21: nit: we could rename the `nodeID` param to `prefix`

pkg/repo/fs.go
```diff
@@ -360,7 +359,7 @@ func (fsr *FsRepo) WriteRunInfo(ctx context.Context, summaryShellVariablesString
 	}
 
 	return runInfoPath, nil
-	// TODO previous behaviour put it in these places, we may consider creating a symlink later
```
frrist, line 363: Isn't this the correct UK English spelling? We have a mix of behaviour vs behavior spelling in the codebase..

pkg/config/configenv/test.go
```diff
@@ -120,7 +120,7 @@ var TestingRequesterConfig = types.RequesterConfig{
 		ProbeExec:           "",
 	},
 	JobStore: types.StorageConfig{
-		Type: types.InMemory,
+		Type: types.BoltDB,
```
frrist, line 123: Are we sure we want the testing defaults to go on disk?

frrist (approved):
LGTM

pkg/repo/fs.go
frrist: Until recently I hadn't realized/realised so many words could be spelled/spelt differently. Not sure what the best practice/practise ought to be, perhaps just a good sense of humor/humour while we analyze/analyse this and maneuver/manoeuvre things? (sorry, I stop now...)

pkg/config/configenv/test.go
rossjones: I think so. If we get any evidence that they slow down the tests dramatically, we can always revert back (or better still, improve the performance).  One thing I think we're likely to not see in the tests, but will in long-running request nodes, is a reduction in RAM usage.  As we're no longer holding all of the state in non-aligned structs in memory.

PR #2780, opened 2023-08-31
Use a temporary folder for the devstack repo

rossjones:
Whenever we create a devstack, the result is lots of files stored in the repo for each compute and requester node started that never get cleaned up. This PR changes the devstack to use a temporary directory for its repository.  This means we can auto cleanup the entire repository when the devstack is torn down.

TODO:

It would still be useful to force a specific location for the repo but without the burden of a default setting. e.g. if --repo is specified, we should use it, but we don't want the default value at all.

### DIFF
--- a/cmd/cli/devstack/devstack.go
+++ b/cmd/cli/devstack/devstack.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 	"strconv"
 
-	"github.com/spf13/viper"
 	"k8s.io/kubectl/pkg/util/i18n"
 
 	"github.com/bacalhau-project/bacalhau/cmd/util/flags/configflags"
@@ -144,8 +143,18 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
 
 	cm := util.GetCleanupManager(ctx)
 
-	fsRepo, err := repo.NewFS(viper.GetString("repo"))
+	repoPath, _ := os.MkdirTemp("", "")
+	cm.RegisterCallback(func() error {
+		// We need to clean up the repo when the node shuts down, but we can ONLY
+		// do this because we know it is a temporary directory.
+		return os.RemoveAll(repoPath)
+	})
+
+	fsRepo, err := repo.NewFS(repoPath)
 	if err != nil {
+		return err
+	}
+	if err = fsRepo.Init(); err != nil {
 		return err
 	}
 	if err := fsRepo.Open(); err != nil {
--- a/cmd/cli/devstack/devstack.go
+++ b/cmd/cli/devstack/devstack.go
@@ -143,12 +143,10 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
 
 	cm := util.GetCleanupManager(ctx)
 
+	// We need to clean up the repo when the node shuts down, but we can ONLY
+	// do this because we know it is a temporary directory.
 	repoPath, _ := os.MkdirTemp("", "")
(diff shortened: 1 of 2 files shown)


frrist:
👀 I wonder what map needs a lock...
```bash
=== RUN   TestServeSuite/TestNoTimeoutSetOrApplied/job_in_Completed_has_timeout_30s_after_configuring_<nil>_and_specifying_30s
15:00:36.315 | DBG pkg/ipfs/node.go:230 > Closing IPFS node QmSCPk6wMFJ56MZX8je12EWQBvXHzcCm7xggzb8QVbvyVi
15:00:36.315 | DBG pkg/ipfs/node.go:230 > Closing IPFS node QmSCPk6wMFJ56MZX8je12EWQBvXHzcCm7xggzb8QVbvyVi
    serve_test.go:108: Command to execute: ["serve" "--repo" "/tmp/1693494035759349718" "--peer" "none" "--private-internal-ipfs" "--port" "38517"]
15:00:36.316 | ERR pkg/node/requester.go:285 > failed to shutdown job info pubsub error="context canceled"
15:00:36.316 | DBG pkg/util/error_cleanup.go:19 > Context canceled: node info pub sub error="context canceled"
15:00:36.316 | DBG pkg/routing/node_info_publisher.go:110 > stopped publishing node info
15:00:36.316 | DBG pkg/publicapi/server.go:197 > API server closed for host 0.0.0.0 on . [NodeID:QmVYN3g9]
15:00:36.316 | DBG pkg/ipfs/node.go:346 > node 'QmSCPk6wMFJ56MZX8je12EWQBvXHzcCm7xggzb8QVbvyVi' failed to serve ipfs api error="accept tcp4 0.0.0.0:46881: use of closed network connection"
15:00:36.316 | DBG pkg/requester/housekeeping.go:86 > stopped housekeeping task
15:00:36.321 | DBG pkg/telemetry/tracing.go:26 > OLTP tracing endpoints are not defined. No traces will be exported
15:00:36.321 | DBG pkg/telemetry/metrics.go:22 > OLTP metrics endpoints are not defined. No metrics will be exported
fatal error: concurrent map iteration and map write
```

cmd/cli/devstack/devstack.go
```diff
@@ -144,10 +143,20 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
 
 	cm := util.GetCleanupManager(ctx)
 
-	fsRepo, err := repo.NewFS(viper.GetString("repo"))
+	repoPath, _ := os.MkdirTemp("", "")
+	cm.RegisterCallback(func() error {
+		// We need to clean up the repo when the node shuts down, but we can ONLY
+		// do this because we know it is a temporary directory.
+		return os.RemoveAll(repoPath)
+	})
```
frrist, line 151: Since this command blocks until its done we could just defer this, trying to move away from the cleanup manager

frrist (approved):

frrist (commented):
On second thought, will this complicate the process of rerunning a devstack with the same keys for deterministic environment variables?

cmd/cli/devstack/devstack.go
rossjones: Ooh good call

rossjones:
> On second thought, will this complicate the process of rerunning a devstack with the same keys for deterministic environment variables?

Yeah, I think this is why I'd like to get --repo working but without the default value (of ~/.bacalhau).

frrist:
My comment about determinism is non-blocking. Could we introduce a repo specific flag on the devstack called `--stack-repo` or something?

PR #2784, opened 2023-09-04
Add option to specify location of devstack config

rossjones:
By default, the devstack config is now stored in a temporary folder, which is removed when the devstack is shut down.

There are cases, however, where it is useful for the devstack config to persist beyond the live of the devstack, allowing for inspection of the values in the repo, or access to the databases.

By specifying `--stack-repo PATH` it is now possible for the devstack to create a config repo in a specific location that is not removed when the devstack closes.

### DIFF
--- a/cmd/cli/devstack/devstack.go
+++ b/cmd/cli/devstack/devstack.go
@@ -11,7 +11,7 @@ import (
 	"github.com/bacalhau-project/bacalhau/cmd/util/flags/configflags"
 	"github.com/bacalhau-project/bacalhau/pkg/config"
 	"github.com/bacalhau-project/bacalhau/pkg/node"
-	"github.com/bacalhau-project/bacalhau/pkg/repo"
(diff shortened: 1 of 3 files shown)


cmd/cli/devstack/devstack.go
```diff
@@ -250,3 +256,30 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
 	cmd.Println("\nShutting down devstack")
 	return nil
 }
+
+// getRepo will return a config repo that given a directory, will
+// create, init (if necessary) and open the repository before
+// returning it.
+func getRepo(repoPath string) (*repo.FsRepo, error) {
```
frrist, line 263: Could we reuse this method instead?
https://github.com/bacalhau-project/bacalhau/blob/main/pkg/setup/setup.go#L69

frrist (approved):

PR #2785, opened 2023-09-04
Auto cert for TLS certificates

rossjones:
Using the `--autocert [hostname]` parameter to the CLI (in the serve and devstack commands), a certificate will be obtained automatically from Lets Encrypt. The provided hostname can be a comma-separated list of hostnames, but they should all be publicly resolvable as Lets Encrypt will attempt to connect to the server to verify ownership (using the [ACME HTTP-01](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) challenge). On the very first request this can take a short time whilst the first certificate is issued.

As a result of the Lets Encrypt verification step, it is necessary for the server to be able to handle requests on port 443. This typically requires elevated privileges, and rather than obtain these through a privileged account (such as root), we can instead use setcap to grant the executable the right to bind to ports <1024.

```
sudo setcap CAP_NET_BIND_SERVICE+ep ./bacalhau 
```

A cache of ACME data will be held in the config repository, by default ~/.bacalhau/autocert-cache, and this will be used to manage renewals to avoid rate limits. 

This also sets the groundwork for user-supplied/generated certificates in future PRs.


simonwo (approved):
LGTM!

frrist (commented):
LGTM!

PR #2792, opened 2023-09-06
Support IPFS private swarm keys

### LINKED ISSUE #2549: bacalhau get should allow pulling from non public IPFS
https://bacalhauproject.slack.com/archives/C052BAZJVEU/p1686067941236939?thread_ts=1686059168.126689&cid=C052BAZJVEU

simonwo:
Bacalhau will now respond to `BACALHAU_IPFS_SWARM_KEY` environment variable or the `--ipfs-swarm-key` flag which should be the path of a swarm key file used to connect to a private swarm. When these flags are used, Bacalhau will copy the swarm key into any IPFS repo it uses, including in the client. This will allow clients to download data using `bacalhau get` from private IPFS swarms.

This also makes Bacalhau IPFS nodes only bootstrap from the private swarm and not try to connect to public nodes when operating in private mode.

Resolves #2549.


frrist:
I have disabled auto merge as there are some issues here I want us to address before this goes in automatically.

pkg/config/config.go
```diff
@@ -86,9 +86,13 @@ func Get[T any](key string) (T, error) {
...
 	}
 
+	var val T
 	val, ok := raw.(T)
 	if !ok {
-		return zeroValue[T](), fmt.Errorf("value not of expected type, got: %T", raw)
+		err := ForKey(key, &val)
+		if err != nil {
+			return zeroValue[T](), fmt.Errorf("value not of expected type, got: %T", raw)
+		}
```
frrist, line 95: what is this change doing?

frrist (commented):
I think we should avoid introducing the `CommaSeparatedStringSlice` type in this change. As an example, `--node-type` and `BACALHAU_NODE_TYPE` face a similar issue, as will all other string slice flags. If we are going to make the change I think it should be done in a PR to follow this isolated to only said change. There are many instances we'll need to address. Alternately we could leave it unaddressed and dictate in documentation that env vars are space separated and flags are comma separated.

Aside from this I think it looks good, although with the multiple issues being addressed here it's hard to tell what is related to `CommaSeparatedStringSlice` and what is related to the original issue, which I don't fully comprehend, but trust that you do.

I have a couple questions related to config package changes that I'd like to understand before merging.

frrist:
I am also concerned about how much IPFS-related config we are baking into our own config. Are there any alternatives to using IPFS for downloading results? I am concerned we are adding technical debt that will be hard to undo later.

pkg/ipfs/utils.go
```diff
@@ -60,3 +62,24 @@ func ParsePeersString(peers []string) ([]peer.AddrInfo, error) {
...
+	}
+	defer closer.CloseWithLogOnError(sourcePath, sourceFile)
+
+	destinationFile, err := os.Create(destinationPath)
+	if err != nil {
+		return err
+	}
+	defer closer.CloseWithLogOnError(destinationPath, destinationFile)
+
+	_, err = io.Copy(destinationFile, sourceFile)
```
rossjones, line 79: Whilst copy is pretty robust, if you follow frrst's advice above, it'd probably be worth checking the write-count against the fstat size.

rossjones:
> we should avoid introducing the `CommaSeparatedStringSlice` type in this change

I can't see that change in this PR, was it removed or am I missing something obvious?

cmd/cli/docker/docker_run.go
```diff
@@ -92,13 +93,20 @@ func NewCmd() *cobra.Command {
...
+	}
+
 	dockerRunCmd := &cobra.Command{
 		Use:     "run [flags] IMAGE[:TAG|@DIGEST] [COMMAND] [ARG...]",
 		Short:   "Run a docker job on the network",
 		Long:    runLong,
 		Example: runExample,
 		Args:    cobra.MinimumNArgs(1),
 		PreRun:  util.ApplyPorcelainLogLevel,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
simonwo: `PreRunE` is the preferred cobra way of handling functions that return an error.

pkg/config/config.go
simonwo: If `Get` is used to get a custom struct then it will incorrectly return a `map[string]interface{}`. This change makes it correctly return a struct, so that you can do `Get[types.IpfsConfig](...)`

pkg/ipfs/node.go
```diff
@@ -253,27 +192,36 @@ func (n *Node) Close(ctx context.Context) error {
...
+		destinationPath := filepath.Join(repoPath, "swarm.key")
+		err = CopyFile(ipfsConfig.SwarmKeyPath, destinationPath)
+		if err != nil {
+			return nil, nil, "", fmt.Errorf("failed to copy swarm key: %w", err)
+		} else {
+			log.Ctx(ctx).Debug().Str("Path", destinationPath).Msg("Copied IPFS private swarm key")
+		}
+	}
+
+	repo, err := fsrepo.Open(repoPath)
```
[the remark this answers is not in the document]
simonwo: Indeed! It feels like a half-finished feature.

pkg/config/types/slice.go
```diff
@@ -0,0 +1,21 @@
...
+// this type to correctly select the unmarshaler.
+type CommaSeparatedStringSlice []string
+
+// UnmarshalText implements encoding.TextUnmarshaler.
+func (s *CommaSeparatedStringSlice) UnmarshalText(text []byte) error {
+	*s = strings.Split(string(text), ",")
+	return nil
+}
+
+var _ encoding.TextUnmarshaler = (*CommaSeparatedStringSlice)(nil)
```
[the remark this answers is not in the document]
simonwo: Yes I removed this whilst you were reviewing after discovering the issue. I've gone back to a simpler getter pending either a fix to that bug or a more wholesale approach here.

pkg/model/downloader.go
```diff
@@ -9,14 +9,12 @@ const (
 	DownloadCIDsFolderName   = "raw"
 	DownloadFolderPerm       = 0755
 	DownloadFilePerm         = 0644
-	DefaultIPFSTimeout       = 5 * time.Minute
+	DefaultTimeout           = 5 * time.Minute
```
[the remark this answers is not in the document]
simonwo: I have renamed. There is probably a future PR around moving the `DownloaderSettings` into config which I don't want to bite off right now.

pkg/downloader/ipfs/downloader_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/bacalhau-project/bacalhau/pkg/util/closer"
+	"github.com/stretchr/testify/require"
+)
+
+const testSwarmKey = "/key/swarm/psk/1.0.0/\n/base16/\n463ff859373f8e89dd23e7d5429864c84283d961148dc311d120534780549ec3\n"
+
+func writeSwarmKey(t *testing.T) string {
+	file, err := os.CreateTemp(t.TempDir(), "swarm.key")
+	require.NoError(t, err)
+	defer closer.CloseWithLogOnError(file.Name(), file)
```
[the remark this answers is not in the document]
simonwo: I think I prefer the current code to the suggestion, but thanks!

simonwo:
> > we should avoid introducing the `CommaSeparatedStringSlice` type in this change
> 
> I can't see that change in this PR, was it removed or am I missing something obvious?

Yes, it was already removed whilst Forrest was reviewing.

simonwo:
> I am also concerned about how much IPFS-related config we are baking into our own config. Are there any alternatives to using IPFS for downloading results? I am concerned we are adding technical debt that will be hard to undo later.

Come and contribute to [this discussion](https://github.com/bacalhau-project/bacalhau/discussions/2752) about the future of `bacalhau get`.

simonwo:
> I think we should avoid introducing the CommaSeparatedStringSlice type in this change. As an example, --node-type and BACALHAU_NODE_TYPE face a similar issue, as will all other string slice flags. If we are going to make the change I think it should be done in a PR to follow this isolated to only said change. There are many instances we'll need to address. Alternately we could leave it unaddressed and dictate in documentation that env vars are space separated and flags are comma separated.

Yeah, I only saw the issues with this after creating the PR, so removed it whilst you were reviewing! I didn't want to change to be only space separated because use of this var is quite widespread amongst existing user base, so have put a getter in until we roll out a widespread fix.

cmd/cli/devstack/devstack.go
```diff
@@ -166,7 +167,7 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
 	}
 
 	// make sure we don't run devstack with a custom IPFS path - that must be used only with serve
-	if config.GetServeIPFSPath() != "" {
+	if path, err := config.Get[string](types.NodeIPFSServePath); err == nil && path != "" {
```
frrist, line 170: I think we should return the error if it's not nil as it could indicate some malformed config. Otherwise looks great.

cmd/cli/docker/docker_run.go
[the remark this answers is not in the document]
frrist: I prefer it too, its a lot simpler! We can probably leave this as is, but fwiw a lot of CLI tests override the `util.Fatal` method for asserting CLI test failures: https://github.com/bacalhau-project/bacalhau/blob/main/cmd/cli/wasm/wasm_run_test.go#L22 and this change could interfere with that.

cmd/cli/serve/serve_test.go
```diff
@@ -208,13 +209,32 @@ func (s *ServeSuite) TestDefaultServeOptionsConnectToLocalIpfs() {
 	})
 	s.Require().NoError(err)
 
-	swarmAddresses, err := client.SwarmAddresses(s.ctx)
-	s.NoError(err)
-	// TODO(forrest): [correctness] this test is bound to flake if an IPFS local node
```
frrist, line 213: <3

cmd/util/flags/configflags/ipfs.go
```diff
@@ -4,12 +4,19 @@ import "github.com/bacalhau-project/bacalhau/pkg/config/types"
...
 		DefaultValue:         Default.Node.IPFS.SwarmAddresses,
 		Description:          "IPFS multiaddress to connect the in-process IPFS node to - cannot be used with --ipfs-connect.",
 		EnvironmentVariables: []string{"BACALHAU_IPFS_SWARM_ADDRESSES"},
 	},
+	{
+		FlagName:             "ipfs-swarm-key",
+		ConfigPath:           types.NodeIPFSSwarmKeyPath,
+		DefaultValue:         Default.Node.IPFS.SwarmKeyPath,
+		Description:          "Optional IPFS swarm key required to connect to a private IPFS swarm",
+		EnvironmentVariables: []string{"BACALHAU_IPFS_SWARM_KEY"},
```
frrist, line 18: This value is optional, right now this setup results in 2 environment variables for this value.
1. `BACALHAU_NODE_IPFS_SWARMKEYPATH` which is derived from the `ConfigPath`
2. the one defined here.

Since this is a new env var would could drop this line and just use the automatic env var from 1. What do you think? The one you've defined is a lot easier on the eyes.

pkg/config/config.go
frrist: 🔥 love it.

pkg/downloader/ipfs/downloader_test.go
```diff
@@ -0,0 +1,140 @@
+//go:build integration || !unit
```
frrist, line 1: ❤️

frrist (approved):

cmd/cli/docker/docker_run.go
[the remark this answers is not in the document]
simonwo: I've updated these to call util.Fatal

cmd/util/flags/configflags/ipfs.go
simonwo: Yeah I do think the manually defined one is more "natural". I think I'm in favour of just having both. An automatic one is still useful for situations where we need to programmatically define env vars.

PR #2796, opened 2023-09-08
Remove using cancel when restarting long-running jobs

rossjones:
When trying to restart a long-running job, we previously attempted to Cancel it, and then call Run.

Once async executors are implemented, we will be able to call just Run() and it will do the correct thing if the underlying process is still running from before the compute node shutdown.

### DIFF
--- a/pkg/compute/startup.go
+++ b/pkg/compute/startup.go
@@ -91,15 +91,9 @@ func (s *Startup) failExecution(ctx context.Context, execution store.LocalExecut
 }
 
 func (s *Startup) runExecution(ctx context.Context, execution store.LocalExecutionState) error {
-	// We want to ensure this execution is running.  If we just call .Run, there's a
-	// chance we'll end up with two copies running if the previous version is alive
-	// for whatever reason.  We'll call Cancel here so that we can be sure that it
-	// is not running before we ask for it to be run.
-	// TODO: Find a better way of either:
-	// * Findout out whether the underlying process (e.g. docker) is still running, or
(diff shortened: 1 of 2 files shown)


pkg/compute/startup.go
```diff
@@ -91,15 +91,9 @@ func (s *Startup) failExecution(ctx context.Context, execution store.LocalExecut
...
-	// TODO: Find a better way of either:
-	// * Findout out whether the underlying process (e.g. docker) is still running, or
-	// * Have .Run be idempotent for the execution id without relying on the store.
-	_ = s.execBuffer.Cancel(ctx, execution)
-
+	// We want to ensure this 'live' execution is running and rather than go through
+	// multiple steps trying to determine whether the executor or underlying process
+	// is still running, we will just call Run() and expect it to do the correct thing.
 	log.Ctx(ctx).Info().Msgf("Re-running execution %s after restart", execution.Execution.ID)
 	return s.execBuffer.Run(ctx, execution)
```
frrist, line 98: This change is relying on the new error assertion here, yeah?
https://github.com/bacalhau-project/bacalhau/blob/f5cfbf0b9f1b2205205e21ee0f7a2f2eca4e76fe/pkg/compute/executor.go#L285-L305

rossjones: Yup. No longer trying to cleanup and start a fresh run, now just calling Run() and expecting it to either success or fail because it's running.  If it fails for something else, there's little we can really do.

frrist (approved):

PR #2800, opened 2023-09-11
Adds manual TLS certificate support to requester

rossjones:
Adds manual (as opposed to automatic ACME) TLS certificate support to the requester node.  

By default, if none of the following options are used, the server will continue to serve its API endpoints over HTTP. We should default to HTTPS with a special opt-out, but we want to support some level of backwards compability. For the next major version, this should setting should be reversed and HSTS implemented.

In cases where ACME/autocert is not suitable, or not desirable, it is possible to obtain certificates using the traditional route - create and submit a CSR (Certificate Signing Request) to a Certificate Authority. Or, if so desired a self-signed certificate can be created.

The server certificate and key are required, and the filenames passed to the `--tlscert` and `--tlskey` parameters respectively.


scripts/make-certs.sh
```diff
@@ -0,0 +1,56 @@
...
+distinguished_name = req_distinguished_name
+
+[req_distinguished_name]
+CN = localhost
+
+[v3_ext]
+keyUsage = critical,digitalSignature,keyEncipherment
+extendedKeyUsage = critical,serverAuth,clientAuth
+subjectAltName = DNS:localhost
+subjectAltName = IP:0.0.0.0
```
simonwo, line 28: What does 0.0.0.0 mean in this context? For servers it just specifies listen-on-all-interfaces, right?

cmd/cli/serve/serve.go
```diff
@@ -236,6 +236,10 @@ func serve(cmd *cobra.Command) error {
 		// with the other data in the requesterConfig.
 		nodeConfig.RequesterAutoCert = config.ServerAutoCertDomain()
 		nodeConfig.RequesterAutoCertCache = config.GetAutoCertCachePath()
+
+		cert, key := config.GetRequesterCertificateSettings()
+		nodeConfig.RequesterTLSCertificateFile = cert
+		nodeConfig.RequesterTLSKeyFile = key
```
simonwo, line 242: Is this actually specific to just the requester node? The compute node also has APIs, right?

If yes, is there any reason to actually limit it to the requester node? Can't we just use it for the whole node, regardless of type?

simonwo (approved):

scripts/make-certs.sh
rossjones: Yeah, it still means 'this host' and the general interpretation is list on all interfaces.  It's because I found I was binding/requesting both to 127.0.0.1 and 0.0.0.0 arbitrarily, and so we cover both cases.  In addition we use localhost in case that's defined in someone's host file.

cmd/cli/serve/serve.go
rossjones: We absolutely should where we use HTTP, but comms between the requester/compute nodes is currently happening over libp2p which secures itself.  And comms between the executors and the compute nodes are across domain sockets. 

tldr; Yes, but in stages and only when we know what we're TLSing

simonwo: I'm talking about the compute node HTTP API endpoints, yes. E.g. https://github.com/bacalhau-project/bacalhau/blob/715951da29483a823d8c7fe7c88da1c831e9812a/pkg/publicapi/server.go#L58-L63

These work for compute nodes as well, right?

rossjones: > These work for compute nodes as well, right?

Currently, yes. And they should be served over TLS, but I don't know just yet whether:

* they should be using requester-node generated certs that as part of using mTLS, or 
* using self-signed certs with really broad/vague caveats/identities and no validity checks, or 
* publicly generated certs (via ACME or CSR) because they have public names

So yes, will add them, but later when we know how they're going to communicate.

PR #2805, opened 2023-09-12
fix failing to start when HOME is ureachable

wdbaruni:
`bacalhau serve` and other CLI commands uses `$HOME` as the default repo directory. Instead of finding `$HOME` IF the `--repo` flag or `BACALHAU_PATH` env is missing, it tries to find `$HOME` preemptively and set it as a default value to the cobra flag, and panics if it doesn't find it.

This means all commands will fail if `$HOME` is unreachable even if the user tries to set the repo manually. This PR fixes that


wdbaruni:
merging without approval to prepare the pre-release for game day. can fix in main if there are issues @frrist

Issue #1768, opened 2023-01-25
Use `ROOT_DIR` as `BACALHAU_DIR` if it is passed


simonwo:
Station will pass us a location to use for persistent data storage:

> A module's executable gets passed the following environment variables:
>
>    FIL_WALLET_ADDRESS The user's Filecoin wallet address
>    ROOT_DIR The long-lived working directory on disk. The module must store all of its files inside (subdirectories of) this directory. The directory isn't expected to be backed up or shared across machines in any way.

So we need to use the `ROOT_DIR` in place of the `BACALHAU_DIR` in our configuration, when we are in "Station mode".

PR #2810, opened 2023-09-13
ensure the repo config value is always set to a usable value

frrist:
Modifies the behavior of the `defaultRepo` method which determines an appropriate directory for storing the repository.

Priority order:
1. If the environment variable BACALHAU_DIR is set, use it.
2. If the environment variable FIL_WALLET_ADDRESS is set, it tries to use ROOT_DIR.
3. The user's home directory.
4. User-specific configuration directory.

Each step logs a debug message on failure and proceeds to the next. The function returns the chosen directory path or an error if no suitable directory is found.

### DIFF
--- a/cmd/cli/root.go
+++ b/cmd/cli/root.go
@@ -49,7 +49,12 @@ func NewRootCmd() *cobra.Command {
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			repoDir, err := config.Get[string]("repo")
 			if err != nil {
-				panic(err)
+				util.Fatal(cmd, fmt.Errorf("failed to read --repo value: %w", err), 1)
+			}
+			if repoDir == "" {
+				// this error indicates `defaultRepo` was unable to find a default location and the user
+				// didn't provide on.
+				util.Fatal(cmd, fmt.Errorf("bacalhau repo not set, please use BACALHAU_DIR or --repo"), 1)
 			}
 			if _, err := setup.SetupBacalhauRepo(repoDir); err != nil {
 				util.Fatal(cmd, fmt.Errorf("failed to initialize bacalhau repo at '%s': %w", repoDir, err), 1)
@@ -84,10 +89,12 @@ func NewRootCmd() *cobra.Command {
 			ctx.Value(util.SystemManagerKey).(*system.CleanupManager).Cleanup(ctx)
 		},
 	}
-	// ensure the `repo` key always gets a usable default value.
+	// ensure the `repo` key always gets a usable default value, warn if it's not.
 	defaultRepo, err := defaultRepo()
 	if err != nil {
-		util.Fatal(RootCmd, err, 1)
+		RootCmd.Printf("WARNING: %s\n"+
+			"cannot determine default repo location: "+
+			"BACALHAU_DIR or --repo must be set to initialize a node.\n\n", err)
 	}
 	RootCmd.PersistentFlags().String("repo", defaultRepo, "path to bacalhau repo")
 	if err := viper.BindPFlag("repo", RootCmd.PersistentFlags().Lookup("repo")); err != nil {
--- a/cmd/cli/root.go
+++ b/cmd/cli/root.go
@@ -7,6 +7,12 @@ import (
 	"strings"
 
 	"github.com/bacalhau-project/bacalhau/cmd/cli/agent"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/job"
+	"github.com/bacalhau-project/bacalhau/cmd/cli/node"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/bacalhau-project/bacalhau/cmd/cli/cancel"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/create"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/describe"
@@ -14,23 +20,19 @@ import (
 	"github.com/bacalhau-project/bacalhau/cmd/cli/docker"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/get"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/id"
-	"github.com/bacalhau-project/bacalhau/cmd/cli/job"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/list"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/logs"
-	"github.com/bacalhau-project/bacalhau/cmd/cli/node"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/serve"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/validate"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/version"
 	"github.com/bacalhau-project/bacalhau/cmd/cli/wasm"
 	"github.com/bacalhau-project/bacalhau/cmd/util"
 	"github.com/bacalhau-project/bacalhau/cmd/util/flags/configflags"
+	"github.com/bacalhau-project/bacalhau/pkg/config"
 	"github.com/bacalhau-project/bacalhau/pkg/logger"
 	"github.com/bacalhau-project/bacalhau/pkg/setup"
 	"github.com/bacalhau-project/bacalhau/pkg/system"
 	"github.com/bacalhau-project/bacalhau/pkg/telemetry"
-	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
-	"go.opentelemetry.io/otel/trace"
 )
 
 //nolint:funlen
@@ -44,7 +46,7 @@ func NewRootCmd() *cobra.Command {
 		Short: "Compute over data",
 		Long:  `Compute over data`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			repoDir, err := setup.GetBacalhauRepoPath()
+			repoDir, err := config.Get[string]("repo")
 			if err != nil {
 				panic(err)
 			}
@@ -81,8 +83,11 @@ func NewRootCmd() *cobra.Command {
 			ctx.Value(util.SystemManagerKey).(*system.CleanupManager).Cleanup(ctx)
 		},
 	}
-
-	RootCmd.PersistentFlags().String("repo", "", "path to bacalhau repo")
+	defaultRepo, err := defaultRepo()
+	if err != nil {
+		panic(err)
+	}
+	RootCmd.PersistentFlags().String("repo", defaultRepo, "path to bacalhau repo")
 	if err := viper.BindPFlag("repo", RootCmd.PersistentFlags().Lookup("repo")); err != nil {
 		util.Fatal(RootCmd, err, 1)
 	}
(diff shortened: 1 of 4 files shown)


cmd/cli/util.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// UserConfigDir provides the root directory for user-specific configuration data.
+	// Unix: $XDG_CONFIG_HOME or $HOME/.config, Darwin: $HOME/Library/Application Support,
+	// Windows: %AppData%, Plan 9: $home/lib.
+	userDir, err := os.UserConfigDir()
+	if err == nil {
+		return filepath.Join(userDir, defaultBacalhauDir), nil
+	} else {
+		log.Warn().Err(err).Msg("Failed to find user config dir. Trying user cache directory next.")
+	}
```
rossjones, line 60: I think this should be used before ~/.bacalhau as it's a configured choice of where to store config files.  And when configured like this I think it should be `path/bacalhau` so, for Linux, `$HOME/.config/bacalhau`.  

I am studiously avoiding the question "is everything in the repo actually config".

cmd/cli/util.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// UserCacheDir provides the root directory for user-specific cached data.
+	// Unix: $XDG_CACHE_HOME or $HOME/.cache, Darwin: $HOME/Library/Caches,
+	// Windows: %LocalAppData%, Plan 9: $home/lib/cache.
+	userCache, err := os.UserCacheDir()
+	if err == nil {
+		return filepath.Join(userCache, defaultBacalhauDir), nil
+	} else {
+		log.Warn().Err(err).Msg("Failed to find user cache dir. Defaulting to temp directory.")
+	}
```
rossjones, line 70: This feels like it might be wrong. `XDG_CACHE_HOME` suggests Cache and it would be painful to have repos disappear from underneath a running system.

cmd/cli/util.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Check if the directory exists.
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+		return "", fmt.Errorf("temp directory (%s) does not exist", tempDir)
+	} else if err != nil {
+		return "", err
+	}
+
+	// else we use the temp dir
+	return tempDir, nil
```
rossjones, line 84: Should we risk leaving stuff lying around in temp to accidentally get used again in future through misconfiguration.  If we are going to choose a temp directory, then should we choose an anonymous temp directory (created for us and therefore unique on each run), or is this intended as a last resort (because will it really ever get past config dir or home dir?

wdbaruni: not comfortable about this as well

cmd/cli/util.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// UserHomeDir gets the user's home directory.
+	// On Unix/macOS: $HOME, Windows: %USERPROFILE%, Plan 9: $home.
+	userHome, err := os.UserHomeDir()
+	if err == nil {
+		return filepath.Join(userHome, defaultBacalhauDir), nil
+	} else {
+		log.Warn().Err(err).Msg("Failed to find user home dir. Trying user config directory next.")
+	}
```
wdbaruni, line 50: I don't like that we are logging a warning when trying to set default value for a repo even when a user is not relying on the default value as is passing a value for `--repo` flag

cmd/cli/util.go
```diff
@@ -0,0 +1,85 @@
...
+	// TempDir returns the default temp directory.
+	// Unix: $TMPDIR or /tmp, Windows: %TMP%, %TEMP%, %USERPROFILE% or Windows directory, Plan 9: /tmp.
+	tempDir := os.TempDir()
+
+	// Check if the directory exists.
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+		return "", fmt.Errorf("temp directory (%s) does not exist", tempDir)
+	} else if err != nil {
+		return "", err
+	}
```
wdbaruni, line 81: this will panic if we can't figure out a default value for `--repo` that users can't resolve even if they pass a value for `--repo` flag

cmd/cli/util.go
frrist: While I agree in principle, this change would may come as a significant surprise for most users. As a user on OSX (darwin) changing the order of these statements will result in my repo being initialized in `/Users/frrist/Library/Application\ Support/.bacalhau/` by default if I do not provide a `--repo` flag.

cmd/cli/util.go
frrist: Agreed. will remove this option from the defaults.

cmd/cli/util.go
frrist: Was using temp as a last resort, but I agree the tradeoff probably isn't worth it. Will remove.

frrist:
> Are we sure about this priority order?

The `defaultRepo` methods sole purpose is to determine where the `.bacalhau` directory will be initialized/read in the event a user doesn't supply the `--repo` flag. Its returned value is then used to set the default for the `--repo` flag. For example:
```shell
$ env BACALHAU_DIR=/some/path/to/repo bacalhau --help
...
      --repo string             path to bacalhau repo (default "/some/path/to/repo")
```
```shell
$ env FIL_WALLET_ADDRESS=/tmp ROOT_DIR=/station bacalhau --help
...
      --repo string             path to bacalhau repo (default "/station")
```
```shell
$ env HOME=/not/my/home bacalhau --help
...
      --repo string             path to bacalhau repo (default "/not/my/home/.bacalhau")

```
```shell
$ env XDG_CONFIG_HOME=/some/config/path bacalhau --help
...
     --repo string             path to bacalhau repo (default "/some/config/path")
```
```shell
$ bacalhau --help
... 
      --repo string             path to bacalhau repo (default "/Users/frrist/.bacalhau")
```
If a user provides a repo flag, the default value shown in the help text will be overridden by the value provided by the user. I will make this more clear in the code comments of the `defaultRepo` method.

- One problem we are facing is not all environments provide a value for `$HOME`, for example our production environment. When users in these environments try and initialize bacalhau without providing an explicit value to `BACALHAU_DIR` or `--repo` bacalhau doesn't know where the repo should go, and fails.
- The second issue we have is that `defaultRepo` is evaluated before any value provided to `--repo` is inspected. Resulting in the below behavior on `v1.0.4`:
```shell
env HOME= ./bin/darwin_arm64/bacalhau --help
panic: failed to get user home dir: $HOME is not defined

goroutine 1 [running]:
github.com/bacalhau-project/bacalhau/cmd/cli.NewRootCmd()
        github.com/bacalhau-project/bacalhau/cmd/cli/root.go:86 +0x4fc
github.com/bacalhau-project/bacalhau/cmd/cli.Execute({0x107b00ae0?, 0x140006b0dc0})
        github.com/bacalhau-project/bacalhau/cmd/cli/root.go:143 +0x28
main.main()
        github.com/bacalhau-project/bacalhau/main.go:37 +0x12c
```

I think the behavior we'd prefer is something like:
```bash
$ env HOME= bacalhau --help
cannot determine default repo location. BACALHAU_DIR or --repo must be set
```
```shell
$ env HOME= bacalhau serve
cannot determine default repo location. BACALHAU_DIR or --repo must be set
bacalhau repo not set, please use BACALHAU_DIR or --repo
```

I will make this change, along with changes required by feedback above.

frrist:
> Also what do you mean that FIL_WALLET_ADDRESS is trying to use ROOT_DIR? Do we need to use FIL_WALLET_ADDRESS as one of our default repos? Is station still using us? If yes, then they should set BACALHAU_DIR or --repo instead of us picking up FIL_WALLET_ADDRESS

I am not sure on this, the behavior existed prior to the config changes and I left it unmodified to maintain compatibility. I think https://github.com/bacalhau-project/bacalhau/issues/1768 and https://github.com/bacalhau-project/bacalhau/pull/1983 have more context, but I am unsure if station is still something we plan to support. cc @simonwo

cmd/cli/util.go
frrist: I have changed these to `DEBUG`

cmd/cli/root.go
```diff
@@ -44,7 +47,7 @@ func NewRootCmd() *cobra.Command {
 		Short: "Compute over data",
 		Long:  `Compute over data`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			repoDir, err := setup.GetBacalhauRepoPath()
+			repoDir, err := config.Get[string]("repo")
```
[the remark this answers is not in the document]
frrist: Behavior is now:
```
env HOME= ./bin/darwin_arm64/bacalhau serve --repo=./back
WARNING: cannot determine default repo location. BACALHAU_DIR or --repo must be set to initialize a node.
12:44:48.912 | INF pkg/repo/fs.go:187 > Initializing repo at './back' for environment 'production'
...
```

cmd/cli/util.go
frrist: I have modified the root command to display a message if this method returns an error, but to also continue attempting to run the command, ultimately failing if `--repo` wasn't set by a user and its default value could not be determined.

cmd/cli/root.go
```diff
@@ -81,8 +89,14 @@ func NewRootCmd() *cobra.Command {
 			ctx.Value(util.SystemManagerKey).(*system.CleanupManager).Cleanup(ctx)
 		},
 	}
-
-	RootCmd.PersistentFlags().String("repo", "", "path to bacalhau repo")
+	// ensure the `repo` key always gets a usable default value, warn if it's not.
+	defaultRepo, err := defaultRepo()
```
rossjones, line 93: By ensuring there's _always_ a value, does this mean that devstack will no longer be able to default to a /tmp folder?

frrist: The devstack uses a [different repo flag](https://github.com/bacalhau-project/bacalhau/blob/main/cmd/cli/devstack/devstack.go#L144) from the rest of the code with a [default value](https://github.com/bacalhau-project/bacalhau/blob/main/cmd/cli/devstack/devstack.go#L60) of `""`. It then uses whatever this value is, falling back to Temp if one isn't specified, to [create a repo](https://github.com/bacalhau-project/bacalhau/blob/main/cmd/cli/devstack/devstack.go#L157-L169). This behavior will continue to work as expect despite this change.

cmd/cli/root.go
```diff
@@ -81,8 +89,14 @@ func NewRootCmd() *cobra.Command {
...
+	if err != nil {
+		RootCmd.Printf("WARNING: %s\n"+
+			"cannot determine default repo location: "+
+			"BACALHAU_DIR or --repo must be set to initialize a node.\n\n", err)
+	}
```
simonwo, line 98: > I don't like that we are logging a warning when trying to set default value for a repo even when a user is not relying on the default value as is passing a value for --repo flag

I'm with Walid, I don't think we need to emit this if the user is being explicit.

simonwo (approved):

cmd/cli/root.go
frrist: This warning exists for the (rare) case when a user doesn't have `BACALHAU_DIR`, `HOME`, `FIL_WALLET_ADDRESS` + `ROOT_DIR`, or `XDG_CONFIG_HOME` set and doesn't specify a repo path via `--repo`. Without this warning and in the event this case is true users would only see the message:
 `failed to read --repo value: value not found for repo`.

PR #2812, opened 2023-09-13
fix: plurality of ipfs-swarm-addrs in help text

frrist:
Flag name is now `--ipfs-swarm-addrs` instead of `--ipfs-swarm-addr`. We were displaying the old flag name in the help text after the improvements from #2792 landed

### DIFF
--- a/cmd/cli/serve/serve.go
+++ b/cmd/cli/serve/serve.go
@@ -297,7 +297,7 @@ func serve(cmd *cobra.Command) error {
 		cmd.Println()
 		cmd.Println("To connect another node to this private one, run the following command in your shell:")
 		cmd.Printf(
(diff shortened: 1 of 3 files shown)


rossjones (approved):

simonwo (approved):
Would it be better to reference the flag name directly, as we do here? https://github.com/bacalhau-project/bacalhau/blob/ad1c0807884351f3db73319a1d7d55afe2d4f7c2/cmd/cli/devstack/devstack.go#L172-L178

frrist:
> Would it be better to reference the flag name directly, as we do here? 

Yes!! I didn't know we could do that. Will also add some help methods to make this a little cleaner.

PR #2813, opened 2023-09-13
fix: don't panic when constructing a compute config

frrist:
return an error instead of panicking if creating a compute config fails. Use multierror for config validation.
Added a TODO here, we probably don't need to fix this now, but worth looking into in a follow on. https://github.com/bacalhau-project/bacalhau/blob/e168bf54fea6f05cf41041c84d56954fd868ffc0/pkg/test/scenario/suite.go#L93

closes: https://expanso.atlassian.net/browse/GDAY-90

### DIFF
--- a/pkg/test/compute/resourcelimits_test.go
+++ b/pkg/test/compute/resourcelimits_test.go
@@ -9,13 +9,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/bacalhau-project/bacalhau/pkg/model"
-	"github.com/bacalhau-project/bacalhau/pkg/models"
-	"github.com/bacalhau-project/bacalhau/pkg/models/migration/legacy"
 	sync "github.com/bacalhau-project/golang-mutex-tracer"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+	"github.com/bacalhau-project/bacalhau/pkg/models/migration/legacy"
 
 	"github.com/bacalhau-project/bacalhau/pkg/devstack"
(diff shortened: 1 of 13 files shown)


pkg/test/compute/resourcelimits_test.go
```diff
@@ -130,14 +131,16 @@ func (suite *ComputeNodeResourceLimitsSuite) TestTotalResourceLimits() {
 		parsedResources, err := resourcesConfig.ToResources()
 		require.NoError(suite.T(), err)
 
+		computeConfig, err := node.NewComputeConfigWith(node.ComputeConfigParams{
+			TotalResourceLimits:           *parsedResources,
+			IgnorePhysicalResourceLimits:  true,                // in case circleci is running on a small machine
```
rossjones, line 136: From what I can tell this is only ever used when validating the resources in the config, and to cope with low resources on our CI provider.  Should we instead set/test the limits to what we have available on CI? Maybe not as part of this PR, but at some point.

It occurs 22 times in 12 different files and exists to skip validation of config during CI.  Seems like the type of thing we'd want to test in CI?

pkg/test/compute/resourcelimits_test.go
```diff
@@ -291,18 +294,20 @@ func (suite *ComputeNodeResourceLimitsSuite) TestParallelGPU() {
...
 	}
 
+	computeConfig, err := node.NewComputeConfigWith(node.ComputeConfigParams{
+		TotalResourceLimits: models.Resources{
+			CPU:    1,
+			Memory: 1 * 1024 * 1024 * 1024,
+			Disk:   1 * 1024 * 1024 * 1024,
+			GPU:    1,
+		},
+		IgnorePhysicalResourceLimits: true, // we need to pretend that we have GPUs on each node
```
rossjones, line 304: Aha - the real reason.

rossjones (approved):
LGTM

simonwo (approved):

PR #2814, opened 2023-09-14
fix: imporve error handling and logging for GetSystemGPUs

frrist:
Since some hosts may have the `nvidia-smi` installed but in a misconfiguration state (like I did!) we instead we provide a warning showing the args to the command we tried and its response in the event its execution fails.
Reasons for command failure could be: drivers are missing, the smi can't communicate with the drivers, drivers are wrong version, lost cats in the server room, etc. 

I left a TODO for consideration as an alternative to shelling out to the `nvidia-smi` - we could instead us https://github.com/NVIDIA/gpu-monitoring-tools for inspection. Unfortunately, it seems to requires CGO which traditionally been a deal breaker.

I have also added a log line that will notify users if nvidia-smi is not installed:
```
17:06:37.686 | INF pkg/compute/capacity/system/provider.go:86 > cannot inspect system GPUs: nvidia-smi not installed. GPUs will not be used.
```

motivation: https://expanso.atlassian.net/browse/GDAY-90


rossjones:
Will this result in unexpected behaviour.  If I set up a machine to use GPU and it doesn't because of this, then there's a chance I'll find out, not through the log warning, but at runtime when debugging why GPU jobs aren't working.  

I think if we set up a node to use a GPU, and for whatever reason (e.g. configuration) it doesn't work, then we should bail rather than try and continue.

simonwo:
Yes, I think it would be better to halt on boot, but just not as a panic.

frrist:
We discussed this in today's standup. A point that came up was we are now [checking GPU resource availability by default](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/config_compute.go#L144), meaning users with a poorly-configured nvidia-smi will either need to correctly configure the binary before running bacalhau, or remove it from their path to start a bacalhau. (I expect this is a very small (just me) subset of our users)

Given that, should we move forward with this PR, or instead prevent GPU inspection unless a user specifically asks bacalhau to check?

simonwo (approved):
I think let's move forward with this – it's better to just detect what we can and log issues than to require manual configuration.

Issue #2298, opened 2023-04-04
Deprecate support for short JobIDs


frrist:
### Description
To simplify and streamline job tracking, we propose to remove the feature that allows referencing jobs using an 8-character short JobID in the job store logic.

### Background
Discussion in Slack thread: https://filecoinproject.slack.com/archives/C02RLM3JHUY/p1680567849557099

### Proposed Solution
- Deprecate the usage of 8-character short JobID.
- Enforce referencing jobs using their full JobID only.
#### Benefits
- Improved consistency in job management.
- Reduced confusion and enhanced user experience (cli only shows full jobID).

### Where to Begin
- delete this: https://github.com/bacalhau-project/bacalhau/blob/534960f7fcbef5a71d44dc23ee9883b7dcc638f7/pkg/model/utils.go#L22-L29 and remove usage in code base.

rossjones:
I think we should try and understand why it's there in the first place, and what problem it's intended to solve.  
If it's push back against using v4 UUID strings (because they are unwieldy or because of problems with space in the terminal) , then we could look at using somethink like Stripe API IDs (they use then like cus_9s6XKzkNRiz8i3 with a type and then ID).

aronchick:
Honestly, my biggest thing is that it feels like we're optimizing for making things easy for _us_ instead of for end users.

BUT, if we offered things like tab completion, and IDs with types (agree with @rossjones here), etc, i think that makes things easier.

I'm surprised it's that big an issue - why aren't we converting the ID to long the moment we get it from the user and then convert it immediately?

wdbaruni:
> Honestly, my biggest thing is that it feels like we're optimizing for making things easy for us instead of for end users.

It is not a bad thing to build something that is maintainable :). I believe we are all trying to find a balance here between maintainably and user experience, and that is why @frrist is bringing this up.

> I'm surprised it's that big an issue - why aren't we converting the ID to long the moment we get it from the user and then convert it immediately?

Converting a short ID back to a long ID isn't straightforward. We will have to do a lookup in the data store to find all jobs with short ID prefix. This means we will have to design the data store or add indexes that enables constant or logarithmic lookups of short IDs, and handle collisions.

[...]

PR #2816, opened 2023-09-14
fix: Allow boltdb to process short job IDs in requests

rossjones:
Currently boltdb will only work with long IDs.  Now if provided a short ID it scan for a prefix and find the (ideally) single matching Job ID. This mirrors the behaviour of the inmemory store.

### DIFF
--- a/pkg/bacerrors/duplicatejob.go
+++ b/pkg/bacerrors/duplicatejob.go
@@ -1,0 +1,64 @@
+package bacerrors
+
+import (
+	"fmt"
+)
+
+type DuplicateJob GenericError
+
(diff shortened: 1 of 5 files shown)


frrist (approved):

frrist:
This isn't a hill I want to die one, nor is this really the right place to even mention said hill. But can we just deprecate this (shortID's) feature? There's a historical issue regarding it here: https://github.com/bacalhau-project/bacalhau/issues/2298. Based on the associated slack thread, it seems the majority of our team supports its deprecation. Just feels like unnecessary complexity. (search `in:#bacalhau job id short` in Filecoin slack and see how many engineering hours have been wasted with it)

PR #2821, opened 2023-09-14
Fix CLI specification of wasm environment vars

rossjones:
Previously when specifying env vars, using A=B resulted in

```
creating job: wasm env vars invalid: array must have an even number of elements
```

and relying on them being specified on the command line as "A,B" which seems counter-intuitive (and is probably a bug).

This PR changes parsing of the `-e` flag to accept a list of A=B specifiers

### DIFF
--- a/cmd/cli/wasm/wasm_run.go
+++ b/cmd/cli/wasm/wasm_run.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/bacalhau-project/bacalhau/pkg/models/migration/legacy"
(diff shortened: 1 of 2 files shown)


frrist (approved):
At some point in the future it would be nice to unify way we provide env vars to both docker and wasm. Currently wasm doesn't really care if they come in a map or a slice: https://github.com/bacalhau-project/bacalhau/blob/v1.0.4/pkg/executor/wasm/handler.go#L100

rossjones:
Would definitely like to find a better way of handling environment variables, I'm still a fan of moving them into the top level spec as I think they're going to be a common theme for configuring code at runtime regardless of the platform (and regardless of whether they're called env vars or properties or config).

PR #2822, opened 2023-09-14
refactor: allow ipfs-connect and ipfs-swarm-peer interop

frrist:
- permits bacahlau to instruct its remote ipfs node to form connections with other ipfs nodes on the ipfs network.
- allows IPFS downloader to use a remote ipfs node as its client, falling back to a in-process node if no client is requested.

### DIFF
--- a/cmd/cli/serve/util.go
+++ b/cmd/cli/serve/util.go
@@ -3,11 +3,9 @@ package serve
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/host"
-	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 
@@ -157,24 +155,12 @@ func SetupIPFSClient(ctx context.Context, cm *system.CleanupManager, ipfsCfg typ
 	}
 
 	if len(ipfsCfg.SwarmAddresses) != 0 {
-		log.Ctx(ctx).Info().Int("count", len(ipfsCfg.SwarmAddresses)).Msg("attempting to connect to ipfs swarm address peers")
 		maddrs, err := ipfs.ParsePeersString(ipfsCfg.SwarmAddresses)
 		if err != nil {
 			return ipfs.Client{}, err
 		}
-		var wg sync.WaitGroup
-		for _, pi := range maddrs {
-			wg.Add(1)
-			go func(peerInfo peer.AddrInfo) {
-				defer wg.Done()
-				if err := client.API.Swarm().Connect(ctx, peerInfo); err != nil {
-					log.Ctx(ctx).Warn().Err(err).Msgf("failed to connect to peer %s with addresses %s", peerInfo.ID, peerInfo.Addrs)
-				} else {
-					log.Ctx(ctx).Info().Msgf("connected to peer %s with addresses %s", peerInfo.ID, peerInfo.Addrs)
-				}
-			}(pi)
-		}
-		wg.Wait()
+
+		client.SwarmConnect(ctx, maddrs)
 	}
 
 	return client, nil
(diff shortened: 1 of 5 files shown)


cmd/cli/serve/util.go
```diff
@@ -152,6 +150,15 @@ func SetupIPFSClient(ctx context.Context, cm *system.CleanupManager, ipfsCfg typ
...
 
+	if len(ipfsCfg.SwarmAddresses) != 0 {
+		maddrs, err := ipfs.ParsePeersString(ipfsCfg.SwarmAddresses)
+		if err != nil {
+			return ipfs.Client{}, err
+		}
+
+		client.SwarmConnect(ctx, maddrs)
+	}
+
```
frrist, line 161: Instructs the remote ipfs client to connect to swarm peers specified by flag, envvar or config.

pkg/downloader/download.go
```diff
@@ -7,9 +7,10 @@ import (
 	"os"
```
frrist, line 7: formatting changes

pkg/downloader/ipfs/downloader.go
```diff
@@ -29,10 +33,31 @@ func (d *Downloader) IsInstalled(context.Context) (bool, error) {
...
+			}
+			client.SwarmConnect(ctx, maddrs)
+		}
+		return client, nil
+	}
 
+	log.Ctx(ctx).Debug().Msg("creating ipfs node")
 	if d.node == nil {
-		node, err := ipfs.NewNode(ctx, d.cm)
+		node, err := ipfs.NewNodeWithConfig(ctx, d.cm, cfg)
```
frrist, line 60: modifies the ipfs downloader to connect to a remote node if one is provided in the config, falling back to the usual in-process ipfs node otherwise as it did before.

pkg/ipfs/client.go
```diff
@@ -138,6 +141,23 @@ func (cl Client) SwarmAddresses(ctx context.Context) ([]string, error) {
...
+// It spawns a goroutine for each peer connection. In the event of a connection failure,
+// a warning log containing the error and peer details is generated.
+func (cl Client) SwarmConnect(ctx context.Context, peers []peer.AddrInfo) {
+	var wg sync.WaitGroup
+	for _, p := range peers {
+		wg.Add(1)
+		go func(ctx context.Context, p peer.AddrInfo) {
+			defer wg.Done()
+			if err := cl.API.Swarm().Connect(ctx, p); err != nil {
+				log.Ctx(ctx).Warn().Err(err).Stringer("peer", p).Msg("failed to connect to peer")
```
frrist, line 154: We might want to return an error here and let the caller handle it, thoughts?

pkg/ipfs/node.go
```diff
@@ -64,16 +64,7 @@ type Node struct {
 	apiAddresses []string
 }
 
-func NewNode(ctx context.Context, cm *system.CleanupManager) (*Node, error) {
```
frrist, line 67: was only used by the IPFS downloader. The downloader now calls `NewNodeWithConfig` directly.

pkg/ipfs/node.go
```diff
@@ -93,6 +84,7 @@ func NewNodeWithConfig(ctx context.Context, cm *system.CleanupManager, cfg types
 		}
 	}()
 
+	// TODO if cfg.PrivateInternal is true do we actually want to connect to peers?
 	if err = connectToPeers(ctx, api, ipfsNode, cfg.GetSwarmAddresses()); err != nil {
 		log.Ctx(ctx).Error().Msgf("ipfs node failed to connect to peers: %s", err)
 	}
```
frrist, line 90: I left a comment in https://expanso.atlassian.net/browse/ENG-109?focusedCommentId=10143 about this. Feels like we should **not** be establishing connections with other nodes if we want a "private" ipfs node. However, that would make downloading results from compute nodes which are not running in the same process basically impossible. Perhaps we could rename the two ipfs related flags:
- `ipfs-internal`: spins up an in-process node for bacalhau.
- `ipfs-remote/external`: tells bacalhau to use an remote/external ipfs node.

simonwo (approved):

PR #2825, opened 2023-09-15
ENG-116 Fix downloads when in private internal mode.

simonwo:
Private internal mode is used to run an ephemeral IPFS node only for downloading results from jobs. It turns out that the config needs bootstrap nodes to be defined for this to work. This commit keeps the bootstrap nodes in place but applies the other local-only config as per the kubo "test" profile.

Resolves ENG-116.


pkg/ipfs/node.go
```diff
@@ -339,6 +339,9 @@ var localOnlyProfile = config.Profile{
 		c.Swarm.RelayClient.Enabled = config.False
 		c.Swarm.RelayService.Enabled = config.False
 		c.Swarm.Transports.Network.Relay = config.False
+		c.Swarm.DisableNatPortMap = true
+		c.Addresses.Gateway = []string{}
+		c.Discovery.MDNS.Enabled = false
```
rossjones, line 344: mdns discovery looks potentially useful for bootstrapping in private networks in future, will make a note.

rossjones (approved):

PR #2828, opened 2023-09-16
fix: fail CI if go generate results in changes

frrist:
- very similar to our go mod tidy check. We need to ensure we have not modified code that requires codegen to be run after the change. e.g. changes to the config structure, or changes to our mocks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,6 +351,7 @@ jobs:
             if [[ -n $(git status --porcelain) ]]; then
               echo "Error: changes found after running go generate"
               git status --porcelain
+              git diff
               exit 1
             fi
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -334,21 +334,33 @@ jobs:
...
-      - run:
-          name: Cache Precommit
-          command: |
-            cp .pre-commit-config.yaml pre-commit-cache-key.txt
-            poetry run python --version --version >> pre-commit-cache-key.txt
+      #- run:
+          #name: Cache Precommit
+          #command: |
+            #cp .pre-commit-config.yaml pre-commit-cache-key.txt
+            #poetry run python --version --version >> pre-commit-cache-key.txt
```
simonwo, line 341: Why commented out now?

simonwo (approved):

PR #2831, opened 2023-09-18
add service and daemon job schedulers

wdbaruni:
This PR enables scheduling long running jobs. It introduces a dedicated scheduler for daemon jobs, and reuses batch scheduler for service jobs as they share a lot of similarities.

Had to do changes in the BaseExecutor to enable running an already started execution (e.g. during restart) without having conflicts while updating the execution state. However, there are still issues with docker executor engine (an possibly wasm) where it complains that a container with the same name is still running. This is a separate issue the needs fixing before releasing this feature. To test things working properly, I ran a service job, stopped bacalhau, deleted the container and then started bacalhau again and the job restarted.

ENG-16


pkg/orchestrator/scheduler/batch_service_job.go
```diff
@@ -3,45 +3,42 @@ package scheduler
 import (
```
rossjones, line 3: Should probably rename the test `batch_job_test.go` as well

wdbaruni: actually `batch_job_test.go` only tests batch jobs against the shared scheduler, and `service_job_test.go` only tests service jobs against the same scheduler. I tried created a common test suite between both that has the common test scenarios and then implement the specific test scenarios under `batch_job_test` and `service_job_test`, but that didn't work well as the schedulers treat `completed` executions differently based on the job type and most test scenarios were slightly different

rossjones (approved):
LGTM although we're probably near the point now where we need to document the different job types somewhere.

PR #2832, opened 2023-09-18
Simple Version Check and Update Notification

js-ts:


cmd/cli/version/version.go
```diff
@@ -105,12 +108,48 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+		return
+	}
+
+	if currentVersion != serverResponse.Version {
+		fmt.Println("A new version is available!")
+		fmt.Println("Update by running the following command:")
+		fmt.Println("curl -sL https://get.bacalhau.org/install.sh | bash")
+	} else {
+		fmt.Println(serverResponse.Message)
+	}
```
simonwo, line 154: Don't we want to print the server response message if the software is out of date, and nothing else otherwise?

cmd/cli/version/version.go
```diff
@@ -75,10 +78,10 @@ func runVersion(cmd *cobra.Command, oV *VersionOptions) error {
 
 	err := oV.Run(ctx, cmd)
 	if err != nil {
-		return fmt.Errorf("error running version: %w", err)
+		log.Ctx(ctx).Error().Err(err).Msgf("error running version")
 	}
 
-	return err
+	return nil
 }
```
simonwo, line 85: I think we still want the `version` command to print an error if it's usual client-server version check doesn't work. Otherwise, a user running `bacalhau version` will just not see any output and not be able to debug why it didn't work.

cmd/cli/version/version.go
```diff
@@ -105,12 +108,48 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
 	return output.OutputOne(cmd, columns, oV.OutputOpts, versions)
 }
+
+type ServerResponse struct {
+	Version string `json:"version"`
+	Message string `json:"message"`
+}
+
+func checkForUpdates(ctx context.Context, currentVersion string) {
+	resp, err := http.Get("http://35.238.214.16/version")
```
simonwo, line 128: Can we use the domain name here?

cmd/cli/version/version.go
```diff
@@ -105,12 +109,67 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
 		serverVersion, err := util.GetAPIClient(ctx).Version(ctx)
 		if err != nil {
 			log.Ctx(ctx).Error().Err(err).Msgf("could not get server version")
-			return err
+			// Fail silently
+		} else {
+			versions.ServerVersion = serverVersion
+			columns = append(columns, serverVersionColumn)
 		}
+	}
```
simonwo, line 117: You've put this back to what it was before. It needs to return the error if you can't get the requester (server) version, but not the update version.

cmd/cli/version/version.go
```diff
@@ -105,12 +109,67 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("Failed to unmarshal the server response.")
+		return
+	}
+
+	if currentClientVersion != serverResponse.Version {
+		fmt.Println(serverResponse.Message)
+	} else {
+		fmt.Println("Your Bacalhau Version is latest")
+	}
```
simonwo, line 174: You are still checking the version in the client. Do that in the server, and just print the message in the client if the server returns one. You don't need to print anything if the version is up to date.

cmd/cli/version/version.go
```diff
@@ -105,12 +109,67 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("Failed to parse URL.")
+		return
+	}
+
+	q := u.Query()
+	q.Set("clientVersion", currentClientVersion)
+	if currentServerVersion != "" {
+		q.Set("serverVersion", currentServerVersion)
+	}
```
simonwo, line 147: Can we also add the other things in here? E.g. runtime.GOOS, runtime.GOARCH, config.GetUserId(), etc

cmd/cli/version/version.go
```diff
@@ -105,12 +109,67 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+
+	var serverResponse ServerResponse
+	err = json.Unmarshal(body, &serverResponse)
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("Failed to unmarshal the server response.")
+		return
+	}
+
+	if currentClientVersion != serverResponse.Version {
+		fmt.Println(serverResponse.Message)
```
simonwo, line 171: This will break output when using `--output=json` or CSV or Yaml – the contents of standard out will no longer be valid. 

I think you should add the results of the update check to the `Versions` struct above and then it will print out in the correct format always.

cmd/cli/version/version.go
```diff
@@ -104,13 +107,71 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
 	return output.OutputOne(cmd, columns, oV.OutputOpts, versions)
 }
+
+type ServerResponse struct {
+	Version string `json:"version"`
+	Message string `json:"message"`
+}
+
+func checkForUpdates(ctx context.Context, currentClientVersion, currentServerVersion, os, arch, userID string) {
+	u, err := url.Parse("http://update.bacalhau.org/version")
```
frrist, line 138: Where does the code live that backs this endpoint, or is this being served by our public requester node?

We should probably make endpoint part of the config file, see `pkg/config`. 
 - This will make testing easier as we could replace the endpoint with a stubbed local version while testing this command. 
 - It's also not guaranteed that all bacalhau networks run the same version, e.g. staging and production are on different version - v1.0.4 and v1.0.3 respectively.

cmd/cli/version/version.go
```diff
@@ -104,13 +107,71 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+
 	return output.OutputOne(cmd, columns, oV.OutputOpts, versions)
 }
+
+type ServerResponse struct {
+	Version string `json:"version"`
+	Message string `json:"message"`
+}
+
+func checkForUpdates(ctx context.Context, currentClientVersion, currentServerVersion, os, arch, userID string) {
```
frrist, line 137: Let's modify this method to return an error, that way callers can decide how to handle different error cases.

cmd/cli/version/version.go
```diff
@@ -104,13 +107,71 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+	}
+
+	var serverResponse ServerResponse
+	err = json.Unmarshal(body, &serverResponse)
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("Failed to unmarshal the server response.")
+		return
+	}
+
+	fmt.Println(serverResponse.Message)
```
frrist, line 176: Rather than print the response here, we should return the `ServerResponse`. This will allow callers to determine how the update message is displayed to the user. e.g. as json, yaml, columns, text, etc.

cmd/cli/version/version.go
```diff
@@ -104,13 +107,71 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+	versions.UserID = "some-user-id" // Replace this with your method to get UserID
 
-		versions.ServerVersion = serverVersion
-		columns = append(columns, serverVersionColumn)
+	var serverVersion string
+	if versions.ServerVersion != nil {
+		serverVersion = versions.ServerVersion.GitVersion
 	}
 
+	checkForUpdates(ctx, versions.ClientVersion.GitVersion, serverVersion, versions.OperatingSystem, versions.Architecture, versions.UserID)
```
frrist, line 127: Alternatively, this method could just accept a `ctx` and `version` - pulling the required fields from version to meet the needs of the query. This would simplify the functions signature.

cmd/cli/version/version.go
```diff
@@ -104,13 +107,71 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
 	if !oV.ClientOnly {
 		serverVersion, err := util.GetAPIClient(ctx).Version(ctx)
 		if err != nil {
-			log.Ctx(ctx).Error().Err(err).Msgf("could not get server version")
-			return err
+			return fmt.Errorf("error running version: %w", err)
+		} else {
+			versions.ServerVersion = serverVersion
+			columns = append(columns, serverVersionColumn)
 		}
```
frrist, line 114: Can drop the `else` here since the above `if` is returning the error.

frrist (commented):
Good stuff. 

Is it safe to assume this is the beginning of the implementation described here: https://www.notion.so/expanso/Auto-Updates-Check-Metrics-Download-08cf7bcf6bd14011bdadcc23fc48db00?d=d5548392118b41849b28698b35901743#107583b5c46646f1a911ba17ba4b39fe

cmd/cli/version/version.go
```diff
@@ -30,21 +34,24 @@ import (
...
 
-// Versions is a struct for version information
 type Versions struct {
-	ClientVersion *models.BuildVersionInfo `json:"clientVersion,omitempty"`
-	ServerVersion *models.BuildVersionInfo `json:"serverVersion,omitempty"`
+	ClientVersion  *models.BuildVersionInfo `json:"clientVersion,omitempty"`
+	ServerVersion  *models.BuildVersionInfo `json:"serverVersion,omitempty"`
+	OperatingSystem string                   `json:"operatingSystem,omitempty"`
+	Architecture    string                   `json:"architecture,omitempty"`
+	UserID          string                   `json:"userID,omitempty"`
```
simonwo, line 46: What is the point of adding these fields to the struct? They are set but never read from anywhere? Plus, the `BuildVersionInfo` struct already contains OS and ARch fields:
https://github.com/bacalhau-project/bacalhau/blob/f1a8337869a6a947b5685a1297d6a505800ed57b/pkg/models/buildversion.go#L18-L19

How about instead you just add a `LatestVersion` field? And make the server return a `BuildVersionInfo` struct?
```suggestion
	LatestVersion   *models.BuildVersionInfo `json:"latestVersion,omitempty"`
```

cmd/cli/version/version.go
```diff
@@ -91,26 +98,95 @@ var serverVersionColumn = output.TableColumn[Versions]{
 	Value:        func(v Versions) string { return v.ServerVersion.GitVersion },
 }
 
-// Run executes version command
+var updateMessageColumn = output.TableColumn[Versions]{
+	ColumnConfig: table.ColumnConfig{Name: "Update Message"},
+	Value:        func(v Versions) string { return v.UpdateMessage },
+}
```
simonwo, line 104: :+1:

cmd/cli/version/version.go
```diff
@@ -91,26 +98,95 @@ var serverVersionColumn = output.TableColumn[Versions]{
...
+    return output.OutputOne(cmd, columns, oV.OutputOpts, versions)
+}
+
+
+type ServerResponse struct {
+	Version string `json:"version"`
+	Message string `json:"message"`
+}
+
+func checkForUpdates(ctx context.Context, currentClientVersion, currentServerVersion, os, arch, userID string) string {
```
simonwo, line 153: If you just pass in a `model.BuildVersionInfo` instead of `currentClientVersion, os, arch` you will still get all of the fields you need:
```suggestion
func checkForUpdates(ctx context.Context, currentClientVersion, currentServerVersion *model.BuildVersionInfo, userID string) string {
    os := currentClientVersion.GOOS
    arch := currentClientVersion.GOARCH
```

cmd/cli/version/version.go
```diff
@@ -144,29 +144,27 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
 
 type ServerResponse struct {
-	Version *model.BuildVersionInfo `json:"version"`
+	Version *models.BuildVersionInfo `json:"version"`
 	Message string `json:"message"`
 }
 
-func checkForUpdates(ctx context.Context, currentClientVersion, currentServerVersion *model.BuildVersionInfo, userID string) string {
-    os := currentClientVersion.GOOS
-    arch := currentClientVersion.GOARCH
```
simonwo: You've undone the suggestions that you accepted in earlier commits...

cmd/cli/version/version.go
```diff
@@ -104,13 +119,84 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+		if err != nil {
+			log.Ctx(ctx).Warn().Err(err).Msg("Failed to perform update check")
+		} else {
+			versions.UpdateMessage = updateCheck.Message
+			versions.LatestVersion = updateCheck.Version
+			columns = append(columns, latestVersionColumn)
+
+			// Print the update message only if --output flag is not used
+			if oV.OutputOpts.Format == output.TableFormat {
+				fmt.Println(updateCheck.Message)
```
rossjones, line 143: I believe the `cmd` var here, which is a pointer to a `cobra.Command` and so you can use https://pkg.go.dev/github.com/spf13/cobra#Command.Println which will correctly ensure it goes to the correct place.

cmd/cli/version/version.go
rossjones: Also, we need to be sure not to hard-code `http` for the URL once the host is configured.

cmd/cli/version/version.go
```diff
@@ -104,13 +119,84 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+}
+
+func checkForUpdates(
+	ctx context.Context,
+	currentClientVersion, currentServerVersion *models.BuildVersionInfo,
+	clientID string,
+) (*serverResponse, error) {
+	u, err := url.Parse("http://update.bacalhau.org/version")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse URL")
```
rossjones, line 165:
```suggestion
		return nil, errors.Wrap(err, "failed to parse URL for update host")
```

a little bit more context gives us more of a clue of where to look when we see the error.  With more generic error messages they might appear more than once and leave us looking in the wrong place.

cmd/cli/version/version.go
```diff
@@ -104,13 +119,84 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+	}
+
+	q := u.Query()
+	q.Set("clientVersion", currentClientVersion.GitVersion)
+	if currentServerVersion.GitVersion != "" {
+		q.Set("serverVersion", currentServerVersion.GitVersion)
+	}
+	q.Set("operatingSystem", currentClientVersion.GOOS)
+	q.Set("architecture", currentClientVersion.GOARCH)
+	q.Set("userID", clientID)
```
rossjones, line 175: Are we confident that it is okay to send this?  Over http (the default) every host/proxy the request travels through will see the query string.

cmd/cli/version/version.go
```diff
@@ -104,13 +119,84 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+
+	u.RawQuery = q.Encode()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build HTTP request")
```
rossjones, line 181:
```suggestion
		return nil, errors.Wrap(err, "failed to build HTTP request for update check")
```

cmd/cli/version/version.go
```diff
@@ -104,13 +119,84 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read response body")
+	}
+
+	var updateCheck serverResponse
+	err = json.Unmarshal(body, &updateCheck)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal the server response")
```
rossjones, line 198:
```suggestion
		return nil, errors.Wrap(err, "failed to unmarshal the server response when checking for updates")
```

rossjones (approved):
What'd be really useful, is a good description for this PR so that it's obvious what we're reviewing.  Just some context for what the situation is today, and why the change is being introduced (what will change). From the title alone the `and` makes me suspicious that it's two PRs in one and with no description, puts pressure on the reviewer to work out what the PR is actually trying to do as well as review how it is doing it.

I've left a few commands about adding a little more context to error messages, to help us find the cause, and a query about whether it is worth logging in the CLI if we never output it (by default). Don't feel that every comment needs a change, please feel free to discuss in the comments if you disagree.

Clearly won't work because of the absence of update.bacalhau.org, and I think I agree with Frrst that it should probably be in configuration with a default (and correct support for HTTPS). I don't know if this is next on your list...

I think it'd also be good to support initiatives like https://consoledonottrack.com/ so that we don't do these checks, or remove the client id from the query string, for users that don't wish this data to be sent.

cmd/cli/version/version.go
```diff
@@ -104,13 +119,84 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
...
 		columns = append(columns, serverVersionColumn)
+
+		clientID, err := config.GetClientID()
+		if err != nil {
+			return fmt.Errorf("error getting UserID: %w", err)
+		}
+
+		updateCheck, err := checkForUpdates(ctx, versions.ClientVersion, versions.ServerVersion, clientID)
+		if err != nil {
+			log.Ctx(ctx).Warn().Err(err).Msg("Failed to perform update check")
```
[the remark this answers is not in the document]
simonwo: The logs won't show by default, but will show if the users specifies e.g. `LOG_LEVEL=DEBUG` which is a common thing for devs to do. And we can ask users to rerun the command with the variable if they complain that the update check isn't working

cmd/cli/version/version.go
simonwo:
```suggestion
				cmd.Println(updateCheck.Message)
```

simonwo:
@rossjones The server is set up already! 

```
% curl 'http://update.bacalhau.org/version?clientVersion=v1.0.3'
```
```json
{"version":{"major":"1","minor":"0","gitversion":"v1.0.3","gitcommit":"","builddate":"0001-01-01T00:00:00Z","goos":"","goarch":""},"message":""}
```

The code is at https://github.com/bacalhau-project/get.bacalhau.org/tree/main/server.

cmd/cli/version/version.go
simonwo: There is nothing particularly sensitive in the query string. Far more sensitive things will be in our job specs, which are also HTTP by default (still!) 🙃 

But I agree that HTTPS would be preferable. The easiest way maybe seems to be to use your autocert thing... would that be easy to integrate into the server code?

PR #2833, opened 2023-09-18
Check for running containing on docker start

rossjones:
When using the docker executor, the start function checks its map to see if it has any containers currently running for the provided execution. This will avoid dupes.  During a restart however, that map is empty and so the only way to check if there is a running docker container, is to look for it, which this PR attempts to do.

### DIFF
--- a/pkg/executor/docker/executor.go
+++ b/pkg/executor/docker/executor.go
@@ -112,38 +112,51 @@ func (e *Executor) Start(ctx context.Context, request *executor.RunCommandReques
 		Str("jobID", request.JobID).
 		Msg("starting execution")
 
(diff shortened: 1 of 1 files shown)


frrist (approved):
Nice!

Eventually I think we could move to a design where the state for an execution is stored by the compute node on execution creation (perhaps we are already at this point with the execution store?). Said state would be dynamic depending on the type of execution, for the case of docker we'd store the containerID, along with other bits. This pattern would permit the executor plugin to maintain in-memory state of its running executions (as it does now), and allow the compute node to "recover" an executions state into the plugins state in the event an executor-plugin is restarted.

PR #2838, opened 2023-09-18
Fix output volume being added twice

simonwo:
We don't need to always add the outputs directory – if the user specifies a different directory then we just don't add it. It's only a default.


simonwo:
This is premature, clearly the existing behaviour is that we always add outputs from the docker run command.

PR #2843, opened 2023-09-19
set min bacalhau version to v1.0.4

wdbaruni:
There is a changed in `bprotocol` request specs with v1.0.4. Setting a min version will allow the requester to not communicate or schedule jobs on older compute nodes, and avoid running into unnecessary scheduling failures


pkg/node/config_defaults.go
```diff
@@ -40,7 +40,7 @@ var DefaultRequesterConfig = RequesterConfigParams{
 	OverAskForBidsFactor:               3,
 
 	MinBacalhauVersion: models.BuildVersionInfo{
-		Major: "0", Minor: "3", GitVersion: "v0.3.26",
+		Major: "1", Minor: "0", GitVersion: "1.0.4",
```
simonwo, line 43: Does the Git version definitely not need a `v` in front of it??

simonwo (approved):

pkg/node/config_defaults.go
wdbaruni: oh it is needed. Thats what you get when you do a late night change :S
Fixed in https://github.com/bacalhau-project/bacalhau/pull/2843/commits/0d8023d9f9786a12a89d6e036de0e24b2112313f

PR #2848, opened 2023-09-20
Use correct path when updating legacy local dir storage spec

rossjones:
Previously was using target directory for the SpecConfig and so the check whether the folder was in the allowed list never succeeded.

e.g. when running devstack like this 

```
bacalhau devstack --allow-listed-local-paths "/tmp/data"
```

and running a job like this 

```
bacalhau docker run -i src=file:/tmp/data,dst=/inputs/data ubuntu /bin/sh -- -c "ls -al /inputs/data"
```

it would fail as the storage spec would decode the SpecConfig and try to compare `/tmp/data` against `/inputs/data`.

It now does the correct thing.

### DIFF
--- a/pkg/models/migration/legacy/from.go
+++ b/pkg/models/migration/legacy/from.go
@@ -155,7 +155,7 @@ func FromLegacyStorageSpec(legacy model.StorageSpec) (*models.SpecConfig, error)
 		res = &models.SpecConfig{
 			Type: models.StorageSourceLocalDirectory,
 			Params: localdirectory.Source{
-				SourcePath: legacy.Path,
+				SourcePath: legacy.SourcePath,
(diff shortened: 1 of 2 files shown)


simonwo:
Would we be able to add a test for this? 😬

simonwo (approved):

wdbaruni (approved):

wdbaruni:
ahh .. sorry about this. I wanted to add tests the migrate legacy, to new and back to legacy and assert it equals to the original legacy job, but looks like I missed that :(

rossjones:
> How urgent do we think they are needed given that most things seem to be working fine?

I'll add a couple in the morning to cover this case in this PR, and then we can add them as/if they become a necessity.

PR #2854, opened 2023-09-22
Enable TLS support for CLI

rossjones:
Adds support for making TLS requests to the requester node for both v1 and v2 API clients.

By default, all communication with the requester API will remain HTTP as switching to HTTPS would be a breaking change. Making TLS calls can be activated with the command line flag `--tls` or by setting the environment variable `BACALHAU_API_TLS=1`.

If the server is using a self-signed certificate, it will be necessary to either accept insecure requests (no certificate verification) or to provide a CA certificate file.

By specifying --cacert [filename.crt] the client will use a provided CA certificate to verify the certificate when communicating with the requester node. This certificate (often called ca.crt) must be provided by the certificate issuer (which may be the cluster owner in a self-signed scenario). In addition to the command line flag, it is possible to set the `BACALHAU_API_CACERT` environment variable to point to the file.

The least preferable option, but possibly the easiest is to use the `--insecure` flag to make HTTPS requests without verifying the certificate. You may also set `BACALHAU_API_INSECURE`.


rossjones:
@simonwo @frrist is it possible to get a review on this PR in its current state please and once we've agreed it is (or isn't) the correct approach I can fix it up.

cmd/util/api.go
```diff
@@ -10,12 +10,24 @@ import (
 )
 
 func GetAPIClient(ctx context.Context) *client.APIClient {
-	return client.NewAPIClient(config.ClientAPIHost(), config.ClientAPIPort())
+	legacyTLS := client.LegacyTLSSupport(config.ClientTLSConfig())
```
frrist, line 13: This is just type casting, right?

cmd/util/flags/configflags/api.go
```diff
@@ -19,6 +19,28 @@ Ignored if BACALHAU_API_HOST environment variable is set.`,
...
+	{
+		FlagName:     "cacert",
+		DefaultValue: Default.Node.ClientAPI.ClientTLS.CACert,
+		ConfigPath:   types.NodeClientAPIClientTLSCACert,
+		Description: `The location of a CA certificate file when self-signed certificates
+	are used by the server`,
+		EnvironmentVariables: []string{"BACALHAU_API_CACERT"},
+	},
+	{
+		FlagName:             "insecure",
```
frrist, line 38: Naming is hard, maybe this could be called `tls-unverified`? I find the idea of insecure TLS a bit confusing..

pkg/publicapi/client/v2/options.go
```diff
@@ -37,11 +40,47 @@ type Options struct {
 
 	// Headers is a map of headers to add to all requests.
 	Headers http.Header
+
+	// TLSOptions provides info on how we want to use TLS
+	TLS TLSOptions
+}
+
+type TLSOptions struct {
```
frrist, line 48: I think this, and the above `Optons` structure should really be named `Config*` (as you have rightly called the variable of this type in some places), the functions that mutate it are the actual _options_. 

Additionally the `OptionsFn` should probably be made to return errors so we can avoid panicking - I know you are just building on this and didn't introduce the change, but if it's not too much trouble I'd like to see this debt paid down as a part of this work.

Generally is abnormal (imho) usage of [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)...

pkg/publicapi/client/v2/options.go
```diff
@@ -106,11 +145,42 @@ func resolveHTTPClient(config *Options) {
...
+	tr := &http.Transport{}
+
+	if !config.TLS.UseTLS {
+		return tr
+	}
+
+	if config.TLS.CACert != "" {
+		caCert, err := os.ReadFile(config.TLS.CACert)
+		if err != nil {
+			panic("invalid ca certificate provided")
```
frrist, line 159: :disappointed: wish we didn't need to panic here, regardless if we keep the panic or not let's be sure to include the error in the message.

pkg/publicapi/client/v2/options.go
```diff
@@ -37,11 +40,47 @@ type Options struct {
...
+}
+
+type TLSOptions struct {
+	// UseTLS denotes whether to use TLS or not
+	UseTLS bool
+	// Insecure activates TLS but does not verify any certificate
+	Insecure bool
+	// CACert specifies the location of a self-signed CA certificate
+	// file
+	CACert string
```
frrist, line 55: Would it be possible to trim this down to two fields and remove the `UseTLS` field? I think `UseTLS` is used to determine the URI schema, right? Could the schema be inferred from the other two values?

frrist (approved):
I trust this "does the thing" and am happy to move forward and address comments in a follow on. Otherwise it would be nice to pay down some of the debt I mentioned, assuming I have identified the debt correctly.

pkg/publicapi/client/v2/options.go
rossjones: It can't be determined unfortunately because the Insecure bool being false doesn't infer any knowledge of whether to use TLS connections or not.  e.g. Insecure=True could be interpreted as use TLS but don't verify, but Insecure=False would mean use TLS and verify and that would be the default value.

cmd/util/flags/configflags/api.go
rossjones: Yeah, maybe `tls-verify=true` (default) or `tls-verify=false`?

cmd/util/api.go
rossjones: Yes, they have the same structure so structural typing allows us to just convert from one to another, and if I recall correctly this was a scoping thing to avoid import cycles.

pkg/publicapi/client/v2/options.go
rossjones: I've just tried to add the appropriate error returns all the way back up the stack, but it gets more difficult the higher up you go.

rossjones: I ended up removing this by checking the existence of the file denoted by the setting at startup.  It's still there in case the file goes missing in the meantime, but no longer used as an error message mechanism.

PR #2863, opened 2023-09-27
ENG-167 - Flaky test: Allow different target states

rossjones:
The `publish-partial-failure` test, expects the end result of the test to be Fail, Fail, Cancelled. However, sometimes, the test expecting to be cancelled actually completes in time, leaving us with Fail, Fail, Complete. Any attempt to Cancel the completed task will fail as it is a bad state transition, and the object revision will have changed leaving the call to Cancel to fail with a revision error from the job store.

To "resolve" this, we instead allow the result for this to to be one of:

* Fail, Fail, Complete
* Fail, Fail, Cancel

In the absence of a usable Range type, the test introduces an IntMatch type which contains a simple list of numbers allowing us to later provide a single int in the Match method to see if it exists in the list of integers.  We _could_ have achieved the same with []int{} throughout the test cases, but this felt a little cleaner (and more testable).

Also manually tested on Windows.

### DIFF
--- a/pkg/test/requester/retries_test.go
+++ b/pkg/test/requester/retries_test.go
@@ -5,11 +5,14 @@ package requester
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/rs/zerolog"
 	"github.com/stretchr/testify/suite"
+	"golang.org/x/exp/slices"
 	"k8s.io/apimachinery/pkg/selection"
 
 	"github.com/bacalhau-project/bacalhau/pkg/models/migration/legacy"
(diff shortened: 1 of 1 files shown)


rossjones:
> Indulge me... can't we just make that one test case use a job that will never complete? E.g. `sleep 300`?

We can add _another_ test to show that for long-running jobs the behaviour we get is Fail, Fail, Cancel, but I don't think we should hide the current behaviour in that case. It's one that's likely to occur in production and so we should cover it - in the absence of clear semantics on what we expect to happen. 

Annoyingly it's not one we can test with a mock clock, so we'll have to _actually_ sleep for a period of time - probably only 30s realistically.

simonwo (approved):

rossjones:
I've got a test to exercise the alternative suggested, (have one sleep 30s), although it introduces docker (we should replace with simpler sleeping wasm module).  This looks like it might surface a race condition, so will push to a separate branch so that I don't replace one flaky test with another.

PR #2869, opened 2023-09-28
[circleci] Build swagger reference - this is an automatic commit [skip ci]

enricorotundo:


rossjones (approved):
Change in case of some properties in line with recent changes

PR #2881, opened 2023-10-04
Text corrections

criadoperez:
Text corrections on several files


rossjones (approved):
Great, thanks for the PR!

PR #2884, opened 2023-10-09
make  `name` optional in job spec and use jobID by default

wdbaruni:
ENG-169

### DIFF
--- a/pkg/orchestrator/transformer/job.go
+++ b/pkg/orchestrator/transformer/job.go
@@ -47,10 +47,22 @@ func DefaultsApplier(defaults JobDefaults) Job {
 	return JobFn(f)
 }
 
+// RequesterInfo is a transformer that sets the requester ID and public key in the job meta.
 func RequesterInfo(requesterNodeID string, requesterPubKey model.PublicKey) Job {
 	f := func(ctx context.Context, job *models.Job) error {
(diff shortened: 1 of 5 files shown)


pkg/orchestrator/transformer/job.go
```diff
@@ -55,3 +56,14 @@ func RequesterInfo(requesterNodeID string, requesterPubKey model.PublicKey) Job
 	}
 	return JobFn(f)
 }
+
+// NameOptional is a transformer that sets the job name to the job ID if it is empty.
+func NameOptional() Job {
```
rossjones, line 61: The use of `Job` as an interface that `JobFn` implements is a  bit confusing here, because the name Job is so overloaded.  I realise it is unrelated to this PR but refactoring it to something less used would be really useful.

rossjones (approved):

pkg/orchestrator/transformer/job.go
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/2884/commits/48e4456224f689dde451ac9cbe4231e066a35dd7

I initially thought `transformer.JobTransformer` is redundant and that `transformer.Job` is enough, but I don't think that was the right call

PR #2897, opened 2023-10-10
Use non-p2p multiaddress for peer

rossjones:
BLOCKER:  The port used to http discovery (1234) is NOT the same port used by the P2P address (1235) so we need to figure out how to find that.
---- 

Currently we need to use a P2P multiaddress at compute node startup, which requires that is know the ip address, the ip port and the peer id of the requester node.

This PR allows the user to specify a HTTP or HTTPS multiaddress (e.g. /tcp/127.0.0.1/ip/1234/http) that points to the requester node. This address is then used to call the node info endpoint to retrieve the peerid, before using it to construct a p2p multiaddress. The aim of the PR is to make it easier to bootstrap when the ip of the requester node is known but not its peer id.

### Before 

```
bacalhau serve --node-type=compute --private-internal-ipfs --peer=/ip4/192.168.0.93/tcp/1235/p2p/QmaMaJvWQnBsX6WyH5t2GLJkZarbyX6dYJLCvWMpk5SD1L
```

### After 

```
bacalhau serve --node-type=compute --private-internal-ipfs --peer=/ip4/192.168.0.93/tcp/1234/http
```


pkg/libp2p/utils.go
```diff
@@ -75,3 +81,77 @@ func GeneratePrivateKey(numBits int) (crypto.PrivKey, error) {
...
+		// Strip the http(s) from the end of the address
+		if addr, err := multiaddr.NewMultiaddr(fmt.Sprintf("/%s", scheme)); err != nil {
+			return address, err
+		} else {
+			address = address.Decapsulate(addr)
+		}
+
+		// Swap the port to a hard-coded value until we have a better mechanism
+		// for specifying the P2P listen port
+		parts = strings.Split(address.String(), "/")
```
wdbaruni, line 139: Multiaddr has Split and Join functions that might be more helpful and safer than dealing with their string values

rossjones: Fixed in 9deb50ec

PR #2900, opened 2023-10-11
fix migration of new jobs submitted with local directory

wdbaruni:
localDirectory storage source allows an optional parameters ReadWrite, but handling and migrating a job spec where ReadWrite is not passed fails with nil pointer exception. This PR fixes that

### DIFF
--- a/pkg/models/migration/legacy/to.go
+++ b/pkg/models/migration/legacy/to.go
@@ -160,11 +160,14 @@ func ToLegacyStorageSpec(storage *models.SpecConfig) (model.StorageSpec, error) 
 			URL:           storage.Params["URL"].(string),
 		}, nil
 	case models.StorageSourceLocalDirectory:
-		return model.StorageSpec{
+		storageSpec := model.StorageSpec{
 			StorageSource: model.StorageSourceLocalDirectory,
 			Path:          storage.Params["SourcePath"].(string),
-			ReadWrite:     storage.Params["ReadWrite"].(bool),
-		}, nil
+		}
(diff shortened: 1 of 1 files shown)


pkg/models/migration/legacy/to.go
```diff
@@ -160,11 +160,14 @@ func ToLegacyStorageSpec(storage *models.SpecConfig) (model.StorageSpec, error)
...
 	case models.StorageSourceLocalDirectory:
-		return model.StorageSpec{
+		storageSpec := model.StorageSpec{
 			StorageSource: model.StorageSourceLocalDirectory,
 			Path:          storage.Params["SourcePath"].(string),
-			ReadWrite:     storage.Params["ReadWrite"].(bool),
-		}, nil
+		}
+		if storage.Params["ReadWrite"] != nil {
+			storageSpec.ReadWrite = storage.Params["ReadWrite"].(bool)
```
rossjones, line 168: It might be safer here to use the two value type assertion before the cast, to make sure the underlying interface is a bool. At the moment if the interface{} in storage.Params["ReadWrite"] is somehow a string, it'll panic.

```
if b, ok := storage.Params["ReadWrite"].(bool); ok {
    storageSpec.ReadWrite = b 
}
```

rossjones (approved):

pkg/models/migration/legacy/to.go
wdbaruni: fixed in https://github.com/bacalhau-project/bacalhau/pull/2900/commits/f4d2598eca6de26e4bdbafecddd8aaf9e31c734d

PR #2909, opened 2023-10-13
Run update check when any client command is executed

### LINKED ISSUE #2893: Automatically perform update checks on every command
In v1.1.1 we added an update check feature to `bacalhau version` command. This is useful but users don't run that command very often.

Now we want to run the check for every command. As ever, the update check should:

- print out a message at the end of the command if an update is required
- fail silently, no error message should be printed if the update check fails
- terminate silently if the command completes before the update check is finished

simonwo:
This commit automatically kicks off a background update check whenever any client command is executed and prints the result at the end of the command. This means that update notifications will be reported as part of normal client usage rather than just when requested using the version command.

Resolves #2893.


js-ts:
image
Is there a need to print the update command if the clientVersion != LatestVersion ?

simonwo:
> image Is there a need to print the update Text for every command if the clientVersion != LatestVersion ?

Users don't run `bacalhau version` very much and so are unlikely to see an update. This means that they will see the update as part of their normal workflow.

In the next PR, we will introduce a mechanism to only do this once per day (or whatever period) to print the message less often.

cmd/cli/job/list.go
```diff
@@ -66,7 +66,6 @@ func NewListCmd() *cobra.Command {
 		Short:   listShort,
 		Long:    listLong,
 		Example: listExample,
-		PreRun:  util.ApplyPorcelainLogLevel,
```
rossjones, line 69: No PreRunE and PostRunE for list?

pkg/repo/fs.go
```diff
@@ -213,6 +217,9 @@ func (fsr *FsRepo) Init() error {
 	if err != nil {
 		return err
 	}
+
+	// TODO this should be a part of the config.
+	telemetry.SetupFromEnvs()
```
rossjones, line 222: Does this need to be in Open _and_ Init?

pkg/version/update.go
```diff
@@ -0,0 +1,65 @@
...
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+	"github.com/pkg/errors"
+)
+
+const endpoint = "http://update.bacalhau.org/version"
```
rossjones, line 14: Can we have a comment/todo to move this to configuration somewhere, otherwise we're at risk of doing a release just to satisfy a move to https or a diff domain?

rossjones (approved):
Looks good, but when I tried this manually with devstack it tells me there's an update for every command.  And if I were being picky, the text output at the end of the main command's output is too squashed up, could do with some vertical whitespace.  Have approved but would be nice to fix the devstack issue

pkg/version/update.go
simonwo: The reasons for this are:

* We do actually want everyone to always use this domain, and it shouldn't be configurable
* It's actually easier to migrate a hard-coded constant than a configuration value right now because we don't have any config migraiton capability

If the concern is about testing, there's a few ways we can do that even with a hard-coded value (e.g. a mock http client).

pkg/repo/fs.go
simonwo: I think our conclusion was maybe? And we're not confident enough to consolidate this.

simonwo:
Discussed with Ross and comments are above. Broadly we're happy to merge this as long as we get the "once a day" PR out next. So I'll pick that up ASAP.

PR #2912, opened 2023-10-14
[WIP] Skeleton of the python executor

rossjones:


simonwo:
Making each pluggable executor define its own Docker client seems like a lot of duplicate effort, and we will end up with a lot of half-implemented Docker clients of variable quality. Is there a way we can instead allow a pluggable executor to just call an RPC on the host to trigger a new job? That way we can easily implement executors that just run docker commands by just writing a job spec and using our existing reasonably high qualtiy client. (This is worth investing in, because there will be a lot of these)

rossjones:
> allow a pluggable executor to just call an RPC on the host to trigger a new job

There's definitely been some discussion around direct access to a compute node's API, but currently we can only communicate with the compute node via the [bprotocol](https://github.com/bacalhau-project/bacalhau/tree/main/pkg/transport/bprotocol) over libp2p. We also need to work out what the implications are for scheduling where the requester node knows nothing about the detached docker job, and the compute node needs to deal with two different job types (when updating state and performing restarts etc).  

As an alternative, perhaps there's a job type that is just a transformation to an existing one.  e.g. we submit a python job type, and it transparently gets transformed into a docker job somewhere in the requester node? 

All these unanswered questions are why I haven't touched this for a couple of weeks (except to bring up to date) whilst I finish the design doc.

rossjones:
Closing PR but keeping branch around for now.

PR #2924, opened 2023-10-16
Changes the Publisher interface to take an Execution rather than an ExecutionID

rossjones:
PublishResults requires an executionID as a parameter, and so this PR changes it to accept an Execution instead, allowing for better substitutions elewhere (e.g. nodeID in s3 prefix)

### DIFF
--- a/pkg/publisher/fanout/publisher.go
+++ b/pkg/publisher/fanout/publisher.go
@@ -136,7 +136,7 @@ func (f *fanoutPublisher) ValidateJob(ctx context.Context, j models.Job) error {
 // PublishResult implements publisher.Publisher
 func (f *fanoutPublisher) PublishResult(
 	ctx context.Context,
-	executionID string,
+	execution models.Execution,
 	job models.Job,
 	resultPath string,
 ) (models.SpecConfig, error) {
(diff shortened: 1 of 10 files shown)


pkg/publisher/fanout/publisher.go
```diff
@@ -136,15 +136,15 @@ func (f *fanoutPublisher) ValidateJob(ctx context.Context, j models.Job) error {
 // PublishResult implements publisher.Publisher
 func (f *fanoutPublisher) PublishResult(
 	ctx context.Context,
-	executionID string,
+	execution models.Execution,
 	job models.Job,
```
wdbaruni, line 140: we can stop passing the job since the execution also holds job instance. Also I've been passing pointers of large objects instead of a copy. I understand that also has its own set of drawbacks and potentially impacting data integrity, but it is not that harmful when we know that the source (boltdb) is giving us a copy and not the source object

wdbaruni (approved):
LGTM, minor suggestion

pkg/publisher/fanout/publisher.go
rossjones: Resolved in d330f8f0

Issue #2929, opened 2023-10-17
JobSelectionPolicy is un-used for the requester node


frrist:
Both the Compute and Requester Node support a JobSelectionPolicy. At present, the Requester node accepts a policy in its configuration but fails to apply/use it anywhere. The task here is to re-enable support for the JobSelectionPolicy in the Requester Node. This issue was discovered while investigating https://github.com/bacalhau-project/bacalhau/issues/2898 will will follow the fix there.

frrist:
This issue was discovered while investigating [#2898](https://github.com/bacalhau-project/bacalhau/issues/2898) will will follow the fix there.

frrist:
The removal of the JobSelectionPolicy from the requester was made while upgrading to the evaluation broker via: https://github.com/bacalhau-project/bacalhau/pull/2681/files#diff-db482b8c1643fdf50b3648bd443784dfa31ceac24f6c6994f421c42f73d419eaL132-L137 

I believe we'll need to incorporate the previous logic of the JobSelectionPolicy into the Evaluation Brokers `NodeRanker` to preserve existing behavior.

aronchick:
Stale?

simonwo:
I think it's still an issue and we should still look to restore this feature. Stopping jobs that don't meet policy should be done as early as we can, i.e. in the Requester.

wdbaruni:
we can close this as we are moving more towards config files instead of cli flags to configure nodes #4014

PR #2930, opened 2023-10-17
fix: separate flags for compute & requester job selection

### LINKED ISSUE #2898: Job selection probes are misconfigured for Compute and Requester Nodes
> Hi. With Bacalhau 1.1.0, I don't see `--job-selection-probe-http` cli arg working This worked fine in v1.0.3.
> I'm running with `LOG_LEVEL='debug'`. Is there a particular log line that I could look for that shows that probe endpoint being called?
>
> I can see the log shows flag being passed in with correct endpoint:
> `os.Args: [bacalhau serve --node-type requester --ipfs-connect <ipfs-conenct> --private-internal-ipfs=false --labels owner=labdao --job-selection-accept-networked --job-selection-data-locality anywhere --peer none --job-selection-probe-http http://<endpoint>]`

frrist:
- Compute and Requester nodes each have their own JobSelectionPolicy. This change separates the flags for each configured value.
- The Requester JobSelectionPolicy is currently unused and a TODO has been added (see code comment) to address it via: https://github.com/bacalhau-project/bacalhau/issues/2929
- closed #2898

### DIFF
--- a/cmd/util/flags/configflags/job_selection.go
+++ b/cmd/util/flags/configflags/job_selection.go
@@ -3,34 +3,72 @@ package configflags
 import "github.com/bacalhau-project/bacalhau/pkg/config/types"
 
 var JobSelectionFlags = []Definition{
+	//
+	// Requester Job Selection Policy
+	//
+	// TODO: https://github.com/bacalhau-project/bacalhau/issues/2929
+	// The flags are currently settable on the requester, however the requester does not support a JobSelectionPolicy
+	// setting the flags is a noop until the above issue is fixed.
 	{
-		FlagName:     "job-selection-data-locality",
+		FlagName:     "requester-job-selection-data-locality",
 		ConfigPath:   types.NodeRequesterJobSelectionPolicyLocality,
 		DefaultValue: Default.Node.Requester.JobSelectionPolicy.Locality,
 		Description:  `Only accept jobs that reference data we have locally ("local") or anywhere ("anywhere").`,
 	},
 	{
-		FlagName:     "job-selection-reject-stateless",
+		FlagName:     "requester-job-selection-reject-stateless",
 		ConfigPath:   types.NodeRequesterJobSelectionPolicyRejectStatelessJobs,
 		DefaultValue: Default.Node.Requester.JobSelectionPolicy.RejectStatelessJobs,
 		Description:  `Reject jobs that don't specify any data.`,
 	},
 	{
-		FlagName:     "job-selection-accept-networked",
+		FlagName:     "requester-job-selection-accept-networked",
 		ConfigPath:   types.NodeRequesterJobSelectionPolicyAcceptNetworkedJobs,
 		DefaultValue: Default.Node.Requester.JobSelectionPolicy.AcceptNetworkedJobs,
 		Description:  `Accept jobs that require network access.`,
 	},
 	{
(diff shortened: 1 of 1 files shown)


cmd/util/flags/configflags/job_selection.go
```diff
@@ -3,34 +3,72 @@ package configflags
...
 		Description:  `Use the result of a exec an external program to decide if we should take on the job.`,
 	},
+	//
+	// Compute Job Selection Policy
+	{
+		FlagName:     "compute-job-selection-data-locality",
+		ConfigPath:   types.NodeComputeJobSelectionLocality,
+		DefaultValue: Default.Node.Compute.JobSelection.Locality,
+		Description:  `Only accept jobs that reference data we have locally ("local") or anywhere ("anywhere").`,
+	},
```
wdbaruni, line 49: this breaks backward compatibility and will need to update our documentation with the new flags. Also will be confusing for users when running hybrid nodes where defining `compute-job-selection-accept-networked` won't be enough as they will have to set `requester-job-selection-accept-networked` as well

The requirement here is to have a single flag that enables the job selection policy regardless if the node is a requester, compute or hybrid. I believe the quick fix for now as I understood during yesterday's meeting is to have the flag set compute node config instead of the requester, until we come up with a better solution of a single flag that sets multiple configs

simonwo: Agreed, we should be the ones doing the hard work to use the flag in two places, not the user.

wdbaruni:
The issue was blocking my use case as I needed network access in my nodes. This fix seemed to work
https://github.com/bacalhau-project/bacalhau/pull/2936

cmd/util/flags/configflags/job_selection.go
frrist: I'm guessing the worry about backwards compatibility isn't a deal-breaker here, since this isn't about changing the protocol, just the user experience. If compatibility is a showstopper, we might as well close this now. But if not, let's continue below:

Agreed that if we accept this change I'll need to update the documentation. I am happy to do that if this is approved.

> I believe the quick fix for now as I understood during yesterday's meeting is to have the flag set compute node config instead of the requester

I believe that is what's happening here, there is a single flag (set of flags) to configure the compute nodes policy. There is a different set of flags to configure the requesters policy, but that latter flag is ineffectual due to https://github.com/bacalhau-project/bacalhau/issues/2929

Here's why I chose to use two separate flags:
- It felt messy to use one flag to set up either my compute node, requester node, or both based on the type of node I had. 
  - I'm not 100% sure, but there might come a time when we want a hybrid node with different config values for its job selection policies: what if the requester and compute nodes need different settings for `probe-exec` or `probe-http` to decide if they'll take a job? A single flag won’t cut it then.
  - Also, the `data-locality` feels more like it’s meant for the compute node. Maybe we'll phase it out from the requester node at some point?
- At the moment, both compute and requester nodes have their unique config values. Using viper+cobra, there's no simple way to tie one flag to two different configs. Usually, one flag matches one config value.

But if we're set on using just one flag, we might consider creating a shared config that both node types can use. This would have common settings and flags for both nodes. I think there is a reasonable case for making these separate config values, but I'll implement whatever decision you decide on.

frrist:
Closing based on review comments.

PR #2931, opened 2023-10-17
fix: allow compute timeouts to be configured

### LINKED ISSUE #2870: bacalhau serve --max-timeout is broken
The `--max-timeout` flags value is being ignored.

The flag is bound to the following config filed: https://github.com/bacalhau-project/bacalhau/blob/main/cmd/util/flags/configflags/capacity.go#L57
But isn't used when creating the compute config: https://github.com/bacalhau-project/bacalhau/blob/main/cmd/cli/serve/util.go#L32

Source: https://bacalhauproject.slack.com/archives/C052BAZJVEU/p1695933433654849

frrist:
- The flag `max-timeout` has been renamed to `max-job-execution-timeout` to be more specific to its purpose. (docs change here: https://github.com/bacalhau-project/docs.bacalhau.org/pull/283)
- closes #2870


cmd/util/flags/configflags/timeouts.go
```diff
@@ -0,0 +1,36 @@
...
+		ConfigPath:   types.NodeComputeJobTimeoutsMaxJobExecutionTimeout,
+		DefaultValue: Default.Node.Compute.JobTimeouts.MaxJobExecutionTimeout,
+		Description:  `The maximum execution timeout this compute node supports. Jobs with higher timeout requirements will not be bid on.`,
+	},
+	{
+		FlagName:     "default-job-execution-timeout",
+		ConfigPath:   types.NodeComputeJobTimeoutsJobExecutionTimeoutClientIDBypassList,
+		DefaultValue: Default.Node.Compute.JobTimeouts.DefaultJobExecutionTimeout,
+		Description:  `default value for the execution timeout this compute node will assign to jobs with no timeout requirement defined.`,
+	},
```
simonwo, line 35: FYI this value should also be used by the requester node which will apply a default timeout if the timeout is below the minimum timeout. Same with min timeout above. So we will need to find a way to share these flags sooner rather than later.

simonwo (approved):

simonwo:
You will need to inform the LabDAO users that you have changed the name of the flag – please make a note to do this when you make the release.

cmd/util/flags/configflags/timeouts.go
frrist: Follow on for this here: https://github.com/bacalhau-project/bacalhau/issues/2940

PR #2936, opened 2023-10-18
fix job selection policy flag

wdbaruni:


frrist (approved):
LGTM.

But I'd like to finish discussion in https://github.com/bacalhau-project/bacalhau/pull/2930#discussion_r1366052777 regarding an alternative implementation before merging this change.

wdbaruni:
> But I'd like to finish discussion in https://github.com/bacalhau-project/bacaalhau/pull/2930#discussion_r1366052777 regarding an alternative implementation before merging this change.

This is a user impacting issue. We should move fast, have bias for action, and can continue the discussion after the merge if needed as this fix is not introducing a change in behaviour

PR #2938, opened 2023-10-19
Fix a nil pointer error when a job has no ResourcesConfig.

rossjones:
Some combination of mixing new and old APIs has lead to me having some jobs that have no ResourcesConfig attached to the task, and so when converting to a legacy jobspec, a nil pointer causes a panic.

Pre-emptively does the same with timeouts.

cf `get ae17be23` (or `list` if you have an old enough job)

### DIFF
--- a/pkg/models/migration/legacy/to.go
+++ b/pkg/models/migration/legacy/to.go
@@ -87,6 +87,18 @@ func ToLegacyJobSpec(job *models.Job) (*model.Spec, error) {
 		deal.TargetingMode = model.TargetAll
 	}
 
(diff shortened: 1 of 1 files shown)


wdbaruni:
The thing is, `ResourcesConfig` should never be nil if the Task was normalized right after creation. This PR https://github.com/bacalhau-project/bacalhau/pull/2890 should've fixed most of those issues, but it seems there are other areas I am missing. If you have old jobs in your datastore, then that PR should've normalized the job right after deserialization

rossjones:
Looks like that fix didn't make it into the 1.1.1 release, so it looks like it'll be fixed in 1.1.2

PR #2939, opened 2023-10-20
Add update checking to Bacalhau nodes

### LINKED ISSUE #2896: Long-running commands should do an update check regularly
Following #2894, long-running commands like `bacalhau serve` will run an update check on first boot but then won't subsequently run any more checks. This means we will only do an update check once for compute and requester nodes.

Instead, we should run the update check _every_ 24 hours, and for those commands write any results into the log.

simonwo:
Following #2894, short-running commands like `bacalhau list` will run an update check.

If we just extended this to `bacalhau serve`, nodes would run a check on first boot but then won't subsequently run any more checks.

Instead, we now run the update check every 24 hours, and for those commands write any results into the log. Existing behaviour where version updates are written to the output for client commands is retained.

Resolves #2896.

### DIFF
--- a/cmd/util/version.go
+++ b/cmd/util/version.go
@@ -2,10 +2,7 @@ package util
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"os"
-	"time"
 
 	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
@@ -13,7 +10,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/bacalhau-project/bacalhau/pkg/config"
-	"github.com/bacalhau-project/bacalhau/pkg/config/types"
 	"github.com/bacalhau-project/bacalhau/pkg/models"
 	"github.com/bacalhau-project/bacalhau/pkg/publicapi/client"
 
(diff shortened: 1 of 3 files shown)


cmd/util/version.go
```diff
@@ -72,73 +68,11 @@ var printMessage *string = nil
...
-	if err != nil {
-		return errors.Wrap(err, "error writing update state")
-	}
-
-	return nil
+	version.RunUpdateChecker(
+		cmd.Context(),
+		client.NewAPIClient(config.ClientAPIHost(), config.ClientAPIPort()).Version,
+		func(_ context.Context, ucr *version.UpdateCheckResponse) { printMessage = &ucr.Message },
+	)
```
rossjones, line 75: :heart:

pkg/node/node.go
```diff
@@ -278,6 +278,18 @@ func NewNode(
...
 	})
 
+	// Start periodic software update checks.
+	updateCheckCtx, stopUpdateChecks := context.WithCancel(ctx)
+	version.RunUpdateChecker(
+		updateCheckCtx,
+		func(ctx context.Context) (*models.BuildVersionInfo, error) { return nil, nil },
+		version.LogUpdateResponse,
+	)
+	config.CleanupManager.RegisterCallback(func() error {
```
rossjones, line 288: Probably okay for now, but I believe we're slowly moving away from the cleanup manager's. Can be fixed when the one below is migrated though.

pkg/version/update.go
```diff
@@ -63,3 +72,135 @@ func CheckForUpdate(
...
+	}
+
+	lastCheck, err := readLastCheckTime()
+	if err != nil {
+		log.Ctx(ctx).Warn().Err(err).Msg("Error reading update check state – will perform check anyway")
+		lastCheck = time.UnixMilli(0)
+	}
+
+	// Count down the remaining time between the last check and the next check,
+	// and then reset the ticker to start doing regular periodic checks.
```
rossjones, line 140:
```suggestion
	// Count down the remaining time between the last check and the next check,
	// and then reset the ticker to start doing regular periodic checks. This is fine 
	// because the ticker will not have fired before the initial timer.
```

super-minor suggestion to make it a tiny bit clearer

rossjones (approved):
👍🏻

pkg/node/node.go
simonwo: Yeah if we have defined what we're moving towards then I'm happy to do that instead. I haven't seen anything on that though.

Issue #2895, opened 2023-10-10
Include installation ID in update check


simonwo:
Now that we have update checks, we should include the installation ID (if one is configured) into the data provided by the client.

PR #2943, opened 2023-10-22
userID to updates,create/set userID,rename userID

js-ts:
https://github.com/bacalhau-project/bacalhau/issues/2895


simonwo (commented):
You are referring to the same thing as "Installation ID" and "User ID"... can you make sure we consistently refer to it everywhere as "Installation ID" please?

pkg/repo/fs.go
```diff
@@ -142,6 +143,12 @@ func (fsr *FsRepo) Open() error {
 		config.SetLibp2pKey(cfg.User.Libp2pKeyPath)
 	}
 
+	if cfg.User.InstallationID == "" {
+		ID, _ := config.GetClientID()
+		uuidFromUserID := uuid.NewSHA1(uuid.New(), []byte(ID))
+		config.SetUserID(uuidFromUserID.String())
+	}
+
```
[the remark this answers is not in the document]
simonwo: This is only the default that will be used when an installation ID is not explicitly set. The installation ID is meant to be set by the installer, so the common case is that it will be in config already.

pkg/config/setters.go
```diff
@@ -14,6 +14,10 @@ func SetLibp2pKey(path string) {
 	viper.Set(types.UserLibp2pKeyPath, path)
 }
 
+func SetUserID(path string) {
```
frrist, line 17: Can we call this `SetIntallationID`?

pkg/repo/fs.go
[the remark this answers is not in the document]
frrist: Ahh I see, is the code that explicitly sets the installationID going to follow this work? We'll need to pay close attention to how we persist it since setting the value in the config structure will not result in the file being written to disk/config file.

cmd/cli/version/version.go
```diff
@@ -63,7 +65,10 @@ func NewCmd() *cobra.Command {
 
 func runVersion(cmd *cobra.Command, oV *VersionOptions) error {
 	ctx := cmd.Context()
-
+	INSTALLATIONID := os.Getenv("BACALHAU_USER_INSTALLATIONID")
+	if INSTALLATIONID != "" && config.GetInstallationUserID()==""{
+		config.SetIntallationID(INSTALLATIONID)
+	}
```
simonwo, line 71: You shouldn't need to do this explicitly – the config system should already be picking it up for you. Can you try removing this and doing this test: if you run Bacalhau without an established config file, does the installation ID set by this environment variable automatically get persisted to that config file?

You can test this in a fresh config location using e.g.
```bash
export BACALHAU_DIR=$(mktemp -d)
export BACALHAU_USER_INSTALLATIONID="aaaaa"
bacalhau version
cat "$BACALHAU_DIR/config.yaml"
```

frrist: Setting an environment variable will not cause changes to the config file if that's what's being suggested here?

frrist: But if the environment variable was set before starting bacalhau `config.GetInstallationUserID` will return that value, so we don't need to make the `os.Getenv("BACALHAU_USER_INSTALLATIONID")` call here. Its handled by the config

js-ts: > export BACALHAU_DIR=$(mktemp -d)
> export BACALHAU_USER_INSTALLATIONID="aaaaa"
> bacalhau version
> cat "$BACALHAU_DIR/config.yaml"
The main issue with this is it doesn't show up because the binary that's installed in install.sh isn't build from scratch. so there is no way of knowing it except for merging and releasing the code or modifying the code in install.sh to build from scratch

simonwo: Ah, I was under the impression that the config that is written out on first init of a repo will represent the current state of config set by environment variables. But yeah, this isn't the case. So we do need to set the installation ID persistently somehow.

PR #2944, opened 2023-10-23
fix error when update.json is missing

wdbaruni:


frrist:
Would it be worth moving the initialization of the file to repo instantiation? i.e. this method https://github.com/bacalhau-project/bacalhau/blob/main/pkg/repo/fs.go#L193

simonwo:
@frrist Moving it to the repo init makes sense – does that run even for pre-existing repos? Or do we still have to handle the case where it doesn't exist for users with earlier versions?

wdbaruni:
We want to create the file on existing repos as well. The fix will not log an error and the file will be created after

PR #2956, opened 2023-10-25
Increase the default job resource limits

rossjones:
They are currently too low and causing OOM errors for some users when they use more than 100Mb.  There is some leeway but not very much.

The memory has been increase to 1Gb which is high for simple short-lived script, but sufficient for something more complex or long-running. We should fine tune this when we have more data about usage.

Similarly the CPU share was currently 1/10 of a CPU, and this was causing scripts that should be running quickly (<10s) to take nearly a minute.  The default has been increased to .5 of a CPU. We should consider removing the CPU limit entirely for dockerised workloads and let the operating system share the CPU appropriately.

### DIFF
--- a/pkg/config/configenv/production.go
+++ b/pkg/config/configenv/production.go
@@ -90,8 +90,8 @@ var ProductionComputeConfig = types.ComputeConfig{
 			GPU:    "",
 		},
 		DefaultJobResourceLimits: model.ResourceUsageConfig{
-			CPU:    "100m",
-			Memory: "100Mi",
+			CPU:    "500m",
+			Memory: "1Gb",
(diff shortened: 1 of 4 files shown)


pkg/config/configenv/production.go
```diff
@@ -90,8 +90,8 @@ var ProductionComputeConfig = types.ComputeConfig{
 			GPU:    "",
 		},
 		DefaultJobResourceLimits: model.ResourceUsageConfig{
-			CPU:    "100m",
```
frrist, line 93: Can you port these over to staging, development, and local defaults too?

frrist (approved):
LGTM

PR #2958, opened 2023-10-26
fixes: image manifest cache

Nasfame:
I find that cache is being abused, instead of caching it for multiple reads, with every read there is a corresponding write


pkg/executor/docker/bidstrategy/semantic/image_platform.go
```diff
@@ -61,6 +62,24 @@ func (s *ImagePlatformBidStrategy) ShouldBid(
 		if m != nil {
 			manifest = *m
 		}
+		// Cache the platform info for this image tag for a day. We could cache
```
frrist, line 65: With `ManifestCache` set to evict images after 24 hours, we might end up using outdated images tagged as latest or using semantic versioning if they're updated within that timeframe. This is a rare scenario, but still possible. Additionally, with this change and as the code comment indicates, we're changing behaviour here as we will no longer be extending the eviction policy duration for frequently used images. I am not sure this is the behavior we want to wdyt @rossjones?

Also, there's a TODO to replace this cache with either an LRU (we should consider https://github.com/hashicorp/golang-lru) or potentially an ARC, which the mentioned package also supports. Perhaps addressing this TODO would be a more helpful change at this point?

Nasfame: @frrist I am not able to follow you. 

My change was to prevent caching the same manifest retrieved from Cache. 

@frrist Sorry I am not able to follow your thoughts 

No operations seems to be performed on the manifest, so why cache the same manifest again ? after retrieving from DB.  Its only updating the `ttl`

Nasfame: @frrist ??

Issue #2960, opened 2023-10-27
Dylibso is noticing some tracing issues


aronchick:
Zach (cc'd) is making great progress on getting the Observe SDK traces merged with compute node (host) traces so they both show up in the same sink (in this case Jaeger) using the trace id configured by the host, however we've run into a couple of snags and could use your assistance.

Although some of traces from both the host (looks like mostly background tasks) and the wasm module are making into Jaeger, it's missing the context around the wasm execution and we're getting some otel errors from Bacalhau that look like the following:

```
13:33:47.323 | ERR pkg/telemetry/common.go:18 > Error occurred while handling spans error="traces export: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8"
13:34:37.327 | ERR pkg/telemetry/common.go:18 > Error occurred while handling spans error="traces export: rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8"
```

We're having some trouble unwinding the source of the error as it appears that a [global error error handler](https://pkg.go.dev/go.opentelemetry.io/otel#SetErrorHandler) is being used. Also appears we're getting these errors on the main branch with the Observe SDK disabled as well. The Observe SDK is also reusing the Bacalhau Otel client config so we're reasonably confident both are using the same protocol (e.g. grpc/protobuf).

simonwo:
cc @zshipko @wikiwong

I've recreated the setup locally and checked out your fork. Tl;dr: I can't observe any issue, and the tracing seems to be working as intended (see below screenshot).

![image](https://github.com/bacalhau-project/bacalhau/assets/4951176/2e2b2226-4e24-49b6-b838-046d7ed63c98)

The telemetry errors you are seeing appear to be coming from inside the library that provides our IPFS node. When I remove the IPFS node (by running an external node instead) the errors disappear. So I believe you can safely ignore this – all of the traces from Bacalhau appear to be transmitting correctly.

I can see context around the WASM execution in the form of `pkg/compute.ExecutorBuffer.Run` as shown in the screenshot. Do you not get this? Are you expecting more/different context?

I can also see the Observe SDK traces appearing.

zshipko:
Thank you - using this setup it seems to work correctly!

Before I wasn't seeing the `pkg/compute.ExecutorBuffer.Run` events in jaeger, but maybe it is just a configuration issue. I was also using `bacalhau devstack` for my tests before.

Very excited to see this working!

[...]

PR #2971, opened 2023-10-30
Integrate observe-sdk to get tracing information during Wasm execution

zshipko:
This PR integrates [dylibso/observe-sdk](https://github.com/dylibso/observe-sdk) into bacalhau. 

- ~~I am able to see the wasm traces along with some bacalhau traces but it seems like we're hitting an issue with opentelemetry in bacalhau captured here:~~ https://github.com/bacalhau-project/bacalhau/issues/2960
- ~~We're re-using the `OTEL_EXPORTER_OTLP_PROTOCOL` and `OTEL_EXPORTER_OTLP_ENDPOINT` environment variables, is there another way we should handle the configuration?~~
- I've also added `testdata/wasm/observe/hello.instr.wasm` so you can try it out with some instrumented code


pkg/executor/wasm/handler.go
```diff
@@ -77,7 +82,37 @@ func (h *executionHandler) run(ctx context.Context) {
...
+	conf := &opentelemetry.OTelConfig{
+		ServiceName:        "bacalhau",
+		EmitTracesInterval: time.Second * 1,
+		TraceBatchMax:      100,
+		Endpoint:           endpoint,
+		Protocol:           protocol,
+		AllowInsecure:      allowInsecure, // for localhost in dev via http
+	}
+	var adapter *opentelemetry.OTelAdapter
+	if endpoint != "" {
```
simonwo, line 109: Here, if `endpoint` is non-empty but failed to parse as a valid URL, we will still attempt to pass it to the adapter which will presumably have a bad time. We should only instantiate the adapter if the URL parse succeeded?

pkg/executor/wasm/handler.go
```diff
@@ -77,7 +82,37 @@ func (h *executionHandler) run(ctx context.Context) {
 		h.cancel()
 	}()
 
-	tracingEngine := tracedRuntime{h.runtime}
+	configuredProtocol := strings.ToLower(os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL"))
+	endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
```
simonwo, line 86: I would much prefer we do this by re-using the config that Bacalhau is already providing rather than doing config parsing inline in the executor here.

Ideally you could just consume the [`telemetry.getTraceClient`](https://github.com/dylibso/bacalhau/blob/c7aa118c45d685bb7583330b8ba3b51bf12c483f/pkg/telemetry/tracing.go#L58-L75) and pass it to [`OTelAdapter.UseCustomClient`](https://github.com/dylibso/observe-sdk/blob/b4a765d1f1479c7a8b54d0939f571d28e86ae93e/go/adapter/opentelemetry/adapter.go#L41-L45). But @wikiwong prefers to not do this because it will tie your adapter to the same version of the OTel library that Bacalhau uses. 

For my understanding, isn't that a problem anyway? Can two versions of the same go lib sit side-by-side in the same binary? Won't we always need a match with the version of OTel used by dylibso?

So if we can't use the client directly, can we reuse more of the config options directly from our telemetry module? E.g. the [service  name](https://github.com/dylibso/bacalhau/blob/c7aa118c45d685bb7583330b8ba3b51bf12c483f/pkg/telemetry/common.go#L45), the [`OTEL_EXPORTER_OTLP_*`](https://github.com/dylibso/bacalhau/blob/c7aa118c45d685bb7583330b8ba3b51bf12c483f/pkg/telemetry/constants.go#L10) constants etc...?

pkg/executor/wasm/handler.go
zshipko: Good point, it seems like re-using the client would avoid these types of errors

pkg/executor/wasm/handler.go
zshipko: I don't think it would be a big issue to re-use the existing client, the function is already there so it seems worth testing this approach.

pkg/executor/wasm/handler.go
```diff
@@ -83,33 +80,20 @@ func (h *executionHandler) run(ctx context.Context) {
...
 		EmitTracesInterval: time.Second * 1,
-		TraceBatchMax:      100,
-		Endpoint:           endpoint,
-		Protocol:           protocol,
-		AllowInsecure:      allowInsecure, // for localhost in dev via http
+		TraceBatchMax:      10,
+		// the remaining fields are completed from a system-configured client
+		// by using the `UseCustomClient` method on the adapter below
+	}
+	traceClient, err := telemetry.GetTraceClient()
```
nilslice, line 91: @simonwo - we've made this function visible in order to create another instance of the otel client (which is more consistently configured from the same sources). Initially, we tried to set the existing client to a package-level global and re-use a single instance of it, but it seems that there is some context cancellation on a per-trace basis which caused errors between the wasm spans and the root spans.

pkg/executor/wasm/handler.go
```diff
@@ -83,33 +80,20 @@ func (h *executionHandler) run(ctx context.Context) {
...
-	}
-
-	allowInsecure := false
-	if u, err := url.Parse(endpoint); err == nil {
-		host := u.Hostname()
-		endpoint = host + ":" + u.Port()
-		allowInsecure = host == "localhost" || host == "127.0.0.1"
-	}
-
-	conf := &opentelemetry.OTelConfig{
```
nilslice: all of this is now being provided by the `telemetry` package `GetTraceClient` function. there's more detail in [another comment](https://github.com/bacalhau-project/bacalhau/pull/2971/commits/4066ee99b3550d0896e4baaad39cb64d842a0cc2#r1391801813).

nilslice:
Ok, we think this is ready for further review and testing. Thanks!

pkg/executor/wasm/handler.go
nilslice: Although maybe it would be better to put as much of our adapter initialization code in the `telemetry` package and construct adapters there to be used in the `tracedRuntime` struct.

neuronicnobody:
@simonwo PR for documentation updates is here https://github.com/bacalhau-project/docs.bacalhau.org/pull/288

simonwo (approved):
OK! The code looks great to me and I've tested that it is working – I can run instrumented WASM and can see the output in Jaeger, or the WASM also runs fine even if telemetry is not defined.

Please rebase on top of main and then we can merge this!!

simonwo:
Looks like there is a failing WASM test: https://app.circleci.com/pipelines/github/bacalhau-project/bacalhau/7378/workflows/90db56a5-e1f7-4ece-bc6c-00b98718a869/jobs/103619/tests

Also looks like you need to run `go mod tidy` on this branch before committing. Thanks!

wikiwong:
hey @simonwo , thanks for the review!  Zach pushed up some changes, can you please enable the workflow to proceed?

simonwo:
Hm, looks like we are still [seeing errors](https://app.circleci.com/pipelines/github/bacalhau-project/bacalhau/7408/workflows/42b9c974-92c9-457c-9175-c7b659f6a820/jobs/104122/tests) with that test:
```
21:33:16.223 | INF pkg/executor/wasm/handler.go:125 > instantiating wasm modules [entrypoint:_start] [execution:619b3f4b-b7e0-40ee-b9ba-f8cf24b76d6f] [job:test-job]
21:33:16.223 | DBG pkg/executor/wasm/loader.go:44 > Loading WASM module [Path:/var/folders/g1/43z0w1wj4p5dlcfxvdr9zhx80000gn/T/inline-storage1806494036/file1452090962]
21:33:16.229 | DBG pkg/executor/wasm/loader.go:44 > Loading WASM module [Path:/var/folders/g1/43z0w1wj4p5dlcfxvdr9zhx80000gn/T/1700429595020410000/executor_storages/bacalhau-ipfs319504829/QmPympgyrEGEdSJ93rqvQkR71QLuQGdhKQtYztFwxpQsid]
21:33:16.229 | WRN pkg/executor/wasm/handler.go:161 > failed to instantiate entry module [entrypoint:_start] [execution:619b3f4b-b7e0-40ee-b9ba-f8cf24b76d6f] [input_alias:] [input_source:inline] [input_target:] [job:test-job] [volume_source:/var/folders/g1/43z0w1wj4p5dlcfxvdr9zhx80000gn/T/inline-storage1806494036/file1452090962] [volume_target:] [volume_type:Bind]
```

wikiwong:
Thanks, I had a feeling it was still occurring but wanted to see if our latest updates resolved it. Investigating

zshipko:
I just pushed a commit to avoid trying to re-link the observe module but it looks like CI needs to be allowed to run. I can rebase again too.

What's the best way to run this test locally?

zshipko:
@simonwo I fixed an issue in observe-sdk where the module was being re-initialized in some compiled modules. It looks like tests are passing now!

simonwo (approved):
@zshipko Awesome! Please do one final rebase to get on top of main and the PR will merge automatically ⭐

PR #2986, opened 2023-11-03
allow describing nodes with their short IDs

wdbaruni:


pkg/routing/inmemory/inmemory.go
```diff
@@ -77,24 +79,62 @@ func (r *NodeInfoStore) Add(ctx context.Context, nodeInfo models.NodeInfo) error
...
+	var errNotFound routing.ErrNodeNotFound
+	if !errors.As(err, &errNotFound) {
+		return models.NodeInfo{}, err
+	}
+
+	// look for a node with the prefix. if there are multiple nodes with the same prefix, return ErrMultipleNodesFound error
+	var nodeIDsWithPrefix []string
+	for nodeID, infoWrapper := range r.nodeInfoMap {
+		if time.Now().After(infoWrapper.evictAt) {
+			go r.evict(ctx, infoWrapper)
```
rossjones, line 115: This takes variadic params so we could collect the items be evicted and evict them in a single goroutine.

rossjones (approved):
Tiny suggestion, but looks good!

When I tried it manually, if I managed to mess up copying and pasting, I get the following error.

```
❯ ./bin/darwin_arm64/bacalhau node describe Qm
could not get node Qm: Unexpected response code: 500 ({
  "error": "multiple nodes found for nodeID prefix: Qm, matching nodeIDs: [QmfXLhf8zNfackGmUsdxqPf1ccYFby5r9dQgyrBS8Mk7NP QmaNbEfGP3uRCfetTQRmgS9tGrXG5W6Xo4Dof8svR2Xcm1 QmRDsMN56fsDEWm5wr6m8K4qDEPb3A1yvX5MeSWmsXjj7z ...]",
  "message": "Internal Server Error"
})
```

It doesn't really feel like a 500 error, more just an unexpected result

pkg/routing/inmemory/inmemory.go
wdbaruni: Fixed in https://github.com/bacalhau-project/bacalhau/pull/2986/commits/590a38684f703bb33e27ecfeb9e1e739ee198cbe

as for the 500 error, almost all failed calls return 500 in the new APIs. I have a backlog item (still in Jira) to address this and improve error reporting

PR #2993, opened 2023-11-06
fixes: image manifest cache

Nasfame:
Dups: #2958

Edit (@rossjones):

Changes the behaviour of the image manifest cache to only record a manifest in the cache if it is not currently present.  This means that the record will be present for the expiry period from first insertion, not last use.


rossjones:
Why a duplicate PR?  

This PR changes the behaviour of the cache.  Currently items are cached for a period of time (based on label etc) __from the last time they were used__.  Your PR changes this to items being cached for a period of time __from the first time they were used__.  

Is that what we want?

Nasfame:
> Why a duplicate PR?

Yeah well the original PR was raised from remote branch.

> This PR changes the behaviour of the cache.
> 
> Currently items are cached for a period of time (based on label etc) from the __last time__ they were used. Your PR changes this to items being cached for a period of time from the __first time__ they were used.
> 
> Is that what we want?

yeah, well I noticed unnecessary cache writes (for every read, there was a write. It seemed unneccessary to rewrite cached item once more). In the log that was shared over slack: I noticed that "Manifest found in cache", & "Unable to write cache". 

So in my pr change i made sure, cache works as per its standards, the cache is only writthen if the manifest in not found.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/bacalhau-project/bacalhau
 
 go 1.20
 
+//quic-go is tied to a specific go version: look at https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions
```
rossjones, line 5: I don't think we need this comment here, we're aware of the issue

rossjones:
Fair enough, we can always see how it behaves. If you can fix the broken tests I'll review + approve it

go.mod
Nasfame: ok

Nasfame:
> Fair enough, we can always see how it behaves. If you can fix the broken tests I'll review + approve it TestBidsBasedOnImagePlatform is the failing test so you can run it by itself with
> 
> ```
> go test pkg/executor/docker/bidstrategy/semantic/*.go  -v -run TestBidsBasedOnImagePlatform --tags=integration 
> ```

sure, was waiting for your approval

PR #2995, opened 2023-11-08
enable downloading s3 results

wdbaruni:
This PR enables downloading results published to S3 using `bacalhau get <job_id>` command. This works by allowing the Requester node to return a pre-signed URL for the S3 object that is valid for 30 minutes. 

### Run job that publishes to S3
```
→ cat s3-publisher.yaml
Name: Docker Job with S3
Type: batch
Namespace: org1
Count: 1
Tasks:
  - Name: main
    Engine:
      Type: docker
      Params:
        Image: ubuntu:latest
        Entrypoint:
          - /bin/bash
        Parameters:
          - -c
          - echo hello world
    Publisher:
      Type: s3
      Params:
        Bucket: bacalhau-test-datasets
        Key: integration-tests-publisher/walid-manual-test-{jobID}
        Region: eu-west-1
        Compress: true
```
```
bacalhau job run s3-publisher.yaml
```

### Download Results
```
# download results to job-j-cbfb184a directory
bacalhau get j-cbfb184a-82cf-4712-a263-2c3e50b59ffb

# directory structure
→ tree job-j-cbfb184a
job-j-cbfb184a
├── exitCode
├── stderr
└── stdout

0 directories, 3 files

# stdout content
→ cat job-j-cbfb184a/stdout
hello world

```

### Download raw results
```
# delete previous results
rm -rf  job-j-cbfb184a

# download raw results
bacalhau get j-cbfb184a-82cf-4712-a263-2c3e50b59ffb --raw

# directory structure

→ tree job-j-cbfb184a
job-j-cbfb184a
└── raw
    └── bacalhau-test-datasets.s3.eu-west-1.amazonaws.com_integration-tests-publisher_walid-manual-test-j-cbfb184a-82cf-4712-a263-2c3e50b59ffb.tar.gz

1 directories, 1 files
```

## Requirements:
- Requester node must have s3.GetObject permission to the S3 objects to be able to generate pre-signed URLs

## Limitations:
- Only works when publishing compressed results to S3 using `Compress: true` as it is not possible to generate pre-signed URLs to a prefix of objects. I am planning to make uploading a single compressed object the default behaviour in an upcoming PR
- Doesn't support downloading single objects such as `bacalhau get j-cbfb184a-82cf-4712-a263-2c3e50b59ffb/stdout`


pkg/config/types/requester.go
```diff
@@ -19,6 +19,7 @@ type RequesterConfig struct {
 
 	EvaluationBroker EvaluationBrokerConfig `yaml:"EvaluationBroker"`
 	Worker           WorkerConfig           `yaml:"Worker"`
+	S3SignedURL      S3SignedURLConfig      `yaml:"S3SignedURL"`
```
rossjones, line 22: We should _eventually_ move this into StorageConfig before we start adding more configs here.

pkg/downloader/http/downloader.go
```diff
@@ -2,89 +2,126 @@ package http
...
-			item.CID, item.Target,
-		)
-
-		innerCtx, cancel := context.WithDeadline(ctx, time.Now().Add(httpDownloader.Settings.Timeout))
-		defer cancel()
-
-		return fetch(innerCtx, item.CID, item.Target)
-	}()
+// FetchResult downloads the result of a computation and saves it to a local file.
+func (httpDownloader *Downloader) FetchResult(ctx context.Context, item downloader.DownloadItem) (string, error) {
```
rossjones, line 38: I know it's not the time to do it, but this whole downloader package feels like it should be integrated with the publishing for the storage providers.

pkg/lib/gzip/gzip.go
```diff
@@ -0,0 +1,96 @@
...
+
+const DefaultMaxDecompressSize = 100 * 1024 * 1024 * 1024 // 100 GB
+
+// Decompress takes the path to a .tar.gz file and decompresses it into the specified directory.
+func Decompress(tarGzPath, destDir string) error {
+	return DecompressWithMaxBytes(tarGzPath, destDir, DefaultMaxDecompressSize)
+}
+
+// DecompressWithMaxBytes takes the path to a .tar.gz file and decompresses it into the specified directory.
+// It enforces a maximum decompressed file size to prevent decompression bombs.
```
rossjones, line 21:
```suggestion
// It enforces a maximum decompressed file size (per file) to prevent decompression bombs.
```

rossjones (approved):
Couple tiny suggestions, but LGTM

pkg/config/types/requester.go
wdbaruni: To understand better, are you suggesting something like:
```
type RequesterConfig struct {
	...
	
	StorageConfig    StorageConfig          `yaml:"StorageConfig"`
}


type StorageConfig struct {
	S3SignedURLConfig S3SignedURLConfig `yaml:"S3SignedURLConfig"`
}

type S3SignedURLConfig struct {
	Disabled   bool     `yaml:"S3SignedURLDisabled"`
	Expiration Duration `yaml:"S3SignedURLExpiration"`
}

Comment (rossjones) [reply pkg/config/types/requester.go:22]: Yes, just so that requester doesn't end up with lots of top level configs for different providers.

Comment (wdbaruni) [reply pkg/config/types/requester.go]: changed in https://github.com/bacalhau-project/bacalhau/pull/2995/commits/cbfb539d6dd5e6c954e411afc5c452030702d90a. The config structure now looks like:
```
    Requester:
        ...
        Worker:
            WorkerCount: 8
            WorkerEvalDequeueTimeout: 5s
            WorkerEvalDequeueBaseBackoff: 1s
            WorkerEvalDequeueMaxBackoff: 30s
        StorageProvider:
            S3:
                PreSignedURLDisabled: false
                PreSignedURLExpiration: 30m0s
```

Let me know if you have other thoughts, specially around `PreSignedURLDisabled`. I don't like the name, but I wanted to make it enabled by default

File: pkg/downloader/http/downloader.go [hunk 4]
```diff
@@ -2,89 +2,126 @@ package http
...
 		return err
 	}
 
 	return nil
 }
+
+func checkHTTPResponse(resp *http.Response, url string) error {
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return nil
+	}
```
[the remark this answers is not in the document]
wdbaruni: added a TODO in https://github.com/bacalhau-project/bacalhau/pull/2995/commits/cbfb539d6dd5e6c954e411afc5c452030702d90a

pkg/config/types/requester.go
rossjones: No, that looks good, seems best to default to on.  And apologies I didn't realise it'd lead to renaming the other config structs (jobstore) too :D 

Looks good, we should merge....

PR #2998, opened 2023-11-09
change to orchestrator endoint file to allow cross origin comms

XandraMcC:


pkg/publicapi/endpoint/orchestrator/endpoint.go
```diff
@@ -33,6 +34,7 @@ func NewEndpoint(params EndpointParams) *Endpoint {
 	// JSON group
 	g := e.router.Group("/api/v1/orchestrator")
 	g.Use(middleware.SetContentType(echo.MIMEApplicationJSON))
+	g.Use(echo_middleware.CORS())
```
wdbaruni, line 37: this will only take effect for orchestrator APIs. If you want to enable CORS in other APIs, such as `/api/v1/agent`, then you either need to enable CORS under each API or at the root level here https://github.com/bacalhau-project/bacalhau/blob/main/pkg/publicapi/server.go#L100

XandraMcC: Do you think that is the way we should do it? at the moment I am only using orchestrator (and requestor for old jobs). If yes, I can add at the route level

wdbaruni: That depends on your use case. It makes sense to limit it for now to the orchestrator's APIs if thats all what you need. The orchestrator APIs can also query old jobs, so you won't need to use the requester's APIs. Approving

wdbaruni (approved):

pkg/publicapi/endpoint/orchestrator/endpoint.go
XandraMcC: If I need more than the orchestrator I will change in future

PR #3001, opened 2023-11-10
Serve webui

XandraMcC:


cmd/cli/serve/serve.go
rossjones, line 1: We be running with `bacalhau serve --node-type=requester,compute --web-ui`.    Perhaps we _should_ specify it as a node type?  So it would be `--node-type=requester,web` instead?  Either way, I'm not a fan of `web-ui`, maybe just `web`?

cmd/cli/serve/serve.go
```diff
@@ -173,6 +174,11 @@ func serve(cmd *cobra.Command) error {
 		return err
 	}
 
+	startWebUI, err := config.Get[bool](types.NodeWebUI)
+    if err != nil {
+        return err
+    }
+
```
rossjones, line 181: Formattings a bit wonky.  If your editor supports plugins, you can get them to run `go fmt` for you whenever it saves a file, and that'll save you having to indent stuff yourself.

cmd/cli/serve/serve.go
```diff
@@ -235,6 +241,7 @@ func serve(cmd *cobra.Command) error {
 		IsComputeNode:         isComputeNode,
 		IsRequesterNode:       isRequesterNode,
 		Labels:                getNodeLabels(autoLabel),
+		WebUI:			   	   startWebUI,
```
rossjones, line 244: More wonky

cmd/cli/serve/serve.go
XandraMcC: Speaking with @wdbaruni he mentioned to run the dashboard we would need requester & compute

If so, we could do `bacalhau serve --node-type=requester,compute,web` OR `bacalhau serve --node-type=requester,compute,dashboard`

cmd/cli/serve/util.go
```diff
@@ -167,6 +168,27 @@ func SetupIPFSClient(ctx context.Context, cm *system.CleanupManager, ipfsCfg typ
 	return client, nil
 }
 
+func StartWebUIServer() (error) {
+     // Define command to start the React server
```
rossjones, line 172: I don't think this should be the way we launch our web ui as it's adding a big dependency on deployments that they need to have node/npm installed and ideally we wouldn't do that.  

We _should_ have a build step that uses NPM to build a deployable standalone version of the content/JS/CSS.   Then we can embed that into this binary (something like https://harsimranmaan.medium.com/embedding-static-files-in-a-go-binary-using-go-embed-bac505f3cb9a) and then implement a webserver in the app to serve that embedded content.

cmd/cli/serve/serve.go
rossjones: web would save some typing :D

cmd/cli/serve/serve.go
```diff
@@ -250,6 +258,17 @@ func serve(cmd *cobra.Command) error {
 		nodeConfig.RequesterTLSKeyFile = key
 	}
 
+	// Start up Dashboard
+	if startWebUI {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
```
rossjones, line 264: If you check the end of this function, you should see 

```
<-ctx.Done() // block until killed
```

which suggests there is already a context you can use.  And at the start of the function you can see it defined.
So, short version, you don't need to create another one here, you can use the one that was already created (just delete these two lines)

cmd/cli/serve/serve.go
```diff
@@ -235,6 +242,7 @@ func serve(cmd *cobra.Command) error {
 		IsComputeNode:         isComputeNode,
 		IsRequesterNode:       isRequesterNode,
 		Labels:                getNodeLabels(autoLabel),
+		WebUI:                 startWebUI,
```
frrist, line 245: Looks like this is unused. Can it be removed from the `NodeConfig`?

pkg/node/node.go
```diff
@@ -59,6 +59,8 @@ type NodeConfig struct {
 	NodeInfoPublisherInterval   routing.NodeInfoPublisherIntervalConfig
 	DependencyInjector          NodeDependencyInjector
 	AllowListedLocalPaths       []string
+	WebUI						bool
```
frrist, line 62: drop this too.

cmd/cli/serve/serve.go
```diff
@@ -250,6 +258,17 @@ func serve(cmd *cobra.Command) error {
 		nodeConfig.RequesterTLSKeyFile = key
 	}
 
+	// Start up Dashboard
+	if startWebUI {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		err := StartWebUIServer(ctx)
```
frrist, line 266: This method is blocking since it waits for the command to complete or the context to be canceled (same is true for the simplified version of the method I mentioned below). We'll need to either call this method via a go routine, or make it non-blocking else the node will never start when the webUI is enabled. We can't merge this until this issue is addressed.

Typically methods prefixed with `Start*` imply they are non-blocking - they start something and then return, conversely methods prefixed with `Run*` are usually blocking. The method name should reflect its behavior.

cmd/cli/serve/serve.go
frrist: I think this should be a separate command all together. The `serve` command and method are complicated enough as is, and the logic for starting the webui has no dependencies on bacalhau from what I can tell. Let's move this to a new command, e.g. `bacalhau webui`. In the new command we can intercept the processes stdout and stderr and display it in the terminal for logging purposes.

frrist (changes requested):
Good stuff!

Couple issues need to be addressed before merging.
- Method to start webui is blocking, and will prevent the node from starting when enabled.
- `NodeConfig` received unnecessary change, just revert it.
- I don't think this belongs on the `serve` command, should be a new command.

cmd/cli/serve/util.go
```diff
@@ -167,6 +168,39 @@ func SetupIPFSClient(ctx context.Context, cm *system.CleanupManager, ipfsCfg typ
...
+		<-done
+		return fmt.Errorf("server stopped due to context cancellation")
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("server stopped unexpectedly: %w", err)
+		}
+		return nil
+	}
+}
+
```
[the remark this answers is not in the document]
rossjones: This was originally a go-routine, so I think it's the 'go' that is missing when it is called.

cmd/cli/serve/serve.go
wdbaruni: - I am not in favour of defining the ui as a node type. It is just an interface for the requester node. I am also not in favour of defining a separate command `bacalhau webui` as I don't see a common case where users would run the ui on a separate fleet than the requester. We can explore this when there is demand for it
- @XandraMcC to clarify my comment, the ui mainly needs a requester node as it will be calling APIs exposed by the requester node. My comment was for your development and testing, it might make things easier to run a hybrid node `--node-type=requester,compute` so that you can submit and execute jobs. 
- Changing the flag to `--web` can be misleading as it doesn't tell anything about a user interface. `--ui` or `--webui` can be good choices.
- My recommendation is to do something like `bacalhau server --node-type=requester --ui`. I'll also be happy with making the ui enabled by default if the node is a requester node. 
- I am sure the `ui` will need additional configuration, such as the port number. My recommendation is to start pushing users to use the config file instead of adding more flags to `bacalhau serve` command.

cmd/cli/serve/serve.go
wdbaruni: why is lint not complaining?

cmd/cli/serve/serve.go
XandraMcC: A few open questions:

1. Looks like we are in favour of having it as a separate flag. I think `webui` works as it's shorter, so: `bacalhau server --node-type=requester --webui` (`ui` is confusing as the CLI is also a ui)
2. No need for compute flag (just requester)
3. I think enabled by default may not be great, should be an option
4. Still working of the port configuration. Will be a default port for now

cmd/cli/serve/util.go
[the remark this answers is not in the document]
frrist: We'll need a way to return the errors from the routine then, perhaps we could just log them?

cmd/cli/serve/serve.go
frrist: I would guess because the value on the struct is being used since we write to it

cmd/cli/serve/serve.go
frrist: > I am also not in favour of defining a separate command bacalhau webui as I don't see a common case where users would run the ui on a separate fleet than the requester

Some reasons could be:
- They don't want to take down their bacalhau node every-time the UI updates, which could happened independently to updates of bacalhau itself. If the only way they can start the UI is through `serve` command they're a bit limited.
- They may want to isolate the bacalhau process from the UI process, rather than run it as a child process.
- They may want to collect logs and metrics from the UI process, or see its logs in the terminal. Right now they will all be intermixed.

> I am sure the ui will need additional configuration, such as the port number. My recommendation is to start pushing users to use the config file instead of adding more flags to bacalhau serve command.

This seems like further motivation to separate it from the `serve` command. It would reduce the complicity of arguments passes to `serve` and make it easier to configure the UI using flags.

If we really want the UI to run with `serve` we should:
1. Provide a separate command to start and configure the UI, e.g. `bacalhau webui`
2. In the `serve` command we call the new bacalhau command directly, sorta like we do now, but with the below instead:
```go
exec.Command(os.Executable(), "webui")
```
This gives us the best of both worlds, users have the option to simply run it via the `serve` command and get a UI. Or run it via it's own command independent of bacalhau.

cmd/cli/serve/util.go
```diff
@@ -168,9 +168,10 @@ func SetupIPFSClient(ctx context.Context, cm *system.CleanupManager, ipfsCfg typ
 	return client, nil
 }
 
+
 func StartWebUIServer(ctx context.Context) error {
 	// Define command to start React server
-	cmd := exec.Command("npm", "run", "build")
+	cmd := exec.CommandContext(ctx, "npm", "run", "build")
```
rossjones, line 174: Isn't this something that should happen in the Makefile as a build step, and then the output of that could be embedded in the binary and served from here over http?

cmd/cli/serve/serve.go
wdbaruni: As I said, I don't see a _common_ case where users would run the ui on a separate fleet than the requester, and the reasons you mentioned don't sound common that _most_ users will do. 

Lets keep things simple, allow our users to run the requester and the webui in a single command. and simplify their onboarding. If we hear demand for running the webui separately, then we can always introduce `bacalhau webui` command in the future. It is not a one way door decision.

PR #3006, opened 2023-11-13
Improve bid strategy error messages

### LINKED ISSUE #2526: do our best to give more sensible error messages around docker registry
Example:
```
- Status: 'Could not inspect image "bacalhauproject/duckdb-log-processor:v0.17"
  - could be due to repo/image not existing, or registry needing authorization:
```

Image didn't exist (my fault) - we should try to make this friendly.

### LINKED ISSUE #2544: Error for submitting job with networking is very confusing
Job yaml:
```
Job:
  APIVersion: V1beta1
  Spec:
    Deal:
      Concurrency: 4
    Docker:
      Entrypoint:
        - /bin/bash
        - -c
        - python3 /process.py aperitivo_logs.log.1 archive-bucket "SELECT * FROM log_data WHERE message LIKE '%[SECURITY]%' ORDER BY '@timestamp'"
      Image: docker.io/bacalhauproject/duckdb-log-processor:v1.0.6
    Engine: Docker
    Network:
      Type: Full
    inputs:
      - Name: file:///var/log/logs_to_process
        SourcePath: /var/log/logs_to_process
        StorageSource: LocalDirectory
        path: /var/log/logs_to_process
```

Error:
```
❯ cat job.yaml | bacalhau create
Job successfully submitted. Job ID: 14745a35-7b29-4b81-bf00-5acabf086200

	Communicating with the network  ................  done ✅  0.1s
	   Creating job for submission  ................  done ✅  0.0s
	   Job waiting to be scheduled  ................  err  ❌  0.0s

Error submitting job: job rejected: command `/terraform_node/apply-http-allowlist.sh` returned non-zero exit code 1
Job Results By Node:
```

### LINKED ISSUE #2913: Fix error message for "networking enabled"
I just ran this job:

The error came back:

```
	   Creating job for submission  ................  err  ❌  2.0s

Error submitting job: Job rejected: networking is enabled: false
Job Results By Node:
• Node QmXGrMuX, QmeVvpXC, Qma7dgtw, QmTyCF46, QmYHLuZq, QmNkP5ef, QmXZhend, QmVjG3Sc, QmXevyMC, QmVw5y5c, Qmbh79tK, QmaXFzuP, QmZX4DRh, QmeJr2aX, QmcAbmNZ, QmSwrDKC: Job rejected: networking is enabled: false. execution failed
```

I ASSUME this means that networking is disabled on all these nodes? Is there a better way we can do/name this? E.g. 
```
Error submitting job: No nodes fit the following requirement:
'network=full'
```

```
$ bacalhau error --full
<YAML FOLLOWS ...>
Nodes Checked:
- QmXGrMuX
- QmeVvpXC
...
- QmSwrDKC
<... /YAML>
```

Right now, we're obviously just passing through go errors, which is not the best user experience.

simonwo:
This makes the error messages returned by bid strategies more user friendly; resolves #2913, resolves #2544, resolves #2526.

Also adds some minor efficiency buffs and removes some outdated code (see commit messages).


pkg/bidstrategy/semantic/networking.go
```diff
@@ -17,13 +16,19 @@ func NewNetworkingStrategy(accept bool) *NetworkingStrategy {
 	return &NetworkingStrategy{accept}
 }
 
+const (
+	docsLink        = "(see https://docs.bacalhau.org/next-steps/networking)"
```
rossjones, line 20: I like this, we should definitely keep it, would be nice if we could do it with more errors too.

I think long-term using a hard-coded link might one day cause problems (either because it'll be wrong or we'll be managing redirects forever), so it might be worth a ticket to come up with a neat way of referencing the docs from the code.

rossjones (approved):
LGTM. Minor comment on a confusing (to me) change in the http probe.

pkg/bidstrategy/semantic/external_http.go
```diff
@@ -59,15 +64,8 @@ func (s *ExternalHTTPStrategy) ShouldBid(
...
-		return bidstrategy.BidStrategyResponse{
-			ShouldBid: false,
-			Reason:    fmt.Sprintf("url `%s` returned %d status code", s.url, resp.StatusCode),
-		}, nil
-	}
-
-	if resp.Header.Get("Content-Type") != "application/json" {
-		return bidstrategy.NewShouldBidResponse(), nil
+	if resp.StatusCode >= http.StatusBadRequest || resp.Header.Get("Content-Type") != "application/json" {
+		return bidstrategy.NewBidResponse(resp.StatusCode < http.StatusBadRequest, statusCodeReason, s.url, resp.StatusCode), nil
```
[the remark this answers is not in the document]
simonwo: I'm not super keen to keept his structure - this is probably an over reach of my LOC conservatism. I'll put it back.

Ah indeed, the extra JSON path is not documented. I'll add something to the docs about that.

pkg/bidstrategy/semantic/networking.go
simonwo: FWIW I think the discipline of maintaining redirects is no bad thing and is good for a healthy Web for other reasons. We should make it easy on ourselves though.

PR #3008, opened 2023-11-13
Show node ranking failures in the CLI (again)

simonwo:
We used to show a list of nodes that had failed ranking when not enough nodes were found to run the job. At some point we regressed and no longer showed those errors, leading to lots of confusion around why jobs aren't running. This commit restores the view of node ranking failures being passed to the CLI.

### DIFF
--- a/pkg/orchestrator/selection/selector/node_selector.go
+++ b/pkg/orchestrator/selection/selector/node_selector.go
@@ -33,7 +33,7 @@ func (n NodeSelector) AllNodes(ctx context.Context) ([]models.NodeInfo, error) {
 }
 
 func (n NodeSelector) AllMatchingNodes(ctx context.Context, job *models.Job) ([]models.NodeInfo, error) {
-	filteredNodes, err := n.rankAndFilterNodes(ctx, job)
+	filteredNodes, _, err := n.rankAndFilterNodes(ctx, job)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/orchestrator/selection/selector/node_selector.go
```diff
@@ -42,45 +42,46 @@ func (n NodeSelector) AllMatchingNodes(ctx context.Context, job *models.Job) ([]
 	return nodeInfos, nil
 }
 func (n NodeSelector) TopMatchingNodes(ctx context.Context, job *models.Job, desiredCount int) ([]models.NodeInfo, error) {
-	filteredNodes, err := n.rankAndFilterNodes(ctx, job)
+	possibleNodes, rejectedNodes, err := n.rankAndFilterNodes(ctx, job)
 	if err != nil {
 		return nil, err
 	}
-	if len(filteredNodes) < desiredCount {
+	if len(possibleNodes) < desiredCount {
```
frrist, line 49: `possibleNodes` contains the same content as `filteredNode` did, right? And were just returning the rejected nodes here for better error messages?

frrist (approved):

pkg/orchestrator/selection/selector/node_selector.go
simonwo: Yeah that's it. I renamed it because in my head `filtered` could mean both "filtered in" and "filtered out" and I wanted to make it clearer which was which

Issue #3004, opened 2023-11-12
Don't persist defaults in BACALHAU_DIR/config.yaml


wdbaruni:
When we initialize Bacalhau configuration for the very first time, we create a `config.yaml` file with default config values under `BACALHAU_DIR` ( usually `~/.bacalhau`). 

The problem is changing these defaults will only take affect with new users/hosts that run the CLI for the very first time. Existing users who did not explicitly override the configs will have to take a manual action to apply the new configs, which will result in undesired behaviours, such as https://github.com/bacalhau-project/bacalhau/issues/3003.

The right behaviour should be for `config.yaml` to be empty by default, and be a place for users to define their own configs. This should make `config.yaml` be less noisy, and enable updating defaults without user intervention 

The mitigation for now is for users to:
- If only using default configs, then delete `~/.bacalhau/config.yaml` 
- If overriding some configs, then delete entries inside `~/.bacalhau/config.yaml` that weren't provided by you

wdbaruni:
ideally, we should bump the repo version from 1 to 2, and cleanup the config.yaml on behalf of the user as part of the migration.

Issue #3003, opened 2023-11-10
`bacalhau` fails to `get`


dtbuchholz:
## Summary

Using `bacalhau` will successfully start up, `run` jobs, and `describe` them. However, the `get` command hangs and exits early. This is also true when doing [local dev](https://github.com/bacalhau-project/bacalhau/blob/main/docs/running_locally.md) with `bacalhau devstack`

## Details
You can test this in either production or dev env:
1. Set up the dev env: `bacalhau devstack`
2. Copy the env vars in a separate window:
   ```
   export BACALHAU_NODE_IPFS_SWARMADDRESSES=/ip4/127.0.0.1/tcp/58669/p2p/QmPPGBfhzWYfA6hBhkfskoxNQMhgDUDwCt8y4s8E6mT39d
   export BACALHAU_NODE_CLIENTAPI_HOST=0.0.0.0
   export BACALHAU_NODE_CLIENTAPI_PORT=58658
   export BACALHAU_NODE_LIBP2P_PEERCONNECT=/ip4/127.0.0.1/tcp/58651/p2p/QmPZUMAf6VkygJZ6C4JaxonNXFFUJKLp2LHQ7t1ojm9Jag,/ip4/127.0.0.1/tcp/58661/p2p/QmUfykjRmiQhMUZdAXva2rKa5jySnc5g3BWF1xWrywS4J7,/ip4/127.0.0.1/tcp/58666/p2p/QmdEbzT6BvqMRXDn1spTU5v3wmSNo1JUfHq3XwYPtdqJ66,/ip4/127.0.0.1/tcp/58671/p2p/QmRtpY9DA8HHQA8riM7ifFf5yk73NY2wmPpwRUcBftJQ29
   ```
3. Run a simple job (success): `bacalhau docker run ubuntu echo "hello devstack test"`
    <details>
    <summary>output</summary>
    <pre>
    Using default tag: latest. Please specify a tag/digest for better reproducibility.
    Job successfully submitted. Job ID: a9f4255a-075d-44a0-ba3b-445c73987fc6
    Checking job status... (Enter Ctrl+C to exit at any time, your job will continue running):
    
	    Communicating with the network  ................  done ✅  0.0s
	       Creating job for submission  ................  done ✅  0.5s
	                   Job in progress  ................  done ✅  0.0s
    
    To download the results, execute:
	    bacalhau get a9f4255a-075d-44a0-ba3b-445c73987fc6
    
    To get more details about the run, execute:
[...]

Issue #2941, opened 2023-10-21
API for managing config


alabdao:
It would be great to have an endpoint where we can update configs such as adding/updating peer, ipfs info, etc.
This would be useful for environments where peers are not known at startup and can be added later by interacting with the API.
Similar to what IPFS provides, for example, [add swarm peers](https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-swarm-peering-add).

frrist:
#3012 didn't add support for a command like `ipfs swarm peers` to reopening.

wdbaruni:
As we are moving away from embedded ipfs node, the work in #3012 should be enough to close this issue

PR #3009, opened 2023-11-13
refactor: no longer write a config file by default

frrist:
### Why
This PR addresses #3004 by preventing a config file from being written to the bacalhau repo when the node is initialized. With this change, bacalhau will no longer write a config file in the bacalhau repo. 

The initial motivation for this work spawned from https://github.com/bacalhau-project/bacalhau/issues/3003 and https://github.com/bacalhau-project/bacalhau/pull/2999. When the demo networks were re-deployed the PeerIDs and IP addresses of the bacalhau bootstrap nodes and IPFS nodes changed. While the code's default values for these entities were updated accordingly, users with pre-existing Bacalhau nodes encountered issues. Their local config files continued to reference outdated bootstrap and swarm peer values, failing to align with the newly established defaults. (Reminder: values in a config file will override the default values as outlined in our [docs](https://docs.bacalhau.org/running-node/configuration#configuring-a-bacalhau-node)). Users with these outdated config values are unable to bootstrap to the network and/or retrieve content over the network.

This change will not fix issues previous users are experiencing as they still hold an invalid config. But it will prevent issues like this happening again in the since when we change defaults now users without a config will receive those new values when they update their version.

### Decision
**The decision we need to make here is: Do we want to follow this change with a PR that implements config file migration code?**

If we implement a migration:
- Users won't need to manually edit their config file if it contains outdated values, bacalhau will do it for them in this specific case.

If we don't implement a migration:
- Users with a config file will need to edit the file:
   - removing values of `Node.IPFS.SwarmPeers`.
   - removing values of `Node.BootstrapAddresses` if and only if they intend to use the `serve --peers=config` flag which forces bacalhau to use the bootstrap addresses defined in the config.

Migration v No Migration:
The migration for this would entail adding code to the bacalhau config package that explicitly checks the loaded config file for Bootstrap addresses and Swarm addresses such as these:
- https://github.com/bacalhau-project/bacalhau/blob/1a225c03c10cdb8fed7bd99b802b8e9d7a54ac8e/pkg/config/configenv/production.go#L59-L70
- https://github.com/bacalhau-project/bacalhau/blob/1a225c03c10cdb8fed7bd99b802b8e9d7a54ac8e/pkg/config/configenv/production.go#L33-L36
In the event they are present it will replace them with the correct values, otherwise it's a noop.


dtbuchholz:
note: i can confirm this fixed the issue i opened with #3003. i manually removed `Node.IPFS.SwarmPeers` entirely and replaced the values for `Node.BootstrapAddresses` with those outlined above. tested with both `bacalhau` and `bacalhau devstack`—thanks for the quick resolution!

pkg/config/getters.go
```diff
@@ -161,3 +164,144 @@ func GetStoragePath() string {
...
+	if err != nil {
+		return err
+	}
+
+	if isNested {
+		val, ok := compositeValue[key]
+		if !ok {
+			// NB(forrest): this case should never happen as we ensure all configuration values
+			// have a corresponding key via code gen. If this does occur it represents an error we need to debug.
+			return fmt.Errorf("CRITICAL ERROR: invalid configuration detected for key: %s. Config value not found", key)
```
rossjones, line 234: Where this eventually unrolls to ForKey() the error we set here is ignored and "value not of expected type, got: %T" return instead.  Should we log the error as well if it is important?

pkg/config/getters.go
```diff
@@ -161,3 +164,144 @@ func GetStoragePath() string {
...
+func Get[T any](key string) (T, error) {
+	raw := viper.Get(key)
+	if raw == nil {
+		return zeroValue[T](), fmt.Errorf("value not found for %s", key)
+	}
+
+	var val T
+	val, ok := raw.(T)
+	if !ok {
+		err := ForKey(key, &val)
```
rossjones, line 181: There's a lot of code associated with this path through `Get[T]` do you have an ide of how often we expect to take this path?

rossjones (commented):
I guess my only concern now is that we don't know what a config file looks like if there isn't a default one (or even a commented out one) to crib from.  Should we have something in https://docs.bacalhau.org/?  I went to take a look but couldn't find anything.   

> Do we want to follow this change with a PR that implements config file migration code?

I think yes, as long as it only replaces previously good (now bad) defaults.  I wonder whether we should have a general migration strategy so that we have a known way of providing migrations between releases?  e.g. specific code that is run going from 1.1.4 to 1.1.5, although I suspect this would be too big a burden unless we said it was only for 1.x.x (e.g. we have to maintain compat and remove nothing until we get to 2.x)

pkg/config/getters.go
rossjones: To answer my own question.  Of the 7 times this func (`Get`) is called, 6 times it is `Get[string]` and once it is `Get[bool]`.

wdbaruni:
> I think yes, as long as it only replaces previously good (now bad) defaults. I wonder whether we should have a general migration strategy so that we have a known way of providing migrations between releases

That is my recommendation as well. We have different versioning for our repos not related to releases, and can be found under `~/.bacalhau/repo.version`. The current version is 1, and this change can bump the version to 2 or something similar. Having a general migration strategy between repo versions is my recommendation as well during my discussion with Forrest yesterday.

For this specific migration, we should load the existing config, compare each value with current and previous defaults, and only write back the config if the values don't match. The goal is not only to replace `Node.IPFS.SwarmPeers`, but delete all default values

frrist:
> I guess my only concern now is that we don't know what a config file looks like if there isn't a default one (or even a commented out one) to crib from

Yeah, a concern I share. We can either continue to generate one as we do now, but with everything commented out (sshd, postgres, etc. do this). Or we can add a `bacalhau config` command that spits our the default config to stdout (e.g. `bacalhau config create > config.yaml`) . There is some appetite for a config API in https://github.com/bacalhau-project/bacalhau/issues/2941 so I am leaning towards the latter option. I am hesitant to put a default config in the docs at this point in time given how frequently we add new config values. Docs could quickly become outdated if we don't stay on top of config changes. Generating a default config on demand seems simpler to me.

pkg/config/getters.go
frrist: Great catch. I've adjusted the error message there to include the original error. I will also add a log line here.

rossjones:
> we can add a bacalhau config command that spits our the default config to stdout (e.g. bacalhau config create

I think `bacalhau config defaults` might be useful, but I'm not sure about priority as I think it might be lower priority than other issues (such as the migration).  Is it worth putting into a ticket to get picked up later?

wdbaruni:
> I think `bacalhau config defaults` might be useful, but I'm not sure about priority as I think it might be lower priority than other issues (such as the migration). Is it worth putting into a ticket to get picked up later?

I agree with Ross. Not a priority and better to wait until we get more signals on how useful it is

frrist:
Anything blocking this from being approved? The code that adds a migration can follow this work.

rossjones (approved):

Issue #1758, opened 2023-01-24
Compute node panics when configured resource usage is unavailable


simonwo:
https://github.com/filecoin-project/bacalhau/blob/a7f006ff7736325832a365addbb0218cf7cc76d2/pkg/node/config_compute.go#L173-L177

If I pass more resources to the compute node config than I have available (e.g. `--limit-job-gpus 4` when I only have 1 GPU), I receive a panic.

This is probably better than no error at all :) But why does this need to be a panic? (Assuming that panics are bad/scary for users.) Can't this interface just support an error?

Issue #2980, opened 2023-11-01
memory limits seem weird?


aronchick:
Here's why my resource limits look like:
```
            JobResourceLimits:
                CPU: ""
                Memory: ""
                Disk: ""
                GPU: ""
            DefaultJobResourceLimits:
                CPU: 8000m
                Memory: 8Gi
                Disk: ""
                GPU: ""
```

But i'm getting this error - i don't understand?

```
validating compute config: 1 error occurred:
	* default job resource limits {CPU:8 Memory:8589934592 Disk:0 GPU:0} exceed job resource limits {CPU:1.6 Memory:6652533145 Disk:23600719462 GPU:0}
```

rossjones:
Is that suggesting that you requested defaults that are higher than availability on the machine?  The second set of numbers show ~6GB ram which seems like it might be something your machine might have.  I guess two observations:

* "exceed job resource limits" is a confusing message given the name of the previous config blog
* If we set the limits higher than the available resources, maybe a warning but carry on anyway?

simonwo:
I think this is partly a naming issue - `DefaultJobResourceLimits` might be better named `DefaultResources`. This is the limit that jobs will get when they don't specify any resource limits themselves.

So in your case, you are saying "by default I want jobs to have 8 CPUs and 8Gibibytes of RAM" and Bacalhau is saying "you only have 1.6 CPUs and 6GB of RAM, so your jobs won't be accepted if they use these defaults".

PR #3012, opened 2023-11-14
feat: add a command to generate a default config and list keys

frrist:
# Why
We need:
1. A simple way for users to see what their configuration is as we no longer have a file by default and instead depend on code constants.
2. A user friendly way to edit specific parts of the bacalhau configuration system. 
3. A way to stop depending on default values we set for resources based config fields. We can now generate a sane set of default
4. there are numerous issues related to capacity configuration.

# How
```bash
Interact with the bacalhau configuration system.

Usage:
  bacalhau config [command]

Available Commands:
  auto-resources Auto set compute resources values in the config.
  default        Show the default bacalhau config.
  list           List all config keys.
  set            Set a value in the config.
```
## `bacalhau config default`
prints the default configuration to standard out, fairly simple.

## `bacalhau config list` 
lists all config keys are their values in key sorted order, e.g:
| KEY| VALUE|
|-----|-------|                                    
| metrics.eventtracerpath       |                                  /dev/null   |                             
| metrics.libp2ptracerpath           |                             /dev/null    |                            
| node.allowlistedlocalpaths                |                      []               |                        
| node.bootstrapaddresses        |                                 []   |    
| node.clientapi.port                          |                  1234                                     |
| ......so on........|. ......run it your self and see............|

## `bacalhau config set`
Edit the config file for you, it accepts a key, like the ones returned by `list` and a value to assign that key. `bacalhau set` will validate types, and will complain if you try and assign something the wrong type, like a bool as a string, string as an int, float as a log mode, log mode as a job policy, etc.
Example usage:
```
# There isn't a config file yet, as shown
$ cat ~/.bacalhau/config.yaml
cat: /Users/frrist/.bacalhau/config.yaml: No such file or directory

# Now we write some field, this will create a file with the single entry
$ bacalhau config set node.clientapi.host 0.0.0.0
$ cat ~/.bacalhau/config.yaml
node:
    clientapi:
        host: 0.0.0.0

# Setting a port to a host string fails with helpful error
$ bacalhau config set node.clientapi.port 0.0.0.0
Error: setting "node.clientapi.port": invalid integer value: "0.0.0.0": strconv.ParseInt: parsing "0.0.0.0": invalid syntax

# It wants an int
$ bacalhau config set node.clientapi.port 1234

# we have a log mode in our config, a custom type validation works here too!
$ bacalhau config set node.loggingmode foobar
Error: setting "node.loggingmode": invalid log mode value: "foobar": "foobar" is an invalid log-mode (valid modes: 

# now use a valid type
$ bacalhau config set node.loggingmode json

# similar to before, we have a StorageTyep with validation
$ bacalhau config set node.requester.jobstore.type mississippi
Error: setting "node.requester.jobstore.type": invalid storage type value: "mississippi": StorageType: unknown type
$ bacalhau config set node.requester.jobstore.type inmemory

# after applying all the above config commands we are left with a config containing only the things we set.
$ cat ~/.bacalhau/config.yaml
node:
    clientapi:
        host: 0.0.0.0
        port: 1234
    loggingmode: json
    requester:
        jobstore:
            type: InMemory
```

## `bacalhau cofing auto-resources` (Better name welcome)
This will auto generate a configuration file based on your system resources. Setting fields related to capacity, job resources, and queuing:
```
Auto set compute resources values in the config.

Usage:
  bacalhau config auto-resources [flags]

Flags:
      --default-job-percentage int   TODO I DON'T KNOW WHAT THIS DOES FIX ME (default 75)
  -h, --help                         help for auto-resources
      --job-percentage int           Percentage expressed as a number from 1 to 100 representing per job amount of resource the system can be using at one time for a single job (values over 100 will be rejected) (default 75)
      --queue-job-percentage int     Percentage expressed as a number from 1 to 100 representing the total amount of resource the system can queu at one time in aggregate for all jobs (values over 100 are accepted) (default 150)
      --total-percentage int         Percentage expressed as a number from 1 to 100 representing total amount of resource the system can be using at one time in aggregate for all jobs (values over 100 will be rejected) (default 75)
```

Example:
```
$ bacalhau config auto-resources
16:16:36.626 | INF pkg/compute/capacity/system/provider.go:106 > cannot inspect system GPUs: nvidia-smi not installed. GPUs will not be used.

$ cat ~/.bacalhau/config.yaml
node:
    compute:
        capacity:
            defaultjobresourcelimits:
                cpu: 7500m
                disk: 555 GB
                gpu: "0"
                memory: 39 GB
            jobresourcelimits:
                cpu: 7500m
                disk: 555 GB
                gpu: "0"
                memory: 39 GB
            queueresourcelimits:
                cpu: 15000m
                disk: 1.1 TB
                gpu: "0"
                memory: 155 GB
            totalresourcelimits:
                cpu: 7500m
                disk: 555 GB
                gpu: "0"
                memory: 39 GB
```


cmd/cli/root.go
```diff
@@ -157,6 +158,9 @@ func NewRootCmd() *cobra.Command {
 	RootCmd.AddCommand(id.NewCmd())
 	RootCmd.AddCommand(devstack.NewCmd())
 
+	// config command...obviously
```
rossjones, line 161: So obvious it might not need a comment :P

rossjones (approved):
LGTM.  

I'm not sure about the list command, I don't know how useful that will be. Perhaps if it showed (default) next to those that were default values (unchanged) or an option to only list default, or non-default items

frrist:
The idea of `list` is to pair it with a `set` command. Since we've stopped producing a default config on startup, we now need to perform some chicanery to include installationIDs in the config file: https://github.com/bacalhau-project/get.bacalhau.org/commit/867c0001220a9b24b8b461290832642f2e1b9ad4.

`set` would accept a config key (returned from list) and a value, then persist it to the config file on the users behalf.

simonwo (approved):
Great, awesome! Love that we're making good use of the common output stuff.

pkg/compute/capacity/system/provider.go
```diff
@@ -54,6 +54,26 @@ func (p *PhysicalCapacityProvider) GetAvailableCapacity(ctx context.Context) (mo
 	}, nil
 }
 
+func (p *PhysicalCapacityProvider) GetTotalCapacity() (models.Resources, error) {
```
frrist, line 57: The other method `GetAvailableCapacity` in this file is applying 80% to all values. I needed something that returned 100% to do maths right.

pkg/model/resource_usage_parser.go
```diff
@@ -5,39 +5,59 @@ import (
...
+	"github.com/dustin/go-humanize"
 )
 
-func ParseResourceUsageConfig(usage ResourceUsageConfig) ResourceUsageData {
-	return ResourceUsageData{
-		CPU:    ConvertCPUString(usage.CPU),
-		Memory: ConvertBytesString(usage.Memory),
-		Disk:   ConvertBytesString(usage.Disk),
-		GPU:    ConvertGPUString(usage.GPU),
+func ParseResourceUsageConfig(usage ResourceUsageConfig) (ResourceUsageData, error) {
```
frrist, line 11: Rather than ignore the error this method now exposes it to the caller.

pkg/model/resource_usage_parser.go
```diff
@@ -55,11 +75,7 @@ func convertBytesStringWithError(val string) (uint64, error) {
 	if val == "" {
 		return 0, nil
 	}
-	mem, err := datasize.ParseString(sanitizeBytesString(val))
-	if err != nil {
-		return 0, err
-	}
-	return mem.Bytes(), nil
+	return humanize.ParseBytes(val)
```
frrist, line 78: https://github.com/dustin/go-humanize can round trip when parsing (string -> bytes and bytes -> string) and works with SI and IEC units natively. As the name suggests it provides a more human readable value when expressed as a string.

cmd/cli/config/set.go
```diff
@@ -0,0 +1,160 @@
...
+	//calling `Get` on this instance will return a default value from the config structure that we can type assert on.
+	curValue := viperSchema.Get(key)
+
+	// we need special handling for the nodetype since its just a string, but carries implicit requirements in the config.
+	if strings.EqualFold(types.NodeType, key) {
+		if !validNodeTypes(values) {
+			return fmt.Errorf("setting: %q, invalid node type value: %q, must be one of: 'requester' 'compute' 'requester compute'", key, values)
+		}
+	}
+	switch curValue.(type) {
```
frrist, line 64: This bit of code is why the code gen changed - needed multiple viper instances to make this work.

frrist:
Not sure why the S3 test are failing/timeout-ing.... don't think I touched that.

cmd/cli/config/util.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func singleValueOrError(v ...string) (string, error) {
+	if len(v) != 1 {
+		return "", fmt.Errorf("expected single value got %d from %q", len(v), v)
+	}
+	return v[0], nil
+}
+
+// TODO for the node type methods below. Make the NodeType a concrete type so we can make type assertions on it.
```
rossjones, line 30: We already have `Requester` and `Compute` types in pkg/node which are disjoint types (they should probably be implementations of some interface). I'm not sure we need to add more abstractions, we could just use `getNodeType` which is in cmd/cli/serve/util.go after changing it to take a slice of strings (rather than allow it to dig into config itself).

pkg/compute/capacity/system/provider.go
rossjones: Could we not supply the multiplier to GetAvailableCapacity? So that current calls will pass `0.8` (hopefully a well named constant so we know why 0.8, and this PR can pass 1.0?

pkg/config/types/gen_viper/generate.go
```diff
@@ -57,21 +57,24 @@ func generateSetDefaults(t reflect.Type, prefix string, path []string, value ref
...
-				defaultValue = "cfg." + constantPath + ".String()"
-			}
+			/*
+				// Check if the field implements String() method
+				stringerMethod := fieldValue.MethodByName("String")
+				if stringerMethod.IsValid() {
+					defaultValue = "cfg." + constantPath + ".String()"
+				}
+
+			*/
```
rossjones, line 75: I realise it's commented out but what about ..

```go
_, ok := interface{}(valueField).(fmt.Stringer)
if ok { ... }
```

or there's reflect.TypeOf().Implements()

rossjones (commented):
Couple of comments/suggestions, haven't had a chance to try this manually yet

cmd/cli/config/set.go
```diff
@@ -0,0 +1,160 @@
...
+	showCmd := &cobra.Command{
+		Use:   "set",
+		Args:  cobra.MinimumNArgs(2),
+		Short: "Set a value in the config.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return setConfig(args[0], args[1:]...)
+		},
+	}
+	return showCmd
+}
```
simonwo, line 30: Doesn't seem to work for update configuration.
```
% go run . config list | grep 'updateconfig.skipchecks'
 updateconfig.skipchecks                                         false                                    
% go run . config set updateconfig.skipchecks true     
Error: unsupported type <nil> for key: "updateconfig.skipchecks"
Usage:
  /var/folders/xz/qrd28pyx1095q2nzvhd9wb_c0000gn/T/go-build2679611526/b001/exe/bacalhau config set [flags]

Flags:
  -h, --help   help for set

Global Flags:
      --api-host string         The host for the client and server to communicate on (via REST).
                                Ignored if BACALHAU_API_HOST environment variable is set. (default "bootstrap.production.bacalhau.org")
      --api-port int            The port for the client and server to communicate on (via REST).
                                Ignored if BACALHAU_API_PORT environment variable is set. (default 1234)
      --log-mode logging-mode   Log format: 'default','station','json','combined','event' (default default)
      --repo string             path to bacalhau repo (default "/Users/simon/.bacalhau")

unsupported type <nil> for key: "updateconfig.skipchecks"
exit status 1
```

cmd/cli/config/set.go
simonwo: Agreed, a data-first approach would be better here and would avoid a lot of the repetition. We even already have some of this available as flags, which could be adapted to work in both places. https://github.com/bacalhau-project/bacalhau/blob/bfc436a9afb4b62134321f390a94883130ed39c8/cmd/util/flags/types.go#L296-L303

cmd/cli/config/util.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, item := range slice {
+		parts := strings.SplitN(item, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("expected 'key=value', receieved invalid format for key-value pair: %s", item)
+		}
+		key, value := parts[0], parts[1]
+		result[key] = value
+	}
+	return result, nil
+}
```
simonwo, line 21: Already have this in https://github.com/bacalhau-project/bacalhau/blob/bfc436a9afb4b62134321f390a94883130ed39c8/cmd/util/flags/types.go#L172-L180

pkg/model/resource_usage_parser.go
simonwo: Please check if this resolves #1758.

cmd/cli/config/list.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	return nil
+}
+
+var listColumns = []output.TableColumn[configListEntry]{
+	{
+		ColumnConfig: table.ColumnConfig{Name: "Key"},
+		Value: func(s configListEntry) string {
+			return fmt.Sprintf("%s", s.Key)
```
simonwo, line 65: `Key` is already a string??

simonwo (commented):
Could do with some testing here, to give a bit of confidence that it actually works for all config values. Maybe time to introduce sharness?

cmd/cli/config/set.go
frrist: `go run . config list | grep 'updateconfig.skipchecks'` doesn't return anything since that isn't a valid key.
I think what we mean is: `go run . config list | grep 'update.skipchecks'` which returns:
` update.skipchecks                                               false`

To continue with the example then:
```
go run . config set update.skipchecks true
frrist@oak ~/W/s/g/b/bacalhau (frrist/cmd/config)> cat ~/.bacalhau/config.yaml
update:
    skipchecks: true
```
This is good feedback though, we should return a better error in the event a user passes an invalid key.

simonwo: Hm, what I pasted above is literally the output from my terminal. Is it possible it is returning any YAML key (erroneously) contained in my config, even ones that aren't relevant?

simonwo: Haha yes it is 😂
```
% echo "OOGABOOGA: yes" >> ~/.bacalhau/config.yaml
% go run . config list | grep oogabooga
 oogabooga                                                       yes 
```
We should probably only be listing config keys that are recognised?

cmd/cli/config/auto.go
```diff
@@ -0,0 +1,143 @@
...
+		return fmt.Errorf("failed to calculate system physical resources: %w", err)
+	}
+
+	if err := setResources(types.NodeComputeCapacityTotalResourceLimits, settings.TotalPercentage, physicalResources); err != nil {
+		return err
+	}
+	if err := setResources(types.NodeComputeCapacityJobResourceLimits, settings.JobPercentage, physicalResources); err != nil {
+		return err
+	}
+	if err := setResources(types.NodeComputeCapacityDefaultJobResourceLimits, settings.DefaultPercentage, physicalResources); err != nil {
```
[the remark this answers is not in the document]
frrist: Here is the only place we use this value: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/compute.go#L128. The calculator then merges the `DefaultJobResourceLimits` with a tasks `ResourcesConfig`. If any value in the task (job) resource config is zero then the default is substituted in. So it serves as the default on compute nodes for jobs that don't specify resources.

`JobResourceLimits` is used in https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/compute.go#L171 and https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/compute.go#L202. It represents the max resources for a job a compute node bids on. 

Sorry if you already knew that, but writing it out was helpful for me. 

>  If this is true and we set the default to 75% aren't we forcing sequential execution

Yes, in the event we (a compute node) receive a job that doesn't specify any resources, we set its default limit to 75% of our system. For what its worth in this [code](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/config_compute.go#L153) we are using 80% if unspecified (but I am trying to move us away from that code because too many configs).

Do you think we should have a default of less than 75% here, maybe 45%? My worry with lower values is that it could lead to jobs take a really long time to run. On the other hand, this command is totally optional, and I am adding it to help users mitigate issues like: https://github.com/bacalhau-project/bacalhau/issues/2980

------------------------

Separately, it feels strange for jobs to be submitted to the network that don't specify resource requirements. Perhaps it would be worth making `DefaultJobResourceLimits` a value defined by the client submitting the job.

cmd/cli/config/auto.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return nil
+}
+
+func setResources(key string, percentageValue int, resources models.Resources) error {
+	percentage := float64(percentageValue) * .01
+	if err := setConfig(fmt.Sprintf("%s.CPU", key), k8sresource.NewCPUFromFloat(resources.CPU*percentage).ToString()); err != nil {
+		return err
+	}
+	if err := setConfig(fmt.Sprintf("%s.Memory", key), humanize.Bytes(uint64(float64(resources.Memory)*percentage*percentage))); err != nil {
```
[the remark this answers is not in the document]
frrist: whoops, good catch, will fix.

cmd/cli/config/list.go
frrist: yup, will just return it.

cmd/cli/config/list.go
```diff
@@ -0,0 +1,74 @@
...
+		Wide:       false,
+	}
+	listCmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all config keys.",
+		Args:  cobra.MinimumNArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return list(cmd, o)
+		},
+	}
```
simonwo, line 30: One more thing, please add the `cmd.util.ClientPreRunHooks` and `cmd.util.ClientPostRunHooks` to all of these client commands.

pkg/compute/capacity/system/provider.go
frrist: Yeah, I'd actually made that change initially using options but decided against it. I don't know what 80% of a GPU is, for example. In my opinion `PhysicalCapacityProvider` should only ever provide values representing 100% of the system, and allow callers to make their own decision. But, I don't want to unwind that all here, nor do I know if my opinion would be accepted - thus a new method. 

Care if I just leave this new method as is?

pkg/config/types/gen_viper/generate.go
frrist: Yeah that would be better if this code was sticking around, but deleting the commented stuff in commit to follow.

cmd/cli/config/auto.go
```diff
@@ -0,0 +1,143 @@
...
+			"can be using at one time for a single job (values over 100 will be rejected)")
+	flags.IntVar(&settings.DefaultPercentage,
+		"default-job-percentage",
+		defaultDefaultJobPercentage,
+		"TODO I DON'T KNOW WHAT THIS DOES FIX ME")
+	flags.IntVar(&settings.QueuePercentage,
+		"queue-job-percentage",
+		defaultQueuePercentage,
+		"Percentage expressed as a number from 1 to 100 representing the total amount of resource the system "+
+			"can queu at one time in aggregate for all jobs (values over 100 are accepted)")
```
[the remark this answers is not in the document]
frrist: Yeah, I think we are still waiting for Queue documentation to land? We certainly added queuing at some point.

I gave an explication of how I _think_ this works here: https://bacalhauproject.slack.com/archives/C052BAZJVEU/p1698433427914199?thread_ts=1698419478.577109&cid=C052BAZJVEU

> The [LocalTracker](https://github.com/bacalhau-project/bacalhau/blob/d05980906ccf83a9741ff00a6fc474afafc97a50/pkg/compute/capacity/trackers.go#L31) is used by the [ExecutorBuffer](https://github.com/bacalhau-project/bacalhau/blob/d05980906ccf83a9741ff00a6fc474afafc97a50/pkg/compute/executor_buffer.go#L113) to queue work based on resources configured in the LocalTracker which receives its config from QueueResourceLimits. The decision wrt bidding is made via the [AvaliableCapacityStrategy](https://github.com/bacalhau-project/bacalhau/blob/d05980906ccf83a9741ff00a6fc474afafc97a50/pkg/bidstrategy/resource/capacity_available_strategy.go#L31). This all occurs in the compute node internals

So setting this as:
```
  QueueResourceLimits:
                CPU: "64000m"
                Memory: "64Gi"
                Disk: "1Ti"
                GPU: ""
```
means bacalhau the compute node will queue up to 64 cores, 64 gb, and/or 1 tb of Jobs before rejecting a job due to lack of capacity.

cmd/cli/config/set.go
frrist: So something like this?
```
type Parser[T any] func(...string) (T, error)

type ValueParser[T any] struct {
	parser Parser[T]
}
```
```go
	case types.StorageType:
		vp := &ValueParser[types.StorageType]{
			parser: func(values ...string) (types.StorageType, error) {
				value, err := singleValueOrError(values...)
				if err != nil {
					return 0, err
				}

				var out types.StorageType
				if err := flags.StorageTypeFlag(&out).Set(value); err != nil {
					return 0, err
				}
				return out, nil
			},
		}
```
then later after the switch:
```go
		if val, err := vp.parser(values...); err != nil {
			return err
		} else {
			viperWriter.Set(key, val)
		}
```
I think this might save us 10-15LOC at the cost of greater cognitive overhead. @rossjones how strongly do you feel about trying to shorten the `switch`? If you have a better example than than this I'd be happy to implement it, but based on what I tried this doesn't feel simpler, just shorter.

frrist: @simonwo 
> We even already have some of this available as flags, which could be adapted to work in both places

Love it, adapting to use these

cmd/cli/config/set.go
frrist: ahh wow great catch, hand't considered that case. Thankfully is easy to address

cmd/cli/config/list.go
frrist: huh, hadn't realized we needed these hooks on every command. Could we eventually move them to the root commands persistant pre run hook? Regardless will add them here.

cmd/cli/config/set.go
rossjones: > I think this might save us 10-15LOC at the cost of greater cognitive overhead.

I sorta meant something like this which saves 60+ lines and I think might be cognitively simpler if you think of it in terms of "what parser do I need for this type"?

```go
(string) (any, error)
	var parser parserFunc

	switch curValue.(type) {

	case string:
		parser = func(v string) (any, error) { return v, nil } // identity by default
	case bool:
		parser = func(v string) (any, error) { return strconv.ParseBool(v) }
	case int, int8, int16, int32, int64:
		parser = func(v string) (any, error) { return strconv.ParseInt(v, 10, 64) }
	case float32, float64:
		parser = func(v string) (any, error) { return strconv.ParseFloat(v, 64) }
	case types.Duration:
		parser = func(v string) (any, error) { return time.ParseDuration(v) }
	case model.JobSelectionDataLocality:
		parser = func(v string) (any, error) { model.ParseJobSelectionDataLocality(v) }
	case logger.LogMode:
		parser = func(v string) (any, error) { logger.ParseLogMode(v) }
	case types.StorageType:
		parser = func(v string) (any, error) { types.ParseStorageType(v) }
	default:
		return fmt.Errorf("unsupported type %T for key: %q", curValue, key)
	}

	value, err := singleValueOrError(values...)
	if err != nil {
		return fmt.Errorf("setting %q: %w", key, err)
	}

	configValue, err := parser(value)
	if err != nil {
		return fmt.Errorf("setting %q: invalid boolean value: %q: %w", key, value, err)
	}
	viperWriter.Set(key, configValue)
```

pkg/model/resource_usage_parser.go
frrist: Now we get an error:
```
frrist@oak ~/W/s/g/b/bacalhau (frrist/cmd/config)> cat ~/.bacalhau/config.yaml
node:
    compute:
        capacity:
            defaultjobresourcelimits:
                cpu: 7500m
                disk: 564 GB
                gpu: "4"
                memory: 52 GB
            jobresourcelimits:
                cpu: 7500m
                disk: 564 GB
                gpu: "4"
                memory: 52 GB
            queueresourcelimits:
                cpu: 15000m
                disk: 1.1 TB
                gpu: "4"
                memory: 103 GB
            totalresourcelimits:
                cpu: 7500m
                disk: 564 GB
                gpu: "4"
                memory: 52 GB
frrist@oak ~/W/s/g/b/bacalhau (frrist/cmd/config)> ./bin/darwin_arm64/bacalhau serve
14:43:08.663 | INF pkg/compute/capacity/system/provider.go:106 > cannot inspect system GPUs: nvidia-smi not installed. GPUs will not be used.
validating compute config: 1 error occurred:
        * total resource limits {CPU:7.5 Memory:52000000000 Disk:564000000000 GPU:4} exceed physical resources {CPU:8 Memory:54975581388 Disk:600723724697 GPU:0}
```

cmd/cli/config/set.go
frrist: This has been fixed.

cmd/cli/config/list.go
frrist: addressed.

cmd/cli/config/util.go
frrist: using this now.

cmd/cli/config/set.go
frrist: I see, I've incorporated this change.

cmd/cli/config/util.go
frrist: Removed the TODO.

cmd/cli/config/auto.go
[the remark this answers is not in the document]
rossjones: Thanks, that's really useful.  I guess it'd be equally useful for us to have this information about resource usage for compute nodes in the docs somewhere.

> Separately, it feels strange for jobs to be submitted to the network that don't specify resource requirements. 

I agree, but it can be hard for users to know/guess what their code is likely to use.  For CPUs [this blog post](https://home.robusta.dev/blog/stop-using-cpu-limits) has some interesting advice wrt limiting access to CPU in K8s.

FWIW today, the default for docker is no limits on CPU or RAM, so it will use all of the resources available to docker (or at least make them available) unless we set values.  Wasm has a 4gb limit of RAM (by virtue of being 32bit).

cmd/cli/config/list.go
simonwo: They are hooks for "client" commands only, i.e. not `serve` or `devstack`. For that reason they can't be on the root.

pkg/config/types/gen_viper/generate.go
```diff
@@ -57,21 +57,20 @@ func generateSetDefaults(t reflect.Type, prefix string, path []string, value ref
 			// Inserting the viper.SetDefault for the nested struct path.
 			constantNameForStruct := strings.ReplaceAll(newPrefix, ".", "")
 			defaultValueForStruct := "cfg." + strings.Join(newPath, ".")
-			fmt.Fprintf(writer, "\tviper.%s(%s, %s)\n", methodName, constantNameForStruct, defaultValueForStruct)
+			fmt.Fprintf(writer, "\tp.Viper.%s(%s, %s)\n", methodName, constantNameForStruct, defaultValueForStruct)
```
rossjones, line 60: is that an accidental `\t -> \tp.` ?

pkg/config/types/gen_viper/generate.go
```diff
@@ -57,21 +57,20 @@ func generateSetDefaults(t reflect.Type, prefix string, path []string, value ref
...
-			stringerMethod := fieldValue.MethodByName("String")
-			if stringerMethod.IsValid() {
-				defaultValue = "cfg." + constantPath + ".String()"
+			typeDurationToTimeDurationMethod := fieldValue.MethodByName("AsTimeDuration")
+			if typeDurationToTimeDurationMethod.IsValid() {
+				defaultValue = "cfg." + constantPath + ".AsTimeDuration()"
 			}
 
-			fmt.Fprintf(writer, "\tviper.%s(%s, %s)\n", methodName, constantName, defaultValue)
+			fmt.Fprintf(writer, "\tp.Viper.%s(%s, %s)\n", methodName, constantName, defaultValue)
```
rossjones, line 73: same q as above

rossjones (approved):

pkg/config/types/gen_viper/generate.go
frrist: This is intentional: https://github.com/bacalhau-project/bacalhau/pull/3012/files#diff-5c467aa6d981a438438edace2b47a85fc4586b732c1f0af74bb10491c5db8d77
The code generated method now accepts an optional viper instance to set values on.

pkg/config/types/gen_viper/generate.go
frrist: https://github.com/bacalhau-project/bacalhau/pull/3012#discussion_r1407922234

PR #3042, opened 2023-11-22
Linting fixes

rossjones:
Fixes a few accumulated linter errors that can make merging problematic unless you skip verification.

As we've got nearly 20 counts of gocyclo (high cyclomatic complexity), and 20+ funlen (long  functions) we might consider taking some time to address them.  I also run a default revive (another linter) against the codebase, and it suggest there are some changes we could make.

```
❯ revive ./... | wc -l
    2690
```

### DIFF
--- a/pkg/s3/middleware/drop_accept_encoding.go
+++ b/pkg/s3/middleware/drop_accept_encoding.go
@@ -29,6 +29,8 @@ const (
 // do not support the Accept-Encoding header.
 // It does this by dropping the Accept-Encoding header before the request is signed, and restoring it after the request is signed.
 // https://stackoverflow.com/questions/73717477/gcp-cloud-storage-golang-aws-sdk2-upload-file-with-s3-interoperability-creds/74382598#74382598
+//
+//nolint:lll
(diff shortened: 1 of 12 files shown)


pkg/s3/middleware/drop_accept_encoding.go
```diff
@@ -53,6 +55,8 @@ func DropAcceptEncoding(o *s3.Options) {
 
 // dropAcceptEncodingHeader is a middleware function that removes the Accept-Encoding header from the request.
 // This is necessary for compatibility with certain S3-like storage providers.
+//
+//nolint:lll
```
simonwo, line 59: Can we just split this across multiple lines rather than turning off the lint?

simonwo (approved):

pkg/s3/middleware/drop_accept_encoding.go
rossjones: We can. I tried before adding it but obviously broke the line at the wrong point and Go was particularly fussy about the line-break.  It does work though when you break after the opening ( of the function parameters.  Also replaced the long stackoverflow link with a short version

PR #3044, opened 2023-11-23
WebUI with ability to spin up using the serve command

XandraMcC:
This PR contains both the code for the WebUI (built with React & Typescript) and the code to spin up the dashboard with the `--web-ui` config flag. All React code is located in `./webui`.

Command & response to spin up dashboard with `serve` command an `--web-ui` flag:
Screenshot 2023-11-23 at 11 59 57

Please see screenshots below for a view of what the WebUI looks like:
Screenshot 2023-11-23 at 12 00 28
Screenshot 2023-11-23 at 12 00 37
Screenshot 2023-11-23 at 12 00 45


simonwo (commented):
So I'm happy that this is ready to be merged now, but won't approve it because I've contributed some of the code and it seems improper somehow.

There are a few issues we should be dealing with in future PRs:

- The jobs list makes a lot of assumptions: for a start it only really covers Docker jobs, and not jobs of any other types.
- The data structures we are using also assume that every job is a Docker job. 
- The job list shows a job name for batch jobs, which isn't a thing, IIRC we removed this in the actual design.
- There are lots of things that we aren't matching from the design, we seem to have invented new columns that weren't there previously.
- Job labels don't seem to show up.
- We shouldn't be maintaining our own copies of the API data strucutres – we already generate these using swagger for the Python SDK, and should be able to do the same thing here. 
- For that matter, we should extract out the Bacalhau API calls into a proper Javascript SDK, and then just reference that in our webui...

Makefile
```diff
@@ -169,6 +169,22 @@ build-ci: build-bacalhau install-plugins
...
+
+.PHONY: build-webui
+build-webui: ${WEB_BUILD_FILES}
+
+webui/build:
+	mkdir -p $@ && touch $@/.keep
+
+$(WEB_BUILD_FILES): $(WEB_SRC_FILES) webui/build
+	cd webui && npm run build
+
```
rossjones, line 187: There's no `npm install` step here, so on a fresh build this will fail to compile.

```
Makefile:186: warning: overriding commands for target `webui/build'
Makefile:183: warning: ignoring old commands for target `webui/build'
make: Circular webui/build <- webui/build dependency dropped.
cd webui && npm run build

> webui@0.1.0 build
> react-scripts build

sh: react-scripts: command not found
make: *** [webui/build] Error 127
```

Annoyingly `npm install` always seems to go to the internet, so it would be good to have if `node_modules` doesn't exist or the `package.json` has changes, then build.

Perhaps for now having a check for it the node_modules exists and if not, running `npm install`? That way we only have to do it once, and can delete the folder to rebuild if we add new modules.

cmd/cli/serve/serve.go
```diff
@@ -173,6 +178,11 @@ func serve(cmd *cobra.Command) error {
 		return err
 	}
 
+	startWebUI, err := config.Get[bool](types.NodeWebUI)
+	if err != nil {
+		return err
+	}
```
rossjones, line 184: This is okay for now, but in future we'll probably want the web ui to be available on a configured port instead of the default port (80).  So it might be worth writing a ticket to capture this need in a future change?

webui/public/robots.txt
```diff
@@ -0,0 +1,3 @@
+# https://www.robotstxt.org/robotstxt.html
+User-agent: *
+Disallow:
```
rossjones, line 3: Just to double-check this was intended.

This gives complete access to robots, whereas `Disallow: /` will block all access.

webui/webui.go
```diff
@@ -0,0 +1,79 @@
...
+			// If nothing matches, just serve the index page. This allows the
+			// app to modify the browser URL to refer to client side "pages"
+			// (e.g. `/jobs`) which the user can then successfully refresh,
+			// share or bookmark. If this wasn't here, the user would have to
+			// enter the app manually at `/` every time.
+			w.Write(indexPage)
+		}
+	})
+
+	server := &http.Server{
```
rossjones, line 69: With no setting for `Addr` here it will always be on port 80.  This is fine for now, but I suspect we'll need a little more flexibility later on (for instance when we support TLS for this) and so it's worth adding a ticket for a future change to add this (via the --web-ui parameter perhaps?).

rossjones (approved):
Looks good! Couple of comments around the port it listens on (just a ticket required) but the biggest issue is that it breaks when building on a clean install.  If we can fix that up I think we're good to go.

cmd/cli/serve/serve.go
XandraMcC: Agreed, will make a ticket for this

webui/public/robots.txt
XandraMcC: Thanks, added `/`

webui/webui.go
XandraMcC: Will add this to configuration ticket (same as https://github.com/bacalhau-project/bacalhau/pull/3044#discussion_r1404243551)

XandraMcC: https://github.com/bacalhau-project/expanso-planning/issues/133

cmd/cli/serve/serve.go
XandraMcC: https://github.com/bacalhau-project/expanso-planning/issues/133

PR #3051, opened 2023-11-27
Temporarily hide "View" button until Job Detail and Node Detail pages are available

XandraMcC:


webui/src/services/bacalhau.ts
```diff
@@ -6,7 +6,6 @@ class BacalhauAPI {
   apiClient: AxiosInstance
 
   constructor(baseURL: string) {
-    console.log(baseURL);
```
simonwo, line 9: 😅

simonwo (approved):

PR #3053, opened 2023-11-28
Tell StorageProviders where to store data

rossjones:
Currently storage providers create a temporary folder underneath the system configured storage path, where the folder include the provider's name.  This means all of the data is indexed on the provider, rather than the execution. Most of the StorageProvider implementations determine these paths when they are first created.

This PR switches to telling the provider where to store the data, so that it is keyed by job/execution instead (also within the configured data folder). This provides a single location for all inputs affecting an execution (or all local executions for a Job).


pkg/compute/executor.go
```diff
@@ -216,7 +221,13 @@ func (e *BaseExecutor) Start(ctx context.Context, execution *models.Execution) (
 		return
 	}
 
-	args, cleanup, err := PrepareRunArguments(ctx, e.Storages, execution, resultFolder)
+	executionStorage := filepath.Join(e.storageDirectory, fmt.Sprintf("%s/%s", execution.JobID, execution.ID))
```
simonwo, line 224: Why not just `filepath.Join(e.StorageDirectory, execution.JobID, execution.ID)` ?

pkg/compute/executor.go
```diff
@@ -216,7 +221,13 @@ func (e *BaseExecutor) Start(ctx context.Context, execution *models.Execution) (
...
+	if err := os.MkdirAll(executionStorage, os.ModePerm); err != nil {
```
simonwo, line 225: This sets 777 permissions, is that what we want? Would we prefer 755 or 500?

pkg/executor/util/utils.go
```diff
@@ -84,18 +81,6 @@ func NewStandardStorageProvider(
...
-	if err != nil {
-		return nil, err
-	}
-
-	cm.RegisterCallback(func() error {
-		if err := os.RemoveAll(dir); err != nil {
-			return fmt.Errorf("unable to clean up S3 storage directory: %w", err)
-		}
-		return nil
-	})
```
simonwo, line 97: So where does cleanup happen now?

simonwo (approved):

frrist (approved):

pkg/executor/util/utils.go
rossjones: Other than tests, PrepareStorage is only ever called from ParallelPrepareStorage, and ParallelPrepareStorage is only used by the base executor (and the wasm executor) and each time returns a function that will call ParallelCleanup.  This is a step in the direction of us cleaning up the resources when they're no longer used so that storage is responsible for fetching the data, but the user is responsible for cleaning it up.

pkg/compute/executor.go
rossjones: Have set to 0755 and a local constant.  I've left the exec bit for now, meaning users can download a binary and run it if they wish (within the container theyre in).  Is this something we want to allow?

PR #3061, opened 2023-12-01
Support multiple AMD and Nvidia GPUs for Docker jobs

### LINKED ISSUE #1665: Bacalhau only supports the use of 1 GPU
Only 1 GPU can be used at a time: https://github.com/filecoin-project/bacalhau/blob/dev/pkg/executor/docker/executor.go#L225
The [Docker API expects](https://docs.docker.com/config/containers/resource_constraints/#expose-gpus-for-use) you to specify the exact GPU you want to run on by ID or by device number.

At the time of writing this code I didn’t know how to or whether we wanted to support maintaining the state of what GPUs are in use by that particular node. Maybe there’s a way now, but that was considered future work. Hence the todo.

In summary, only GPU 0 can be used and the docs are correct.

In the future, we need to support both running multiple jobs on multiple GPUs and single jobs on multiple GPUs. But this is very tied to how GPU jobs are specified and state is maintained so there's probably some dependent work that is required first. See the main GPU improvement issue for this.

simonwo:
This PR adds two major features:

1. Support for AMD GPUs (previously we only supported Nvidia)
2. Support for multiple GPUs per job (previously we only ever attached 1 GPU, even if multiple were requested)

This has been facilitated by having access to a server with some AMD GPUs for testing, with thanks to @JvD007.

Resolves #1665.


pkg/compute/capacity/system/gpu/amd_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	gpus := resources.GPUs
+	require.Len(t, gpus, 4)
+
+	for index, gpu := range gpus {
+		require.Equal(t, models.GPUVendorAMDATI, gpu.Vendor)
+		require.Equal(t, uint64(index), gpu.Index)
+		require.Equal(t, "Instinct MI210", gpu.Name)
+		require.Equal(t, uint64(65520), gpu.Memory)
+	}
```
rossjones, line 69: Because all of the cards are identical, how do we know that it didn't just parse the first card entry into all of the positions in the array? Might be best to vary say gpu.Memory to be related to the index so that we know it's parsed the right cards into the right slots.

pkg/models/resource.go
```diff
@@ -195,6 +224,9 @@ func (r *Resources) Merge(other Resources) *Resources {
 	if newR.GPU <= 0 {
 		newR.GPU = other.GPU
 	}
+	if len(newR.GPUs) <= 0 {
```
rossjones, line 227: I was about to comment that len(array) is unlikely to ever be less than 0, but then I fell down the rabbit hole of "why does Go return signed values for len()", so I guess this is fine.

rossjones (approved):
Looks great, guess we also need doc updates to the places where we say "nvidia only"?

pkg/compute/capacity/system/gpu/amd_test.go
JvD007: I can add an other AMD GPU to the server, a MI100 GPU. The server is part of a Composable Infrastructure, if you want I can add more GPU's to server. Let me know

PR #3065, opened 2023-12-04
More gracefully handle bad tool output

simonwo:
It turns out that `rocm-smi` will print non-JSON output but still return a zero error code in the case of an error – naughty! This has prompted us to be more uniform with how we handle tool errors – any error generated by a GPU detector tool will be ignored, rather than errors that they explicitly flag. This will allow the node to come online with zero GPUs in this case, and the reason for this will be in the log.

### DIFF
--- a/pkg/compute/capacity/tool.go
+++ b/pkg/compute/capacity/tool.go
@@ -3,12 +3,11 @@ package capacity
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io"
 	"os/exec"
 
 	"github.com/bacalhau-project/bacalhau/pkg/models"
-	"github.com/rs/zerolog/log"
+	"github.com/pkg/errors"
 )
 
 // ToolBasedProvider will run an external tool and parse the results in a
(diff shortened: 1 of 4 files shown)


rossjones (approved):

pkg/compute/capacity/tool.go
```diff
@@ -27,28 +26,22 @@ type ToolBasedProvider struct {
 func (tool *ToolBasedProvider) GetAvailableCapacity(ctx context.Context) (models.Resources, error) {
 	toolPath, err := exec.LookPath(tool.Command)
 	if err != nil {
-		// If the tool is not installed, we can't know the number of GPUs.
-		// It is not an error to assume zero.
-		log.Ctx(ctx).Info().Msgf("cannot inspect system %s: %s not installed. %s will not be used.", tool.Provides, tool.Command, tool.Provides)
-		return models.Resources{}, nil
+		// If the tool is not installed, we can't know the available resources.
+		return models.Resources{}, errors.Wrapf(err, "tool %q is not installed or not on PATH", tool.Command)
```
frrist, line 30: In the future it might be a good practice to be consistent around our usage of `fmt.Errorf("%w", err)` vs `errors.Wrap`. At the moment we are using both, but not a big deal.

frrist (commented):
LGTM (this already merged but left some comments anyways)

PR #3066, opened 2023-12-04
Webui job detail

XandraMcC:
Following PR has new Job Detail page which is accessed through the Jobs Dashboard page via the "View" button.
- This page can also be accessed via: http://localhost:3000/JobDetail/{JobID}
- This page contains 2 components: `CliView` and `JobInfo` which help represent the data
- This page also makes use of the `Container`, `ActionButton`, and `Table` global components
- The page loads both the `describeJob()` & `jobExecution()`

### One Execution:
One Execution

### Multiple Executions:
Multiple Executions


simonwo (commented):
Looks good, although it's a little difficult to spot the changed things amongst all the linter fixes – in future it would be better to separate out linter fixes into a seperate PR. No matter!

Some comments:
* can you render the timeout(s) using a Moment duration element to get a human readable duration – 1800 is in seconds, and Moment will render it automatically as e.g. 30 minutes.
* Same for the created/modified dates, and any or all dates actually.
* What does the `Image` row display when the executor is not Docker?
* The `GPU` field, despite being a string, will normally just be a number. So we'll just get `GPU Details: 1` which doesn't look quite right? Can you test with a GPU job plz?
* TimeoutDeadline and Timeout seem to print the same thing?
* What happens if the STDOUT or STDERR is very long? There is a cap of 2000 lines. 
* Why don't we display the executions table all the time?

XandraMcC:
> * can you render the timeout(s) using a Moment duration element to get a human readable duration – 1800 is in seconds, and Moment will render it automatically as e.g. 30 minutes.

Done

> * Same for the created/modified dates, and any or all dates actually.

Done

> * What does the `Image` row display when the executor is not Docker?

`{ label: "Image", value: job.Tasks[0].Engine.Params.Image },` returns a string, so will return an empty string if no image info in API. Not sure what it should return if not a Docker Image.

> * The `GPU` field, despite being a string, will normally just be a number. So we'll just get `GPU Details: 1` which doesn't look quite right? Can you test with a GPU job plz?

Wilco

> * TimeoutDeadline and Timeout seem to print the same thing?

Timeout deadline was in the design but I cannot find it in the API?

> * What happens if the STDOUT or STDERR is very long? There is a cap of 2000 lines.

Done, added `limitLines`: 
Screenshot 2023-12-05 at 12 09 05

> * Why don't we display the executions table all the time?

If there is only 1 execution, this info is displayed in the Execution Record, Standard Output, and Standard Error components with no need to choose between them via the table

simonwo (approved):

PR #3067, opened 2023-12-04
Webui fix job labels to display properly

olgibbons:
Currently the web labels are not displaying in their column on the dashboard. This is chiefly because 'bacalhau docker run -l myLabel....' will assign the label as a key with an empty value. 

YAML files can pass both a key and value for the label, eg: {OS: MacOS}. To accommodate both scenarios, the code pass the key in the event of an empty string, and a value in the case of a key/value pair.

We might want to look at how we pass labels to the API so that it has consistent behaviour for all scenarios.

Fixes  https://github.com/bacalhau-project/expanso-planning/issues/135


webui/src/pages/JobsDashboard/JobsTable/JobsTable.tsx
```diff
@@ -40,13 +40,25 @@ function parseData(jobs: Job[]): ParsedJobData[] {
...
 }
 
+function createLabelArray(label: {[key: string]: string}): string[] {
+  const labelArray: string[] = []
+  for (const [key, value] of Object.entries(label)) {
+    if (value === "") {
+      labelArray.push(key)
+    } else {
+      labelArray.push(value)
+    }
```
simonwo, line 57: If the label is `{key: value}` then the key is significant. E.g. if the label is `{least_favourite_food: banana}` it would be bad to just print `banana`. So come up with some way to include both the key and the value when the key is not blank.

webui/src/pages/JobsDashboard/JobsTable/JobsTable.tsx
```diff
@@ -76,7 +88,7 @@ const JobsTable: React.FC<TableProps> = ({ data }) => {
                 <ProgramSummary data={jobData.tasks} />
               </td>
               <td className={styles.jobType}>{jobData.jobType}</td>
-              <td className={styles.label}>{jobData.label}</td>
+              <td className={styles.label}>{jobData.label.map((label) => <ul>{label}</ul>)}</td>
```
simonwo, line 91: This is a weird use of <ul>... do you want to just use a <span>?

simonwo (commented):
Generally LGTM!

webui/src/pages/JobsDashboard/JobsTable/JobsTable.tsx
olgibbons: sure thing!

simonwo:
Final comment: have you tested with multiple labels? It looks like they stack on top of each other, whereas they should probably sit side-by-side.

image

You'll need to replace `display:  flex; justify-content: center` with just `display: inline-block` in the Labels scss.

simonwo (approved):

PR #3069, opened 2023-12-05
Bulk update OpenTelemetry dependencies.

simonwo:
To fix issues:

- https://github.com/bacalhau-project/bacalhau/security/dependabot/116
- https://github.com/bacalhau-project/bacalhau/security/dependabot/117
- https://github.com/bacalhau-project/bacalhau/security/dependabot/118


pkg/telemetry/logging_trace.go
```diff
@@ -7,9 +7,11 @@ import (
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/otel/baggage"
 	oteltrace "go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/embedded"
 )
 
 type loggingTracerProvider struct {
+	embedded.TracerProvider
```
frrist, line 14: I assume this is the reason for this change? Or is it something else?

> Implementers of the [OpenTelemetry trace API](https://pkg.go.dev/go.opentelemetry.io/otel/trace) can embed the relevant type from this package into their implementation directly. Doing so will result in a compilation error for users when the [OpenTelemetry trace API](https://pkg.go.dev/go.opentelemetry.io/otel/trace) is extended (which is something that can happen without a major version bump of the API package).

- https://pkg.go.dev/go.opentelemetry.io/otel/trace/embedded

simonwo: Yes. I don't quite understand what this achieves or how it works, but doing this seems to be the intended solution.

PR #3072, opened 2023-12-06
fix: correctly parse resource config

frrist:
This change modifies `models.ResourceConfig` to use `go-humanize` instead of `datasize` for more accurate parsing of human-readable strings into bytes. `go-humanize` can parse mebibyte strings and megabytes strings (ditto that for the rest of the SI units). Also updated the 'String' method of 'ResourceConfig' to reflect these changes.

I noticed this issue when `bacalhau config auto-resources` produced a config that failed to parse. The bacalhau config system uses `go-humanize` for parsing values, and had modified `model.ResourceConfig` to do the same. When https://github.com/bacalhau-project/bacalhau/pull/3061 landed it replaced usage of `model.ResourceConfig` with `models.ResourceConfig` which broke the behavior implemented in https://github.com/bacalhau-project/bacalhau/pull/3012. The change resolves this breakage.

Additionally, when making this change I noticed the `models.ResourceConfig.Normalize()` method erroneously converted Mebibytes (Mi/MiB) to Megabyte (MB) when parsing the string representation. The `datasize` package doesn't distinguish these types. The `model.ResourceConfig` implementation did not have this behavior. `go-humanize` does distinguish between these types. I have removed the incorrect behavior from `models.ResourceConfig.Normalize()` and update the test for it.

### DIFF
--- a/pkg/executor/docker/executor_test.go
+++ b/pkg/executor/docker/executor_test.go
@@ -5,6 +5,7 @@ package docker
 import (
 	"context"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


pkg/models/resources_test.go
```diff
@@ -0,0 +1,30 @@
...
+		{"42MB", 42000000},  // 42 Megabytes
+		{"42MiB", 44040192}, // 42 Mebibyte
+		{"42mb", 42000000},  // 42 Megabytes
+		{"42mib", 44040192}, // 42 Mebibyte
+	}
+
+	for _, p := range tests {
+		cfg, err := NewResourcesConfigBuilder().Memory(p.in).Build()
+		require.NoError(t, err)
+		// TODO I think this is where the bug occurs
```
frrist, line 24: See previous commit https://github.com/bacalhau-project/bacalhau/pull/3072/commits/f6f2f5a274c16f8cd8ec5bdd2debbc33ccb29124

pkg/executor/docker/executor_test.go
```diff
@@ -237,26 +244,39 @@ func (s *ExecutorTestSuite) TestDockerResourceLimitsMemory() {
 		s.T().Skip("Resource limits don't apply to containers running on Windows")
 	}
 
-	// this will give us a numerator and denominator that should end up at the
-	// same 0.1 value that 100m means
-	// https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_monitoring_and_updating_the_kernel/using-cgroups-v2-to-control-distribution-of-cpu-time-for-applications_managing-monitoring-and-updating-the-kernel#proc_controlling-distribution-of-cpu-time-for-applications-by-adjusting-cpu-bandwidth_using-cgroups-v2-to-control-distribution-of-cpu-time-for-applications
```
frrist, line 242: This is only relevant to the CPU test case above. Removed from here (it appears to have been copied)

PR #3075, opened 2023-12-07
Webui settings

XandraMcC:


webui/src/context/TableSettingsContext.tsx
```diff
@@ -1,41 +1,55 @@
...
+  showStatus?: boolean;
+  // Nodes Table
+  showNodeId?: boolean;
+  showNodeName?: boolean;
+  showNodeType?: boolean;
+  showEnv?: boolean;
+  showInputs?: boolean;
+  showOutputs?: boolean;
+  showVersion?: boolean;
+  showAction?: boolean;
```
simonwo, line 21: We added customisable columns so that we could get more node details available in the table. Currently we're missing labels and capacity. Could you (add a new ticket to) add those in plz?

webui/src/pages/Settings/Settings.tsx
```diff
@@ -26,51 +28,58 @@ const Settings = () => {
...
 
+  const jobsOptions: { label: string; key: keyof TableSettings }[] = [
+    { label: "Job ID", key: "showJobId" },
+    { label: "Name", key: "showJobName" },
+    { label: "Created", key: "showCreated" },
+    { label: "Program", key: "showProgram" },
+    { label: "Job Type", key: "showJobType" },
+    { label: "Label", key: "showLabel" },
+    { label: "Status", key: "showStatus" },
+  ];
```
simonwo, line 39: Can we use a similar technique to avoid hard-coding the `TableSettingsContext` i.e. not have to list each config as a separate variable?

simonwo (approved):

PR #3076, opened 2023-12-07
Adds a very early `bacalhau exec` command

rossjones:
Adds an exec command which can be used for submitting custom job types. Because these custom job types could be anything, then we have to take steps to cater for lots of potential edge cases.

This initial version supports duckdb and python, with both templates currently being embedded in the binary. Ideally we should retrieve these from somewhere, either on disk or from a requester node (which can serve the templates based on the translators installed).

Although this PR will submit a job specification, we expect it to fail as there are not yet any translators installed to turn them into docker or wasm jobs.

### Implementation

The command after exec is used as the Command: value in the Engine Params.  Remaining parameters that are not bacalhau parameters are passed as the Engine.Params["Arguments"] string array.

If the user wishes to pass any parameters that clash with bacalhau (say -i) they must do so after using --

Some commands will need to have code to run.  To support this we have added --code which allows the user to specify a single file or a folder of files.  As long as the total size is less than 10Mb the contents will be added as an inline storage to the job specification to be mounted at /code with an alias of 'code'.

### Example commands 

`bacalhau exec -i src=...,dst=... duckdb "select * from /inputs/data.csv"`

Sends a request where Command: duckdb, Arguments: ["select * from /inputs/data.csv"].

`bacalhau exec python app.py -x -y --code=.`

Sends a request where Command: python, Arguments: ["app.py", "-x", "-y"] and where . is added as an inline tgz.


cmd/cli/exec/templates.go
```diff
@@ -0,0 +1,69 @@
...
+	    "Type":"python",
+		"Params":{
+			"Version": "{{or (index . "version") "3.11"}}"
+		}
+	}
+ }]
+}`
+
+	return strings.TrimSpace(t)
+}
```
simonwo, line 51: Could we do these as separate files (`templates/python.json`) and then `go:embed` them? Writing a map builder that just consumes the contents of an `fs.FS` (starting with an `embed.FS`) feels like a natural way to add support for templates in the repo later.

simonwo (approved):
Looks cool!

cmd/cli/exec/exec.go
```diff
@@ -0,0 +1,194 @@
...
+		Execute a specific job type.
+`))
+
+	//nolint:lll // Documentation
+	getExample = templates.Examples(i18n.T(`
+		# Execute the app.py script with Python
+		bacalhau exec python app.py
+
+		# Run a duckdb query against a CSV file
+		bacalhau -i src=...,dst=/inputs/data.csv duckdb "select * from /inputs/data.csv"
```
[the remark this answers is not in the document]
rossjones: Fixed in 799b33c9e

cmd/cli/exec/exec.go
```diff
@@ -0,0 +1,194 @@
...
+func PrepareJob(cmd *cobra.Command, cmdArgs []string, unknownArgs []string, options *ExecOptions) (*models.Job, error) {
+	var err error
+	var jobType, templateString string
+	var job *models.Job
+
+	// Determine the job type and lookup the template for that type. If we
+	// don't have a template, then we don't know how to submit that job type.
+	jobType = cmdArgs[0]
+	if templateString, err = Template(jobType); err != nil {
+		return nil, fmt.Errorf("the job type '%s' is not supported", jobType)
```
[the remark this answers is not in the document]
rossjones: Good catch, thanks. Fixed in 799b33c9e

cmd/cli/exec/templates.go
rossjones: Implemented in 799b33c9e
We can replace the loader with something that implements fs.ReadDirFS if we want to load these from a configured location in future (assuming also we have some way of installing them).

rossjones: Templated JSON turns out to not be valid JSON, and it isn't possibly to change the templates to use different '.  Renamed to tpl to avoid lint-issues but can rename back later

wdbaruni:
I have a minor comment about the CLI name. All recent CLIs follow a `bacalhau <resource> <command>` pattern, such as `bacalhau job list`, `bacalhau job run`, `bacalhau node describe`, ... I am also planning to introduce new CLI resources in the future, specifically commands to interact with executions that will follow the pattern `bacalhau execution <commadn>`.

I don't have a clear suggestion, but I am hoping if we can do something similar here. If not, then my main concern is that this command might conflict with `bacalhau execution`

rossjones:
I'm not sure there's an obvious resource to name this after, at least there isn't one that's obvious to me.  I'm more than happy to rename it once we've an idea of what we want to call it, and ideally before I get around to documenting it at the end of the other related work.  I've added something to the agenda for tomorrow's call to discuss with everyone's input.

PR #3089, opened 2023-12-11
Implement new and improved terraform-base deployment

frrist:
Documentation: https://www.notion.so/expanso/Demo-Marketplace-Deployment-10136faf7b424425bc74acbf647b9606


simonwo (commented):
This broadly looks good to me... will this mean we have new two terraform folders now? `terraform` and `tf`? Are we aiming to replace the former?

frrist:
> will this mean we have new two terraform folders now? terraform and tf? 

I think it depends. The features we have in _our_ deployments to development/staging/production are specific to our needs as operators of a public network. e.g. limiting the HTTP endpoints jobs can query when executed by a compute node, promtail agent, ipfs running as a separate service, per-defined IP addresses (for DNS), hard-coded network size, lack of auth, etc. 

The terraform in `tf` which will be used for marketplace deployments doesn't/cannot have some of these custom things. 

I think the ideal scenario would be us using the same code for our networks as users use for MP deployments, perhaps we can merge these things together in the future? WDYT?

simonwo (approved):
Yes, I think in the future it would be nice to use one as a base for the other. But we can work forward to that from here.

Maybe a more descriptive folder name would be good? I.e. `marketplace` or something?

frrist:
> Maybe a more descriptive folder name would be good? I.e. marketplace or something?

Done. merging once CI passes.

PR #3092, opened 2023-12-12
Remove building Bacalhau as part of Docker image build

simonwo:
This meant that our new Web UI build steps were not being picked up. This is not really necessary as we can just reuse the output that was created from our CI build step, rather than the rigmarole of trying to capture the right files and recreate the build process within the Dockerfile.

This change also means we can build the docker image on PRs as well, to check that it works.

There are some other refactors here – we now build and test as separate jobs so that we don't need to wait for tests to pass before pushing binaries to a release, and also generate all of our binaries from a single cross-compiling step (because we don't use CGO and hence can compile for all platforms from anywhere).

As a result of this refactor, the build time in CI is reduced from 17m to 14m... every little helps, I guess.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ BINARY_NAME := ${BINARY_NAME}.exe
 CC = gcc.exe
 endif
 
-BINARY_PATH = bin/${GOOS}_${GOARCH}${GOARM}/${BINARY_NAME}
+BINARY_PATH = bin/${GOOS}/${GOARCH}${GOARM}/${BINARY_NAME}
 
 TAG ?= $(eval TAG := $(shell git describe --tags --always))$(TAG)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -192,7 +192,7 @@ $(WEB_INSTALL_GUARD): webui/package.json
 	cd webui && npm install
 
 export GENERATE_SOURCEMAP := false
-$(WEB_BUILD_FILES): $(WEB_SRC_FILES) $(WEB_INSTALL_GUARD)
+${WEB_BUILD_FILES} &: $(WEB_SRC_FILES) $(WEB_INSTALL_GUARD)
```
rossjones, line 195: The &: was a good find!

rossjones (approved):

Makefile
simonwo: Yeah, I thought `make` did this by default, but it turns out it doesn't! TIL!

PR #3093, opened 2023-12-12
Custom job type docker images

rossjones:
This PR contains images for running the first two custom job types, python and duckdb.  The images currently build and the python tests work well (to test the launcher process), but they may require changes in future in response to the job translation work.

The duckdb image is reasonably straight-forward, it downloads the appropriate version of duckdb (for the target platform) and makes it available for use.

The python image is slightly more complex in that it expects code to be mounted at /code, and a launcher process (launcher.py) is responsible for moving the core, determining where the requirements are defined, and then executing the program using the arguments provided by the CLI on submission.

May require further changes in response to the job translation (and will require network access for requirements fetching).


docker/custom-job-images/python/test/multi-file-poetry/README.md
simonwo, line 1: Empty?

docker/custom-job-images/Makefile
```diff
@@ -0,0 +1,25 @@
...
+duckdb: 
+	@$(MAKE) -C duckdb build 
+
+duckdb-local: 
+	@$(MAKE) -C duckdb local	
+
+.PHONY: local python duckdb
+
+
+python-test:
```
simonwo, line 22: Should we br running this as part of CI?

simonwo (approved):

docker/custom-job-images/Makefile
rossjones: We could, but we'll also need to look at how we push the images.  At the moment it requires a manual push to docker hub.

docker/custom-job-images/python/test/multi-file-poetry/README.md
rossjones: Helpfully added by `poetry new`. Now removed.

Issue #138, opened 2022-06-08
122. Update scale testing framework


lukemarsden:
With new tests for the cases in this phase.

PR #3098, opened 2023-12-13
Webui fix job name and

olgibbons:
Job name and ID were often showing duplicate information
depending on the job type. The new 'job' column will
just show the shortened ID in the case of a batch job,
and the name for non-batch jobs.

The Node name column has been removed, as nodes are
currently unnamed.

Resolves issue #138


olgibbons:
I think there are some references to the node name that need to be removed

webui/src/pages/JobsDashboard/JobsTable/JobsTable.tsx
```diff
@@ -32,12 +33,18 @@ function parseData(jobs: Job[]): ParsedJobData[] {
...
+    const jobName = job.Name;
 
+    if (jobType === "batch") {
+      job.Name = jobShortID;
+    } else {
+      job.Name = jobName;
+    }
     return {
-      id: getShortenedJobID(job.ID),
+      id: job.ID,
```
simonwo, line 47: What is the `id` field used for now?

olgibbons: I've removed this now

simonwo (approved):

PR #3102, opened 2023-12-13
refactor: allow node info store ttl configuration

### LINKED ISSUE #3083: Node evicition policy cannot be configured in requester node store
The TTL is hard-coded to 10 Mins https://github.com/bacalhau-project/bacalhau/blob/main/pkg/node/node.go#L184
This may be a sane default for some use cases, but probably not all. We should allow this value to be configured by operators.

frrist:
- closes #3083
- `bacalhau serve` accepts a flag called `--node-info-store-ttl` which expects a duration and is used to configure NodeInforStore TTL. The default value remains 10mins. 

I am a bit torn on this PR, on one hand we can configure the TTL NodeInfoStore - something we need today. On the other hand this feels like its adding some debt to configure structures we'd like to move away from. Additionally, I'd assume the NodeInfoStore will need a more robust configuration than what's provided here, e.g. in the event it becomes a persisted store. Feedback welcome on the addition of new config fields.


pkg/devstack/devstack.go
```diff
@@ -106,9 +106,9 @@ func Setup(
 		isComputeNode := (totalNodeCount - i) <= computeNodeCount
 		log.Ctx(ctx).Debug().Msgf(`Creating Node #%d as {RequesterNode: %t, ComputeNode: %t}`, i+1, isRequesterNode, isComputeNode)
 
-		//////////////////////////////////////
+		// ////////////////////////////////////
 		// IPFS
-		//////////////////////////////////////
+		// ////////////////////////////////////
```
rossjones, line 111: weird whitespace?

rossjones (approved):
LGTM. It's a positive move in a direction that gives us more control, we can address issues with any change in cluster management once that'd decided.

PR #3103, opened 2023-12-13
refactor: remove unused WebUI field from NodeConfig

frrist:
Its not used so nixing it.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -5,11 +5,6 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


simonwo (approved):
I think we're about to add back in some Web UI config to control the port, but probably makes sense to do that in a dedicated struct anyway, so +1 to this.

rossjones:
I've removed it over in #3132 although there is some strange behaviour with the webui configflags :|

simonwo:
Made redundant by #3132.

PR #3108, opened 2023-12-14
Translation converting custom task types to docker tasks

rossjones:
Adds a translation layer to the orchestrator, allowing for custom job types to be converted to known job types (docker/wasm) for execution on a compute node. The translation most consists of specifying the docker image to use, putting the engine parameters in the correct place, and adding metadata to the newly created job. 

This initial PR only support `python` and `duckdb` with special support for installing python dependencies implemented within the image. 

By default translations in the requester node are disabled, but can be enabled with `--requester-job-translation-enabled`


cmd/cli/exec/exec.go
```diff
@@ -120,6 +121,14 @@ func PrepareJob(cmd *cobra.Command, cmdArgs []string, unknownArgs []string, opti
...
 	jobType = cmdArgs[0]
 
+	for i := range cmdArgs {
+		// If any parameters were quoted, we should make sure we try and add
+		// them back in after they were stripped for us.
+		if strings.Contains(cmdArgs[i], " ") {
+			cmdArgs[i] = fmt.Sprintf(`"%s"`, cmdArgs[i])
+		}
+	}
+
```
simonwo, line 131: What if `cmdArgs[...]` was `'hello "world"!'` i.e. single quotes.

I feel like trying to handle shell escaping is going to be painful. Can we just ensure these parameters are passed to the program directly without any further shell interpretation?

pkg/orchestrator/endpoint.go
```diff
@@ -60,6 +64,30 @@ func (e *BaseEndpoint) SubmitJob(ctx context.Context, request *SubmitJobRequest)
...
+		return nil, err
+	}
+
+	// If we have translated the job (i.e. at least one task was translated) then we will switch
+	// to using the translated job after we have saved it in the jobstore.
+	//
+	// TODO: We want to return to the user the JobID of the original job, not the translated job.
+	// But this means the evaluation will not match as we only have an evaluation for the translated
+	// job. We need to think about how we can handle this or whether we have to move the translation
+	// to the compute node (where it will end up being run anyway).
```
simonwo, line 81: So what's the impact of the TODO today? It sounds like a `described` translated job will have no evaluations, executions or results?

Can we introduce job versioning? And the translated job can just be a subsequent version of the original, under the same ID? 

I would be in favour of porque no los dos, where compute node can advertise custom job types if it has them and the requester can just pass them along, but also the requester can perform the translation which is easier to manage in a big cluster.

pkg/translation/translators/python.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	builder := original.
+		ToBuilder().
+		Meta(models.MetaTranslatedBy, "translators/python").
+		Engine(dkrSpec)
+
+	original.Network = &models.NetworkConfig{
+		Type:    models.NetworkFull,
+		Domains: PythonPackageDomains,
```
simonwo, line 48: The `Domains` flag has no effect on `NetworkFull`. Do you actually want `NetworkHTTP`?

simonwo (commented):
Is there anything in `/docs` to be updated? 🙏🏻 

I could do with more comments somewhere on the overall flow of jobs and the externally visible effects of job translation, i.e. 

- is the translated job the same ID as the original job?
- what does calling describe return (today and intended)?
- what does calling get on a completed translated job return (today and intended)?
- etc

pkg/translation/translators/python.go
rossjones: I do, but I've been unable to get it to work with devstack, and I meant to switch this back and then investigate.  

Trying a `docker run` with a task that makes HTTP requests seems to work okay with NetworkHTTP, but when I run my task it fails - but running my task with NetworkFull succeeds.

I wonder if it is the gateway image that might be outdated?

pkg/orchestrator/endpoint.go
rossjones: Yes, the original (exec python) job will have no evaluations, or execution or results, it exists solely to act as a record for the newly translated (docker run python) job.  The translated job contains a metadata reference back to the original job, but this is only one way - we can only save a job once, after which we only update it's state.  

We already have version and revision field in the job model. Revisions are used to check sequence/ordering when interacting with the jobstore, but Versions are mostly unused at present.  We _could_ start making use of them but I suspect it won't be a short piece of work as we'll need to introduce new jobstore functionality for updating jobs (we can't currently) and changes to the interface for retrieving different versions. 

We _could_ ensure the metadata reference is bidirectional (by late saving the origin job) and using that in calls to get/describe to retrieve the origin job specification?

cmd/cli/exec/exec.go
rossjones: Yeah, I'll take another look, I think it was giving me problems at the Python end but will double-check

rossjones: We do need to quote these strings, as they are used as args later. e.g.

```shell
python -c "import this"
```

needs to translate to 

```shell
launcher.py -- -c "import this"
```

Rather than DIY I've used `shellescape` which is based on the Python version of the same name.

pkg/translation/translators/python.go
rossjones: And after many hours of debugging, I hadn't added all the hosts I thought I had. :(

simonwo (approved):

PR #3109, opened 2023-12-14
Fix error message when missing plugin path

rossjones:
Was using the wrong value in the error message.


pkg/repo/util.go
```diff
@@ -54,7 +54,7 @@ func validateRepoConfig(cfg types.BacalhauConfig) error {
 	if exists, err := fileExists(cfg.Node.ExecutorPluginPath); err != nil {
 		return err
 	} else if !exists {
-		return fmt.Errorf("executor plugin path does not exist at: %q", cfg.User.Libp2pKeyPath)
+		return fmt.Errorf("executor plugin path does not exist at: %q", cfg.Node.ExecutorPluginPath)
```
frrist, line 57: :facepalm:

frrist (approved):

PR #3115, opened 2023-12-14
Don't auto-generate Go-based docs using Swagger anymore

simonwo:
These were only needed by the swagger embedded server which we don't use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,22 +82,6 @@ $(warning "Pre-commit is not installed in .git/hooks/pre-commit. Please run 'mak
 endif
 
 ################################################################################
-# Target: swagger-docs
-################################################################################
-.PHONY: swagger-docs
-swagger-docs:
-	@echo "Building swagger docs..."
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -269,6 +253,25 @@ dist/${PACKAGE}.tar.gz: ${BINARY_PATH} | dist/
 dist/${PACKAGE}.tar.gz.signature.sha256: dist/${PACKAGE}.tar.gz | dist/
 	openssl dgst -sha256 -sign $(PRIVATE_KEY_FILE) -passin pass:"$(PRIVATE_KEY_PASSPHRASE)" $^ | openssl base64 -out $@
 
+################################################################################
+# Target: swagger-docs
+################################################################################
```
frrist, line 258: I thought the intent was to remove this?

simonwo: We still generate the `swagger.json` file as its used to generate our Python SDK. But we no longer generate the `docs/docs.go` file.

rossjones (approved):

PR #3122, opened 2023-12-15
Appeasing linter

rossjones:
More appeasing of the linter for various 'violations'.  Also fixes a complaint about cyclomatic complexity by the judicious extraction of functions.

### DIFF
--- a/pkg/publicapi/endpoint/requester/endpoints_cancel.go
+++ b/pkg/publicapi/endpoint/requester/endpoints_cancel.go
@@ -16,20 +16,20 @@ import (
 	"github.com/pkg/errors"
 )
 
-//	@ID				pkg/requester/publicapi/cancel
-//	@Summary		Cancels the job with the job-id specified in the body payload.
-//	@Description	Cancels a job specified by `id` as long as that job belongs to `client_id`.
-//	@Description	Returns the current jobstate after the cancel request has been processed.
-//	@Tags			Job
-//	@Accept			json
-//	@Produce		json
-//	@Param			CancelRequest	body		legacymodels.CancelRequest	true	" "
-//	@Success		200				{object}	legacymodels.CancelResponse
-//	@Failure		400				{object}	string
-//	@Failure		401				{object}	string
-//	@Failure		403				{object}	string
-//	@Failure		500				{object}	string
-//	@Router			/api/v1/requester/cancel [post]
+// @ID				pkg/requester/publicapi/cancel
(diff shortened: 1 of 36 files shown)


pkg/publicapi/endpoint/requester/endpoints_cancel.go
```diff
@@ -16,20 +16,20 @@ import (
...
+// @Tags			Job
+// @Accept			json
+// @Produce		json
+// @Param			CancelRequest	body		legacymodels.CancelRequest	true	" "
+// @Success		200				{object}	legacymodels.CancelResponse
+// @Failure		400				{object}	string
+// @Failure		401				{object}	string
+// @Failure		403				{object}	string
+// @Failure		500				{object}	string
+// @Router			/api/v1/requester/cancel [post]
```
simonwo, line 32: These are not going to survive a round trip because atm CI calls `swag fmt`. Can you look at if we actually need to do that? One thing I will say about `swag fmt` is it lines up all the comments on these headers whereas regular fmt does not.

simonwo:
How come our CI lint step doesn't pick these up?

pkg/publicapi/endpoint/requester/endpoints_cancel.go
rossjones: Will take a look, but my editor go formats on save so I'm likely to be creating a lot of this noise :(

rossjones:
> How come our CI lint step doesn't pick these up?

It definitely running golangci-lint but it's not finding anything.  I'll try and SSH in to one and check.

pkg/publicapi/endpoint/requester/endpoints_cancel.go
simonwo: Yeah more I mean, we can remove `swag fmt` if it's not strictly necessary

rossjones: Removed in b0607970952b46398d42f5e837edc36dfd6708e8

pkg/storage/copy/copy.go
```diff
@@ -37,11 +37,14 @@ func CopyOversize(
 	specs []*models.InputSource,
 	srcType, dstType string,
 	maxSingle, maxTotal datasize.ByteSize,
-) (modified bool, err error) {
+) (bool, error) {
+	var modified bool
+	var err error
```
simonwo, line 42: sigh

simonwo (approved):

PR #3123, opened 2023-12-15
Speed up the failing ipfs downloader test

rossjones:
Currently the 'fail' sub-test completes by reaching a deadline, which is set to 120 seconds.  There is little point waiting 120s when the success path completes in ms.

Sets the context provided to timeout after 5 seconds instead of 120s.

### DIFF
--- a/pkg/downloader/ipfs/downloader_test.go
+++ b/pkg/downloader/ipfs/downloader_test.go
@@ -130,9 +130,13 @@ func TestDownloadFromPrivateSwarm(t *testing.T) {
 	config.Set(cfg)
 
 	t.Run("download failure without swarm key", func(t *testing.T) {
+		// This fails by timing out, but does so after 2minutes. We should give it
+		// 5 seconds to find the file, which is plenty given success takes ms.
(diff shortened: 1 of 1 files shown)


pkg/downloader/ipfs/downloader_test.go
```diff
@@ -130,9 +130,13 @@ func TestDownloadFromPrivateSwarm(t *testing.T) {
 	config.Set(cfg)
 
 	t.Run("download failure without swarm key", func(t *testing.T) {
```
frrist, line 132: :thinking: All we are really doing here is validating we configured IPFS correctly and validating it's behavior with a swarm key configured. Would asserting the IPFS daemon is configured as expected be sufficient, rather than attempting to download a file we can't actually download?

frrist (approved):
:fire:

pkg/downloader/ipfs/downloader_test.go
rossjones: I think it's probably okay that we're at least testing that timeouts are honoured if we specify them in the context.

PR #3124, opened 2023-12-15
Fix non-working test to save 30s

rossjones:
Initially the test would block waiting for stdin, would not run, and then timeout. 
The result was that all tests here take 1 minute, without this test, 25s. 

It appears that at the full size (700+bytes) the first test was failing to read from stdin (even with ReadAll). 
Shortening the size of the file makes it readable so the text now runs and finishes much faster than when it was waiting for the test to timeout.

### DIFF
--- a/cmd/cli/create/create_test.go
+++ b/cmd/cli/create/create_test.go
@@ -35,10 +35,14 @@ func (s *CreateSuite) TestCreateGenericSubmitBetter() {
 		Name    string
 		Fixture *testdata.FixtureLegacy
 	}{
-		{
-			Name:    "noop json",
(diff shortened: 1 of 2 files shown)


cmd/cli/create/create_test.go
```diff
@@ -80,9 +80,12 @@ func (s *CreateSuite) TestCreateGenericSubmitBetter() {
 		},
 	}
 
+	// Let's do this once
+	canRunS3Test := s3helper.CanRunS3Test()
+
```
frrist, line 85: :clap: :clap: :clap:

frrist (approved):

PR #3132, opened 2023-12-19
Provide flag to set webui port

rossjones:
Currently the webui port is hard-coded to 80 (by it's absence in the ListenAndServe struct), and for some users they may not have enough privileges to use that.

Now using `--web-ui-port=2020` will make the web ui available on port 2020 (http://127.0.0.1:20202) if --web-ui is also set.


cmd/util/flags/configflags/web_ui.go
```diff
@@ -5,8 +5,14 @@ import "github.com/bacalhau-project/bacalhau/pkg/config/types"
...
 		FlagName:     "web-ui",
-		ConfigPath:   types.NodeWebUI,
+		ConfigPath:   types.NodeWebUIEnabled,
 		DefaultValue: false,
 		Description:  `Whether to start the web UI alongside the bacalhau node.`,
 	},
+	{
+		FlagName:     "web-ui-port",
+		ConfigPath:   types.NodeWebUIPort,
+		DefaultValue: 80,
```
rossjones, line 15: This seems to be ignored and not set for the ConfigPath.
e.g. `listenPort, err := config.Get[int](types.NodeWebUIPort)` returns 0 instead of 80.

simonwo (approved):

cmd/util/flags/configflags/web_ui.go
simonwo: This sounds like a reasonably serious problem?

cmd/cli/serve/serve.go
```diff
@@ -286,13 +281,31 @@ func serve(cmd *cobra.Command) error {
...
 	if startWebUI {
+		listenPort, err := config.Get[int](types.NodeWebUIPort)
+		if err != nil {
+			return err
+		}
+
+		// If we get the default value for an int, use 80 instead.
+		if listenPort == 0 {
+			listenPort = 80
+		}
```
simonwo, line 299: Some day it would be nice to be able to set 0 to mean "let the kernel pick a free port"

cmd/util/flags/configflags/web_ui.go
frrist: weird, let me check this out then.

frrist: I think we forgot to run https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/types/config.go#L3

frrist (changes requested):
Don't merge until https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/types/config.go#L3 has been run. 99% sure its the reason for the failure to get the port.

cmd/util/flags/configflags/web_ui.go
frrist: Issue is a mixture of the above and not setting the defaults for our various configs will push a change to fix this in a sec.

cmd/cli/serve/serve.go
frrist, line 299: we can remove this now. The default is used from the config via 138c8e9dbc61aac9126f5ae4e980fa2e045b6b31

frrist (approved):
assuming 138c8e9dbc61aac9126f5ae4e980fa2e045b6b31 looks alright to everyone LGTM.

cmd/cli/serve/serve.go
simonwo: Ah. We could add a linter step to make sure that all generate commands are up to date (i.e. don't produce a diff)

frrist: https://github.com/bacalhau-project/bacalhau/pull/3133#pullrequestreview-1789760388
😄😄😄

frrist: fwiw That wouldn't have completely caught the problem here, the other part that needed to change was updating our default config structures and using the defaults for the value default value. But would be good to have regardless.

PR #3133, opened 2023-12-19
Run code generation

rossjones:
Whenever I run code gen to generate config settings, it ends up with lots and lots of unrelated changes in the mocks where `interface{}` is being replaced with `any`. Running it now in the hope that future runs of `make generate` won't flip them back.


frrist (approved):
LGTM!

At some pointer it might be worth including a `go generate` check in CI to detect discrepancies like this. For example if an interface was extended and there wasn't a mock for it. Or if the config stricture gets a new filed and the developer forgets to run `go generate`.

PR #3135, opened 2023-12-20
Retries jobs that failed to schedule, up to a user-configurable deadline

alaric-rd:
The change to line 37 of `capacity_available_strategy.go` is a bug fix - as we were passing in the usage struct rather than a pointer to a usage struct, the nice String method defined on pointers to those structs wasn't being invoked in the error message and so it used an ugly default rendering.

Other than that, the logic is pretty simple. I added a new retry strategy that looks at the scheduling deadline and creation time of the job (and changed the retry strategy interface slightly to pass the job rather than just the job ID, to save looking it up again in the job store). Then made the `previous_executions` ranker stop ruling out nodes we'd failed to get a bid from before. Then added a field to the scheduler plan for a new evaluation, added a planner to enqueue that evaluation if found, and made the batch/service job scheduler differentiate retry and failure cases and make a delayed evaluation if it's time to retry.

The rest is just plumbing to let users specify a scheduling deadline.

Still to do:

 - Add command line flags to all the other ways to run a job, that sets a scheduling deadline? That should be useful for users.
 - Audit all the places people look at jobs and make sure something sensible is displayed. `plan.DeferEvaluation` does `p.DesiredJobState = JobStateTypePending` but perhaps we need a new job state type for "queued" or something. However, I've not seen that state displayed in any UI I've looked at - instead I see ones that match `ExecutionState` values (AskForBid and all that).

## Summary by CodeRabbit

- **New Features**
	- Introduced job scheduling and retry settings to enhance job creation and execution processes.
	- Added a new `ReschedulingPolicy` to manage job retries and timeouts more effectively.
	- Implemented a new `DeadlineStrategy` for retry logic, focusing on job scheduling deadlines.
	- Updated job ranking logic based on execution history to improve node selection.

- **Enhancements**
	- Enhanced job evaluation methods to support deferring evaluations, allowing for more flexible job scheduling.
	- Improved handling of rejected job executions with updated retry mechanisms and failure handling strategies.

- **Bug Fixes**
	- Fixed issues related to job rescheduling and retry logic to ensure more reliable job execution.

- **Refactor**
	- Refactored job creation and scheduling components to support new scheduling and retry settings.
	- Updated internal models and data structures to align with new job scheduling and retry features.


simonwo (approved):

pkg/orchestrator/scheduler/batch_service_job.go
```diff
@@ -89,19 +91,42 @@ func (b *BatchServiceJobScheduler) Process(ctx context.Context, evaluation *mode
...
+	} else {
+		// create new executions if needed
+		remainingExecutionCount := desiredRemainingCount - execsByApprovalStatus.activeCount()
+		if remainingExecutionCount > 0 {
+			allFailed := existingExecs.filterFailed().union(lost)
+			var placementErr error
+			if len(allFailed) > 0 && !b.retryStrategy.ShouldRetry(ctx, orchestrator.RetryRequest{Job: &job}) {
+				placementErr = fmt.Errorf("exceeded max retries for job %s", job.ID)
+				b.handleFailure(nonTerminalExecs, allFailed, plan, placementErr)
+				return b.planner.Process(ctx, plan)
```
wdbaruni, line 122: unnecessary as we are checking for placementErr in line 126

pkg/orchestrator/scheduler/batch_service_job.go
```diff
@@ -162,8 +187,23 @@ func (b *BatchServiceJobScheduler) placeExecs(ctx context.Context, execs execSet
 	return nil
 }
 
+func (b *BatchServiceJobScheduler) handleRetry(plan *models.Plan, job *models.Job, failures int) {
+	// delay = RetryDelay * RetryDelayGrowthFactor ^ failures
+	// ...but never more than MaximumRetryDelay
+	delay := gomath.Min(
+		float64(job.RetryDelay)*gomath.Pow(job.RetryDelayGrowthFactor, float64(failures)),
+		float64(job.MaximumRetryDelay))
+
```
wdbaruni, line 196: exponential backoff usually multiplies the backoff duration with the growth factor instead of using powers

pkg/orchestrator/scheduler/batch_service_job.go
```diff
@@ -89,19 +91,42 @@ func (b *BatchServiceJobScheduler) Process(ctx context.Context, evaluation *mode
...
+	if (rejectedExecutions > 0) && evaluation.TriggeredBy != models.EvalTriggerDefer {
+		// If we had failed executions due to bid rejections in the
+		// past, then we should retry. This causes the scheduler to be
+		// nivoked again after the retry delay; when that happens,
+		// evaluation.TriggeredBy will equal models.EvalTriggerDefer so
+		// the test above checks for that to make us not retry *again*
+		// after we've retried.
+		b.handleRetry(plan, &job, rejectedExecutions)
+		return b.planner.Process(ctx, plan)
+	} else {
```
wdbaruni, line 113: I am not sure I follow this. An evaluation arriving at the scheduler should be processed as is. If an delayed evaluation was enqueued, then the evaluation would arrive at the scheduler after the backoff duration.

pkg/orchestrator/scheduler/batch_service_job.go
```diff
@@ -162,8 +187,23 @@ func (b *BatchServiceJobScheduler) placeExecs(ctx context.Context, execs execSet
 	return nil
 }
 
+func (b *BatchServiceJobScheduler) handleRetry(plan *models.Plan, job *models.Job, failures int) {
+	// delay = RetryDelay * RetryDelayGrowthFactor ^ failures
+	// ...but never more than MaximumRetryDelay
```
wdbaruni, line 192: I think this logic should fall under the `retryStrategy` where instead of returning true of false, it should return a response that says do not retry or do retry with a backoff value. This decouples the calculation of the backoff window from the scheduler into the retry strategy

wdbaruni (changes requested):
Nice PR. General comments:
1. The change relies on the `WaitUntil` feature in the evaluations to work as expected. Lets make sure this is actually the case and that the evaluation broker is blocking the evaluation until `WaitUntil` period is reached. Though we need to understand how other evaluations related to the same job are handled, such as if the user updates the job, or one of the compute nodes sends an update 
2. We should add unit tests for the retry policies and scheduler, specially that we already have good coverage and test helpers for `batch_service_job` to verify the generated plan matches the expectations. In this case, we need to verify new evaluations were added with expected backoff, and make sure no evaluation was enqueud when not needed

pkg/orchestrator/scheduler/batch_service_job.go
alaric-rd: Yeah, that's what I started with, but I couldn't see a particularly nice place to store the current backoff duration. Mutating the job didn't seem to fit with the philosophy of the system, which seems to avoid mutating the job (there's only a few, limited, kinds of updates accepted by the Job Store), so I felt that using exponentiation to re-compute the result of those repeated multiplications from scratch each time was more in keeping with how the rest of the scheduler worked, purely by considering the history of past executions.

pkg/orchestrator/scheduler/batch_service_job.go
alaric-rd: Yes. I'll improve the comment - this is because a retry involves two evaluations: one that diagnoses the situation as a failure that should be retried (which calls `b.handleRetry` to schedule that retry evaluation). When the retry evaluation is then processed, after the delay, it finds itself in exactly the same situation, because nothing else has changed - but it can observe that it is a retry evaluation (`evaluation.TriggeredBy == models.EvalTriggerDefer`) so it should actually plan new execution attempts rather than deferring itself once more.

pkg/orchestrator/scheduler/batch_service_job.go
alaric-rd: I've made a new ticket for this as it'll require a bit of rearrangement!

pkg/orchestrator/scheduler/batch_service_job.go
wdbaruni: As explained in our last meeting, the evaluation broker is deduping evaluations for the same job, and there is no guarantee that the deferred evaluation is the one the will be served. Relying on the `TriggeredBy` field to make scheduling decisions might result in unexpected behaviour.

**Second,** `rejected` executions is not the only reason that we want to retry scheduling. Failed executions, or no live nodes at all are reasons where we want to queue the job and retry at a later time

**Third,** we shouldn't delay a job if there are `rejected` or `failed` executions, but we should delay if the scheduler couldn't find any new node at allocate the execution to, and the retry policy still says we can retry. If I am following the code correctly, this is what your scheduler will do:
1. found 10 nodes, ask 1 node to bid
2. node rejects
3. backoff and retry later instead of asking the remaining 9 nodes immediately

Though the expected behaviour is:
1. found 10 nodes, ask 1 node to bid
2. node rejects
3. found 9 nodes, ask 1 node to bid
4. ...
5. found 0 nodes, backoff and retry later

We used to ask more than one node in each iteration, but that is no [longer working](https://github.com/bacalhau-project/expanso-planning/issues/597)

This means the node ranking should exclude nodes that were asked before, and somehow reconsider them later when retrying after backoff.

pkg/orchestrator/scheduler/batch_service_job.go
```diff
@@ -92,13 +94,30 @@ func (b *BatchServiceJobScheduler) Process(ctx context.Context, evaluation *mode
...
+
 	// create new executions if needed
 	remainingExecutionCount := desiredRemainingCount - execsByApprovalStatus.activeCount()
 	if remainingExecutionCount > 0 {
-		allFailed := existingExecs.filterFailed().union(lost)
+		allFailed := existingExecs.filterFailed().union(lost).union(rejected)
 		var placementErr error
-		if len(allFailed) > 0 && !b.retryStrategy.ShouldRetry(ctx, orchestrator.RetryRequest{JobID: job.ID}) {
+		if len(allFailed) > 0 && !b.retryStrategy.ShouldRetry(ctx, orchestrator.RetryRequest{Job: &job}) {
 			placementErr = fmt.Errorf("exceeded max retries for job %s", job.ID)
```
wdbaruni, line 118: the retryStrategy is mainly invoked for failed executions to evaluate if the execution should retried on other nodes, or if the job is bad and we shouldn't retry any longer. `lost` are included here as well as the job might be the cause of taking the node down. Rejection on the other hand is different as the node rejects a job without even executing it, such as if the required engines are not supported, doesn't fit the total capacity, or no capacity currently available

pkg/orchestrator/scheduler/batch_service_job.go
alaric-rd: > As explained in our last meeting, the evaluation broker is deduping evaluations for the same job, and there is no guarantee that the deferred evaluation is the one the will be served. Relying on the `TriggeredBy` field to make scheduling decisions might result in unexpected behaviour.

I think it might end up having evaluations that do nothing, and might end up waiting longer than it needs to if a `WaitUntil` evaluation is lost because an immediate one overwrites it in dedupe (what exactly is the logic by which the evaluation broker decides which one to keep when deduping?), but I think it should always make progress: any new evaluation will create a delayed evaluation that will then make progress.

The question is, what could we do better?

> **Second,** `rejected` executions is not the only reason that we want to retry scheduling. Failed executions, or no live nodes at all are reasons where we want to queue the job and retry at a later time

Yeah. To refactor the retry strategy, I had to combine the handling of rejected, failed, and lost executions so I think this is now OK. But I only made it delay if there were rejected executions, because we didn't originally have a delay for the other cases.

> **Third,** we shouldn't delay a job if there are `rejected` or `failed` executions, but we should delay if the scheduler couldn't find any new node at allocate the execution to, and the retry policy still says we can retry. If I am following the code correctly, this is what your scheduler will do:

Originally the code worked more like that - `createMissingExecs` used to return an error if it couldn't find a suitable node - but this stopped working when I rebased onto `main` after Christmas :-(

So. What can we do better? The tricky thing is that the scheduler has little information to go on: the desired state of the job and the execution history. When it's triggered because an execution was rejected by a node, if it decides it needs to wait before trying, it creates a new execution with a `WaitUntil`, but when THAT execution occurs, how does it know to do anything different to the original execution?

Perhaps rather than looking at `TriggeredBy`, it could look at the timestamps in the execution history. Perhaps if the most recently rejected execution is less than the current retry delay ago we should just create a delayed evaluation (delayed just enough to wait until the current retry delay later than the rejection time of the last execution, so if an evaluation rocks up after 5s for some reason and we only need to try again in 8s, we create one delayed by 3s). If the evaluation broker dedupes by always keeping the latest evaluation for a given job ID, this might work better?

What do you think?

> We used to ask more than one node in each iteration, but that is no [longer working](https://github.com/bacalhau-project/expanso-planning/issues/597)

Hmmm, I thought we created as many executions as we wanted in one go, from looking at how `createMissingExecs` is used and works!

> This means the node ranking should exclude nodes that were asked before, and somehow reconsider them later when retrying after backoff.

Ok, I made the node ranking prefer nodes that haven't been asked before, and accept (at a lower priority) nodes that have been asked before, under the assumption that this meant the fresh nodes would be tried in preference but if we rant out of them we'd start looking again at previously tried nodes.

Should we try to NOT introduce delays if there are "fresh" nodes still to try, only delaying when we have to go back to previously-tried nodes?

That would require deeper refactoring of the batch/service job scheduler, i think. I'm nervous about going into deeper depths of yak-shaving as this thing has dragged on a lot already. Would it be good to get something basic and working merged into main, by me writing tests and documentation for what I have, then make it more clever and less wasteful in new branches?

pkg/models/job.go
```diff
@@ -114,6 +114,23 @@ type Job struct {
...
+	SchedulingTimeout int64 `json:"SchedulingTimeout"`
+
+	// How many seconds to initially wait between retries. This increases each retry, up to MaximumRetryDelay
+	BaseRetryDelay int64 `json:"BaseRetryDelay"`
+
+	// The maximum delay between retries
+	MaximumRetryDelay int64 `json:"MaximumRetryDelay"`
+
+	// The growth factor: how much longer to wait before retrying each time
+	RetryDelayGrowthFactor float64 `json:"RetryDelayGrowthFactor"`
```
wdbaruni, line 133: As I pointed out in our last meeting, how will these timeouts play with the `TimeoutConfig` we have under the task https://github.com/bacalhau-project/bacalhau/blob/3f1733a84d3ae48d61b59f75c19d679fba8745c0/pkg/models/task.go#L40?

`TimeoutConfig` currently only have `ExecutionTimeout` value. While the name might indicate it is a timeout for when an execution start and doesn't cover scheduling or when an execution is rescheduled somewhere else, the implementation is actually an end to end timeout from when the job is first created as shown here https://github.com/bacalhau-project/bacalhau/blob/3f1733a84d3ae48d61b59f75c19d679fba8745c0/pkg/requester/housekeeping.go#L72

The reason we have `TimeoutConfig` under `Task` instead of `Job` is that once we support multiple executions per job, we should allow different timeout values for each task. Also the plan is to add a timeout for pulling the inputs, a timeout for publishing the results, and make execution timeout only for actual execution.

Also, the compute nodes do timeout the execution on their own if it takes longer than `ExecutionTimeout` and report that to the requester. The reason we have a Housekeeper in the Orchestrator side is to make sure lost jobs are eventually marked as failed, such as when the compute node is no longer responsive and didn't report that its execution was timedout to the orchestrator. 

Since we are adding a new `SchedulingTimeout` here, my recommendations are:
- Change Housekeeping to treat execution timeout from when the execution started instead of when the job was created. My guess it that the housekeeper should fail the execution instead of the job and enqueue an evaluation. Then it should be the responsibility of the scheduler to make a decision on whether to retry on another node or not. 
- Clarify what `SchedulingTimeout` is. Is it end to end from the the job was created? Or is it how long the job stays idle in the queue waiting for allocation? I think it makes sense to have both types of timeouts in our config.

It would be nice to do some of these changes on separate PRs as this one keeps getting bigger. Example introduce and merge the housekeeper changes first

pkg/orchestrator/interfaces.go
```diff
@@ -95,21 +95,21 @@ type NodeDiscoverer interface {
...
 type NodeSelector interface {
 	// AllNodes returns all nodes in the network.
 	AllNodes(ctx context.Context) ([]models.NodeInfo, error)
 	// AllMatchingNodes returns all nodes that match the job constrains and selection criteria.
-	AllMatchingNodes(ctx context.Context, job *models.Job) ([]models.NodeInfo, error)
+	AllMatchingNodes(ctx context.Context, job *models.Job, retryDelay time.Duration) ([]models.NodeInfo, error)
 	// TopMatchingNodes return the top ranked desiredCount number of nodes that match job constraints
 	// ordered in descending order based on their rank, or error if not enough nodes match.
-	TopMatchingNodes(ctx context.Context, job *models.Job, desiredCount int) ([]models.NodeInfo, error)
+	TopMatchingNodes(ctx context.Context, job *models.Job, retryDelay time.Duration, desiredCount int) ([]models.NodeInfo, error)
```
wdbaruni, line 109: if we follow my advice to encode the delay in the execution instead of the history, then we can pass the list of previous executions to the node rankers. We already have a ranker that pulls previous executions from the job store, and might make more sense to do it once and pass it to all rankers

pkg/orchestrator/planner/re_evaluator.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+	"github.com/bacalhau-project/bacalhau/pkg/orchestrator"
+)
+
+// ReEvaluator is a planner implementation that emits events based on the job state.
+type ReEvaluator struct {
+	id               string
```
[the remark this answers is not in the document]
wdbaruni: is this used somewhere?

pkg/orchestrator/scheduler/batch_job_test.go
```diff
@@ -60,11 +61,71 @@ func TestBatchSchedulerTestSuite(t *testing.T) {
...
+	s.jobStore.EXPECT().RecordJobDeferral(gomock.Any(), job.ID, gomock.Any(), "Deferring rescheduling for 10s")
+
+	// No nodes are suitable, as we're within the retry delay of the last failure
+	nodeInfos := []models.NodeInfo{}
+
+	s.mockNodeSelection(job, nodeInfos, job.Count, 10*time.Second)
+
+	matcher := NewPlanMatcher(s.T(), PlanMatcherParams{
+		Evaluation:         evaluation,
+		NewEvaluationDelay: time.Duration(10 * time.Second),
```
wdbaruni, line 86: better to create an evaluation with expected type and delay, and pass that to the matcher to verify that a new matching evaluation has been enqueued

pkg/models/job.go
alaric-rd: Yeah, I didn't put this in with the TimeoutConfig because (a) it's about the whole job not just one task and (b) even thought one of the four tunables here *is* a timeout, it's specifically the timeout for retrying scheduling so it seemed nice to put it together with the other things that tune scheduling retries. I think you're right about changing the meaning of the ExecutionTimeout - a user who expects their job to take ten minutes of CPU should be able to set an execution timeout of ten minutes plus a bit of safety margin, and not need to effectively add scheduling time into the execution timeout as well; and a job that happens to get scheduled quickly doesn't deserve to have extra execution time! And that should indeed be a separate change.

pkg/models/execution.go
```diff
@@ -209,6 +209,16 @@ func (e *Execution) IsDiscarded() bool {
...
+// IsRejected returns true if the execution has been rejected.
+func (e *Execution) IsRejected() bool {
+	switch e.ComputeState.StateType {
+	case ExecutionStateAskForBidRejected, ExecutionStateBidRejected:
+		return true
+	default:
+		return false
+	}
+}
+
```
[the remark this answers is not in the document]
alaric-rd: That was probably a misunderstanding on my part - I'll fix it!

pkg/orchestrator/scheduler/batch_job_test.go
alaric-rd: Done

pkg/jobstore/boltdb/store.go
```diff
@@ -30,6 +30,7 @@ const (
 	BucketJobEvaluations   = "evaluations"
 	BucketJobHistory       = "job_history"
 	BucketExecutionHistory = "execution_history"
+	BucketDeferralHistory  = "deferral_history"
```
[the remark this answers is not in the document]
alaric-rd: Done

pkg/jobstore/boltdb/store.go
```diff
@@ -839,6 +850,34 @@ func (b *BoltJobStore) updateJobState(tx *bolt.Tx, request jobstore.UpdateJobSta
 	return b.appendJobHistory(tx, job, previousState, request.Comment)
 }
 
+func (b *BoltJobStore) RecordJobDeferral(ctx context.Context, jobID string, waitUntil time.Time, comment string) error {
```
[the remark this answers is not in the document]
alaric-rd: I've made deferrals a job-level history entry (and update the "legacy printer" to just skip those for now as I don't know how it should handle them; that was the one test that failed after the change, but I fear there may be other code that assumes a job-level history entry is always a job state change).

pkg/jobstore/boltdb/store.go
```diff
@@ -839,6 +850,34 @@ func (b *BoltJobStore) updateJobState(tx *bolt.Tx, request jobstore.UpdateJobSta
...
+	historyEntry := models.JobHistory{
+		Type:  models.JobHistoryTypeJobSchedulingDeferral,
+		JobID: jobID,
+		SchedulingDeferral: &models.SchedulingDeferral{
+			DeferredUntil: waitUntil,
+		},
```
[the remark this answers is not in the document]
alaric-rd: The scheduler uses the execution history to know when a node can be reconsidered again (that's in the previous_executions.go ranker). The job history is consulted by the scheduler just to get a count of deferrals to compute the exponential backoff: this is better than looking at the executions as there isn't a 1:1 relationship between execution history and the scheduler deferring things.

wdbaruni:
Closing this PR and will pick up the work at a later time

PR #3154, opened 2024-01-03
3151 fix tests when s3 credentials are valid but dont have access to the test bucket

aronchick:


pkg/test/logstream/stream_address_test.go
```diff
@@ -17,6 +19,13 @@ import (
 )
 
 func (s *LogStreamTestSuite) TestStreamAddress() {
+	// Test to see if we're running on a Mac where docker doesn't work
+	// because the loopback for Docker is different than the loopback
+	// on Linux and skip the test if we are
+	if runtime.GOOS == "darwin" {
+		s.T().Skip("Skipping test on Mac")
+	}
+
```
simonwo, line 28: This test passes fine for me on a Mac M1. So I'd rather work out what is wrong with your set up than disable it for everyone.

docs/DEVELOPMENT.md
simonwo, line 1: Please can you add this information to a proper page in the documentation site? See the "Developing" section which is where our developer-focused docs live now.

webui/tests/test_bacalhau_running.py
simonwo, line 1: What is this for? Should it be part of this PR?

simonwo (changes requested):

pkg/test/logstream/stream_address_test.go
aronchick: bizarre - this fails every time for me on a mac, but works fine on linux. also, it's clearly pulling from the bad docker IP space

docs/DEVELOPMENT.md
aronchick: k, fixed

webui/tests/test_bacalhau_running.py
aronchick: It was from another PR that i chained into this one.

pkg/test/logstream/stream_address_test.go
simonwo: Can you explain your logic more? I can't see anywhere in the test code is dealing with loopback, so does this mean the whole feature just doesn't work for you? If so, we should probably deal with that problem – if we skip the test, we'll never come back to it.

What version of Docker Desktop are you using? I'm at 4.22.1.

aronchick: Docker Desktop: 4.25.2 (129061)

I reliably get this error 100% of the time locally, and 0% when testing on linux.

Error:
```
17:55:53.702 | INF pkg/executor/docker/handler.go:169 > removing container [container:0c9dc29ecc9562fe549b60b38cbaae7913512a21ccff95743d95d9505b471d30] [execution:01235017-c46b-4334-b4f5-c54e2f0b0efd] [job:e072faf7-35ed-4be3-886d-14f99e648d99]
    stream_address_test.go:88:
        	Error Trace:	/Users/daaronch/code/bacalhau/pkg/test/logstream/stream_address_test.go:88
        	Error:      	Received unexpected error:
        	            	logstreamclient failed to open stream: failed to dial QmRWhPvxNz71Nsrs6udwo5WxoWXMMdbCGBZbGLyRL1t1X8:
        	            	  * [/ip4/192.0.2.42/tcp/61919] dial tcp4 0.0.0.0:62044->192.0.2.42:61919: i/o timeout
        	Test:       	TestLogStreamTestSuite/TestStreamAddress
```


I figured skipping on Mac was fine because we'd catch any potential errors on other platforms (the functionality of this test requires spinning up a separate container, we're not testing in any functionality that should be specific to a mac, nor anything in a topology that an end user would actually do

aronchick: Ok, i give up. I removed it. It'll just fail on my mac, but it passes in CI.

simonwo: Hm, I've upgraded to the same Docker version and it's still passing for me. It looks from the error like the test requester node is unable to contact the test compute node, both of which shouldn't be inside Docker? Perhaps @rossjones can spot a problem from that error message.

simonwo: The lines are still here in this PR @aronchick

aronchick: done.

simonwo (approved):

PR #3157, opened 2024-01-03
adding testing bacalhau running script

aronchick:


simonwo (commented):
This PR looks identical to #3154 ??

aronchick:
> This PR looks identical to #3154 ??

The issue is that we are slow at approving PRs, i started using stacked branches so each PR was small, but worked when done in order - https://andrewlock.net/working-with-stacked-branches-in-git-is-easier-with-update-refs/ - this minimizes pain, but requires we do things in order (which is not clear to me how to signal). This one IS different, but only very small so it looks the same.

simonwo (commented):
Needs an update to remove stuff already committed/reviewed.

simonwo (approved):

aronchick:
https://github.com/bacalhau-project/bacalhau/issues/3219

PR #3158, opened 2024-01-03
200 for api root

aronchick:


simonwo (commented):
... I can't really see the wood for the trees here because there is a mix of changes from main and from #3154 and friends. Please can you check if this PR is required and just rebase it on top of main instead of using merge commits, which make the diff hard to read? Thanks.

simonwo (approved):

PR #3173, opened 2024-01-05
Add initial docs for exec command

rossjones:
Adds initial docs for the exec command.

### DIFF
--- a/docs/docs/references/cli/exec/index.md
+++ b/docs/docs/references/cli/exec/index.md
@@ -5,12 +5,13 @@ sidebar_label: exec
 
 ## Description
 
-The `bacalhau exec` command provides ...
+The `bacalhau exec` command allows for the specification of jobs to be executed from the command line,
+without the need for a job specification file (see [job run](../job/run/)). 
 
 ## Usage
 
 ```shell
-bacalhau exec [flags] [job-type] 
+bacalhau exec [flags] [job-type] arguments
 ```
 
 ## Flags
--- a/docs/docs/references/cli/exec/index.md
+++ b/docs/docs/references/cli/exec/index.md
@@ -1,0 +1,165 @@
+---
+sidebar_label: exec
+---
+# Command: `exec`
+
+## Description
+
+The `bacalhau exec` command provides ...
+
+## Usage
+
+```shell
+bacalhau exec [flags] [job-type] 
+```
+
+## Flags
+
+- `-h`, `--help`:
+    - Description: Displays help information for the `exec` sub-command.
+
+- `--code`: 
+    - Includes the specified code in the job. This can be a single file, or a directory containing many files.  There is a limit of 10Mb on the size of the uploaded code.
+
+- `-f`, `--follow`:
+    - Description: If provided, the command will continuously display the output from the job as it runs.
+
+- `--wait`
+	- Description: Wait for the job to finish. Use --wait=false to return as soon as the job is submitted.
+
+- `--wait-timeout-secs`
+	- Description: When using --wait, how many seconds to wait for the job to complete before giving up.
+
+- `--node-details`
+	- Description: Print out details of all nodes (overridden by --id-only).
+
+- `--id-only`:
+    - Description: On successful job submission, only the Job ID will be printed.
+
+- `-p`, `--publisher`
+	- Description: Where to publish the result of the job.
+
+- `-i`, `--input`
+    - Description: Mount URIs as inputs to the job. Can be specified multiple times. Format: src=URI,dst=PATH[,opt=key=value]
+        ### Examples:
+        **Mount IPFS CID to /inputs directory**
+
+        `-i ipfs://QmeZRGhe4PmjctYVSVHuEiA9oSXnqmYa4kQubSHgWbjv72`
+
+        **Mount S3 object to a specific path**
+
+        `-i s3://bucket/key,dst=/my/input/path`
+
+        **Mount S3 object with specific endpoint and region**
+
+        `-i src=s3://bucket/key,dst=/my/input/path,opt=endpoint=https://s3.example.com,opt=region=us-east-1`
+
+- `-o`, `--output`
+    - Description: name:path of the output data volumes. 'outputs:/outputs' is always added unless '/outputs' is mapped to a different name.
+
+- `-e`, `--env`
+    - Description: The environment variables to supply to the job (e.g. --env FOO=bar --env BAR=baz)
+
+- `--timeout`
+    - Description:  Job execution timeout in seconds (e.g. 300 for 5 minutes)
+
+- `-l`, `--labels`
+    - Description: List of labels for the job. Enter multiple in the format '-l a -l 2'. All characters not matching /a-zA-Z0-9_:|-/ and all emojis will be stripped.
+
+- `-s`, `--selector`
+    - Description: Selector (label query) to filter nodes on which this job can be executed, supports '=', '==', and '!='.(e.g. -s key1=value1,key2=value2). Matching objects must satisfy all of the specified label constraints.
+
+
+## Global Flags
+
+- `--api-host string`:
+    - Description: Specifies the host used for RESTful communication between the client and server. The flag is disregarded if the `BACALHAU_API_HOST` environment variable is set.
+    - Default: `bootstrap.production.bacalhau.org`
+
+- `--api-port int`:
+    - Description: Specifies the port for REST communication. If the `BACALHAU_API_PORT` environment variable is set, this flag will be ignored.
+    - Default: `1234`
+
+- `--log-mode logging-mode`:
+    - Description: Sets the desired log format. Options are: `default`, `station`, `json`, `combined`, and `event`.
+    - Default: `default`
+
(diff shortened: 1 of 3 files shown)


docs/docs/references/cli/exec/index.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Examples
+
+### Running python tasks
+
+1. **Basic Usage**:
+
+   **Command**:
+   ```shell
+   → bacalhau exec python -- -c "import this"
```
aronchick, line 101: i think we should not require '-c' or '--' - just wrap those ourselves

docs/docs/references/cli/exec/index.md
```diff
@@ -0,0 +1,166 @@
...
+   Complex is better than complicated.
+   Flat is better than nested.
+   ....
+   ```
+
+2. **Single file Python**:
+
+   **Command**:
+   ```shell
+   → bacalhau exec --code=app.py python app.py
```
aronchick, line 120: how about bacalhau exec python app.py?

docs/docs/references/cli/exec/index.md
```diff
@@ -0,0 +1,166 @@
...
+   ```
+
+   where app.py is 
+   ```python
+   """
+   pip install colorama
+   """
+
+   from colorama import Fore
+   print(Fore.RED + "Hello World")
```
aronchick, line 130: not sure about color as an example - just do Fibonacci?

docs/docs/references/cli/exec/index.md
```diff
@@ -0,0 +1,166 @@
...
+### Running duckdb queries
+
+1. **Basic Usage**:
+
+   **Command**:
+   ```shell
+   → cat describe.sql
+     DESCRIBE TABLE '/inputs/world-cities_csv.csv';
+
+   → bacalhau exec --code=describe.sql -i src=https://datahub.io/core/world-cities/r/world-cities.csv,dst=/inputs duckdb -- -init /code/describe.sql
```
aronchick, line 151: same deal here - just provide the file without all the other details (still need -i)

docs/docs/topic-guides/custom-job-types.md
```diff
@@ -0,0 +1,90 @@
...
+
+```python
+"""
+pip install colorama 
+"""
+
+from colorama import Fore 
+```
+
+If a `pyproject.toml` is found in a directory specified by `--code`, then it is used to install the dependencies.  If the file has a poetry section, then `poetry install` is used instead.
```
aronchick, line 62: fascinating ... how? shouldn't we have a flag - e.g. "--requirements" or something?

docs/docs/references/cli/exec/index.md
rossjones: Python supports a lot of flags though, and so if we dont tell bacalhau to stop processing flags (--) then we'll end up having to disable the bacalhau flags in favour of the ones supplied by the user when using python. One of those being -i.

I don't think this use case is likely to be common, there's not much mileage in running code like this, it was just more indicative of how to pass flags to python rather than to bacalhau.

docs/docs/references/cli/exec/index.md
rossjones: Can do, but the point of this example was really to show the auto-requirements for single file scripts.  So maybe I can find something that does both.

docs/docs/references/cli/exec/index.md
rossjones: In this case, the `-init /code/describe.sql` are the flags being passed to duckdb.
We can change the docker image to include the -init but that would mean we can no longer also do `duckdb -- -c select name from table`

docs/docs/topic-guides/custom-job-types.md
rossjones: So we check, in order:

*  If it's a single file, check for pip install commands in the module doc (by parsing the python with the ast package)
* Look for a pyproject.toml and if it has a poetry.tool section, use poetry. If not, pip install it.
* If there's a requirements.txt, install it
* Use any setup.py

docs/docs/references/cli/exec/index.md
```diff
@@ -0,0 +1,166 @@
...
+
+- `--code`: 
+    - Includes the specified code in the job. This can be a single file, or a directory containing many files.  There is a limit of 10Mb on the size of the uploaded code.
+
+- `-f`, `--follow`:
+    - Description: If provided, the command will continuously display the output from the job as it runs.
+
+- `--wait`
+	- Description: Wait for the job to finish. Use --wait=false to return as soon as the job is submitted.
+
```
frrist, line 30: While my opinion deviates from our usual practice - I'll share it anyways: I have a predisposition against boolean flags that are enabled by default because using them often requires explicitly disabling them. 

Consider the difference between using a flag like `exec --wait=false <commands>` vs. `exec --no-wait <commands>`. Preference for the latter since the presence of the `--no-wait` flag implies an active choice to skip waiting, with its default state being inactive (false). This means if the flag is used, it's assumed to be true. However, I agree that using negative terms in flags can sometimes lead to confusion due to the inherent double negatives.

docs/docs/references/cli/exec/index.md
```diff
@@ -0,0 +1,166 @@
...
+- `--node-details`
+	- Description: Print out details of all nodes (overridden by --id-only).
+
+- `--id-only`:
+    - Description: On successful job submission, only the Job ID will be printed.
+
+- `-p`, `--publisher`
+	- Description: Where to publish the result of the job.
+
```
frrist, line 42: Would love an example similar to `--input` here if it's not too much trouble.

docs/docs/topic-guides/custom-job-types.md
```diff
@@ -0,0 +1,90 @@
...
+
+Once a job has been translated, for example: Python -> Docker then the new job ID is returned to the client as if it were a normal job. The newly translated job may have a dependency on an external container which performs extra-steps during deployment, it is up to the translator to ensure it specifies the correct container or wasm module.
+
+
+## Initial job types 
+
+The initial implementation of custom job types includes job types for Python and DuckDB. Both of these custom job types support the `--code` parameter to simplify the code being sent to  the requester node - for python this is likely python code, for duckdb it is expected to be sql.  Command line parameters that follow the job type are used as the parameters to the translator, so `python /code/app.py /inputs/data.txt` will result in parameters that are `['python', '/code/app.py', '/inputs/data.txt']`.  Any flags that need to be passed to the python process, for instance should app.py require -f for the final parameter, must be sent after the bash -- separator.
```
frrist, line 46: Aside from reading these docs, how would a user find the list of custom job types bacalhau supports? Is there anything like a `list` command that returns these? Or maybe it's included in `exec --help`?

docs/docs/topic-guides/custom-job-types.md
```diff
@@ -0,0 +1,90 @@
...
+"""
+pip install colorama 
+"""
+
+from colorama import Fore 
+```
+
+If a `pyproject.toml` is found in a directory specified by `--code`, then it is used to install the dependencies.  If the file has a poetry section, then `poetry install` is used instead.
+
+A `requirements.txt` file will be used with `pip install -r` to install requirements, if found in the specified directory.
```
frrist, line 64: Does this imply the compute nodes running the job need to have networking enabled? And/or that the client submitting the job most request a networking job?

docs/docs/topic-guides/custom-job-types.md
```diff
@@ -0,0 +1,90 @@
...
+2. Ensure the template contains a single task that specifies an engine that can be processed by your translator.
+3. Add any templated values you wish to be filled with full exec parameters. For example, ` "Version": "{{or (index . "version") "3.11"}}"` will add the value from `--version` or default to 3.11
+
+### Requester translator 
+
+1. Add a new implementation of [Translator](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/translation/translation.go#L30-L38) to the [translators directory](https://github.com/bacalhau-project/bacalhau/tree/main/pkg/translation/translators) in the translation package.
+2. Ensure the translation replaces the engine in the task it is provided with a new engine that is either a valid docker or webassembly engine.  
+3. Add your translator to the [standard translations provider](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/translation/translation.go#L45-L52).
```
frrist, line 85: This implies users need to make a code change to bacalhau, correct?

frrist (approved):
lgtm

docs/docs/references/cli/exec/index.md
rossjones: Totally agree, and we should remove the --wait flags from v2 of the CLI

docs/docs/topic-guides/custom-job-types.md
rossjones: Currently, that's true.

docs/docs/topic-guides/custom-job-types.md
rossjones: Added a ticket to address this today

PR #3177, opened 2024-01-06
3148 add swaggo for swagger directory for rest

aronchick:


main.go
```diff
@@ -16,6 +16,57 @@ import (
...
+//// 	@tokenUrl								https://example.com/oauth/token
+//// 	@scope.read								Grants read access
+//// 	@scope.write							Grants write access
+//// 	@scope.admin							Grants read and write access to administrative information
+//
+//// 	@securitydefinitions.oauth2.accessCode	OAuth2AccessCode
+//// 	@tokenUrl								https://example.com/oauth/token
+//// 	@authorizationUrl						https://example.com/oauth/authorize
+//// 	@scope.admin							Grants read and write access to administrative information
+
```
simonwo, line 69: None of this is correct for us

pkg/models/node_info.go
```diff
@@ -63,28 +55,57 @@ func (p mergeProvider) GetLabels(ctx context.Context) map[string]string {
...
+// NodeInfo godoc
+//
+// @ID			NodeInfo
+// @Summary		Information about the node.
+// @Description	Information about the node.
+// @Tags		Ops
+// @Produce		json
+// @Success	200	{object}	NodeInfo
+// @Failure	500	{object}	string
+// @Router		/api/v1/agent/node [get]
```
simonwo, line 79: Swagger comments should be added on the actual method which serves the request, rather than the data structure, no?

pkg/publicapi/endpoint/shared/endpoint.go
```diff
@@ -76,7 +76,7 @@ func (e *Endpoint) id(c echo.Context) error {
 // @Description	each node is connected to a number of peer nodes.
 // @Tags			Utils
 // @Produce		json
-// @Success		200	{object}	[]peer.AddrInfo
+// @Success		200	{object}	[]string
```
simonwo, line 79: [peer.AddrInfo](https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.32.2/core/peer#AddrInfo) is not a string.

aronchick: This is a mechanism to flatten a complex structure into a string - swaggo does this automatically. If we'd like to do this more thoroughly, we can, but I get it was overkill

pkg/models/node_info.go
aronchick: Swagger needs the godoc but you're right (I think) no router here

main.go
aronchick: This is all commented out, fyi

aronchick:
Replaced with this stack of PRs - https://github.com/bacalhau-project/bacalhau/pull/3224

PR #3188, opened 2024-01-08
introduce nats transport

wdbaruni:
This PR introduces a new transport layer for inter-node connectivity based on NATS, in addition to the existing libp2p transport.

## Integration Architecture
NATS is embedded into Bacalhau nodes. Each Orchestrator node (aka Requester node) also runs a NATS server and connects with other Orchestrator nodes to form a NATS cluster. Compute nodes on the other hand only run a NATS client and connect to any of the Orchestrator nodes to listen for job requests and publish their node info to the network.

With this approach, only the Orchestrator nodes need to be reachable, which simplies the network requirements for the Compute nodes. Also, the Compute nodes only need to know the address of only a single Orchestrator node, and will learn about the other Orchestrators at runtime. This simplifies bootstrapping, but also enables the Compute node to failover and reconnect to other Orchestrators at runtime.

For completeness, the Orchestrator nodes also run a NATS client to subscribe for Compute node info, publish job requests and listen for responses.

### NATS Subjects
NATS follows subject based addressing where clients publish and subscribe to subjects, and the servers take care of routing of those messages. These are the current subject patterns used to orchestrate jobs:

- `node.compute.<node-id>.*`
Every compute node subscribes to a topic in the format of `node.compute.<node-id>.*` to handle job requests from the orchestrator, such as `AskForBid` and `BidAccepted`. For example, if the Orchestrator selects node `QmUg1MoAUMEbpzg6jY45zpfNo2dmTBJS6CzZCTMbniYsvC` to run a job, it will send a message to `node.compute.QmUg1MoAUMEbpzg6jY45zpfNo2dmTBJS6CzZCTMbniYsvC.AskForBid/1`, which will land on the correct compute node. The suffix `/1` is just for versioning in case we change the `AskForBid` message format

- `node.orchestrator.<node-id>.*`
Similarly, Orchestrator nodes listen to a dedicate subject to handle compute callbacks, such as `OnRunComplete` and `OnComputeFailure`

- `node.info.<node-id>`
Compute nodes periodically publish their node info to `node.info.<node-id>` subject, whereas Orchestrator nodes listen to `node.info.*` so their can handle node info from all compute nodes

## Running with NATS
NATS is currently opt-in, and libp2p is still the default transport layer. To run a network with NATS, use the following commands:
```
# Run Orchestrator node
bacalhau serve --node-type=requester --use-nats

# Run a second Orchestrator node
bacalhau serve --node-type=requester --use-nats --cluster-peers=<HOST>:6222


# Run a compute node
bacalhau serve --node-type=compute --use-nats --orchestrators=<HOST>:4222

# Run devstack with NATS
bacalhau devstack --use-nats
```

## Testing Done
Deployed and tested in `development` environment

## Remaining Work
1. Logstream support for NATS transport
1. Auth, as currently any node can join the network
1. Devstack tests based on NATS and deployed as part of circleci checks


pkg/libp2p/transport/libp2p.go
```diff
@@ -0,0 +1,203 @@
...
+	callbackProxy     *bprotocol.CallbackProxy
+	nodeInfoPubSub    pubsub.PubSub[models.NodeInfo]
+	nodeInfoDecorator models.NodeInfoDecorator
+}
+
+func NewLibp2pTransport(ctx context.Context,
+	config Libp2pTransportConfig,
+	nodeInfoStore routing.NodeInfoStore) (*Libp2pTransport, error) {
+	// Monkey patch the identify protocol to allow discovering advertised addresses of networks of 3 or more nodes, instead of 5.
+	identify.ActivationThresh = 2
```
rossjones, line 49: Comment mentions three or more nodes, but the ActivationThresh is set to 2.  Is this how the ActivationThresh works?

pkg/models/node_info.go
```diff
@@ -77,7 +85,12 @@ type NodeInfo struct {
 
 // ID returns the node ID
 func (n NodeInfo) ID() string {
-	return n.PeerInfo.ID.String()
+	if n.NodeID != "" {
+		return n.NodeID
+	} else if n.PeerInfo != nil {
+		return n.PeerInfo.ID.String()
+	}
+	return ""
```
rossjones, line 93: Doesn't seem like "" would be valid.

pkg/nats/logger.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// NewZeroLogger creates a new ZeroLogger
+func NewZeroLogger(logger zerolog.Logger) ZeroLogger {
+	return ZeroLogger{
+		logger: logger,
+	}
+}
+
+func (N ZeroLogger) Noticef(format string, v ...interface{}) {
```
rossjones, line 20:
```suggestion
func (z ZeroLogger) Noticef(format string, v ...interface{}) {
```

Capital for the receiver name seems strange, and might be best to choose a letter from the type name?

pkg/nats/proxy/callback_handler.go
```diff
@@ -0,0 +1,82 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	log.Debug().Msgf("ComputeHandler %s subscribed to %s", handler.name, subject)
+	return handler, nil
+}
+
+// handleRequest handles incoming NATS messages.
+func handleCallback(msg *nats.Msg, handler *CallbackHandler) {
```
rossjones, line 49: Should this be a method on *CallbackHandler?

pkg/nats/proxy/constants.go
```diff
@@ -0,0 +1,35 @@
+package proxy
+
+import "fmt"
+
+const (
+	ComputeEndpointSubjectPrefix = "node.compute"
+	CallbackSubjectPrefix        = "node.orchestrator"
+
+	AskForBid       = "AskForBid/1"
```
rossjones, line 9: Is the /1 the function arity, or the version or something else?

pkg/orchestrator/transformer/job.go
```diff
@@ -38,10 +37,9 @@ func DefaultsApplier(defaults JobDefaults) JobTransformer {
 }
 
 // RequesterInfo is a transformer that sets the requester ID and public key in the job meta.
```
rossjones, line 39: Comment now inaccurate as the public key is removed

rossjones (approved):
A few stylistic/readability suggestions, but looks good.

rossjones:
Maybe unrelated, but when I try the command below no ID is shown

```shell
bacalhau --api-host bootstrap.development.bacalhau.org node list
```

pkg/nats/server.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// NewServerManager is a helper function to create a NATS server with a given options
+func NewServerManager(ctx context.Context, opts *server.Options) (*ServerManager, error) {
+	ns, err := server.NewServer(opts)
+	if err != nil {
+		return nil, err
+	}
+	ns.SetLoggerV2(NewZeroLogger(log.Logger), opts.Debug, opts.Trace, opts.TraceVerbose)
+	go ns.Start()
```
frrist, line 31: creating the server should happen separately from starting it, let's make this a method on `ServerManager`.

frrist (commented):
few drive by comments, will continue to review over the week(s).

cmd/cli/devstack/devstack.go
```diff
@@ -147,7 +147,9 @@ func NewCmd() *cobra.Command {
 		&ODs.ConfigurationRepo, "stack-repo", ODs.ConfigurationRepo,
 		"Folder to act as the devstack configuration repo",
 	)
-
+	devstackCmd.PersistentFlags().BoolVar(
+		&ODs.UseNATS, "use-nats", ODs.UseNATS,
+		"Use NATS for messaging")
```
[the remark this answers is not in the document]
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/f4fbdada06f06953d10eb3ceac4b9d16a4ea8b4c

Though my thought of the flag is just something temporary while we fully transition to nats. I am hoping that we focus and support a single type of transport and deprecate libp2p in favour of nats. We can look into supporting more transports in the future if we get strong demand for it.

cmd/cli/serve/serve.go
```diff
@@ -173,42 +173,70 @@ func serve(cmd *cobra.Command) error {
...
 		return err
 	}
 
-	ipfsClient, err := SetupIPFSClient(ctx, cm, ipfsConfig)
+	libp2pCfg, err := config.GetLibp2pConfig()
 	if err != nil {
 		return err
 	}
 
+	if !networkConfig.UseNATS {
```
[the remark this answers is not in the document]
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/f4fbdada06f06953d10eb3ceac4b9d16a4ea8b4c

pkg/compute/endpoint.go
```diff
@@ -142,6 +142,10 @@ func (s BaseEndpoint) CancelExecution(ctx context.Context, request CancelExecuti
 
 func (s BaseEndpoint) ExecutionLogs(ctx context.Context, request ExecutionLogsRequest) (ExecutionLogsResponse, error) {
 	log.Ctx(ctx).Debug().Msgf("processing log request for %s", request.ExecutionID)
+	// TODO: remove this once we support log streaming with nats
```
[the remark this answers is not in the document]
wdbaruni: What I am referring to here is the vending of logs is tightly coupled to libp2p streams. We can decouple this by having the compute node unaware of the transport layer and just return log lines to a caller, which then can write them to either a libp2p stream or to a nats subject.

What you are referring to is another layer between the compute node and the executor, which also needs more abstractions to support executors other than docker

cmd/cli/serve/serve.go
```diff
@@ -173,42 +173,70 @@ func serve(cmd *cobra.Command) error {
...
+	var nodeID string
+	privKey, err := config.GetLibp2pPrivKey()
 	if err != nil {
 		return err
 	}
+	peerID, err := peer.IDFromPrivateKey(privKey)
+	if err != nil {
+		return err
+	}
+	nodeID = peerID.String()
```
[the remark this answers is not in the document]
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/8ab1829ef8d416acf8594559094cd49a1b682b63

cmd/cli/serve/serve.go
```diff
@@ -308,92 +331,124 @@ func serve(cmd *cobra.Command) error {
...
+			cmd.Println()
+			cmd.Println(sb.String())
 		}
-		if err != nil {
-			return err
+	} else {
+		if !networkConfig.UseNATS {
+			cmd.Println("Make sure there's at least one requester node in your network.")
 		}
+	}
```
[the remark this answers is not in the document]
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/8ab1829ef8d416acf8594559094cd49a1b682b63

cmd/cli/serve/util.go
```diff
@@ -188,3 +185,20 @@ func getDisabledFeatures() (node.FeatureConfig, error) {
 func getAllowListedLocalPathsConfig() []string {
 	return viper.GetStringSlice(types.NodeAllowListedLocalPaths)
 }
+
+func getNetworkConfig() (node.NetworkConfig, error) {
```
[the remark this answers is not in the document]
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/425e11d31effd064feb1951d75773e17e29e3e33

pkg/devstack/devstack.go
```diff
@@ -128,57 +135,66 @@ func Setup(
...
+					return nil, err
 				}
-				libp2pPeer = append(libp2pPeer, peerAddr)
+			}
+
+			if isRequesterNode {
+				clusterConfig.ClusterName = "devstack"
+				clusterConfig.ClusterPort = clusterPort
+				orchestratorAddrs = append(orchestratorAddrs, fmt.Sprintf("0.0.0.0:%d", swarmPort))
+				clusterPeersAddrs = append(clusterPeersAddrs, fmt.Sprintf("0.0.0.0:%d", clusterPort))
```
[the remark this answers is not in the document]
wdbaruni: fixed in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/4878d44cb90f143a800f6a53f06588317265bde8

pkg/libp2p/transport/libp2p.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/hashicorp/go-multierror"
+	libp2p_pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/libp2p/go-libp2p/core/host"
+	basichost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	routedhost "github.com/libp2p/go-libp2p/p2p/host/routed"
+	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
+	"github.com/multiformats/go-multiaddr"
+)
+
+const NodeInfoTopic = "bacalhau-node-info"
```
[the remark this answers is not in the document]
wdbaruni: this is libp2p gossipsub topic. I kept the same existing topic name to not break existing networks

pkg/libp2p/transport/libp2p.go
```diff
@@ -0,0 +1,203 @@
...
+// RegisterComputeCallback registers a compute callback with the transport layer.
+func (t *Libp2pTransport) RegisterComputeCallback(callback compute.Callback) error {
+	bprotocol.NewCallbackHandler(bprotocol.CallbackHandlerParams{
+		Host:     t.Host,
+		Callback: callback,
+	})
+	// To enable nodes self-dialing themselves as libp2p doesn't support it.
+	t.callbackProxy.RegisterLocalComputeCallback(callback)
+
+	return nil
```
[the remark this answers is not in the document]
wdbaruni: the libp2p implementation of `TransportLayer` doesn't return an error, but nats implementation does

pkg/libp2p/transport/libp2p.go
wdbaruni: yeah this is how it works. Improved the comment in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/425e11d31effd064feb1951d75773e1

pkg/nats/logger.go
wdbaruni: fixed in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/b88bed4a56822045d404b9d56dbde218bda4040a

pkg/nats/proxy/constants.go
wdbaruni: it is the version. I changed the subject names to `AskForBid/v1` for better clarity in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/52bbc1e8da4b7c7827436969d814ff15ec854ba0

pkg/orchestrator/transformer/job.go
wdbaruni: fixed in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/a3132f0fc047f08c078753c4fc23c85f2be940ea

pkg/nats/proxy/callback_handler.go
wdbaruni: yeah that works since this method doesn't use generics like `compute_handler`. Fixed in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/a15b80d84c7cf2e9354bc939b3219468346014c5

pkg/models/node_info.go
wdbaruni: it doesn't, but I prefer that `ID()` method doesn't return errors as well.  Maybe we just need to make `NodeInfo` implement `Validate` method and make sure it is called whenever it is instantiated. Added a TODO in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/8ffa1caecd1f42d866d80a27e8b4adf292bf7eca

pkg/nats/server.go
```diff
@@ -0,0 +1,74 @@
...
+	"time"
+
+	"github.com/bacalhau-project/bacalhau/pkg/model"
+	"github.com/nats-io/nats-server/v2/server"
+	"github.com/rs/zerolog/log"
+)
+
+const ReadyForConnectionsTimeout = 5 * time.Second
+
+type ServerManagerParams struct {
```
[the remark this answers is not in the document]
wdbaruni: fixed in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/e4319dd58cf59c85a9349b175824d1fbd590a1c9

pkg/nats/server.go
```diff
@@ -0,0 +1,74 @@
...
+	Options *server.Options
+}
+
+// ServerManager is a helper struct to manage a NATS server
+type ServerManager struct {
+	Server *server.Server
+}
+
+// NewServerManager is a helper function to create a NATS server with a given options
+func NewServerManager(ctx context.Context, opts *server.Options) (*ServerManager, error) {
```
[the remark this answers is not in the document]
wdbaruni: used in logging in https://github.com/bacalhau-project/bacalhau/pull/3188/commits/e4319dd58cf59c85a9349b175824d1fbd590a1c9

pkg/config/types/node.go
```diff
@@ -118,3 +120,18 @@ type FeatureConfig struct {
...
+	Port              int                  `yaml:"Port"`
+	AdvertisedAddress string               `yaml:"AdvertisedAddress"`
+	Orchestrators     []string             `yaml:"Orchestrators"`
+	Cluster           NetworkClusterConfig `yaml:"Cluster"`
+}
+
+type NetworkClusterConfig struct {
+	Name              string   `yaml:"Name"`
+	Port              int      `yaml:"Port"`
+	AdvertisedAddress string   `yaml:"AdvertisedAddress"`
```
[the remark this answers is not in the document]
wdbaruni: The `AdvertisedAddress` under `NetworkConfig` is what the requester exposes for compute nodes to join and listen for jobs. In terms of NATS, compute nodes are clients and the requester is a server, this is the server's address.

The `AdvertisedAddress` under `NetworkClusterConfig` is what the requester exposes for other requester nodes to build a NATS cluster. Compute nodes can join any requester node in a cluster and they will still hear about messages that were received and published to other requester nodes.

Both are optional and useful when requesters are in different networks and when the listen address is not the same as the call address. libp2p tries to discover the advertised address dynamically through identity protocol, but it requires cluster of 5 nodes or more. nats requires defining that expliclity, but we only need to do it for requester nodes and not compute nodes

nomad has a similar configuration https://developer.hashicorp.com/nomad/docs/configuration#advertise

pkg/nats/server.go
wdbaruni: yeah I understand the benefits of decoupling the creation and starting of the server, specially when we need to register handlers or work with the server before it is started, such as the case with our http server. At the same time it adds some complexity where users need to remember to always start the server, and we need to handle multiple starts or starting a stopped server.

In this specific case, the server needs to be started immediately as a client is created in the following lines in `NATSTransport` and will fail if the server is not started
https://github.com/bacalhau-project/bacalhau/blob/nats/pkg/nats/transport/nats.go#L115

wdbaruni:
> Maybe unrelated, but when I try the command below no ID is shown
> 
> ```shell
> bacalhau --api-host bootstrap.development.bacalhau.org node list
> ```

Yeah you are right. When calling `node list` using an older client, the node IDs will be empty since the older client relies on the value under `PeerInfo.ID` as the node id, which will be empty for NATS networks. I don't think this will be a blocker

PR #3207, opened 2024-01-10
feat: allow clients to submit docker/wasm jobs with mismatched versions

frrist:
Allows users to configure if their client version must match the server version to submit a job with `docker run` or `wasm run`.
closes [expanso-planning/#242](https://github.com/bacalhau-project/expanso-planning/issues/242)
closes [expanso-planning/#241](https://github.com/bacalhau-project/expanso-planning/issues/241)


cmd/util/version.go
```diff
@@ -24,15 +25,22 @@ type Versions struct {
 }
 
 func CheckVersion(cmd *cobra.Command, args []string) error {
+	strictVersioning, err := config.Get[bool](types.NodeStrictVersionMatch)
+	if err != nil {
+		return fmt.Errorf("DEVELOPER ERROR: please file an issue with this error message: %w", err)
+	}
```
rossjones, line 31: Is there any way we can have an error-free version of config.Get?  I can see why the errors are currently returned, but it might be better for us to have an interface similar to hash where like `Get[T](string) (T, bool)` so we can flag the non-existence without raising an error that we then have to handle.  

If we pass the wrong key here it's a developer error rather than a runtime error so even better would be a compile-time error.

rossjones (approved):
Looks good. My only real complaint is the error message asking the user to report a developer error which would ideally be a compile time problem. Worse case as we're defaulting to false (the zero val for bools) is that we ignore the error.

Not sure I've ever really understood why we're binding so tightly to server versions, as opposed to say, versioning the API and controlling which client works that way.

frrist:
> Not sure I've ever really understood why we're binding so tightly to server versions, as opposed to say, versioning the API and controlling which client works that way.

I don't really understand either. Initially I want to include a version in the job spec and allow the server to decide what to do. But given we have two different job specs (`model` and `models`) this was the easier change. Once we migrate to a single job spec structure we can include version details there.

PR #3220, opened 2024-01-11
Add authentication API endpoints and default using private keys

### LINKED ISSUE #260: Discussion: Should labels exist on a "job" or "jobSpec"?
cc @lukemarsden @binocarlos @Bubblyworld

### LINKED ISSUE #261: Hate to bring this up again, but this is looking like leaking again - non-deterministically failing test in this test.
```
// TODO: Hate to bring this up again, but this is looking like leaking again - non-deterministically failing test in this test.
```

https://github.com/filecoin-project/bacalhau/blob/471cb01f8030b15607de31fc375ba7562735c487/cmd/bacalhau/run_test.go#L114

simonwo:
This PR adds authenticator infrastructure. 

There is a new authn API endpoint allowing supported authn methods to be listed and used. API users should query the endpoint for acceptable methods and respond to one to receive an authz token.

We also implement a challenge-based authentication method that requires the user to sign a phrase with their private key. If they respond with a valid signature, they are issued an access token using the specified policy. The default policy matches the authentication semantics that we have today, namely that any user with a private key can interact. 

This PR also does some rearranging as suggested by Ross in a previous review, and fixes an issue with logging.

Resolves https://github.com/bacalhau-project/expanso-planning/issues/260. 
Resolves https://github.com/bacalhau-project/expanso-planning/issues/261.


pkg/authn/challenge/challenge.go
```diff
@@ -0,0 +1,118 @@
...
+	return challengeAuthenticator{
+		authnPolicy: p,
+		getPhrase:   getPhrase,
+		key:         lo.Must(jwk.New(key)),
+		nodeID:      nodeID,
+		challenge:   policy.AddQuery[policyData](p, authn.PolicyTokenRule),
+	}
+}
+
+// Authenticate implements Authenticator.
```
rossjones, line 58: Is it better to use an interface guard rather than a comment for this (especially as the comment is confusing and might be best as // challengeAuthenticator implements Authenticator

```
var _ Authenticator = (*challengeAuthenticator)(nil)
```

pkg/authn/challenge/challenge.go
```diff
@@ -0,0 +1,118 @@
...
+func (authenticator challengeAuthenticator) Authenticate(req *http.Request) (authn.Authentication, error) {
+	var userInput response
+	err := json.NewDecoder(req.Body).Decode(&userInput)
+	if err != nil {
+		return authn.Authentication{Success: false, Reason: err.Error()}, err
```
rossjones, line 63: Might it be worth a new constructor for authn.Authentication that takes an err and returns this structure (and the err). Failed authentication seems like a reasonably common occurrence and it would be good to ensure consistency.

Literally something like 

```go
func FailedAuthentication(err error) error {
     return authn.Authentication{Success: false, Reason: err.Error()}, err
}
```

pkg/authn/challenge/challenge.go
```diff
@@ -0,0 +1,118 @@
...
+		NodeID:     authenticator.nodeID,
+		ClientID:   system.ConvertToClientID(userKey),
+	}
+
+	result, err := authenticator.challenge(req.Context(), data)
+	if err != nil {
+		return authn.Authentication{Success: false, Reason: err.Error()}, err
+	} else if len(result) != 1 {
+		return authn.Authentication{Success: false, Reason: "signature verified but user credentials rejected"}, nil
+	}
```
rossjones, line 92:
```suggestion
	}
	
	if len(result) != 1 {
		return authn.Authentication{Success: false, Reason: "signature verified but user credentials rejected"}, nil
	}
```

pkg/authn/challenge/challenge.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return authn.Authentication{
+		Success: true,
+		Token:   (result[0].Expressions[0].Value).(string),
```
rossjones, line 96: Feels a little bit the policy stuff is leaking through.  NBD but might be nice if we can hide it in a future PR?

pkg/authn/challenge/challenge_ns_anon.rego
```diff
@@ -0,0 +1,40 @@
...
+now := time.now_ns() / 1000
+
+one_month := time.add_date(time.now_ns(), 0, 1, 0) / 1000
+
+token := io.jwt.encode_sign(
+	{
+		"typ": "JWT",
+		"alg": "RS256",
+	},
+	{
```
rossjones, line 19: Are we missing a subject?  Even the fingerprint might do if we're not tying it to a human (which we might via a idToken in other challenge types?).

pkg/authn/challenge/challenge_ns_no_anon.rego
```diff
@@ -0,0 +1,49 @@
+package bacalhau.authn
+
+import rego.v1
+
+# Implements a policy where only clients with client IDs that match a known list
+# are permitted access. Anonymous users will not be authneticated.
```
rossjones, line 6:
```suggestion
# are permitted access. Anonymous users will not be authenticated.
```

pkg/authn/challenge/challenge_ns_no_anon.rego
```diff
@@ -0,0 +1,49 @@
...
+
+token := token if {
+    input.clientId in allowed_clients
+
+    token := io.jwt.encode_sign(
+	{
+		"typ": "JWT",
+		"alg": "RS256",
+	},
+	{
```
rossjones, line 27: Missing sub?

pkg/authn/types.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"net/http"
+
+	"github.com/bacalhau-project/bacalhau/pkg/lib/provider"
+)
+
+// The rule that authentication policies must implement. If the rule returns a
+// token string, the authentication method has succeeded and passed policy, and
+// the token string will be passed to future API calls. If the rule returns
```
rossjones, line 11: > the token string will be passed to future API calls

Is this always via Authorization header with a "Bearer <token>"?

rossjones (approved):
LGTM but a few minor points/suggestions

pkg/authn/challenge/challenge.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// A base64 encoded PKCS1 public key, as generated by config.encodePublicKey
+	PublicKey string `json:"publicKey"`
+}
+
+type request struct {
+	InputPhrase []byte `json:"inputPhrase"`
+}
+
+type phraser = func() ([]byte, error)
```
frrist, line 37: This could be the `encoding.BinaryMarshaler` interface instead.

pkg/node/factories.go
```diff
@@ -139,3 +129,25 @@ func NewStandardPublishersFactory() PublishersFactory {
 			return provider.NewConfiguredProvider(pr, nodeConfig.DisabledFeatures.Publishers), err
 		})
 }
+
+func NewStandardAuthenticatorsFactory() AuthenticatorsFactory {
```
frrist, line 133: Can we pretty please not use this pattern for this. I would find it way simple to provide the requester node with a map of method names to Authenticators. :pray: begging you.

pkg/publicapi/endpoint/auth/endpoint.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	g := e.router.Group("/api/v1/auth")
+	g.Use(middleware.SetContentType(echo.MIMEApplicationJSON))
+	g.GET("", e.list)
+
+	for _, key := range provider.Keys(ctx) {
+		method := lo.Must(provider.Get(ctx, key))
+		adaptAuthenticator(method, g.Group("/"+key))
+	}
```
frrist, line 32: Please add a comment clarifying what `keys` are. Similarly what is the return method from `get`? Is `method` the `Authenticator` interface?

frrist (approved):
The authentication bits look good.

Not thrilled about the extension of the factory pattern, or the usage of `lo.Must` when accessing providers and requirements. Lots of places things can potentially panic.

pkg/authn/challenge/challenge.go
```diff
@@ -0,0 +1,118 @@
...
+type response struct {
+	PhraseSignature string `json:"phraseSignature"`
+
+	// A base64 encoded PKCS1 public key, as generated by config.encodePublicKey
+	PublicKey string `json:"publicKey"`
+}
+
+type request struct {
+	InputPhrase []byte `json:"inputPhrase"`
+}
```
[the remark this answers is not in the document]
simonwo: Not any serious reason. I realised that I have a preference for camelcased JSON but no idea why, it just seems to be what other people do. I'll change these to TitleCase to better match our other JSON.

pkg/authn/challenge/challenge.go
simonwo: The interface is actually already guarded because `NewAuthenticator` returns the interface and not the type itself, so any interface mismatch would show up in the constructor. These comments are auto-generated by VS Code when you ask it to create method stubs for an interface. I'll just remove them.

pkg/authn/challenge/challenge.go
simonwo: Yes, good idea.

pkg/authn/challenge/challenge.go
simonwo: Ah, I see, you mean leaking through from the query abstraction. Yes.

pkg/authn/types.go
```diff
@@ -0,0 +1,56 @@
...
+	Type MethodType `json:"type"`
+	// Parameters specific to this authentication type. For example, a list of
+	// required information, or minimum acceptable key sizes.
+	Params any `json:"params"`
+}
+
+// Authenticator accepts HTTP requests for user authentications and returns the
+// result of trying to authenticate the credentials supplied by the user.
+type Authenticator interface {
+	provider.Providable
```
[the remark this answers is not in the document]
simonwo: I will work with you on an alternative at https://github.com/bacalhau-project/bacalhau/discussions/3237 :)

pkg/authn/types.go
simonwo: Yes, but that's more of an authz concern. That's how it's currently set out in the draft docs.

pkg/node/factories.go
simonwo: Very happy to do the refactor work if we have a better way to do all of them, for sure.

pkg/node/factories.go
```diff
@@ -139,3 +129,25 @@ func NewStandardPublishersFactory() PublishersFactory {
...
+	return AuthenticatorsFactoryFunc(
+		func(ctx context.Context, nodeConfig NodeConfig) (authn.Provider, error) {
+			privKey, err := config.GetClientPrivateKey()
+			if err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
simonwo: The problem is that we then have to panic-on-error... is that preferable? AFAIK this function is only actually called once on start up.

pkg/node/requester.go
```diff
@@ -302,6 +305,8 @@ func NewRequesterNode(
 		NodeStore:    nodeInfoStore,
 	})
 
+	auth_endpoint.NewEndpoint(ctx, apiServer.Router, authnProvider)
```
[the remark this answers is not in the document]
simonwo: Yeah, I was also initially confused by how this works with the other endpoints. I will rename this one to `BindEndpoint`.

pkg/authn/challenge/challenge.go
simonwo: I've made the query abstraction handle extracting the result from the policy, with the assumption that it's one result per query atm.

PR #3224, opened 2024-01-11
Swagger PRs - 1 - MVP

aronchick:


main.go
```diff
@@ -16,6 +16,57 @@ import (
...
+//// 	@tokenUrl								https://example.com/oauth/token
+//// 	@scope.read								Grants read access
+//// 	@scope.write							Grants write access
+//// 	@scope.admin							Grants read and write access to administrative information
+//
+//// 	@securitydefinitions.oauth2.accessCode	OAuth2AccessCode
+//// 	@tokenUrl								https://example.com/oauth/token
+//// 	@authorizationUrl						https://example.com/oauth/authorize
+//// 	@scope.admin							Grants read and write access to administrative information
+
```
simonwo, line 69: These still refer to example domains and we don't use OAuth2 as standard (yet)

simonwo (approved):

main.go
aronchick: just to be clear, these are all commented out (4 '/'s) - just placeholders for the future.

Issue #248, opened 2022-06-15
Can't name executor that doesn't exist (enums)


lukemarsden:
We want it to be harder for users to name an executor in a job request that doesn't exist. Switch from using strings to strongly typed enums, or some other method for ensuring they are typo-proof.

PR #3230, opened 2024-01-11
Update docker executor for oom

olgibbons:
When a docker container exceeds memory limits (OOM status = true) An informative error message will be relayed to the user.

Have also implemented a test that will trigger OOM errors. NB: The test currently uses the error message as a signal for an error occurring, so if the message is changed, the test may need to be updated as well

Fixes issues #249 #248

### DIFF
--- a/pkg/executor/docker/handler.go
+++ b/pkg/executor/docker/handler.go
@@ -50,6 +50,7 @@ type executionHandler struct {
 	result *models.RunCommandResult
 }
 
+//nolint:funlen
 func (h *executionHandler) run(ctx context.Context) {
 	h.running.Store(true)
 	defer func() {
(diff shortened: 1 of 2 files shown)


pkg/executor/docker/handler.go
```diff
@@ -101,6 +101,23 @@ func (h *executionHandler) run(ctx context.Context) {
...
+				ExitCode: int(containerExitStatusCode),
+				ErrorMsg: err.Error(),
+			}
+			return
+		}
+		if containerJSON.ContainerJSONBase.State.OOMKilled {
+			containerError = errors.New(`memory limit exceeded. Please refer to https://docs.bacalhau.org/getting-started/resources/#docker-executor for more information`)
+			h.result = &models.RunCommandResult{
+				ExitCode: int(containerExitStatusCode),
+				ErrorMsg: fmt.Sprintf(containerError.Error()),
```
rossjones, line 117: You don't need fmt.Sprintf() here as containerError.Error() is already a string. This is the cause of the linter errors you're seeing.

rossjones (approved):
LGTM, feel free to merge when the checks have completed.

PR #3257, opened 2024-01-17
Store source job spec in metadata when translating

rossjones:
Previously we would store two jobs when a translation took place, the original and the translated job. The ID of the translated job is the one we would return. The original job which would never be updated, and just 'be there' when running the list command.  We would store a reference in the newly defined job and all executions would be attached to the newly translated job (likely showing docker instead of python as the engine type).

This PR now encodes the original job as base64 and injects it into the job's metadata should the v2 client want to extract and show it.

### DIFF
--- a/pkg/orchestrator/endpoint.go
+++ b/pkg/orchestrator/endpoint.go
@@ -2,6 +2,7 @@ package orchestrator
 
 import (
 	"context"
+	"encoding/base64"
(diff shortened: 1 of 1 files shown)


alaric-rd (approved):
LGTM and seems much cleaner from an architectural viewpoint than creating two jobs!

PR #3263, opened 2024-01-19
cleanup: Rename NewIPFSPublishers 

### LINKED ISSUE #2555: Fix NewIPFSPublishers

rossjones:
NewIPFSPublishers creates different publishers, not just IPFS base ones,  and returns a PublisherProvider.

Renamed to NewPublisherProvider

Closes #2555

### DIFF
--- a/pkg/publisher/util/utils.go
+++ b/pkg/publisher/util/utils.go
@@ -18,13 +18,7 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/system"
 )
 
-/* TODO(forrest): Fix me(!):
-- method has wrong name, this make S3 publisher, and noop, in addition to IPFS
-Issue: https://github.com/bacalhau-project/bacalhau/issues/2555
-
-*/
-
-func NewIPFSPublishers(
+func NewPublisherProvider(
 	ctx context.Context,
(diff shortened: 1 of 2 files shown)


pkg/publisher/util/utils.go
```diff
@@ -18,13 +18,7 @@ import (
...
 )
 
-/* TODO(forrest): Fix me(!):
-- method has wrong name, this make S3 publisher, and noop, in addition to IPFS
-Issue: https://github.com/bacalhau-project/bacalhau/issues/2555
-
-*/
-
-func NewIPFSPublishers(
+func NewPublisherProvider(
```
frrist, line 21: To align with existing names of `*Providers` should this be `NewStandardPublisherProvider`?

frrist (approved):

pkg/publisher/util/utils.go
rossjones: If there was going to be a non-standard version, it would be worth it, but think this is alright.

PR #3265, opened 2024-01-19
update Networking page content

AlekseyKorzik:


docs/docs/setting-up/networking-instructions/networking.md
```diff
@@ -3,25 +3,29 @@ sidebar_position: 4
 sidebar_label: "Internet Access"
 ---
 
-# Accessing the Internet from Jobs
+# Internet access
```
aronchick, line 6: I wonder if we need to rewrite this more significantly. There's nothing fundamentally wrong with it, but it's framed in a weird way. You can use external docker containers, for example. It's just external network access that is limited - not necessarily to the internet, even to other nodes on the same network.

Further, the `--network` tag needs to be paired with a server with similar flags - do we need to detail that?

docs/docs/setting-up/networking-instructions/networking.md
```diff
@@ -43,7 +47,7 @@ Bacalhau jobs are explicitly prevented from starting other Bacalhau jobs, even i
 
 ## Support for networked jobs on the public network
```
aronchick, line 48: Let's stop referring to this as the "public" network - we should be talking about the "demo" network, we don't want the impression that we're running a public network for everyone for free.

AlekseyKorzik: done

docs/docs/setting-up/networking-instructions/networking.md
AlekseyKorzik: Could you please provide more details for the node/server flag pairing? Seems this is not currently reflected in the documentation

docs/docs/setting-up/networking-instructions/networking.md
```diff
@@ -3,25 +3,29 @@ sidebar_position: 4
...
 
-When submitting a Bacalhau job, you can specify the CID (Content IDentifier) or HTTP(S) URL to download data from. The data will be retrieved before the job starts and made available to the job as a directory on the filesystem. When running Bacalhau jobs, you can specify as many CIDs or URLs as needed using `--input` which is accepted by both `bacalhau docker run` and `bacalhau wasm run`. See [command line flags](https://docs.bacalhau.org/all-flags) for more information.
+When submitting a Bacalhau job, you can specify the CID (Content IDentifier) or HTTP(S) URL to download data from. The data will be retrieved before the job starts and made available to the job as a directory on the filesystem. When running Bacalhau jobs, you can specify as many CIDs or URLs as needed using `--input` which is accepted by both `bacalhau docker run` and `bacalhau wasm run`. See [command line flags](../../dev/cli-reference/all-flags.md) for more information. Make sure the nodes on the network have enough resources to download and process specified data.
```
rossjones, line 21:
```suggestion
When submitting a Bacalhau job, you can specify the CID (Content IDentifier) or HTTP(S) URL to download data from. The data will be retrieved before the job starts and made available to the job as a directory on the filesystem. When running Bacalhau jobs, you can specify as many CIDs or URLs as needed using `--input` which is accepted by both `bacalhau docker run` and `bacalhau wasm run`. See [command line flags](../..all-flags.md) for more information. Make sure the nodes on the network have enough resources to download and process specified data.
```

docs/docs/setting-up/networking-instructions/networking.md
```diff
@@ -3,25 +3,29 @@ sidebar_position: 4
...
+- Using Data Volumes to download the input data and upload the results
+- Using `--network` flag to allow job to access the internet 
+:::info
+For more details about flags and recent updates see the [CLI Guide](../../dev/cli-reference/all-flags.md) or execute `bacalhau help`. Also feel free to contact us on [Slack](https://bacalhauproject.slack.com).
+:::
 
 ## Using Data Volumes
 
-When you submit a Bacalhau job, you'll need to specify the internet locations to download data from and write results to. Both [Docker](../getting-started/docker-workload-onboarding.md) and [WebAssembly](../getting-started/wasm-workload-onboarding.md) jobs support these features.
+If you need to process data located on a certain network resource and/or save the result of the job execution to such resource - you will need to specify the internet locations to download data from and write results to when creating the job. Both [Docker](../../getting-started/docker-workload-onboarding.md) and [WebAssembly](../../getting-started/wasm-workload-onboarding.md) jobs support these features.
```
rossjones, line 19: We might want to skip discussing saving the result of the job in this paragraph as it's mentioned in a paragraph below.  Perhaps this paragraph could just talk about downloading data?

docs/docs/setting-up/networking-instructions/networking.md
```diff
@@ -3,25 +3,29 @@ sidebar_position: 4
...
+When submitting a Bacalhau job, you can specify the CID (Content IDentifier) or HTTP(S) URL to download data from. The data will be retrieved before the job starts and made available to the job as a directory on the filesystem. When running Bacalhau jobs, you can specify as many CIDs or URLs as needed using `--input` which is accepted by both `bacalhau docker run` and `bacalhau wasm run`. See [command line flags](../../dev/cli-reference/all-flags.md) for more information. Make sure the nodes on the network have enough resources to download and process specified data.
 
-You can write back results from your Bacalhau jobs to your public storage location. By default, jobs will write results to the storage provider using the `--publisher` command line flag. See [command line flags](https://docs.bacalhau.org/all-flags) on how to configure this.
+You can write back results from your Bacalhau jobs to your public storage location. By default, jobs will write results to the storage provider using the `--publisher` command line flag. See [command line flags](../../dev/cli-reference/all-flags.md) on how to configure this.
```
rossjones, line 23:
```suggestion
You can write back results from your Bacalhau jobs to your public storage location. By default, jobs will write results to the storage provider using the `--publisher` command line flag. See [command line flags](../../all-flags.md) on how to configure this.
```

docs/docs/setting-up/networking-instructions/networking.md
```diff
@@ -3,25 +3,29 @@ sidebar_position: 4
...
 
-To use these features, the data to be downloaded has to be known before the job starts. For some workloads, the required data is computed as part of the job if the purpose of the job is to process web results. In these cases, networking may be possible during job execution.
 
 ## Specifying Jobs to Access the Internet
 
-To run Docker jobs on Bacalhau to access the internet, you'll need to specify one of the following:
+For some workloads, the required data is computed as part of the job if the purpose of the job is to process web results. In these cases, networking may be possible during job execution. To run Docker jobs on Bacalhau with internet access, you'll need to use a `--netwwork` flag and specify one of the following options:
```
rossjones, line 28:
```suggestion
For some workloads there is a requirement for the job to fetch data at runtime, particularly when it requires access to external APIs. In these cases, networking may be possible during job execution. To run Docker jobs on Bacalhau with internet access, you'll need to use a `--netwwork` flag and specify one of the following options:
```

docs/docs/setting-up/networking-instructions/networking.md
AlekseyKorzik: fixed

docs/docs/setting-up/networking-instructions/networking.md
rossjones: If a user specifies the --network flag, then it will only be run on compute nodes that were started with the `--job-selection-accept-networked ` flag. If no compute nodes were started with that flag, the job will not run.

AlekseyKorzik: Added a note for this case

rossjones (approved):

PR #3278, opened 2024-01-20
Enforcing more prettier cleanup & linting, fixing some SVG issues, and adding msw libraries

aronchick:


webui/webpack.config.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.svg$/,
+        use: ['@svgr/webpack']
+      }
+    ]
+  }
+}
```
simonwo, line 12: We've gone back to packaging SVGs the way we did before, and we didn't need this before... so what's changed that means we need this?

webui/tests/mocks/jobMock.ts
```diff
@@ -138,7 +138,7 @@ export function generateSampleJob(): Job {
     Labels: { ...selectRandomLabels(jobLabels) },
     Tasks: [generateSampleTask()],
     State: {
-      StateType: selectRandomElements(jobStates),
+      StateType: selectRandomElements(jobStates) as string,
```
simonwo, line 141: It would be preferable for `selectRandomElements` to return a consistent `string[]` type and then define a new non-plural function to return a single `string` rather than using type assertions, which are fragile.

webui/webpack.config.js
aronchick: i swapped back because i (personally) really don't like the pattern of importing files directly. It doesn't allow you to do src inline, doesn't allow you to use alt tags, etc. I just think it's bad practice - my opinion!

simonwo (approved):

PR #3289, opened 2024-01-22
Move linters to github actions

### LINKED ISSUE #3288: Move non-execution CI to GA
Given the limited number of runners, we can move some of our tasks to Github Actions to free up a small amount of resource in our main CI.  

To start with we can move all of the linters.

rossjones:
Moves the precommit linters to github actions to free up a little space in CircleCI.  
We also remove `go vet` as it's already part of golangci-lint.

Fixes #3288


.github/workflows/lint.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    # Setup nodejs and build the webui so that we can 
+    # compile the go code that embeds the webui.
+    - name: Setup node
+      uses: actions/setup-node@v4
+      with:
+        node-version: 21
```
simonwo, line 20: I think we have a file in the repo that specifies the correct version to use?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,30 @@
...
+    - run: cd webui && npm install
+    - run: cd webui && npm run build
+
+    - uses: actions/setup-go@v4
+      with:
+        go-version: '1.20.6'
```
simonwo, line 26: As above.

rossjones: When I didn't specify this I got 1.21.x and a complaint asking me to use `go work use`

.github/workflows/lint.yml
rossjones: Yes, .tool-versions says 

```
python 3.10.5
nodejs 21.5.0
```

frrist:
Is it possible to prevent tests from running until the linter passes? i.e. don't waste time on tests if linting fails?

.github/workflows/lint.yml
aronchick: interesting question - should we reuse .tool-version? i suspect yes

aronchick (approved):

.github/workflows/lint.yml
simonwo: Yeah, my point is that the action should be reading the file to know the correct version to load.

simonwo:
> Is it possible to prevent tests from running until the linter passes? i.e. don't waste time on tests if linting fails?

Undesirable – at best, you then need to wait for the linter to run and pass again before knowing what tests failed. There's no issue with wasting time.

.github/workflows/lint.yml
rossjones: Have added this now, and also added go to the list of languages.  Will have to remember to keep .tool-versions in sync with the go.mod now as well.

aronchick (approved):

PR #3291, opened 2024-01-22
docs: Fix dev links in main README.md

adamac:
These links were pointing to some locations that no longer exist.


rossjones (approved):
Looks great, thanks!

PR #3294, opened 2024-01-23
Make ExecuteTestCobraCommand safer to use

### LINKED ISSUE #3293: Make ExecuteTestCobraCommand safer
When using ExecuteTestCobraCommand if you forget to add --api-host or --api-port, you end up testing against the default demo network.  To simplify using these tests, we should allow users to continue to specify them, but if they are absent, then the configured values for the suite should be used.

rossjones:
It is currently too easy to forget to add --api-host and --api-port and then the test may still run, but against the demo network.  This will cause issues later on when the demo network changes.

This PR instead moved the ExecuteTestCobraCommand (and associated) functions onto the BaseSuite which already specified the host and port. If the user does not specify api-host or api-port then the configured values for the suite will be used. This will effectively make it impossible to run tests against the demo network.

Fixes #3293


aronchick (approved):
Nice clean up! I wonder if we should (in the future) ensure that our test suite runs entirely disconnected and/or document what external dependencies/network connectivity is required.

rossjones:
> Nice clean up! I wonder if we should (in the future) ensure that our test suite runs entirely disconnected and/or document what external dependencies/network connectivity is required.

Yeah, absolutely, we should hopefully be able to run local copies of most things and stuff like S3 tests only run if the credentials are available on your laptop.

PR #3299, opened 2024-01-23
Allow authn and authz methods to be configured

simonwo:
This PR exposes configuration options for controlling authn and authz. This actually makes the system usable for controlling access now!

It also includes some of the docs from the docs PR that are relevant for the work at this stage. From this point forward, the docs will track the delivered features.

Resolves https://github.com/bacalhau-project/expanso-planning/issues/341.


cmd/cli/serve/util.go
```diff
@@ -162,12 +162,12 @@ func SetupIPFSClient(ctx context.Context, cm *system.CleanupManager, ipfsCfg typ
...
-	if err := config.ForKey(types.NodeDisabledFeatures, &featureConfig); err != nil {
-		return node.FeatureConfig{}, err
+func getConfig[T any](name string) (T, error) {
+	var obj T
+	if err := config.ForKey(name, &obj); err != nil {
+		return obj, err
 	}
-	return featureConfig, nil
+	return obj, nil
 }
```
frrist, line 171: Nice! Is there any reason to use this over `config.Get`? Perhaps we should modify the existing `Get` method?

docs/docs/dev/auth_flow.md
```diff
@@ -0,0 +1,184 @@
...
+The user agent decides which authentication method to use (e.g. by asking the
+user, or by knowing it has an appropriate key) and operates the flow.
+
+Once all the data for the method has been successfully collected, the user agent
+POSTs the data to the auth endpoint for the method. The endpoint is the base
+auth endpoint plus the name of the method, e.g. `/api/v1/auth/<method>`. So to
+submit data for a "userpass" method, the user agent would POST to
+`/api/v1/auth/userpass`.
+
+## 4. Auth server checks the authn data against a policy
```
frrist, line 97: Step 3?

docs/docs/running-node/auth.md
```diff
@@ -0,0 +1,49 @@
...
+## By default
+
+With no specific authentication configuration supplied, Bacalhau runs in
+"anonymous mode" – which allows unidentified users limited control over the
+system. "Anonymous mode" is only appropriate for testing or evaluation setups.
+
+In anonymous mode, Bacalhau will allow:
+
+- Users identified by a self-generated private key to submit any job, cancel
+  their own jobs, read job lists and use `bacalhau describe`.
```
frrist, line 15: Should this be `bacalhau job describe` instead as we are moving away from the latter?

docs/docs/running-node/auth.md
```diff
@@ -0,0 +1,49 @@
...
+- Users not identified by any key to access other read-only endpoints, such as
+  node or agent information.
+
+## Restricting anonymous access
+
+Bacalhau auth is controlled by policies. Configuring the auth system is done by
+just supplying a different policy file.
```
frrist, line 22: nit:
```suggestion
Bacalhau auth is controlled by policies. Configuring the auth system is done by
supplying a different policy file.
```

docs/docs/running-node/auth.md
```diff
@@ -0,0 +1,49 @@
...
+
+Restricting API access to only users that have authenticated requires specifying
+a new **authorization policy**. You can download a policy that restricts
+anonymous access and install it by using:
+
+    curl -sL https://raw.githubusercontent.com/bacalhau-project/bacalhau/main/pkg/authz/policies/policy_ns_anon.rego -o ~/.bacalhau/no-anon.rego
```
frrist, line 28: we'll need to be careful to never move or renaming `policy_ns_anon.rego`. Maybe it would be better to reference this via a commit or a version? I don't really have a good suggestion here, just worried that it would be easy for docs to become out of data with this reference.

frrist: Or perhaps we could embed this rego policy in the binary? :thinking:

docs/docs/running-node/auth.md
```diff
@@ -0,0 +1,49 @@
...
+
+Once the node is restarted, accessing the node APIs will require the user to be
+authenticated, but by default will still allow users with a self-generated key
+to authenticate themselves.
+
+Restricting the list of keys that can authenticate to only a known set requires
+specifying a new **authentication policy**. You can download a policy that
+restricts key-based access and install it by using:
+
+    curl -sL https://raw.githubusercontent.com/bacalhau-project/bacalhau/main/pkg/authn/challenge/challenge_ns_no_anon.rego -o ~/.bacalhau/challenge_ns_no_anon.rego
```
frrist, line 39: similar comment to above.

frrist (approved):
Good stuff!

Couple nits, otherwise LGTM

cmd/cli/serve/util.go
simonwo: Yeah, not sure why this was here really. `Get` should work fine!

docs/docs/dev/auth_flow.md
simonwo: Oops, there used to be more steps...

docs/docs/running-node/auth.md
simonwo: I will just change it to not mention the explicit command.

docs/docs/running-node/auth.md
simonwo: I guess with a specific commit we'd then need to remember to update this if we make fixes or improvements. Either way it's just a careful review that is needed. Perhaps we can introduce an automated test or something?

Yeah, I don't actually imagine this being the long term user experience for doing this – but I think it's appropriate right now for beta users. I can imagine in the future we'd have explicit commands to set up users and things, or at least a dedicated tool, but we need to know more about what auth types people commonly want to use.

Is there any value in having something like `bacalhau auth install policy_ns_anon` or something? Just seems to be a bit overkill for where we're at?

frrist: > Is there any value in having something like bacalhau auth install policy_ns_anon or something? Just seems to be a bit overkill for where we're at?

seems like a good thing to keep in our back pocket for later, I think you are right - overkill for now.

PR #3316, opened 2024-01-25
[Documentation] add notes about registries and architecture

AlekseyKorzik:
Clarify the architecture and registry details

### DIFF
--- a/docs/docs/setting-up/other-specifications/engines/docker.md
+++ b/docs/docs/setting-up/other-specifications/engines/docker.md
@@ -41,5 +41,5 @@ In this example, the task will be executed inside an Ubuntu 20.04 Docker contain
 
 ## Notes
 
-- Currently, only images published to public registries are supported. Make sure your image is available in one of them, for example DockerHub, ghcr. Locally hosted registries are not supported.
+- Currently, only images published to public registries are supported. Make sure your image is available in one of them, for example DockerHub or ghcr. Private images can be retrieved from docker hub by [specifying your credentials](../../running-node/quick-start-docker.md#authenticate-with-docker-hub).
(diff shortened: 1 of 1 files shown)


docs/docs/setting-up/other-specifications/engines/docker.md
```diff
@@ -37,4 +37,9 @@ Engine:
     WorkingDirectory: "/app"
 ```
 
-In this example, the task will be executed inside an Ubuntu 20.04 Docker container. The entrypoint is overridden to execute a bash shell that runs an echo command. An environment variable MY_ENV_VAR is set with the value myvalue, and the working directory inside the container is set to /app.
\ No newline at end of file
+In this example, the task will be executed inside an Ubuntu 20.04 Docker container. The entrypoint is overridden to execute a bash shell that runs an echo command. An environment variable MY_ENV_VAR is set with the value myvalue, and the working directory inside the container is set to /app.
+
+## Notes
+
+- Currently, only images published to public registries are supported. Make sure your image is available in one of them, for example DockerHub, ghcr. Locally hosted registries are not supported.
```
rossjones, line 44: I misunderstood the question.  It might be worth trying something like ..

```suggestion
- Currently, only images published to public registries are supported. Make sure your image is available in one of them, for example DockerHub, or ghcr.  Private images can be retrieved from docker hub by [specifying your credentials](/running-node/quick-start-docker/#authenticate-with-docker-hub)
```

I've removed the bit about locally hosted registries because we haven't yet confirmed either way whether they work or not.

AlekseyKorzik: fixed

rossjones (approved):

PR #3344, opened 2024-01-30
Improve pagination and filtering of job list command.

rossjones:
Currently we allow users to define limit query params when listing jobs, but we always return an empty next_token (offset), which prevents users from actual pagination and querying the next set of jobs. 

This PR moves label filtering into the jobstore, and introduces a default sort order so that the offset/limit make sense. This allows for proper pagination from the CLI. The next-token is now used by the jobstore, but it only encodes the offset (e.g. it is the offset).  The next-token should really encode more information so that it can be re-constituted with the correct parameters.  Without this (as currently) it relies on the same sort being applied for each request.

### DIFF
--- a/pkg/jobstore/boltdb/store.go
+++ b/pkg/jobstore/boltdb/store.go
@@ -467,12 +467,8 @@ func (b *BoltJobStore) getJobsBuildList(tx *bolt.Tx, jobSet map[string]struct{},
 }
 
 func (b *BoltJobStore) getJobsWithinLimit(jobs []models.Job, query jobstore.JobQuery) []models.Job {
-	offset := query.Offset
-	if offset != 0 {
-		offset = math.Min(uint32(len(jobs)), offset)
-		if offset == uint32(len(jobs)) {
-			return []models.Job{}
-		}
+	if query.Offset >= uint32(len(jobs)) {
+		return []models.Job{}
 	}
 
 	limit := query.Limit
--- a/pkg/jobstore/boltdb/store.go
+++ b/pkg/jobstore/boltdb/store.go
@@ -21,6 +21,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/samber/lo"
 	bolt "go.etcd.io/bbolt"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 const (
@@ -290,19 +291,28 @@ func (b *BoltJobStore) jobExists(tx *bolt.Tx, jobID string) bool {
 }
 
 // GetJobs returns all Jobs that match the provided query
-func (b *BoltJobStore) GetJobs(ctx context.Context, query jobstore.JobQuery) ([]models.Job, error) {
-	var jobs []models.Job
+func (b *BoltJobStore) GetJobs(ctx context.Context, query jobstore.JobQuery) (*jobstore.JobQueryResponse, error) {
+	var response *jobstore.JobQueryResponse
 	err := b.database.View(func(tx *bolt.Tx) (err error) {
-		jobs, err = b.getJobs(tx, query)
+		response, err = b.getJobs(tx, query)
 		return
 	})
-	return jobs, err
+	return response, err
 }
 
-func (b *BoltJobStore) getJobs(tx *bolt.Tx, query jobstore.JobQuery) ([]models.Job, error) {
+func (b *BoltJobStore) getJobs(tx *bolt.Tx, query jobstore.JobQuery) (*jobstore.JobQueryResponse, error) {
 	if query.ID != "" {
 		job, err := b.getJob(tx, query.ID)
-		return []models.Job{job}, err
+
+		if query.Selector != nil {
+			// If we have a selector and it doesn't match, then it isn't an error
+			// just a non-match.
+			if !query.Selector.Matches(labels.Set(job.Labels)) {
+				return &jobstore.JobQueryResponse{}, nil
+			}
+		}
+
+		return &jobstore.JobQueryResponse{Jobs: []models.Job{job}}, err
 	}
 
 	jobSet, err := b.getJobsInitialSet(tx, query)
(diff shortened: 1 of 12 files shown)


pkg/jobstore/boltdb/store.go
```diff
@@ -290,19 +291,28 @@ func (b *BoltJobStore) jobExists(tx *bolt.Tx, jobID string) bool {
...
+
+		if query.Selector != nil {
+			// If we have a selector and it doesn't match, then it isn't an error
+			// just a non-match.
+			if !query.Selector.Matches(labels.Set(job.Labels)) {
+				return &jobstore.JobQueryResponse{}, nil
+			}
+		}
+
+		return &jobstore.JobQueryResponse{Jobs: []models.Job{job}}, err
```
wdbaruni, line 315: I think for historical reasons we used the same implementation for `get` and `list`, where `query.ID` was what differentiated between both. I don't think it makes sense to list jobs using `query.ID` as the ID is supposed to be unique. 

We should either remove `query.ID`, or at least don't add more features to it by allowing `query.ID` and `query.Selector` to mean something. What do you think?

pkg/jobstore/boltdb/store.go
```diff
@@ -325,7 +335,55 @@ func (b *BoltJobStore) getJobs(tx *bolt.Tx, query jobstore.JobQuery) ([]models.J
...
+	listSorter := func(i, j int) bool {
+		switch query.SortBy {
+		case "id":
+			if query.SortReverse {
+				// what does it mean to sort by ID? (rj) it's mostly meaningless and just
+				// gives us a lexicographic sort of the UUIDs.
+				return result[i].ID > result[j].ID
+			} else {
+				return result[i].ID < result[j].ID
+			}
```
wdbaruni, line 348: I can't think of good reasons for users to sort by `id`, specially that it is auto-generated by us. Does it make sense to remove this to simplify our indexing story and pagination story? We might introduce it later when we allow users to provide their `id` where it can mean something

pkg/jobstore/boltdb/store.go
```diff
@@ -325,7 +335,55 @@ func (b *BoltJobStore) getJobs(tx *bolt.Tx, query jobstore.JobQuery) ([]models.J
...
+		default:
+			// We apply modifytime as a default sort so that we can use it for pagination.
+			// Without a known default we won't have a stable sort that makes sense for
+			// offsets/limits.
+			if query.SortReverse {
+				return result[i].ModifyTime > result[j].ModifyTime
+			} else {
+				return result[i].ModifyTime < result[j].ModifyTime
+			}
```
wdbaruni, line 363: I agree `ModifyTime` sounds like a great default for better user experience, but my concern is that it is an always changing value for active jobs and will introduce indexing challenges along with unpredictable user experience.

I think `CreateTime` strikes a good balance between user experience and simplicity. And if that is the case, does it make sense to keep things simple and don't expose `SortBy` option until we have more valid fields to sort by?

pkg/jobstore/boltdb/store.go
```diff
@@ -431,13 +489,23 @@ func (b *BoltJobStore) getJobsBuildList(tx *bolt.Tx, jobSet map[string]struct{},
...
 
 func (b *BoltJobStore) getJobsWithinLimit(jobs []models.Job, query jobstore.JobQuery) []models.Job {
+	offset := query.Offset
+	if offset != 0 {
+		offset = math.Min(uint32(len(jobs)), offset)
+		if offset == uint32(len(jobs)) {
+			return []models.Job{}
+		}
+	}
+
```
wdbaruni, line 499: nit: `if offset >= len(jobs) { return []odels.Job{} }` might be more readable

pkg/jobstore/boltdb/store.go
```diff
@@ -431,13 +489,23 @@ func (b *BoltJobStore) getJobsBuildList(tx *bolt.Tx, jobSet map[string]struct{},
...
 	limit := query.Limit
 	if limit == 0 {
 		limit = uint32(len(jobs))
 	} else {
 		limit = math.Min(uint32(len(jobs)), limit+query.Offset)
 	}
 
+	fmt.Printf("%v : %v\n", query.Offset, limit)
+
```
wdbaruni, line 508: leftover print

pkg/jobstore/types.go
```diff
@@ -5,18 +5,31 @@ import (
...
+	SortReverse bool
+	Selector    labels.Selector
+}
+
+type JobQueryResponse struct {
+	Jobs            []models.Job
+	Offset          uint32 // Offset into the filtered results of the first returned record
+	Limit           uint32 // The number of records to return, 0 means all
+	NextOffset      uint32 // Offset + Limit of the next page of results, 0 means no more results
+	RecordsReturned uint32 // The number of jobs actually held in the Jobs field
```
wdbaruni, line 32: The way I see it, every data store implementation will have its own way to handle pagination and what it fields it requires to efficiently fetch the next set of results.

For that reason I recommend that `NextToken` to be a base64 string that is generated and decoded by the jobstore implementations. In boltdb, it can be an `offset` based on current implementation, along with sorting criteria to make sure the user is getting consistent results. We can add more fields in the future, or using different mechanism than offsets to gurantee consistent paginations.

If the user sends a follow up request with a sorting criteria doesn't match the one embedded in the token, then we can fail the job or ignore the one in the request.

Future implementations might encode a cursor reference and the requester node that handled the original request and has a cursor open to handle subsequent requests.

What do you think?

pkg/jobstore/boltdb/store.go
rossjones: I think we should remove `query.ID` and just fix up the legacy usage of it until we remove it.  I think removing IncludeTags and ExcludeTags would also be good, but might be hard to replicate unless there's a clear mapping from include/exclude to a selector.

pkg/jobstore/boltdb/store.go
rossjones: I think there's still value in sorting by modify time, even if it isn't the default, particularly for just seeing jobs with recent activity.  So maybe just swapping them around for now so that created_at is default and we can expand upon it later?

pkg/jobstore/types.go
rossjones: Yeah sounds good, search criteria, limit and offset in a base64 encoded blob. It _might_ get a bit unwieldy length wise, but perhaps we can more simply encode it without the verbosity of JSON (even | separated + base64 might be okay).

pkg/jobstore/boltdb/store.go
rossjones: Fixed in 30db3a3f0

pkg/jobstore/inmemory/inmemory.go
```diff
@@ -146,25 +136,54 @@ func (d *InMemoryJobStore) GetJobs(ctx context.Context, query jobstore.JobQuery)
...
+	// To be able to extract based on offset and limit, we need the records to be sorted in
+	// the same order for each query.  This is why we sort the list above and why the offset
+	// and limit are applied so late.
+	if query.Offset > 0 && query.Offset <= uint32(len(result)) {
+		result = result[query.Offset:]
+	}
+
+	if query.Limit > 0 && query.Limit <= uint32(len(result)) {
+		result = result[:query.Limit+query.Offset]
+	}
```
wdbaruni, line 177: result already starts from the offset after line 172. I think you only need to do `result[:query.limit]`. Are we still using inmemory store anywhere? Should we get rid of it?

pkg/jobstore/inmemory/inmemory.go
```diff
@@ -146,25 +136,54 @@ func (d *InMemoryJobStore) GetJobs(ctx context.Context, query jobstore.JobQuery)
...
+
+	if query.Selector != nil {
+		var filtered []models.Job
+		for _, j := range result {
+			if query.Selector.Matches(labels.Set(j.Labels)) {
+				filtered = append(filtered, j)
+			}
+		}
+		result = filtered
+	}
```
wdbaruni, line 166: Nice!

pkg/jobstore/inmemory/inmemory.go
rossjones: Yeah, let's delete it in a new PR, boltdb has had enough time to settle in now, and if we're adding NATS KV soon then even this may not be necessary.

wdbaruni (approved):

PR #3346, opened 2024-01-30
Stops linters from 'fixing' code

rossjones:
Currently we run golangci-lint run with the `--fix` flag that allows individual linters to make changes to the files where it finds problems. Ideally the linter should just flag the issue for the developer to fix, using --fix if they wish, but not making that choice on their behalf.


aronchick (approved):
LGTM

PR #3349, opened 2024-01-31
fix: auth docs and config set for Auth

frrist:
Read amended docs for new usage pattern


rossjones (approved):

docs/docs/running-node/auth.md
```diff
@@ -37,8 +43,7 @@ specifying a new **authentication policy**. You can download a policy that
 restricts key-based access and install it by using:
 
     curl -sL https://raw.githubusercontent.com/bacalhau-project/bacalhau/main/pkg/authn/challenge/challenge_ns_no_anon.rego -o ~/.bacalhau/challenge_ns_no_anon.rego
-    bacalhau config set Node.Auth.Methods.ClientKey.Type challenge
-    bacalhau config set Node.Auth.Methods.ClientKey.PolicyPath ~/.bacalhau/challenge_ns_no_anon.rego
+    bacalhau config set Auth.Methods '{Method: ClientKey, Policy: {Type: challenge, PolicyPath: ~/.bacalhau/challenge_ns_no_anon.rego}}'
```
rossjones, line 46: I think you need to escape the { and } with \ on this line (and below) as otherwise it is interpreted by the doc systems as mdx and it fails to parse.

frrist: Looks like you're right, thanks for pushing the change.

simonwo (approved):

PR #3354, opened 2024-01-31
decouple logsstream from transport layer

wdbaruni:
LogStream is tightly coupled to libp2p to stream logs from the compute node to the requester, which makes supporting NATS more challenging https://github.com/bacalhau-project/expanso-planning/issues/413

This PR decouples LogStream and makes fetching logs from the executor and compute node agnostic to how they are published to the requester.

The way this is achieved by utilizing channels for async vending of logs. The compute nodes return a channel for `AsyncResult[models.ExecutionLog]`, which is consumed by the transport layer (`bprotocol` in current implementation) and sent to the requester. A similar implementation is also introduced from the client to the requester node where instead of exposing websockets to the client user, the new SDK just returns a channel of `AsyncResult[models.ExecutionLog]`

This PR also migrates `logs` endpoint from the old APIs to the new version, implements logs SDK, and introduces helper methods to expose and consume additional websockets in the future.


wdbaruni:
@frrist as our golang expert, I would appreciate your review on the use of channels in this PR

frrist (commented):
Left a couple early review comments.

I have also tried this locally. The command:
```bash
bacalhau --api-host=0.0.0.0 docker run --follow ubuntu:latest -- "bash" "-c" 'for i in {1..60}; do echo "Iteration $i"; sleep 1; done'
```
works as expected. 
However both `job logs` and `logs` just return exit code zero with no output.

EDIT: I had assumed the `logs` and `job logs` commands would follow the logs, but follow defaults to `false`. I further assumed I would at least see the logs emitted since the job had started, but we use `time.Now()` if the request doesn't contain `WithHistory` - and there isn't a way to specify that on `job logs` so the value is also `false`. This means the default behavior of the `job logs` command is unfriendly to users - it won't follow the logs, and it will show you the logs that were emitted around the time the command was run, which is racy at best - almost always no output.

The action I would propose taking here is:
- ensure `--history` can be provided to the `job logs` command (currently there is no flag)
- set the default value of `history` to `true`

This will ensure users see at least the logs since the beginning of the execution rather than noting, or a single line if they time it right.

pkg/transport/bprotocol/constants.go
```diff
@@ -6,7 +6,7 @@ const (
 	BidAcceptedProtocolID = "/bacalhau/compute/bid_accepted/1.0.0"
 	BidRejectedProtocolID = "/bacalhau/compute/bid_rejected/1.0.0"
 	CancelProtocolID      = "/bacalhau/compute/cancel/1.0.0"
-	ExecutionLogsID       = "/bacalhau/compute/executionlogs/1.0.0"
+	ExecutionLogsID       = "/bacalhau/compute/execution_logs/1.0.0"
```
frrist, line 9: I think this is going to break compatibility with all previous version of this, any reason to change this?

pkg/publicapi/endpoint/orchestrator/job.go
```diff
@@ -344,3 +347,73 @@ func (e *Endpoint) jobResults(c echo.Context) error {
...
+		JobID:       jobID,
+		ExecutionID: args.ExecutionID,
+		WithHistory: args.WithHistory,
+		Follow:      args.Follow,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to open log stream for job %s: %w", jobID, err)
+	}
+
+	for {
```
frrist, line 408: purely a style nit, but slight preference for:
```suggestion
	for logMsg := range logstreamCh {
		if err = ws.WriteJSON(logMsg); err != nil {
		        return err
		}
	}
```

pkg/logger/dataframe.go
```diff
@@ -28,38 +28,36 @@ const (
...
-	df := DataFrame{}
+	df := &DataFrame{}
 	df.Tag = StreamTag(binary.LittleEndian.Uint32(header))
 	df.Size = int(binary.BigEndian.Uint32(header[4:]))
 	df.Data = make([]byte, df.Size)
 
 	n, err = reader.Read(df.Data)
 	if err != nil {
-		return DataFrame{}, err
+		return nil, err
```
frrist, line 57: This seems to be returning `EOF` before the container is done, related to my comment about `job logs` not working for me locally.

pkg/transport/bprotocol/compute_proxy.go
```diff
@@ -146,5 +149,60 @@ func proxyRequest[Request any, Response any](
...
+	go func() {
+		defer stream.Close() //nolint:errcheck
+		defer close(ch)
+		for {
+			response := new(concurrency.AsyncResult[Response])
+			err = json.NewDecoder(stream).Decode(response)
+			if err != nil {
+				_ = stream.Reset()
+				if err != io.EOF {
+					response.Err = fmt.Errorf("%s: failed to decode response from peer %s: %w", reflect.TypeOf(request), destPeerID, err)
```
frrist, line 197: `%T` should work instead of reflect.

pkg/lib/concurrency/async_result.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// MarshalJSON customizes the JSON representation of AsyncResult
+func (ar AsyncResult[T]) MarshalJSON() ([]byte, error) {
+	type Alias AsyncResult[T]
+	return json.Marshal(&struct {
+		Error string `json:"error,omitempty"`
+		*Alias
+	}{
+		Error: errorToString(ar.Err),
```
frrist, line 21: could also do:
```suggestion
                Error: fmt.Sprintf("%s",ar.Err)
```
instead of a new method.

pkg/compute/logstream/live_streamer.go
```diff
@@ -0,0 +1,78 @@
...
+	"io"
+
+	"github.com/bacalhau-project/bacalhau/pkg/lib/concurrency"
+	"github.com/bacalhau-project/bacalhau/pkg/logger"
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+)
+
+type LiveStreamerParams struct {
+	Reader io.Reader
+	Buffer int
```
rossjones, line 14: This is set but I don't think it's used?

pkg/nats/proxy/compute_proxy.go
```diff
@@ -50,10 +52,10 @@ func (p *ComputeProxy) CancelExecution(
 		ctx, p.conn, request.TargetPeerID, CancelExecution, request)
 }
 
-func (p *ComputeProxy) ExecutionLogs(
-	ctx context.Context, request compute.ExecutionLogsRequest) (compute.ExecutionLogsResponse, error) {
-	return proxyRequest[compute.ExecutionLogsRequest, compute.ExecutionLogsResponse](
-		ctx, p.conn, request.TargetPeerID, ExecutionLogs, request)
+func (p *ComputeProxy) ExecutionLogs(ctx context.Context, request compute.ExecutionLogsRequest) (
+	<-chan *concurrency.AsyncResult[models.ExecutionLog], error) {
+	//TODO implement me
```
rossjones, line 57: Is this coming in a follow up PR or a later commit to this one?

pkg/compute/logstream/completed_streamer.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		ch <- &asyncResult
+	}
+	if scanner.Err() != nil {
+		asyncResult := concurrency.AsyncResult[models.ExecutionLog]{
+			Err: fmt.Errorf("failed to read output: %w", scanner.Err()),
+		}
+		ch <- &asyncResult
+	}
+}
```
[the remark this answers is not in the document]
wdbaruni: Good point. That would allow `process` to exit if the channel was full and the context is done, right?

pkg/config/configenv/dev.go
```diff
@@ -147,6 +147,9 @@ var DevelopmentComputeConfig = types.ComputeConfig{
 		Duration:  types.Duration(1 * time.Hour),
 		Frequency: types.Duration(1 * time.Hour),
 	},
+	LogStreamConfig: types.LogStreamConfig{
+		ChannelBufferSize: 10,
```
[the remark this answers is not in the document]
wdbaruni: this is number of log lines and not bytes. How many `models.ExecutionLog` can fit in the channel before we block if we have a slow channel consumer

pkg/compute/logstream/completed_streamer.go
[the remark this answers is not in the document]
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/3354/commits/80d1eef8857a74bd9ceb967d9958041bbcd42308

cmd/util/logging.go
```diff
@@ -31,142 +33,53 @@ func Logs(cmd *cobra.Command, jobID string, follow, history bool) error {
...
-type Msg struct {
-	Tag          uint8
-	Data         string
-	ErrorMessage string
-}
-
-func readLogoutput(ctx context.Context, cmd *cobra.Command, conn *websocket.Conn) error {
+func readLogoutput(ctx context.Context, logsChannel <-chan *concurrency.AsyncResult[models.ExecutionLog]) error {
 	interrupt := make(chan os.Signal, 1)
 	signal.Notify(interrupt, os.Interrupt)
```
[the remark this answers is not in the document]
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/3354/commits/50012081cddce56e38eb7a30095ff64ff6a9790c

Though I am sure we can cleanup interruption handling from other commands as well

pkg/transport/bprotocol/constants.go
wdbaruni: yeah we have a new protocol for log vending between the compute node and the requester. Meaning we now need both compute and requester nodes to be on the same version for logging to work. We can utilize `NewMinVersionNodeRanker` to make sure requester won't communicate with lower version compute nodes, but I've checked our graphana metrics for how often users call our current logging API, and I barely saw 4 calls the past month. 

This should be a concern since it is not an actively used API, and just a client update is needed

pkg/lib/concurrency/async_result.go
wdbaruni: not efficient, specially that this is going to be called each time a log record is serialized and deserialized when sent over the wire. A new method might be justified here

rossjones (commented):
LGTM. As per Frrst's comments, I believe the intent is that `logs` will show the history and not follow, whilst -f would show the history AND follow so defaults should reflect that.

Haven't approved because of forrest's comments.

pkg/compute/logstream/live_streamer.go
wdbaruni: good catch. fixed in  https://github.com/bacalhau-project/bacalhau/pull/3354/commits/81fe8c5645703d466ae06b2ed828aac4e5fd2bf4

pkg/nats/proxy/compute_proxy.go
wdbaruni: yeah in a follow up PR. I have an open issue for that https://github.com/bacalhau-project/expanso-planning/issues/413

pkg/transport/bprotocol/compute_proxy.go
wdbaruni: Fixed in https://github.com/bacalhau-project/bacalhau/pull/3354/commits/80d1eef8857a74bd9ceb967d9958041bbcd42308

pkg/publicapi/endpoint/orchestrator/job.go
wdbaruni: Fixed in https://github.com/bacalhau-project/bacalhau/pull/3354/commits/80d1eef8857a74bd9ceb967d9958041bbcd42308

cmd/util/logging.go
```diff
@@ -31,142 +33,53 @@ func Logs(cmd *cobra.Command, jobID string, follow, history bool) error {
...
-
-	for {
-		select {
-		case <-done:
-			exiting = true
-			_ = conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
-			return nil
 		case <-ctx.Done():
-			exiting = true
 			return nil
```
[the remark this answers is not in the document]
wdbaruni: Fixed in https://github.com/bacalhau-project/bacalhau/pull/3354/commits/80d1eef8857a74bd9ceb967d9958041bbcd42308

wdbaruni:
I agree with showing the history to be enabled by default, though I hate making a boolean value true by default as we will end up with inconsistent cli and api behaviours. I'll change `history` to `tail`, and make it false by default. Does that make sense?

wdbaruni:
Switched to using `--tail` and also added `--execution-id` to allow users to fetch logs for a specific execution instead of letting the requester chose one
https://github.com/bacalhau-project/bacalhau/pull/3354/commits/c9d07be61cdc23e289dbc1630b74c638284af275

rossjones (approved):

pkg/transport/bprotocol/constants.go
frrist: > yeah we have a new protocol for log vending between the compute node and the requester. 

Could we just bump the protocol version then?

pkg/compute/logstream/completed_streamer.go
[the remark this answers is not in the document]
frrist: > That would allow process to exit if the channel was full and the context is done, right?

If the channel is full it will block (thus my comments about buffers) if ctx is done it exits.

cmd/cli/job/logs.go
```diff
@@ -36,16 +40,31 @@ func NewLogCmd() *cobra.Command {
...
+				Follow:      options.Follow,
+				Tail:        options.Tail,
+			}
+			if err := util.Logs(cmd, opts); err != nil {
 				util.Fatal(cmd, err, 1)
 			}
 		},
 	}
 
+	logsCmd.PersistentFlags().StringVarP(
```
frrist, line 55: ❤️

frrist (approved):

PR #3357, opened 2024-02-01
Configuration changes for webui

aronchick:


webui/src/layout/Header/Header.tsx
```diff
@@ -1,32 +1,29 @@
...
-        <BacalhauLogo className={styles.logo} height="24" />
-        <div className={styles.pageTitle}>{pageTitle}</div>
-        <div className={styles.searchBar}>
-          {/* Placeholder for search bar */}
-          {/* <input type="text" placeholder="Search..." /> */}
-        </div>
+export const Header: React.FC<HeaderProps> = ({ pageTitle, collapsed }) => (
+  <header className={styles.header} data-collapsed={collapsed}>
+    <div className={styles.left}>
+      <SVGImage src="../../images/bacalhau.svg" alt="Bacalhau Icon" />
```
simonwo, line 13: As I wrote in the previous PR, you are using `SVGImage` here and the old style `... as ReactComponent` in different places in the code. Please can you explain why or unify the two methods to a single approach.

webui/src/layout/Sidebar/Sidebar.tsx
```diff
@@ -1,19 +1,22 @@
...
-import { ReactComponent as JobsIcon } from "../../images/jobs-icon.svg";
-import { ReactComponent as NodesIcon } from "../../images/nodes-icon.svg";
-import { ReactComponent as Cogwheel } from "../../images/cogwheel.svg";
+// layout/Sidebar/Sidebar.tsx
+import React from "react"
+import { Link } from "react-router-dom"
+import { ReactComponent as BacalhauIcon } from "../../images/bacalhau.svg"
+import { ReactComponent as JobsIcon } from "../../images/jobs-icon.svg"
+import { ReactComponent as NodesIcon } from "../../images/nodes-icon.svg"
+import { ReactComponent as CogwheelIcon } from "../../images/cogwheel.svg"
```
simonwo, line 7: Here is where you are using the other approach.

webui/src/layout/Header/Header.tsx
aronchick: yes - because i had several requests to break up the PRs - the PR you mention about using the correct type (SVGImage) instead of the ReactComponent is in a later PR. However, every so often later PRs leak through.

webui/src/layout/Sidebar/Sidebar.tsx
aronchick: See above

simonwo (approved):

PR #3374, opened 2024-02-01
3368 - adding svgs in static dir and tests

aronchick:


simonwo (commented):
You presumably need to remove the existing SVGs from `src/images`? Otherwise they'll be doubled up in the repository.

aronchick:
> You presumably need to remove the existing SVGs from `src/images`? Otherwise they'll be doubled up in the repository.

sigh - excellent point (thought I had). Updated!

PR #3382, opened 2024-02-01
3381 - Just one yaml

### LINKED ISSUE #3381: Pick a YAML library
We currently have a few different yaml parsers installed and in use in various places.

We should just pick one and benefit from fewer dependencies. Currently we are dependent, in different places, on different behaviour of the various libraries.  

The marshaller code depends on the k8s version, the authz code and config depend on the go-yaml (gopkg.in/yaml.v3) version.

rossjones:
Currently we're using 3 different YAML parsing libraries of various versions.  This is unnecessary so this PR moves us towards using "gopkg.in/yaml.v3" (go-yaml). Other unused yaml parsers are removed from the list of installed mods.

Resolves #3381

### DIFF
--- a/cmd/cli/config/set_test.go
+++ b/cmd/cli/config/set_test.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/require"
-	"sigs.k8s.io/yaml"
+	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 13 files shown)


aronchick (approved):

rossjones:
It appears different features of each of the yaml libraries were being used and that makes what should be a simple PR more complex.  Will update the ticket about the work but this version of the fix won't work as we are really dependent on the behaviour of the k8s version.

PR #3394, opened 2024-02-02
Select network address based on type

### LINKED ISSUE #3322: environment output should automatically update with external IP (if available) instead of 0.0.0.0

rossjones:
Currently we use 0.0.0.0 a lot when a bind address is required, and as a result we end up binding to all addresses.  This is both confusing and potentially an issue if we can't more closely control which address types to use (e.g. we may end up binding compute nodes to public addresses unintentionally).

This PR provides functionality to iterate the addresses available on the current host and to filter them based on network type.  This means we can filter on public, private, local or linklocal addresses for various use-cases.  

It also changes the configuarion getter for Server API Host to allow the address type to be specified in place of an IP address.  This means we can set Server.API.Host to an IP address as currently, or "public" to get a public address, or "private" to get a private address. Care must be taken as most dev laptops won't have a public address and as the 'production' environment is the default, we can't (and this PR doesn't) default to 'public'.    It also only uses the first address of a type it finds, and so does not handle special cases which may have multiple public addresses.

This should resolve #3322 by allowing us to specify an address type rather than 0.0.0.0 once we are using the appropriate env.

Future changes should allow for a comma separated list to also support:

* "public,private" 
* "public,100.10.10.100"

### DIFF
--- a/pkg/config/getters.go
+++ b/pkg/config/getters.go
@@ -2,16 +2,19 @@ package config
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/multiformats/go-multiaddr"
+	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 
 	"github.com/bacalhau-project/bacalhau/pkg/config/types"
+	"github.com/bacalhau-project/bacalhau/pkg/lib/network"
 	"github.com/bacalhau-project/bacalhau/pkg/logger"
 )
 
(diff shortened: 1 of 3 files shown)


wdbaruni (approved):
Do we know how other similar systems handle this?

rossjones:
> Do we know how other similar systems handle this?

Mostly by being explicit about the IP you want to bind to, k3s for instance auto binds to the first public address unless you specify an explicit address.  I think we've been over-using 0.0.0.0 so that we don't have to think too much about which address to use.  

This really only gets us an easy way to get the first public, or private address for various uses, the best solution would be to be explicit on which IP to bind to.

PR #3400, opened 2024-02-02
deps: remove job info consumer

frrist:
This doesn't appear to be a service we run, and as we move to NATs will no longer work, I think.


wdbaruni (approved):

simonwo:
Can we merge this now? The app doesn't even build anymore.

PR #3404, opened 2024-02-03
First Storyboard web ui tools for JobTable

aronchick:
Storyboard is now working great. In the process, I had to finalize swapping out the remainder of the (out of date) create react app for Webpack, which is also now working properly.


Makefile
```diff
@@ -160,10 +160,11 @@ build-dev: build-ci
 ################################################################################
 # Target: build-webui
 ################################################################################
-WEB_GO_FILES := $(shell find webui -name '*.go')
```
rossjones, line 163: Looks like it's missing the initial W, where WEB_GO_FILES is used by the BINARY_PATH target in this Makefile.

webui/tests/msw/storybookHandlers.ts
```diff
@@ -0,0 +1,35 @@
+// src/mocks/handlers.js
+import { http, HttpResponse } from "msw"
+import { Job } from "../../src/helpers/jobInterfaces"
+import { Node } from "../../src/helpers/nodeInterfaces"
+
+export const JOBS_RETURN_LIMIT = 10
+export const NODES_RETURN_LIMIT = 10
+// const BASE_URL = "https://localhost:1234"
```
rossjones, line 8: Can probably remove if we're not using this in `jobHandlers`

rossjones (approved):
Looks like breakage in the Makefile, but otherwise 👍🏻

Makefile
```diff
@@ -160,10 +160,11 @@ build-dev: build-ci
...
+EB_GO_FILES := $(shell find webui -name '*.go')
```
rossjones, line 163: Looks like it's missing the initial W, where WEB_GO_FILES is used by the BINARY_PATH target in this Makefile.

PR #3407, opened 2024-02-05
NATS based LogStream 

wdbaruni:
This PR introduces a NATS `stream` package that enables response streams patterns. NATS currently supports request/response patterns which is great when expecting a single response per request. However, this is not sufficient to serve job logs where a user initiates a request to stream logs, and the compute node start serving those logs.

## Overview of `stream` package

This package implements a client leveraging NATS for sending and receiving streaming data records. It abstracts NATS connections, subscriptions, and message handling complexities, offering a simplified interface for data streaming. The client supports multiplexing multiple streams over a single NATS subscription, handling responses from different streams using a unique token-based mechanism. Additionally, the package introduces a Writer component, designed to abstract the complexities of data encoding and NATS publishing into a simple, intuitive interface.

## How It Works

- **The Client** part of the package manages dynamic inboxes for each streaming session, facilitating the sending of data and listening for responses on dedicated subjects. It leverages the NATS publish-subscribe model for asynchronous communication, efficiently routing and correlating messages to their respective streams.
  
- **The Writer** component allows for easy publishing of structured data to any NATS subject. It integrates tightly with the Client, utilizing the same connection for streamlined data streaming. The Writer simplifies the publication process, automatically handling data serialization and supporting graceful stream closure with custom codes.

## Multiplexing Streams

To efficiently handle multiple streams, the client uses a single wildcard subscription for all responses. Each request is sent with a unique response subject (derived from a base inbox prefix), with responses routed back to this subject. The client demultiplexes incoming messages by extracting a token from the response subject, identifying the correct stream (or "bucket") for the message. This approach allows managing multiple concurrent streams with minimal overhead, leveraging NATS's lightweight subjects and messaging capabilities.

## Key Features

- **Asynchronous Streaming**: Supports asynchronous data streaming, allowing clients to send data and receive responses without blocking.

- **Context Support**: Integrates with Go's context package for timeouts, cancellation, and deadlines for streaming requests.

- **Multiplexing**: Efficiently multiplexes multiple streams over a single NATS subscription, using unique response subjects for message correlation.

- **Error Handling**: Provides robust error handling, including custom error codes for stream-related errors (e.g., bad data, normal closure).

- **Data Publication**: The Writer simplifies structured data publishing to NATS, with support for automatic serialization and stream closure signals.

## Usage

Initialize the client with a NATS connection and use the provided methods to send streaming requests and handle responses. The client manages the NATS subscription and response routing, simplifying the process of working with streaming data.

### Example

```go
params := stream.ClientParams{Conn: natsConn}
client, err := stream.NewClient(params)
if err != nil {
    log.Fatal(err)
}

ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()

responseChan, err := client.OpenStream(ctx, "subject", []byte("data"))
if err != nil {
    log.Fatal(err)
}

for asyncResult := range responseChan {
    if asyncResult.Err != nil {
        log.Printf("Received an error: %v", asyncResult.Err)
        continue
    }
    log.Printf("Received data: %s", asyncResult.Value)
}

### Remaining Work
- The compute node has no knowledge when the requester is no longer interested to stream logs. We need to implement a heartbeat or a feedback loop to let the compute node stop publishing logs when there are no active consumers
- The requester node has no knowledge if the compute node is still alive and publishing logs. We do have a graceful handling when the compute node reaches the end of the stream where it informs the requester node that no more logs are coming, but we don't handle ungraceful termination of compute nodes

### DIFF
--- a/pkg/lib/concurrency/channel_transformer.go
+++ b/pkg/lib/concurrency/channel_transformer.go
@@ -18,13 +18,7 @@ func AsyncChannelTransform[In any, Out any](
 				if !ok {
 					return // Input channel closed
 				}
-				result := new(AsyncResult[Out])
-				transformedValue, err := transform(msg)
-				if err != nil {
-					result.Err = err
-				} else {
-					result = transformedValue
-				}
+				result := NewAsyncResult[Out](transform(msg))
 
 				select {
 				case output <- result:
--- a/pkg/lib/concurrency/channel_transformer.go
+++ b/pkg/lib/concurrency/channel_transformer.go
@@ -1,0 +1,41 @@
+package concurrency
+
+import "context"
+
+// AsyncChannelTransform copies messages from an input channel, transforms them, and sends them to an output channel.
+func AsyncChannelTransform[In any, Out any](
+	ctx context.Context,
+	input <-chan In,
+	bufferCapacity int,
+	transform func(In) (*AsyncResult[Out], error)) <-chan *AsyncResult[Out] {
+	output := make(chan *AsyncResult[Out], bufferCapacity)
+
+	go func() {
+		defer close(output)
+		for {
+			select {
+			case msg, ok := <-input:
+				if !ok {
+					return // Input channel closed
+				}
+				result := new(AsyncResult[Out])
+				transformedValue, err := transform(msg)
+				if err != nil {
+					result.Err = err
+				} else {
+					result = transformedValue
+				}
+
+				select {
+				case output <- result:
+				case <-ctx.Done():
+					return
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+
+	return output
+}
--- a/pkg/lib/concurrency/channel_transformer.go
+++ b/pkg/lib/concurrency/channel_transformer.go
@@ -1,13 +1,15 @@
 package concurrency
 
-import "context"
+import (
+	"context"
+)
 
 // AsyncChannelTransform copies messages from an input channel, transforms them, and sends them to an output channel.
 func AsyncChannelTransform[In any, Out any](
 	ctx context.Context,
-	input <-chan In,
+	input <-chan *AsyncResult[In],
 	bufferCapacity int,
-	transform func(In) (*AsyncResult[Out], error)) <-chan *AsyncResult[Out] {
+	transform func(In) (Out, error)) <-chan *AsyncResult[Out] {
 	output := make(chan *AsyncResult[Out], bufferCapacity)
 
 	go func() {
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: pkg/lib/concurrency/channel_transformer.go [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+				if !ok {
+					return // Input channel closed
+				}
+				result := new(AsyncResult[Out])
+				transformedValue, err := transform(msg)
+				if err != nil {
+					result.Err = err
+				} else {
+					result = transformedValue
+				}
```
rossjones, line 27: If the `transform` function returned an Out and an error then this code could be...

```suggestion
				result := NewAsyncResult[Out](transform(msg))
```

pkg/nats/proxy/compute_handler.go
```diff
@@ -59,32 +67,33 @@ func handleRequest(msg *nats.Msg, handler *ComputeHandler) {
...
+		processAndRespond(ctx, handler.conn, msg, handler.computeEndpoint.BidAccepted)
 	case BidRejected:
-		processAndRespond(ctx, msg, handler.computeEndpoint.BidRejected)
+		processAndRespond(ctx, handler.conn, msg, handler.computeEndpoint.BidRejected)
 	case CancelExecution:
-		processAndRespond(ctx, msg, handler.computeEndpoint.CancelExecution)
+		processAndRespond(ctx, handler.conn, msg, handler.computeEndpoint.CancelExecution)
 	case ExecutionLogs:
-		processAndRespond(ctx, msg, handler.computeEndpoint.ExecutionLogs)
+		processAndStream(ctx, handler.streamingClient, msg, handler.computeEndpoint.ExecutionLogs)
```
rossjones, line 78: Unless we're expected branches that don't call processAndRespond() perhaps it might be easier to read if each branch just set the handler function (with ExecutionLogs also overriding the connection) and make a single call after the switch?

rossjones (approved):
Couple of minor points, but LGTM.

I think my only potential concern would be that this might be a good opportunity to do less JSON serialising, but perhaps we can look at that in other PRs.

pkg/lib/concurrency/channel_transformer.go
wdbaruni: Wow .. Nice!

pkg/nats/proxy/compute_handler.go
wdbaruni: That would be nice, though not sure how it will work with ExecutionLogs. They call a different method `processAndStream`, which accepts a streaming client instead of a conn. What is even worse is go generics! You can define a variable to hold the different handlers as each expects a different Request and Response type

rossjones: Yeah it may not be worth it in the end if there's still a special case for ExecutionLogs.

wdbaruni:
> I think my only potential concern would be that this might be a good opportunity to do less JSON serialising, but perhaps we can look at that in other PRs.

It also cross my mind while working on the writer. While I can introduce proto when encoding `StreamingMsg`, the payload within the `StreamingMsg` will most likely continue to be json for now as we don't support proto for our models yet. I can take a look at this while deprecating the old models

PR #3414, opened 2024-02-06
fix python ci build

wdbaruni:
CircleCI has been failing because of python builds and tests expecting a python version less than 3.12. Even though we were installing python3.10 in our config using `apt get`, circleci appears to be using `pyenv` and was ignoring the version we installed.

```
ERROR: Package 'flytekit' requires a different Python: 3.12.1 not in '<3.12,>=3.8'
```


rossjones (approved):
👍🏻

PR #3417, opened 2024-02-06
Make publisher optional in CLI

rossjones:
Currently we default to IPFS for the publisher if one is not specified. This commit changes this to allow no publisher to be specified.  It is expected that this will allow us to run jobs with no publisher, and also to have the requester node choose which publisher to use.

### DIFF
--- a/cmd/util/download.go
+++ b/cmd/util/download.go
@@ -34,7 +34,9 @@ func DownloadResultsHandler(
 
 	if len(response.Results) == 0 {
 		// No results doesn't mean error, so we should print out a message and return nil
-		cmd.Println("no results found")
+		cmd.Println("No results found")
+		cmd.Println("You can check the logged output of the job using the logs command.")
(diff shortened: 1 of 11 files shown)


cmd/util/download.go
```diff
@@ -33,7 +33,9 @@ func DownloadResultsHandler(
 	}
 
 	if len(response.Results) == 0 {
-		return fmt.Errorf("no results found")
+		// No results doesn't mean error, so we should print out a message and return nil
+		cmd.Println("no results found")
+		return nil
 	}
```
wdbaruni, line 39: should we give a hint to use `bacalhau logs` instead? or maybe just fallback to fetching the logs if no result was found?

wdbaruni (approved):
LGTM

cmd/util/download.go
rossjones: I think a warning to use logs might be best. I'll add one.

PR #3422, opened 2024-02-06
Remove inmemory jobstore

### LINKED ISSUE #3418: Delete the inmemory job store
This is no longer required as we the boltdb store performs well enough, and as we add another jobstore the inmemory one is just more unused code to maintain.

rossjones:
We no longer need the inmemory version of the jobstore, and so this commit removes it from the codebase.

This should resolve #3418


wdbaruni:
should we also remove the in-memory executor store and inlocal store?

wdbaruni (approved):

PR #3430, opened 2024-02-08
Adds support for specifying a default publisher on the server

rossjones:
As it is no longer mandatory to have a publisher defined for a job (either by the user, or using the default) there is a possibility of confusion when stuff doesn't continue to default to IPFS.

To allow for the old behaviour to continue as a short-term measure, this commit introduces the `--default-publisher` option for the devstack and requester/compute node (launched by serve).

Using the `--default-publisher` flag is similar to specifying a publisher, and so to continue the same behaviour as we currently have where the CLI has a default, we would start the node with `--default-publisher ipfs`.

This has been implemented for both the requester and orchestrator (v1 and v2) job submissions.


wdbaruni:
What would be nice and a generic solution is to allow operators to define job defaults at the requester level, so that jobs submitted with missing fields, will get populated by the default values configured at the requester. 

We are already doing something hacky today for job execution timeouts ([here](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/orchestrator/transformer/job.go#L19)), and what would make things better, enable default publishers, and more future default configs is to allow the requester config to accept a full job spec, and then do a merge in the transformer with the submitted job. 

Actually if we do this right, we might get rid of more job configurations, such as `DefaultJobResourceLimits`

Also, I am hoping we slow down on adding more cli flags and push users more towards the config file. `bacalhau serve --help` is becoming unreadable, with lots of configurations that we don't expect users to change. Maybe we can add a `-c` flag that allows users to set configurations through the cli, but not a flag per config. For example, `bacalhau serve -c "Node.Requester.JobStore.Path=/mypath"`. Something to talk about on the next engineering standup.

rossjones:
@wdbaruni would you like me to change this PR to take a full 'defaults' job spec? My initial concern was working around users no longer having a default publisher in the CLI, but happy to expand the scope if you like.

wdbaruni (approved):

wdbaruni:
@rossjones I approved the PR in case we see value in exposing a flag for a default publisher. Though if we enable users to define default job specs in the requester's config, then it should solve the problem where users won't need to define a publisher when submitting jobs

PR #3431, opened 2024-02-08
Adds a 'local' publisher

rossjones:
Adds a new publisher type that allows for publishing to the local disk of the compute node. This allows users to test out the publisher functionality without needing to set up a remote storage service making it easier to get started with Bacalhau.  Currently this implemention is does not include features such as content lifetimes, or secure access, and so is not recommended for production use yet. This publisher is useful for testing and development purposes.

The local publisher comprises two parts. The publisher is responsible for taking the output from a job, compressing it and moving it to a location where it can be access by the second part.  This part is the HTTP server which delivers the content back to the caller. 

By default the HTTP server listens on port 6001, but this can be changed by setting the `--local-publisher-port` flag on the compute node.  The HTTP server will serve the content from the directory specified by the `local-publisher-directory` flag.  If this is not set, the HTTP server will serve the content from a subdirectory of the configured bacalhau storage directory. Finally the `--local-publisher-address` flag can be used to specify the address that the HTTP server listens on.  The default for this varies by environment, in test and development environments it uses `localhost`, in production environments it uses `public` to obtain a public address. Of course, you can set these values in config in preference to using the command line flags if the defaults are not suitable.


cmd/cli/devstack/devstack.go
```diff
@@ -66,6 +66,7 @@ func NewCmd() *cobra.Command {
 	ODs := newDevStackOptions()
 	IsNoop := false
 	devstackFlags := map[string][]configflags.Definition{
+		"local_publisher":       configflags.LocalPublisherFlags,
```
wdbaruni, line 69: these configs are at the node level and not stack level. Should we drop them from here and just let devstack choose the ports?

cmd/util/flags/configflags/local_publisher.go
```diff
@@ -0,0 +1,24 @@
+package configflags
+
+import "github.com/bacalhau-project/bacalhau/pkg/config/types"
+
+var LocalPublisherFlags = []Definition{
```
wdbaruni, line 5: does it make sense to allow users to disable this publisher? I would assume we want to disable this our demo network and in marketplace setup as well

ops/terraform/main.tf
```diff
@@ -295,6 +295,7 @@ resource "google_compute_firewall" "bacalhau_ingress_firewall" {
       "4001",  // ipfs swarm
       "1234",  // bacalhau API
       "1235",  // bacalhau swarm
+      "6001",  // local publisher download
```
wdbaruni, line 298: I think we shouldn't enable this publisher in our public network, but at the same time we need to test it in different environments 🤷‍♂️

pkg/publisher/local/server.go
```diff
@@ -0,0 +1,87 @@
...
+	readHeaderTimeout = 3 * time.Second
+	readTimeout       = 3 * time.Second
+)
+
+func NewLocalPublisherServer(ctx context.Context, config types.LocalPublisherConfig) *LocalPublisherServer {
+	return &LocalPublisherServer{
+		rootDirectory: config.Directory,
+		address:       resolveAddress(ctx, config.Address),
+		port:          config.Port,
+		stopChan:      make(chan struct{}),
```
wdbaruni, line 31: maybe we can use a sub-context with a cancel fn instead of a separate stopChan

pkg/node/compute.go
```diff
@@ -183,6 +186,21 @@ func NewComputeNode(
...
+		}
+		if ok {
+			log.Ctx(ctx).Info().Msg("**** local publisher is installed and server being started")
+			localPublisherServer = local.NewLocalPublisherServer(ctx, config.LocalPublisher)
+			go localPublisherServer.Start(ctx)
+		}
+	} else {
+		log.Ctx(ctx).Error().Err(err).Msg("local publisher not installed")
+	}
+
```
wdbaruni, line 203: This is a bit difficult to read. Does it make sense to just have the local publisher to be the one running the web server? This allows `IsInstalled()` to return true only if the server is actually running. Right now we are enabling the publisher without knowing if it is running

cmd/util/flags/configflags/local_publisher.go
rossjones: Yes, can't we use `--disabled-publisher`/`types.NodeDisabledFeaturesPublishers` to do that?

pkg/publisher/local/server.go
rossjones: I tried that originally, but the context documentation actually says:

```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx
```

which makes it harder to keep a sub-context around.

pkg/node/compute.go
rossjones: Yes, we can do that, and was actually my initial approach.  I backed it out and I can't remember why, but will put it back.

rossjones: Ah I remember, it was being able to shutdown the server in cleanup as there's no notification to the poublishers that they're being shutdown.

pkg/publisher/local/server.go
rossjones: Actually, never mind, I got it to work as we'd expect with the context.

cmd/util/flags/configflags/local_publisher.go
wdbaruni: I didn't know that!

rossjones:
> Minor comments. How are you going to make this the default publisher instead of IPFS? Are you going to use the newly added `--default-pubisher` flag in the requester side?

Yes.

I've reverted the firewall change here, and will set up the default for the public network in a follow on PR so it is easy to revert when we don't want it.

frrist (commented):
Looks good, I like that we can now read and write data from disk with jobs! :fish: 

I'd like to have some more conversation around the mandatory compression before merging.

frrist (dismissed):
On second pass this is more complicated than I thought - why do we need a server for the compute node to write data to it's disk? 

I feel like this should follow a similar pattern to file inputs which use `types.NodeAllowListedLocalPaths` to determine if the path can be read from. Could we re-use this for writing results as well?

rossjones:
> On second pass this is more complicated than I thought - why do we need a server for the compute node to write data to it's disk? 

The server is necessary for clients to be able to obtain content from the compute nodes. Once the data is published, the user is given a URL to this server.

frrist:
I see, understood. Then my only other comment I'd like more discussion on is the compression by default

wdbaruni:
What is your concern with compression? We need to archive the data to serve it in a single response. Compression sounds obvious to me.

frrist:
As stated here: https://github.com/bacalhau-project/bacalhau/pull/3431#discussion_r1489906299

> As a user publishing the result of a job, I'd be caught off guard by the published data being a tarball when that wasn't what I published. And if I have subsequent jobs reading this data (file inputs) I'll then need to handle the untar-ing in the job, which is unnecessary toil imo.

As a user, if I wanted to publish a tarball I'd go-ahead do that explicitly, but I would _never_ want (or expect) the application to do that without me instructing it to.

> We need to archive the data to serve it in a single response

Do we really _need_ that? Or is it just convenient? Gut check says it's the latter.

If we want a single response surely we could dynamically packaged, transferred, and unpacked the results instead? Isn't this what we already half-do in the case of get for S3 - decompress the data on write?

wdbaruni:
bringing back the discussion we had with S3 publisher. How the publisher stores the data doesn't have to match what the user generated, specially that `bacalhau get` will handle the decompression. Users are not expected to access data directory, and each publisher should store data in its own efficient way for both storage and retrievals. 
I hope we conclude this discussion here as a significant time was already spent while discussing S3 downloader

wdbaruni (approved):
LGTM, but Forrest has some comments/questions before merging

docs/docs/setting-up/other-specifications/publishers/local.md
```diff
@@ -0,0 +1,43 @@
...
+## Local Publisher Parameters
+The local publisher requires no specific parameters to be defined in the publisher specification. The user only needs to indicate the publisher type as "local", and Bacalhau handles the rest. Here is an example of how to set up a Local Publisher in a job specification.
+
+```yaml
+Publisher:
+  Type: local
+```
+
+## Published Result Specification
+Once the job is executed, the results are published to the local compute node, and stored as compressed tar file, which can be accessed and retrieved over HTTP from the command line using the `get` command. TAhis will download and extract the contents for the user from the remove compute node.
```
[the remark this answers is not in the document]
rossjones: There is already a downloader for it, it uses the HTTP one, which will unpack compressed files if that's the URL it is given.  The problem of course with the HTTP one, is that the URL will only point at a single file.  If we don't compress the output of the job, then we'd have to introduce a mandatory manifest for publishers intending for the data to be retrieved via HTTP.  This manifest would have to be the item actually referenced in the job spec, and would have to contain the URL to each and every file written in the job output.  

>  I'd be caught off guard by the published data being a tarball when that wasn't what I published

It _should_ be transparently extracted for you, so the only way you'd know it was compressed would be if you were looking at the job spec and downloading it yourself. I think I'd be surprised if I wrote 1000 files and had to download them all one by one.  

I'm totally okay with the idea of having a mandatory manifest for uncompressed publishing, but given the technical difficulties (and potential unpleasantness of 100s/1000s of individual downloads) I think we should wait for a strong user signal that it is critical and that the transparent unpacking is unsatisfactory.

pkg/config/types/compute.go
```diff
@@ -55,3 +56,9 @@ type LogStreamConfig struct {
 	// How many messages to buffer in the log stream channel, per stream
 	ChannelBufferSize int `yaml:"ChannelBufferSize"`
 }
+
+type LocalPublisherConfig struct {
+	Address   string `yaml:"Address"`
```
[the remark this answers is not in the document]
rossjones: It can be "public", "private" etc.

pkg/node/node.go
```diff
@@ -248,6 +233,15 @@ func NewNode(
 
 	// setup requester node
 	if config.IsRequesterNode {
+		authenticators, err := config.DependencyInjector.AuthenticatorsFactory.Get(ctx, config)
+		if err != nil {
+			return nil, err
+		}
+
+		metrics.NodeInfo.Add(ctx, 1,
```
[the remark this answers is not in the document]
rossjones: Fixed in 07073228d

pkg/publisher/local/publisher.go
```diff
@@ -0,0 +1,88 @@
...
+func NewLocalPublisher(ctx context.Context, directory string, host string, port int) *Publisher {
+	p := &Publisher{
+		baseDirectory: directory,
+		host:          host,
+		port:          port,
+		urlPrefix:     fmt.Sprintf("http://%s:%d", host, port),
+	}
+
+	p.server = NewLocalPublisherServer(ctx, p.baseDirectory, p.host, p.port)
+	go p.server.Start(ctx)
```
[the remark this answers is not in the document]
rossjones: Yeah, I had originally put the http serving part in the compute node setup, but it was too hard to keep the lifetime of the server/publisher in sync and configured.  

We can hopefully remove this if we change how publishers are constructed (perhaps so there is more than a single instance).

pkg/publisher/local/publisher.go
```diff
@@ -0,0 +1,88 @@
...
+	return p
+}
+
+// IsInstalled checks if the publisher is installed and it determines
+// this based on the presence of the base directory and a local server.
+func (p *Publisher) IsInstalled(ctx context.Context) (bool, error) {
+	fileInfo, err := os.Stat(p.baseDirectory)
+	if err != nil {
+		log.Ctx(ctx).Debug().Msg("local publisher not installed because the base directory does not exist")
+		return false, nil
```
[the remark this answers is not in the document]
rossjones: Fixed in 4c69c00fc to make the error more accurate (e.g. we either explain why it isn't installed, or log the error if it is a different issue).

pkg/publisher/local/publisher.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return fileInfo.IsDir(), nil
+}
+
+func (p *Publisher) ValidateJob(ctx context.Context, j models.Job) error {
+	return nil
+}
+
```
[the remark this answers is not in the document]
rossjones: `publisher.Publisher`

pkg/publisher/local/server.go
```diff
@@ -0,0 +1,81 @@
...
+	address       string
+	port          int
+}
+
+const (
+	readHeaderTimeout = 3 * time.Second
+	readTimeout       = 3 * time.Second
+)
+
+func NewLocalPublisherServer(ctx context.Context, directory, address string, port int) *LocalPublisherServer {
```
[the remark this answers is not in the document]
rossjones: I think it's also needed so we can cancel the HTTPD during shutdown.

pkg/publisher/local/server.go
```diff
@@ -0,0 +1,81 @@
...
+	addrs, err := network.GetNetworkAddress(addressType, network.AllAddresses)
+	if err == nil && len(addrs) > 0 {
+		return addrs[0]
+	}
+
+	log.Ctx(ctx).Error().Err(err).Stringer("AddressType", addressType).Msgf("unable to find address for type, using 127.0.0.1")
+	return "127.0.0.1"
+}
+
+func (s *LocalPublisherServer) Start(ctx context.Context) {
```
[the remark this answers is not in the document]
rossjones: Fixed in 10fcbea7f

pkg/util/filecopy/copy.go
```diff
@@ -0,0 +1,82 @@
...
+	if sourceFile, err = os.Open(src); err != nil {
+		return errors.Wrap(err, "failed to open source file")
+	}
+	defer sourceFile.Close()
+
+	var destinationFile *os.File
+	if destinationFile, err = os.Create(dst); err != nil {
+		return errors.Wrap(err, "failed to open target file")
+	}
+	defer destinationFile.Close()
```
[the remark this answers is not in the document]
rossjones: Fixed in be8306294

rossjones:
@frrist have made the relevant changes, and created an issue for a wider discussion about compressed results.  Are you okay for me to merge this now?

frrist:
> How the publisher stores the data doesn't have to match what the user generated

In some cases, absolutely. But surely the user deserves at least one option where they can store the data as they wrote it. Right now we lack this! Let's not make this sound like a binary decision - it's not.

> Users are not expected to access data directory

This just isn't true - maybe for the users _you're_ thinking of?

I think we need to shift our view of bacalhau. It's not just a tool for running a single task and producing data for users to grab. It's about chaining together multiple tasks, where each one takes the output of the previous and adds its own layer. It's a continuous process of tasks building on each other's outputs, not just a one-off data dump.

PR #3434, opened 2024-02-08
Fix too-speedy shutdown in devstack

rossjones:
Currently the main() func and the devstack are sharing a cancellable context, and when it is cancelled by SIGINT, the main function exits before the deferred tasks in devstack can complete. This resulted in devstack not running a second time because of the existing pid or port (or both) files.

Now main has its own context which is cancelled only when the job is complete (or cancelled from SIGINT). This gives devstack enough time to shutdown.

### DIFF
--- a/main.go
+++ b/main.go
@@ -68,6 +68,8 @@ import (
 //// 	@scope.admin							Grants read and write access to administrative information
 
 func main() {
+	localCtx, localCancel := context.WithCancel(context.Background())
+
 	defer func() {
 		// Make sure any buffered logs are written if something failed before logging was configured.
 		logger.LogBufferedLogs(nil)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -68,6 +68,8 @@ import (
 //// 	@scope.admin							Grants read and write access to administrative information
 
 func main() {
+	localCtx, localCancel := context.WithCancel(context.Background())
```
frrist, line 71: I don't think we are on golang 1.22 yet, but maybe add a TODO for switching to https://pkg.go.dev/context#AfterFunc when we get there?

rossjones: I'm not sure in retrospect whether we could have just done this with a normal channel as well.  Missed this when merging, but I'm getting a bit concerned about the number of TODOs we're adding but never resolving.  Might be best to add things to tickets if we think they need doing just so they're visible.

PR #3443, opened 2024-02-11
Generate TLS server certificates, keys and CA 

### LINKED ISSUE #3337: Auto-generate a test certificate every time the test is run
Currently we have certificates and keys stored in the testdata directory. Aside from the potential security issues, the certificates are also generated with one day expiry dates, so certificates and keys need to be generated at compile time.

olgibbons:
TLS certificates were being generated by a bash script and stored in the testdata directory. 

Certificates/keys are now generated during the test suite build and stored in a temp directory, so private keys and certs are always removed once testing is completed

fixes #3337


pkg/util/certificates/certificates.go
```diff
@@ -0,0 +1,103 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"net"
+	"os"
+	"time"
+)
+
+func CreateCertificates(caCertPath, caKeyPath, serverCertPath, serverKeyPath string) error {
```
rossjones, line 16: It might be worth breaking this down into two seperate functions one for the ca cert/key and one for the server.

rossjones (commented):
Looking good. Small suggestion to break up the CreateCertificates function into two and need to fix the lints that failed (see https://github.com/bacalhau-project/bacalhau/actions/runs/7863194232/job/21453532689?pr=3443).  Mostly just need to defined constants for the magic numbers and file modes.

pkg/util/certificates/certificates.go
simonwo: I would also amend the name to clarify that this produces *test* certificates and shouldn't be (accidentally) used to generate real ones.

simonwo (commented):
It also looks like you have some linter errors:

- Make sure you are checking the errors of `pem.Encode`
- Use 0600 permissions for writing secrets.
- You will unfortunately be forced to factor out any "magic numbers" into consts.

pkg/util/certificates/certificates.go
rossjones: On a second look, perhaps they might also be better in pkg/test/utils/certificates rather than pkg/util?

pkg/util/certificates/certificates.go
```diff
@@ -0,0 +1,127 @@
...
+		Certificate: Certificate{
+			certFile:     caCertPath,
+			keyFile:      caKeyPath,
+			certTemplate: ca,
+			certPrivKey:  caPrivKey,
+		},
+	}, nil
+}
+
+func (c *CACertificate) CreateSignedCertificate(CertPath, KeyPath string) (*Certificate, error) {
```
rossjones, line 76: CertPath and KeyPath should probably be certPath and keyPath.

pkg/util/certificates/certificates.go
```diff
@@ -0,0 +1,127 @@
...
+		NotAfter:    time.Now().AddDate(0, 0, 1),
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:    x509.KeyUsageDigitalSignature,
+	}
+
+	certPrivKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		return nil, err
+	}
+	//Sign the certificate with the CA
```
rossjones, line 98: This comment doesn't add much. May as well remove it.

pkg/test/utils/certificates/certificates.go
```diff
@@ -0,0 +1,152 @@
...
+	"net"
+	"os"
+	"time"
+)
+
+const (
+	permissions           = 0600
+	rsaKeySize            = 4096
+	serialNumberLimitBits = 128
+	defaultIP             = "127.0.0.1"
```
rossjones, line 20: Is this used anywhere?

simonwo (approved):

PR #3445, opened 2024-02-12
3444 adding all fake data

aronchick:
adding fake data for testing the UI


rossjones:
Build failure seems to be missing NodesTable.stories

webui/src/stories/NodesTable.stories.ts
```diff
@@ -0,0 +1,26 @@
...
+import { NodesTable } from "../pages/NodesDashboard/NodesTable/NodesTable"
+
+export default {
+  component: NodesTable,
+  title: "NodesTable",
+  tags: ["autodocs"],
+}
+
+export const fullDataGenerator = (numberToGenerate: number): Node[] => {
+  // Create a list of 10 jobs
```
rossjones, line 12: Could remove this as it's not accurate

webui/src/stories/storyData/README.md
```diff
@@ -0,0 +1,11 @@
+This directory is for creating fake data for mocks.
+
+You need to install bun (we're using bun because it's simple and outside of our project toolkit):
```
rossjones, line 3: Is bun a hard requirement?  There are a lot of tools already, and introducing another JS variant seems unnecessary?

rossjones (approved):
Looks fine, but unsure about the need for bun?

frrist (changes requested):
As Ross mentioned, I'd like to understand if/why Bun is required. Been feeling a lot of dependency creep around installing bacalhau.

webui/src/stories/NodesTable.stories.ts
aronchick: Done

webui/src/stories/storyData/README.md
aronchick: removed

aronchick:
> As Ross mentioned, I'd like to understand if/why Bun is required. Been feeling a lot of dependency creep around installing bacalhau.

No longere required. Removed

frrist (approved):
LGTM assuming this does the ting we need for the webUI

PR #3449, opened 2024-02-12
fix nats logstream

wdbaruni:
Fixing a bug in channel transformer where it tries to unmarshal `models.ExecutionLog` where the object is `concurrency.AsyncResult[models.ExecutionLog]`


rossjones (approved):
LGTM. Does this only occur in certain circumstances? Haven't seen this error happen when I was trying nats logstream

PR #3464, opened 2024-02-15
s3 only publish compressed results

wdbaruni:
The conclusion of discussing [Improvements to Bacalhau Get](https://www.notion.so/expanso/Improvements-to-Bacalhau-Get-62602e6678af41c181996253ef424002?pvs=4) was to disable uncompressed results and only support compressed results for now to support `bacalhau get` for all S3 results, and make sure checksum is encoded in each result.


rossjones (approved):

frrist:
Wait, so now I don't even have the option to publish uncompressed data to s3?!

PR #3466, opened 2024-02-15
[Documentation] update the S3 inference page

AlekseyKorzik:


docs/docs/examples/model-inference/S3-Model-Inference/index.md
```diff
@@ -1,40 +1,47 @@
 ---
-sidebar_label: "S3-Model-Inference"
+sidebar_label: "Run Inference on a S3-stored Model"
 sidebar_position: 3
 ---
-# Running Inference on a Model stored on S3
+# Run Inference on a S3-stored Model
```
rossjones, line 5: "a S3" is hard to say and so this reads funny.  We should use "an" or perhaps find another way of saying it than S3-stored?

docs/docs/examples/model-inference/S3-Model-Inference/index.md
```diff
@@ -137,37 +155,23 @@ bacalhau docker run \
...
-  - `/model-pth` is the output directory for the model.
-
-  - `/inputs/js-ts/mnist-test/image.png` is the path to the input image.
-
-The job has been submitted and Bacalhau has printed out the related job id. We store that in an environment variable so that we can reuse it later on.
+```bash
+bacalhau logs ${JOB_ID}
 
-### Viewing the output
+Predicted class: 0
```
rossjones, line 170: Not sure what this is for?

rossjones (approved):
Couple of suggestions.

docs/docs/examples/model-inference/S3-Model-Inference/index.md
AlekseyKorzik: I think the best solution is to go back to the original ` Run Inference on a model stored on S3` wording

docs/docs/examples/model-inference/S3-Model-Inference/index.md
AlekseyKorzik: This is an example of the job output using the script and image presented in the article. We provide examples of output in other articles as well. Should it be removed from here? Or should we add a comment that this is an example of output to make it more clear?

PR #3467, opened 2024-02-15
[Documentation] update the YOLOv5 page

AlekseyKorzik:


docs/docs/examples/model-inference/object-detection-yolo5/index.md
```diff
@@ -4,159 +4,151 @@ sidebar_position: 2
...
+-- /bin/bash -c 'find /inputs -type f -exec cp {} /outputs/yolov5s.pt \; ; python detect.py --weights /outputs/yolov5s.pt --source $(pwd)/data/images --project /outputs')
 ```
+This should output a UUID (like `59c59bfb-4ef8-45ac-9f4b-f0e9afd26e70`), which will be stored in the environment variable `JOB_ID`. This is the ID of the job that was created. You can check the status of the job using the commands below.
 
 
-This should output a UUID (like `59c59bfb-4ef8-45ac-9f4b-f0e9afd26e70`). This is the ID of the job that was created. You can check the status of the job with the following command:
-
-
-## Checking the State of your Jobs
+### Checking the State of your Jobs
```
rossjones, line 74: This section is repeated at line 135 onwards.

rossjones (commented):
There's some duplication of content in here, and mention of "IPFS/S3 storage" which is confusing because that's two things, do I need to use both?  We should either pick one, or find some way of saying "IPFS or S3 storage"

AlekseyKorzik:
Fixed wording for S3/IPFS
Replaced duplicate instructions with a link to the original instructions

rossjones (approved):

PR #3499, opened 2024-02-20
feat: include evaluation history in JobHistory

frrist:
This change includes the evaluations for a job in the JobHistory results, and makes progress towards: https://github.com/bacalhau-project/expanso-planning/issues/389
```
  TIME      LEVEL            EXEC. ID    EVAL. ID  NODE ID   REV.  PREVIOUS STATE     NEW STATE          COMMENT     
 16:00:00  EvaluationLevel              0ea3700a            0                        pending                        
 15:35:17  JobLevel                                         1     Pending            Pending            Job created 
 15:35:17  ExecutionLevel   e-846e871b            QmYEDJRY  1     New                New                            
 15:35:17  ExecutionLevel   e-846e871b            QmYEDJRY  2     New                AskForBid                      
 15:35:17  ExecutionLevel   e-846e871b            QmYEDJRY  3     AskForBid          AskForBidAccepted              
 15:35:17  EvaluationLevel              3aeac3cf            0                        pending                        
 15:35:17  ExecutionLevel   e-846e871b            QmYEDJRY  4     AskForBidAccepted  AskForBidAccepted              
 15:35:17  JobLevel                                         2     Pending            Running                        
 15:35:17  ExecutionLevel   e-846e871b            QmYEDJRY  5     AskForBidAccepted  BidAccepted                    
 15:35:17  ExecutionLevel   e-846e871b            QmYEDJRY  6     BidAccepted        Completed                      
 15:35:17  EvaluationLevel              30b18cc1            0                        pending                        
 15:35:17  JobLevel                                         3     Running            Completed                      
```

A follow on to this change will use the `GetJobHistory` method to optionally pre-fill a watch channel which returns events related to a job. This is necessary because we cannot 'Watch' for job events until the job is created, therefore we are bound to miss events related to the job, thus necessitating this change.

### DIFF
--- a/pkg/jobstore/boltdb/store.go
+++ b/pkg/jobstore/boltdb/store.go
@@ -10,26 +10,28 @@ import (
 	"sync"
 	"time"
 
-	"github.com/bacalhau-project/bacalhau/pkg/bacerrors"
-	"github.com/bacalhau-project/bacalhau/pkg/jobstore"
-	"github.com/bacalhau-project/bacalhau/pkg/lib/marshaller"
-	"github.com/bacalhau-project/bacalhau/pkg/lib/math"
-	"github.com/bacalhau-project/bacalhau/pkg/models"
-	"github.com/bacalhau-project/bacalhau/pkg/util/idgen"
 	"github.com/benbjohnson/clock"
 	"github.com/imdario/mergo"
 	"github.com/rs/zerolog/log"
 	"github.com/samber/lo"
 	bolt "go.etcd.io/bbolt"
 	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/bacalhau-project/bacalhau/pkg/bacerrors"
+	"github.com/bacalhau-project/bacalhau/pkg/jobstore"
+	"github.com/bacalhau-project/bacalhau/pkg/lib/marshaller"
+	"github.com/bacalhau-project/bacalhau/pkg/lib/math"
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+	"github.com/bacalhau-project/bacalhau/pkg/util/idgen"
 )
 
 const (
-	BucketJobs             = "jobs"
-	BucketJobExecutions    = "executions"
-	BucketJobEvaluations   = "evaluations"
-	BucketJobHistory       = "job_history"
-	BucketExecutionHistory = "execution_history"
+	BucketJobs              = "jobs"
+	BucketJobExecutions     = "executions"
+	BucketJobEvaluations    = "evaluations"
+	BucketJobHistory        = "job_history"
+	BucketExecutionHistory  = "execution_history"
+	BucketEvaluationHistory = "evaluation_history"
 
 	BucketTagsIndex        = "idx_tags"        // tag -> Job id
 	BucketProgressIndex    = "idx_inprogress"  // job-id -> {}
@@ -578,6 +580,7 @@ func (b *BoltJobStore) GetJobHistory(ctx context.Context,
 	return history, err
 }
 
+//nolint:gocyclo
 func (b *BoltJobStore) getJobHistory(tx *bolt.Tx, jobID string,
 	options jobstore.JobHistoryFilterOptions) ([]models.JobHistory, error) {
 	var history []models.JobHistory
@@ -612,6 +615,29 @@ func (b *BoltJobStore) getJobHistory(tx *bolt.Tx, jobID string,
 	if !options.ExcludeExecutionLevel {
 		// 	// Get the executions for this JobID
 		if bkt, err := NewBucketPath(BucketJobs, jobID, BucketExecutionHistory).Get(tx, false); err != nil {
+			return nil, err
+		} else {
+			err = bkt.ForEach(func(key []byte, data []byte) error {
+				var item models.JobHistory
+
+				err := b.marshaller.Unmarshal(data, &item)
+				if err != nil {
+					return err
+				}
+
+				history = append(history, item)
+				return nil
+			})
+
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	if !options.ExcludeEvaluationLevel {
+		// Get the evaluations for this JobID
+		if bkt, err := NewBucketPath(BucketJobs, jobID, BucketEvaluationHistory).Get(tx, false); err != nil {
 			return nil, err
 		} else {
 			err = bkt.ForEach(func(key []byte, data []byte) error {
@@ -693,8 +719,10 @@ func (b *BoltJobStore) createJob(tx *bolt.Tx, job models.Job) error {
 		if _, err := bkt.CreateBucketIfNotExists([]byte(BucketJobHistory)); err != nil {
 			return err
 		}
-
 		if _, err := bkt.CreateBucketIfNotExists([]byte(BucketExecutionHistory)); err != nil {
+			return err
+		}
+		if _, err := bkt.CreateBucketIfNotExists([]byte(BucketEvaluationHistory)); err != nil {
 			return err
 		}
 	}
@@ -1028,6 +1056,35 @@ func (b *BoltJobStore) appendExecutionHistory(tx *bolt.Tx, updated models.Execut
 	return nil
 }
 
+func (b *BoltJobStore) appendEvaluationHistory(tx *bolt.Tx, eval models.Evaluation) error {
+	historyEntry := models.JobHistory{
+		Type:             models.JobHistoryTypeEvaluationLevel,
+		JobID:            eval.JobID,
+		EvaluationID:     eval.ID,
+		EvaluationStatus: eval.Status,
+		Comment:          eval.Comment,
+		Time:             time.Unix(0, eval.ModifyTime),
+	}
+
+	data, err := b.marshaller.Marshal(historyEntry)
+	if err != nil {
+		return err
+	}
+
+	// Get the history bucket for this job ID, which involves potentially
+	// creating the bucket (evaluation_history.<jobid>)
+	if bkt, err := NewBucketPath(BucketJobs, eval.JobID, BucketEvaluationHistory).Get(tx, true); err != nil {
+		return err
+	} else {
+		seq := BucketSequenceString(tx, bkt)
+		if err = bkt.Put([]byte(seq), data); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // CreateEvaluation creates a new evaluation
 func (b *BoltJobStore) CreateEvaluation(ctx context.Context, eval models.Evaluation) error {
 	return b.database.Update(func(tx *bolt.Tx) (err error) {
@@ -1068,7 +1125,7 @@ func (b *BoltJobStore) createEvaluation(tx *bolt.Tx, eval models.Evaluation) err
 	if err != nil {
 		return err
 	}
-	return nil
+	return b.appendEvaluationHistory(tx, eval)
 }
 
 // GetEvaluation retrieves the specified evaluation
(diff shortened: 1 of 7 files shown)


pkg/models/job_history.go
```diff
@@ -41,13 +42,15 @@ type StateChange[StateType any] struct {
...
+	JobID            string                           `json:"JobID"`
+	NodeID           string                           `json:"NodeID,omitempty"`
+	ExecutionID      string                           `json:"ExecutionID,omitempty"`
+	EvaluationID     string                           `json:"EvaluationID,omitempty"`
+	EvaluationStatus string                           `json:"EvaluationStatus,omitempty"`
+	JobState         *StateChange[JobStateType]       `json:"JobState,omitempty"`
+	ExecutionState   *StateChange[ExecutionStateType] `json:"ExecutionState,omitempty"`
+	NewRevision      uint64                           `json:"NewRevision"`
+	Comment          string                           `json:"Comment,omitempty"`
+	Time             time.Time                        `json:"Time"`
```
frrist, line 55: I am not particularity keen on extending this structure to contain more information in its present form. The cognitive overhead of remembering which values are set based on the type high. Ideally we will move towards something like:
```go
type Event struct {
	Type    EventType
	JobID   string
	Time    time.Time
	Payload []byte
}
```
in future implementations where `Type` dictates the destination `Payload` decodes into. But I suspect that's a topic for a later date.

pkg/jobstore/boltdb/store.go
```diff
@@ -1028,6 +1056,35 @@ func (b *BoltJobStore) appendExecutionHistory(tx *bolt.Tx, updated models.Execut
...
 
+func (b *BoltJobStore) appendEvaluationHistory(tx *bolt.Tx, eval models.Evaluation) error {
+	historyEntry := models.JobHistory{
+		Type:             models.JobHistoryTypeEvaluationLevel,
+		JobID:            eval.JobID,
+		EvaluationID:     eval.ID,
+		EvaluationStatus: eval.Status,
+		Comment:          eval.Comment,
+		Time:             time.Unix(0, eval.ModifyTime),
+	}
```
frrist, line 1067: It may also be valuable to also include the `TriggeredBy` field in this entry from the evaluation, but I have opted not to as it will further complicate the fields of `JobHistory`

pkg/jobstore/boltdb/store.go
```diff
@@ -1028,6 +1056,35 @@ func (b *BoltJobStore) appendExecutionHistory(tx *bolt.Tx, updated models.Execut
...
 }
 
+func (b *BoltJobStore) appendEvaluationHistory(tx *bolt.Tx, eval models.Evaluation) error {
+	historyEntry := models.JobHistory{
+		Type:             models.JobHistoryTypeEvaluationLevel,
+		JobID:            eval.JobID,
+		EvaluationID:     eval.ID,
+		EvaluationStatus: eval.Status,
+		Comment:          eval.Comment,
+		Time:             time.Unix(0, eval.ModifyTime),
```
frrist, line 1066: @wdbaruni is this the correct value to represent the time an evaluation occurs in history? At what point in a jobs lifecycle do we modify/update evaluations - looks like we don't? If the later is true, then what does this field represent?

pkg/jobstore/boltdb/store.go
```diff
@@ -1028,6 +1056,35 @@ func (b *BoltJobStore) appendExecutionHistory(tx *bolt.Tx, updated models.Execut
 	return nil
 }
 
+func (b *BoltJobStore) appendEvaluationHistory(tx *bolt.Tx, eval models.Evaluation) error {
+	historyEntry := models.JobHistory{
+		Type:             models.JobHistoryTypeEvaluationLevel,
+		JobID:            eval.JobID,
+		EvaluationID:     eval.ID,
+		EvaluationStatus: eval.Status,
```
frrist, line 1064: A quick skim of the code leads me to believe this value is _always_ set to `pending`. Is there a different value I should use to represent the status of an evaluation?

pkg/publicapi/endpoint/requester/endpoints_events.go
```diff
@@ -40,9 +42,13 @@ func (s *Endpoint) events(c echo.Context) error {
...
-	legacyEvents := make([]model.JobHistory, len(events))
-	for i := range events {
-		legacyEvents[i] = *legacy.ToLegacyJobHistory(&events[i])
+	legacyEvents := make([]model.JobHistory, 0, len(events))
+	for _, e := range events {
+		if e.Type != models.JobHistoryTypeJobLevel &&
+			e.Type != models.JobHistoryTypeExecutionLevel {
+			continue
+		}
+		legacyEvents = append(legacyEvents, legacy.ToLegacyJobHistory(e))
```
frrist, line 51: We ignore the new history type in the legacy api calls...? I think this is a legacy api, not sure.
Additionally removed gratuitous use of pointers.

rossjones (commented):
Looks good, but I think it would be good for @wdbaruni to confirm whether evaluations are intended to be immutable (when persisted) as we can then document the structure to that effect?

pkg/jobstore/boltdb/store.go
wdbaruni: We should update the state to allow differentiating between evaluations that should be processed and those that are fully processed, whether successfully or not. The main reason is to allow the orchestrator to populate the in-memory broker with pending evaluations after a restart:
- `pending` is for new and enqueued evaluations that weren't processed yet
- `complete` is for evaluations that were successfully processed by the scheduler. This doesn't mean that the scheduler was successful in finding nodes, but that it didn't face errors 
- `failed` is for evaluations that schedulers failed to process after `maxReceiveCount` retries
- `cancelled` for deduped evaluations
- `blocked` might for delayed evaluations, but maybe those should be `pending` as well

The important thing is we shouldn't update evaluation state to include in-flight or not, as not worth the state update.

If we are missing those state updates, then this is a good catch and we should create an issue to track this

pkg/jobstore/boltdb/store.go
wdbaruni: We are missing an update evaluation method, which is required to update the evaluation state, but also to enrich it with helpful comments about the actions that took place as part of processing the evaluation

pkg/models/job_history.go
wdbaruni: Yeah I agree this is not great, but I also agree with Ross that it is probably better to wait and evaluate better ways to keep track of such events when there is a pressing need.

wdbaruni: Lets take a step back and understand what we have today and what we are missing to be able to show job progress to the user. 

Now the job history and execution history events are already capturing when the job is created, execution progress, and completion. Adding evaluation creation and update events will show as duplicate events in the history as most evaluations are created after a job/execution state is updated.

What job and execution events are not capturing are:
- Reasons behind the creation or stopping of executions. This might be mitigated by enriching the history event and add more information to the comment field.
- Events that did not result in job or execution state update, which won't show in the history. An example is when scheduling fails and we decide to retry later. We don't have this feature yet and is part of this PR https://github.com/bacalhau-project/bacalhau/pull/3135
- Events that are specific to the job request so that when we enable job updates, you only get events about the job update and not old events from the job creation. Again we don't have this feature yet if you don't want to worry about it now, and the mitigation would be to add the EvalID in the job and execution history events instead of adding separate evaluation events

So it sounds like what you need to do to improve the job progress visibility, is to enrich the existing job and execution history events with additional metadata or better comment messages. I don't think adding events for evaluation state updates will help much

frrist: Let's look back at the issue https://github.com/bacalhau-project/expanso-planning/issues/389 this work is attempting to address, which states:

1. The proposal is to use evaluations to track the job progress
2. Expose APIs to query evaluations.
3. enrich the job submission command.
4. Chain evaluations so that we can always reach child evaluations from a parent one.
5. Enrich evaluations with metadata that are useful to understand the scheduler’s decisions. These metadata can be a list of arbitrary messages, such as number of discovered nodes, filtered nodes.
6. We are mainly interested in the evaluation that was created when the job was submitted, along with its executions and any evaluation created while trying to bring those executions to a running or completed state
 
This PR addresses points 1 and 2 - 3 will come in a follow on to this work as will 5.

We have come to an understanding 4 isn't practically feasible given our current approach for creating evaluations -  although https://github.com/bacalhau-project/bacalhau/pull/3500 moves us closer to doing so via post-processing.

However, what I am understanding here is that we want something slightly different. 

-------------------

> What job and execution events are not capturing are:
> - Reasons behind the creation or stopping of executions. This might be mitigated by enriching the history event and add more information to the comment field.

I _think_ this is what the `TriggeredBy` field of an evaluation is for, I have asked about it here: https://github.com/bacalhau-project/bacalhau/pull/3499#discussion_r1496699255

> Events that did not result in job or execution state update, which won't show in the history.

Yea, these would be great to have! But also sounds like a new event type - separate from jobs, executions, or evaluations. 
One way to obtain this information is by tracking the conditions and why they weren't met via the update request sent to the job store. But since this change is already landing in https://github.com/bacalhau-project/bacalhau/pull/3135 I guess it's probably better to wait and evaluate better ways to keep track of such events when the need arises.

> Events that are specific to the job request so that when we enable job updates, you only get events about the job update and not old events from the job creation

I'm not sure I understand this bit. We already update jobs in the job store. Unless you mean allowing users to update their jobs?

> the mitigation would be to add the EvalID in the job and execution history events instead of adding separate evaluation events

Please inform me of how you would like this work done:
- Job and execution history events all have comments associated with them. Evaluations also have comments. Should these fields be merged in the corresponding job or execution events comment? 
- You have alluded to collecting other information from the evaluation in your commends - such as `Status` and `TriggeredBy`. How can this information be represented in the job or execution event?

wdbaruni: My apologies if the goals were not clear. The goal of the issue you are referring to is to improve the job progress visibility when users call `bacalhau docker run` or `bacalhau job run`. Using evaluations was just a proposal, but you found some challenges with that route as you pointed out in the issue, mainly with tracking the lineage. You have actually suggested a good alternatives where the client listens for job store events instead, but needed to be aware of the jobID before submission. 
We discussed in our meeting that you are free to use any other source to achieve better job progress visibility, and that enriching job history along with pagination would work similar to job store events, but are also persisted and can be queried after the fact.

> I think this is what the TriggeredBy field of an evaluation is for, I have asked about it here: https://github.com/bacalhau-project/bacalhau/pull/3499#discussion_r1496699255

Good point, but I don't think this will work as evaluations can be deduped by the evaluation broker where only one evaluation will be processed while others cancelled. This makes sense because the scheduler only compares the desired job state with the actual observed state to make scheduling decisions, and doesn't care about the source or trigger of the evaluation

> Yea, these would be great to have! But also sounds like a new event type - separate from jobs, executions, or evaluations.

Not necessary. To me it sounds like a job level event as it did not touch a specific execution, and the whole job will fail eventually if we didin't schedule the job. If we keep things simple, anything event that is not related to specific execution is most likely a job level event. 

One way for us to add a history event when scheduling is delayed is to actually update the job state from `running` to `pending` as an example. Another way is to expose an API to add history events directly by the scheduler and not only after `state` updates. Though as you pointed out, we don't need to worry about this for now and should be handled with/after https://github.com/bacalhau-project/bacalhau/pull/3135

> I'm not sure I understand this bit. We already update jobs in the job store. Unless you mean allowing users to update their jobs?

Correct, when the user update their job and want to see progress for the update request and not older history events

> Job and execution history events all have comments associated with them. Evaluations also have comments. Should these fields be merged in the corresponding job or execution events comment?

Maybe. Not all job/execution events have a comment. We just need to evaluate each history event to come up with the right solution on what to include in the comment field. 

> You have alluded to collecting other information from the evaluation in your commends - such as Status and TriggeredBy. How can this information be represented in the job or execution event?

Similar to my answer above

I hope this helps, and let me know if I can clarify anything else

frrist:
in favor of: https://github.com/bacalhau-project/bacalhau/pull/3680

PR #3500, opened 2024-02-21
fix: set EvalID in execution

frrist:
- previously this field was never set. It's now set based on the evaluationID in the plan when an execution is created.

### DIFF
--- a/pkg/orchestrator/scheduler/batch_service_job.go
+++ b/pkg/orchestrator/scheduler/batch_service_job.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


frrist:
@wdbaruni aside from relating the execution to the evaluation responsible for it, what is the intent of the EvalID in the execution models?

wdbaruni:
> aside from relating the execution to the evaluation responsible for it, what is the intent of the EvalID in the execution models?

That is the only reason so far. To help operators link executions to which evaluations created them, and to know which executions created by a specific evaluation, when we add evaluation APIs

wdbaruni (approved):

PR #3502, opened 2024-02-21
Updates the experimental slices package

### LINKED ISSUE #3498: Update x/exp/slices
We can't currently apply https://github.com/bacalhau-project/bacalhau/pull/3473 because we need to update how we use exp/slices since it changed behaviour in an update for 0.11 (update golang.org/x/exp, fixing slice)

rossjones:
Recently the `slices.SortFunc` package was updated, and the signature of SortFunc was changed to expect a cmp() int style return value, rather than a boolean.  

This PR updates the exp/slices package (and as a result, others) fixing the issue with sorting in various places.  In particular `model/network` and `models/network` logic was unclear, and also suffered from the CompactFunc only performing a single run over the domain list.  

The newly introduced `util/cmp#Compare` type provides a temporary version of the functions that will be available to us in the builtin cmp package, once we upgrade, which is now feasible.

Fixes #3498


frrist:
Would it be possible to remove usage of the experimental slices package dependency and instead use the stand library sort package?

rossjones:
> Would it be possible to remove usage of the experimental slices package dependency and instead use the stand library sort package?

It's in the standard library in 1.22 so this is only a temporary step to being able to use that. Once this is merged I think we only need to update IPFS (and the associated boxo version) and fix a few things with the boxo/path types to be able to upgrade.

simonwo (approved):

PR #3506, opened 2024-02-22
Remove invalid or expired authorization tokens

simonwo:
This PR allows the Bacalhau CLI (and optionally users of the SDK) to automatically remove tokens that the server says are not valid, e.g. because they have expired or because the server no longer recognises the signing authority.

In particular, this process is transparent to the rest of the system - the authentication happens as part of a normal API call, and the API call doesn't complete until the authentication flow is done. Naturally this is an opt-in for anyone using the Bacalhau codebase as a library because this might not be desired.

This involved splitting the transport parts of the API out from the domain-specific parts so that the authentication code could monitor for invalid auth tokens and invoke the authentication flow.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the authentication flow across the application for better security and user experience.
	- Streamlined API client usage with improved token management.
- **New Features**
	- Implemented a more robust authorization mechanism, including checks for token validity.
- **Bug Fixes**
	- Fixed an issue to ensure expired or invalid tokens are correctly removed, enhancing security.
- **Tests**
	- Added new test cases to validate the improved authentication and authorization mechanisms.
- **Chores**
	- Updated internal utilities and dependencies for better performance and maintainability.

### DIFF
--- a/cmd/util/tokens.go
+++ b/cmd/util/tokens.go
@@ -6,8 +6,10 @@ import (
 
 	"github.com/bacalhau-project/bacalhau/pkg/config"
 	"github.com/bacalhau-project/bacalhau/pkg/config/types"
+	"github.com/bacalhau-project/bacalhau/pkg/publicapi/apimodels"
 	"github.com/bacalhau-project/bacalhau/pkg/storage/util"
 	"github.com/bacalhau-project/bacalhau/pkg/util/closer"
+	"github.com/pkg/errors"
 )
 
 type tokens map[string]string
@@ -17,14 +19,14 @@ func readTokens(path string) (tokens, error) {
 	if os.IsNotExist(err) {
 		return map[string]string{}, nil
 	} else if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error opening file")
 	}
 	defer closer.CloseWithLogOnError(path, file)
 
 	var t tokens
 	err = json.NewDecoder(file).Decode(&t)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error decoding JSON")
 	}
 
 	return t, nil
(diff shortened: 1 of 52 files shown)


simonwo:
@coderabbit review

simonwo:
@coderabbitai review

pkg/publicapi/client/v2/client.go
```diff
@@ -200,3 +212,103 @@ func (c *Client) url(endpoint string) (*url.URL, error) {
...
+		return err
+	}
+
+	// Try the initial request again with our possible new credential. It's ok
+	// if we didn't authenticate because this server might accept
+	// unauthenticated requests.
+	request.SetCredential(t.Credential)
+	return runRequest(request)
+}
+
```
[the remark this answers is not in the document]
simonwo: Hey, it's right!

cmd/util/tokens.go
```diff
@@ -40,20 +42,30 @@ func writeTokens(path string, t tokens) error {
...
+
+	cred := &apimodels.HTTPCredential{
+		Scheme: "Bearer",
+		Value:  t[apiURL],
+	}
+
+	if cred.Value != "" {
+		return cred, nil
+	} else {
+		return nil, nil
```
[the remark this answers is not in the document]
simonwo: I've changed all of the callers to this function. But you're right, more documentation here would not go amiss.

pkg/authz/policy.go
```diff
@@ -116,8 +121,13 @@ func (authorizer *policyAuthorizer) Authorize(req *http.Request) (Authorization,
 		},
 	}
 
+	tokenValid := true
```
rossjones, line 124: Should we always start with the assumption that it's invalid and let the query tell us otherwise? It feels a little safer.

At the moment if the error from allowQuery is not nil, then The `Authorization.TokenValid` will be true. Even if the error is returned with it, we're relying on it being checked. If tokenValid defaulted to false, the error would be returned along with a TokenValid value of false.

pkg/authz/policy_test.go
```diff
@@ -38,9 +39,10 @@ func TestDeniesWhenPolicySaysDeny(t *testing.T) {
 	result, err := authorizer.Authorize(request)
 	require.NoError(t, err)
 	require.False(t, result.Approved)
+	require.True(t, result.TokenValid)
 }
 
-func TestPolicyEvaluatedAgainstHTTPRequest(t *testing.T) {
+func TestPolicyEvaluatedAgainstGoodHTTPRequest(t *testing.T) {
```
rossjones, line 45: It would be good to see a test that triggers the error from NewPolicyAuthorizer (if straight-forward).  Not critical though.

pkg/publicapi/apimodels/auth.go
```diff
@@ -2,8 +2,11 @@ package apimodels
 
 import (
 	"github.com/bacalhau-project/bacalhau/pkg/authn"
+	"github.com/pkg/errors"
```
rossjones, line 5:
```suggestion
	"errors"
```

We can use the standard library version of errors to call errors.New.

rossjones (approved):
Couple of minor points, but looks good to me. Particularly like the new v2 API interface.

pkg/authz/policy_test.go
simonwo: It's not super easy, basically only some kind of mangled input structure or mangled policy can cause it. I've simplified the error condition so hopefully it's even less critical.

PR #3515, opened 2024-02-26
Adds a node approval state to the node info

rossjones:
Adds a new node approval field to the node_info structure where we currently track information about nodes in the cluster.
The approval field currently defaults to unknown, and is not yet used for anything.  This will be used after:

* Persistent node store can pass filters to list()
* There is a registration endpoint
* The CLI can list/approve/reject nodes.

### DIFF
--- a/pkg/models/node_approval.go
+++ b/pkg/models/node_approval.go
@@ -2,38 +2,93 @@ package models
 
 import "fmt"
 
-// NodeApproval is used to denote the approval status of a given
-// node. These values are set based on the approval process for
-// nodes which will auto-approve some, rely on human approval for
-// others as well as rejecting some nodes as inelligible.
-type NodeApproval string
+type NodeApproval struct {
+	approval
+}
+
+type approval int
 
 const (
-	NodeApprovalUnknown  NodeApproval = "unknown"
-	NodeApprovalPending  NodeApproval = "pending"
-	NodeApprovalApproved NodeApproval = "approved"
-	NodeApprovalRejected NodeApproval = "rejected"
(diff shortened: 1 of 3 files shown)


pkg/models/node_approval.go
```diff
@@ -0,0 +1,39 @@
...
+
+import "fmt"
+
+// NodeApproval is used to denote the approval status of a given
+// node. These values are set based on the approval process for
+// nodes which will auto-approve some, rely on human approval for
+// others as well as rejecting some nodes as inelligible.
+type NodeApproval string
+
+const (
```
wdbaruni, line 11: why string instead of iota. I am asking because I keep asking myself the same question, specially that we always marshal to strings when persisting or sending fields

wdbaruni: One value might having a default value other than blank?

pkg/models/node_approval.go
```diff
@@ -0,0 +1,39 @@
...
+	switch s {
+	case "pending":
+		return NodeApprovalPending, nil
+	case "approved":
+		return NodeApprovalApproved, nil
+	case "rejected":
+		return NodeApprovalRejected, nil
+	case "unknown":
+		return NodeApprovalUnknown, nil
+	default:
```
wdbaruni, line 31: what are the possible cases to have "unknown" state?

rossjones: Originally it was going to be the return value for the Parse() method, but then I thought it would be a reasonable state for nodes until I add the approval endpoints.

wdbaruni (approved):

PR #3535, opened 2024-02-27
Node registration when network=nats

rossjones:
Introduces a requester endpoint to allow compute node to register their existence at startup.  

Compute nodes are responsible for supplying the mostly static information about themselves at node startup time.  This will include the operating system, and the available resources etc (models.NodeInfo).  The dynamic data (such as currently used resources) will be sent separately from this registration and so is not necessary during the initial request. 

Upon receiving the registration request the node info is added to the persistent node store. This data is then served via the usual nodeinfostore use-cases.  Missing from the data will be:

* Current resource availability 
* Approval status 
* Any form of liveness info

These are to come in following PRs


pkg/nats/proxy/registration_handler.go
```diff
@@ -0,0 +1,75 @@
...
+
+	subject := registrationSubscribeSubject()
+	_, err := handler.conn.Subscribe(subject, func(m *nats.Msg) {
+		handler.handle(m)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	log.Debug().Msgf("ComputeHandler %s subscribed to %s", handler.nodeID, subject)
```
wdbaruni, line 45: `ComputeHandler` -> `ReigsterationHandler`

pkg/nats/proxy/constants.go
```diff
@@ -5,6 +5,7 @@ import "fmt"
 const (
 	ComputeEndpointSubjectPrefix = "node.compute"
 	CallbackSubjectPrefix        = "node.orchestrator"
+	RegistrationSubjectPrefix    = "node.registration"
```
wdbaruni, line 8: looks like `Register` is just one of the methods under this prefix. Does it make sense to call it `node.management`?

pkg/nats/proxy/constants.go
```diff
@@ -33,3 +36,11 @@ func callbackPublishSubject(nodeID string, method string) string {
 func callbackSubscribeSubject(nodeID string) string {
 	return fmt.Sprintf("%s.%s.>", CallbackSubjectPrefix, nodeID)
 }
+
+func registrationPublishSubject(method string) string {
+	return fmt.Sprintf("%s.%s", RegistrationSubjectPrefix, method)
```
wdbaruni, line 41: I suggest to include the compute node ID's in the subject to allow doing ACL in the future where a compute node can only publish and subscribe to subjects in the pattern `node.registeration.{node_id}.>`

pkg/nats/proxy/registration_handler.go
```diff
@@ -0,0 +1,75 @@
...
+// RegistrationHandlerParams defines parameters for creating a new RegistrationHandler
+type RegistrationHandlerParams struct {
+	NodeID               string
+	Conn                 *nats.Conn
+	RegistrationEndpoint requester.RegistrationEndpoint
+}
+
+// ComputeHandler handles NATS messages for compute operations.
+type RegistrationHandler struct {
+	nodeID   string
```
wdbaruni, line 24: not used

pkg/nats/proxy/registration_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	ctx := context.Background()
+
+	subjectParts := strings.Split(msg.Subject, ".")
+	method := subjectParts[len(subjectParts)-1]
+
+	switch method {
+	case Register:
+		if err := h.processRegistration(ctx, msg); err != nil {
+			log.Ctx(ctx).Error().Msgf("error processing registration: %s", err)
+		}
```
wdbaruni, line 60: better to return the error back to the compute node using NATS requester/response pattern. Otherwise the registration will fail but the compute node will assume it was successful. `concurrency.AsyncResult` might come handy, not sure though.

We have a backlog item to do this for all compute callbacks as well https://github.com/bacalhau-project/expanso-planning/issues/406

Also might be valuable to return a `RegisterResponse` object, which can be empty for now

pkg/node/node.go
```diff
@@ -271,6 +275,12 @@ func NewNode(
 			attribute.StringSlice("node_authenticators", authenticators.Keys(ctx)),
 		)
 
+		// Create a new node manager to keep track of compute nodes connecting
+		// to the network.
+		nodeManager := manager.NewNodeManager(manager.NodeManagerParams{
+			NodeInfo: tracingInfoStore,
+		})
+
```
wdbaruni, line 283: should this be created under requester.go?

wdbaruni (commented):
Having a separate flow and a handshake when new nodes need to register sounds like the right decision. Though there are a lot of cases that we need to consider, including:
1. Compute node shouldn't attempt to register at startup if it is already registered. Mainly we don't want to fail the compute node from starting if it can't reach the requester node at that time, and should continue to run the locally allocated jobs
2. If compute node info changes, how will it know it needs to call the requester node's `Register` (or maybe another `Update`) endpoint? Most likely we will be fine with the compute node figuring out if needs to re-register during startup, but in the future we might have to introduce this at runtime, such as if `docker` daemon was started while the compute node is already running, and we don't want to users to restart the compute node to make docker discoverable 
3. What if the requester node loses state about the compute nodes, and they all must re-register, but they think they are already registered? 

A proposal to address most of these issues might be:
1. Have the compute nodes persist locally if they are registered or not, and only call register at startup if it has never been registered
2. Have a separate periodic request/response API for compute node to share their latest node info with the requester node. The compute node will call this API every 5-10 minutes, and a future improvement can be as soon as it detects a change in its info. This is similar to what we have today, but in minutes and with a request/response pattern so that the requester node can return a meaningful response  if it doesn't know this compute node to trigger a registration flow 
3. The compute node will continue to heartbeat separately in a fire-and-forget pattern and include dynamic info in the heartbeat, such as available resources

Does that make sense?

rossjones:
> Having a separate flow and a handshake when new nodes need to register sounds like the right decision. Though there are a lot of cases that we need to consider, including:
> 
>     1. Compute node shouldn't attempt to register at startup if it is already registered. Mainly we don't want to fail the compute node from starting if it can't reach the requester node at that time, and should continue to run the locally allocated jobs
>     2. If compute node info changes, how will it know it needs to call the requester node's `Register` (or maybe another `Update`) endpoint? Most likely we will be fine with the compute node figuring out if needs to re-register during startup, but in the future we might have to introduce this at runtime, such as if `docker` daemon was started while the compute node is already running, and we don't want to users to restart the compute node to make docker discoverable
>     3. What if the requester node loses state about the compute nodes, and they all must re-register, but they think they are already registered?

I did consider point 1, that they should only register when they are joining the network the first time, but the solution to both 2 AND 3 is to have the compute node register at each startup.  Not only does it mean that we will catch changes in the operating environment (eg. new GPU added) but it will also ensure that should the requester have forgotten about us, we can remind it.  The downside is really that it's potentially wasteful, and we have to handle the potential update at the registration endpoint, but I don't think that is too onerouse.

I'm okay adding a secondary heartbeat with the latest node info, it'd also help with potential failed delivery of the previous registration message, but I don't think we should add any more state to the compute node if we can avoid it.

wdbaruni:
Actually my proposal for the periodic message is a solution to 2 and 3, along with making updating the node info happen at runtime without having to bounce the compute fleet. Though:
1. They shouldn't be treated as registration. An unregistered compute node will not heartbeat unless it is registered
2. They can trigger the registration path and invalidate a previous registration, but shouldn't be treated as an implicit registration 

I understand that it might be redundant to heartbeat node info after registration, but we should always have a way for compute nodes and requester node to sync their state of view of the world in case something changed and went unnoticed. We might even just include the nodeInfo version or checksum in the main heartbeat, and have the requester ask the compute node for its info if it detects it has outdated version

rossjones:
> Actually my proposal for the periodic message is a solution to 2 and 3, along with making updating the node info happen at runtime without having to bounce the compute fleet. Though:
>     1. They shouldn't be treated as registration. An unregistered compute node will not heartbeat unless it is registered 
>     2. They can trigger the registration path and invalidate a previous registration, but shouldn't be treated as an implicit registration

I'm implementing the frequent node info update right now, but what different info is provided in an explicit registration step that isn't delivered in the 5 minute updates? Presumably we also shouldn't send the node info update every 5 minutes until we've had a successful registration?

pkg/nats/proxy/registration_handler.go
rossjones: This is the last remaining part I think before it's ready for full review.

pkg/compute/management_client.go
```diff
@@ -0,0 +1,133 @@
...
+	})
+	if err != nil {
+		return errors.New("failed to register with requester node")
+	}
+
+	if response.Accepted {
+		log.Ctx(ctx).Debug().Msg("register request accepted")
+		if err := m.registrationFile.Set(); err != nil {
+			log.Ctx(ctx).Error().Msgf("failed to record local registration status")
+		}
```
wdbaruni, line 81: should we return an error here?

pkg/compute/registration_file.go
```diff
@@ -0,0 +1,40 @@
...
+// RegistrationFile is a sentinel on disk whose presence
+// is used to denote that a node has successfully registered
+// with the requester. This file is per-node to allow multiple
+// compute nodes using the same shared directory for config.
+type RegistrationFile struct {
+	path string
+}
+
+func NewRegistrationFile(nodeID string) *RegistrationFile {
+	repo, _ := config.Get[string]("repo")
```
wdbaruni, line 24: we should minimize our usage of global configs https://github.com/bacalhau-project/expanso-planning/issues/474

Better to pass the path instead

wdbaruni (approved):
Minor comments

pkg/compute/registration_file.go
rossjones: Done.

PR #3536, opened 2024-02-27
Make it easier to build _just_ the binary

rossjones:
Currently if you want to build the bacalhau binary, you also have to build the webui and plugins as well, even if you've no intention of using them.

This change adds:

### make binary 

Just builds the bacalhau binary (assuming webui is available already for embedding).

### make binary-web
 
Builds the webui ready for embedding in the binary

### make build-bacalhau 

The existing step now calls binary-web and then binary.

### make build 

Does the same as always

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,8 @@ build-webui:
 ################################################################################
 # Target: build-bacalhau
 ################################################################################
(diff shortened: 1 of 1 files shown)


aronchick (approved):
LGTM

wdbaruni (approved):

PR #3539, opened 2024-02-28
Hide NATS auth token if the user supplied it

### LINKED ISSUE #3490: Insecurely prints NATS auth tokens when the user supplied them
E.g.
```
% bacalhau config set node.network.type nats
% bacalhau config set node.network.authsecret kerfuffle
% bacalhau serve --peer=none
To connect to this node from the client, run the following commands in your shell:
export BACALHAU_NODE_CLIENTAPI_HOST=0.0.0.0
export BACALHAU_NODE_CLIENTAPI_PORT=1234
export BACALHAU_NODE_NETWORK_TYPE=nats
export BACALHAU_NODE_NETWORK_ORCHESTRATORS=nats://kerfuffle@127.0.0.1:4222
export BACALHAU_NODE_IPFS_SWARMADDRESSES=/ip4/127.0.0.1/tcp/61949/p2p/QmR7vHpQikS5J1U3VHhp2G3r7PSTnezmHcRVG5h6jz4ego

A copy of these variables have been written to: /var/folders/xz/qrd28pyx1095q2nzvhd9wb_c0000gn/T/tmp.ocBugqCLmd/bacalhau.run
```
The node auth secret "kerfuffle" should not be printed here as it's insecure to print secrets in system logs, and we assume if the user has supplied us the token then they are capable of adding it to the compute nodes themselves.

simonwo:
We shouldn't print auth tokens unless absolutely necessary. Now the system will hide the auth token if the user supplied it explicitly.

Resolves #3490.

### DIFF
--- a/cmd/cli/serve/serve.go
+++ b/cmd/cli/serve/serve.go
@@ -400,16 +400,12 @@ func buildConnectCommand(ctx context.Context, nodeConfig *node.NodeConfig, ipfsC
 
 		switch nodeConfig.NetworkConfig.Type {
 		case models.NetworkTypeNATS:
-			advertisedAddr := nodeConfig.NetworkConfig.AdvertisedAddress
-			if advertisedAddr == "" {
-				advertisedAddr = fmt.Sprintf("127.0.0.1:%d", nodeConfig.NetworkConfig.Port)
-			}
+			advertisedAddr := getPublicNATSOrchestratorURL(nodeConfig)
 
 			headerB.WriteString("To connect a compute node to this orchestrator, run the following command in your shell:\n")
-
 			cmdB.WriteString(fmt.Sprintf("%s=%s ",
 				configflags.FlagNameForKey(types.NodeNetworkOrchestrators, configflags.NetworkFlags...),
-				advertisedAddr,
+				advertisedAddr.String(),
 			))
 
 		case models.NetworkTypeLibp2p:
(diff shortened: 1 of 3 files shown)


cmd/cli/serve/serve.go
```diff
@@ -517,6 +503,25 @@ func buildEnvVariables(ctx context.Context, nodeConfig *node.NodeConfig, ipfsCon
...
+func getPublicNATSOrchestratorURL(nodeConfig *node.NodeConfig) *url.URL {
+	orchestrator := &url.URL{
+		Scheme: "nats",
+		Host:   nodeConfig.NetworkConfig.AdvertisedAddress,
+	}
+
+	// Only display the secret if the user did not set it explicitly.
+	// Else, they should already know it!
+	if secret, err := config.Get[string](types.NodeNetworkAuthSecret); err == nil && secret == "" {
+		orchestrator.User = url.User(nodeConfig.NetworkConfig.AuthSecret)
```
rossjones, line 515: If `nodeConfig.NetworkConfig.AuthSecret` is "" this eventually generates a URL like `://@host:port/`. While this is entirely valid,  it seems like it might be a cause of support requests on unrelated issues.  Perhaps if we don't have a user to set in the URL we should just not set it?

rossjones (approved):
LGTM. One minor question about generating URLs with the @ separator unnecessarily.

cmd/cli/serve/serve.go
simonwo: Hm. If `nodeConfig.NetworkConfig.AuthSecret` starts out at `nil` then this code will make sure it is filled in?
https://github.com/bacalhau-project/bacalhau/blob/3920ce72201feddf0e21cc6856e9a4028df4d0ad/cmd/cli/serve/util.go#L198

But fair enough, we may as well be defensive and check the value here too.

PR #3540, opened 2024-02-28
updating makefile to throw an error if Earthly not installed

aronchick:


wdbaruni (commented):
You have unrelated changes in `circleci/config.yml`. Are they intentional?

aronchick:
> You have unrelated changes in `circleci/config.yml`. Are they intentional?

Just the one - it was a bad structure previously (it was assuming that you were on a branch for asdf). I was seeing the same problems here.

.circleci/config.yml
```diff
@@ -155,14 +155,17 @@ commands:
...
           command: |
+            git config advice.detachedHead false
             if [ ! -d $HOME/.asdf ]; then
               git clone https://github.com/asdf-vm/asdf.git $HOME/.asdf --branch << parameters.asdf_version >>
               cd $HOME/.asdf
             else
               cd $HOME/.asdf
               git fetch origin
-              git pull
+              git fetch --all
```
wdbaruni, line 165: would `git fetch --tags` be enough here?

.circleci/config.yml
```diff
@@ -155,14 +155,17 @@ commands:
...
             fi
+            git checkout master
+            git pull --all
```
wdbaruni, line 168: are these necessary?

.circleci/config.yml
```diff
@@ -277,6 +281,8 @@ jobs:
       - run:
           name: Build tarball
           command: |
+            source "${BASH_ENV}"
+            cat "${BASH_ENV}"
```
wdbaruni, line 285: do we need to cat the envs?

wdbaruni (approved):
Minor comments. Can be merged once addressed

.circleci/config.yml
aronchick: done

.circleci/config.yml
aronchick: Yes. When pulling from asdf with a version, for some reason it puts you headless. I needed to checkout master, and then wanted to confirm everything was up to date (particularly just to ensure breaking caches).

.circleci/config.yml
aronchick: this was for debugging - you're right, don't need to.

PR #3545, opened 2024-02-28
Adds support for filtering of the List command in NodeInfoStores

rossjones:
Currently when listing node info in the NodeInfoStores, it is all or nothing, and there is no way to narrow down the list.

This PR adds support for passing varargs of filter functions so that we can do things like easily filter by approval state, or GPU count, or specific labels.  It is expected that these filter functions will be pre-defined rather than expecting them to be constructed as required.

## Summary by CodeRabbit

- **New Features**
	- Enhanced node information retrieval with advanced filtering capabilities.
- **Tests**
	- Added tests for the new filtering functionality in node listing.
- **Refactor**
	- Updated interfaces and methods to support advanced node information filtering.

### DIFF
--- a/pkg/routing/inmemory/inmemory.go
+++ b/pkg/routing/inmemory/inmemory.go
@@ -119,16 +119,28 @@ func (r *NodeStore) FindPeer(ctx context.Context, peerID peer.ID) (peer.AddrInfo
 	return peer.AddrInfo{}, nil
 }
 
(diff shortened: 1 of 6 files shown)


rossjones:
@coderabbitai review

aronchick (approved):
LGTM - I find it hilarious that the tests are about 3x the code for the thing ;)

PR #3559, opened 2024-03-04
Generate certificates requester

olgibbons:
WIP: Waiting for comment


frrist:
We will not be generating self signed certificates on the requester node, this is the responsibility of the node operator and will need to happen outside of bacalhau.

PR #3565, opened 2024-03-05
Ensure compute node does not open jobstore

### LINKED ISSUE #3564: Nodes opening wrong databases
If you were to run `bacalhau serve --node-type compute` it calls `GetComputeConfig` and `GetRequesterConfig`. The former creates a reference to the execution store, and the latter will create a reference to the job database, which means that if you run a two processes on one machine (a requester and a compute) then the compute will fail trying to open the jobstore typically owned by the requester node.  This is fine if you run a hybrid node as only one process accesses the database.

I'm implementing a workaround to get my PR through, but we will need to fix this properly, either by moving the store creation or through the node creation refactoring.

rossjones:
During startup of a compute only node (using serve) the requester config is retrieved, and as part of that a reference to the jobstore is created.  This means that if we run a separate requester and compute, the compute won't start because of the jobstore is already locked. The bash-scripts avoid this by creating a new repo for the second node.

This PR is a short-term measure to only create the stores (compute and execution) when explicitly asked to, and this is determined by the node type when retrieving the config.

It is expected that this will be fixed in the node setup refactor.

Temporarily resolves #3564

### DIFF
--- a/cmd/cli/devstack/devstack.go
+++ b/cmd/cli/devstack/devstack.go
@@ -208,11 +208,12 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
 		}
 	}
 
-	computeConfig, err := serve.GetComputeConfig(ctx)
+	computeConfig, err := serve.GetComputeConfig(ctx, true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


aronchick (approved):
Lgtm

PR #3573, opened 2024-03-07
Regularly updates node info when using NATS

rossjones:
Uses the foundation build in the node-registration PR and fixes it up so that it will update node info every 5 minutes.  Removes unnecessary debug output.


alaric-rd (approved):
LGTM

PR #3579, opened 2024-03-08
Cancel authn input if the user presses Ctrl+C

### LINKED ISSUE #3411: CLI that accepts auth token does not respect Ctrl+C/D
In the event a user wants to abort inputting a token over the CLI to auth with a bacalhau node the command does not support canceling via Ctrl+c or Ctrl+d.

simonwo:
Previously the system would wait for the user's input and not cancel input if the user pressed Ctrl+C. This is because the signal was intercepted and used to cancel the root context, but this context isn't used by the terminal reading code and so had no effect.

The solution is to run the input in a separate goroutine and select on the routine's output or the context cancellation.

Alas, there's no way to give the same power to Ctrl+D, because interpreting EOF input is very much the purview of the standard library package we're using. But users can still just press Enter to submit empty input.

Resolves #3411.

## Summary by CodeRabbit

- **Refactor**
	- Improved secure handling of user input in authentication processes.
- **Tests**
	- Enhanced node management in testing scripts.
	- Added tests to verify user input can be safely cancelled.

### DIFF
--- a/cmd/util/auth/ask.go
+++ b/cmd/util/auth/ask.go
@@ -45,26 +45,44 @@ func askResponder(cmd *cobra.Command) responder {
 
 			fmt.Fprintf(cmd.ErrOrStderr(), "%s: ", name)
 
-			var input []byte
-			var err error
+			inputChan := make(chan []byte, 1)
+			errChan := make(chan error, 1)
 
-			// If the property is marked as write only, assume it is a sensitive
-			// value and make sure we don't display it in the terminal
-			if subschema.WriteOnly && term.IsTerminal(int(os.Stdin.Fd())) {
-				input, err = term.ReadPassword(int(os.Stdin.Fd()))
-				fmt.Fprintln(cmd.ErrOrStderr())
-			} else {
-				reader := bufio.NewScanner(cmd.InOrStdin())
-				if reader.Scan() {
-					input = reader.Bytes()
+			// Reading a value is not a cancellable operation. So we manually
+			// listen for the context here and write a final byte to the input
+			// if the user cancels, to hopefully get the input read to end.
+			go func(inputCh chan<- []byte, errCh chan<- error) {
+				var input []byte
+				var err error
+
+				// If the property is marked as write only, assume it is a sensitive
+				// value and make sure we don't display it in the terminal
+				if subschema.WriteOnly && term.IsTerminal(int(os.Stdin.Fd())) {
+					input, err = term.ReadPassword(int(os.Stdin.Fd()))
+				} else {
+					reader := bufio.NewScanner(cmd.InOrStdin())
+					if reader.Scan() {
+						input = reader.Bytes()
+					}
+					err = reader.Err()
 				}
-				err = reader.Err()
-			}
 
-			if err != nil {
+				if err != nil {
+					errCh <- err
+				} else {
+					inputCh <- input
+				}
+			}(inputChan, errChan)
+
+			select {
+			case input := <-inputChan:
+				response[name] = string(input)
+			case err = <-errChan:
 				return nil, err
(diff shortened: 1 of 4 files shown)


simonwo:
@coderabbitai review

test/secret_auth.sh
```diff
@@ -33,6 +33,34 @@ testcase_token_works_for_v1_api() {
...
+    subject pgrep 'bacalhau job list'
+    assert_not_equal 0 $status
+    assert_equal '' $stdout
+
+    assert_not_match '[]' "$(cat screenlog.0)"
+    assert_match 'context canceled' "$(cat screenlog.0)"
+
+    rm screenlog.0
+    pkill -9 'bacalhau job list'
+}
```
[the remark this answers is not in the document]
simonwo: They will be executed even if the assertions fail, so that should be fine.

test/bin/bacalhau.sh
```diff
@@ -1,30 +1,32 @@
 export IFS=$'\n\t'
 
+declare -a RUNNING_NODES
```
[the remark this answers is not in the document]
simonwo: I think the presence of this line actually removes unexpected behaviour. See http://redsymbol.net/articles/unofficial-bash-strict-mode/.

cmd/util/auth/ask.go
```diff
@@ -45,26 +45,46 @@ func askResponder(cmd *cobra.Command) responder {
...
-					input = reader.Bytes()
+			inputChan := make(chan []byte, 1)
+			errChan := make(chan error, 1)
+
+			// Reading a value is not a cancellable operation. So we manually
+			// listen for the context here and write a final byte to the input
+			// if the user cancels, to hopefully get the input read to end.
+			go func(inputCh chan<- []byte, errCh chan<- error) {
+				defer close(inputCh)
+				defer close(errCh)
```
rossjones, line 56: Probably safest not to close these channels, we only need to do that to let the receiver know there are no more messages coming.  If we close them here, I'm worried about an ordering of instructions like

* Write to inputCh 
* Close inputCh
* Select on inputCh 

If we close them outside the goroutine, we might accidentally write to a closed channel.  
So safest to let the GC dispose of them.

test/secret_auth.sh
```diff
@@ -33,6 +33,35 @@ testcase_token_works_for_v1_api() {
...
 }
 
+testcase_input_can_be_cancelled() {
+    setup
+
+    # Run bacalhau job list in a screen session, else the auth code will detect
+    # that a TTY is not in use and won't use the hanging code
+    CMD='bacalhau job list --output=json'
+    rm -f screenlog.0
+    screen -L -d -m bacalhau job list --output=json
```
rossjones, line 43: Is it safer to use `$BACALHAU` here so that we can use it locally (at least for me this might invoke the system installed version in /usr/local/bin

rossjones (approved):
LGTM, couple minor comments.

cmd/util/auth/ask.go
simonwo: Ok, I went and checked my understanding about channels, and I think it would be fine to select on a closed channel as long as we used the `value, ok := <-chan` idiom to check for channel closure. 

But also, just allowing them to go out of scope should work fine here too because we will always push something to one of the channels, so I'll just do that.

test/secret_auth.sh
simonwo: The built `bacalhau` is selected by virtue of appearing first on the `$PATH`... do you mean that `screen` might ignore the contents of `$PATH`?

rossjones: Ah no, if it's first on path I'm less worried.

PR #3584, opened 2024-03-11
More documentation for WASM loader.

### LINKED ISSUE #3577: Document WASM auto-resolving

simonwo:
Resolves #3577.

### DIFF
--- a/pkg/executor/wasm/loader.go
+++ b/pkg/executor/wasm/loader.go
@@ -19,8 +19,18 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/system"
 )
 
-// ModuleLoader handles the loading of WebAssembly modules from storage.PreparedStorage
-// and the automatic resolution of required imports.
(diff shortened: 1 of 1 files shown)


simonwo:
@wdbaruni It looks like the complex functionality that I thought needed documenting was deliberately broken by #2675. Given that there the feature doesn't exist anymore, there isn't much to do here.

However, it looks like the module loading is actually fundamentally broken. I don't think, for example, it would work to upload multiple WASM modules as input, alongside some input data, despite it advertising that this is meant to be supported. What would you like to do about that? So far I have just left a comment about the bug; a bigger change would be to try to fix it (and test it properly) or remove the ability to have multi-module WASM jobs completely (I actually estimate both are about the same amount of work). 

Let me know if you'd like me to pick up either of those tasks. I would guess that this being broken won't cause an issue for any real user for quite some time, maybe longer than the lifetime of the whole WASM feature.

aronchick (approved):
Excellent commentary - should this be in the docs somewhere?

PR #3587, opened 2024-03-11
Fixes a race condition starting server for local publisher

rossjones:
It's possible that the local publisher could be attempt to shutdown the server before it's even started.  This PR moves the server creation outside of the go routine where it listens. In addition to this it removes the panic() and writes the response to an error channel instead should it fail to start the server.


aronchick (approved):
Nicely found - LGTM once tests pass.

simonwo (approved):

PR #3594, opened 2024-03-12
Approve/Reject node membership from CLI

rossjones:
This PR adds the ability to approve/reject node membership to the cluster, based on each node's approval state.  The approve/reject commands are similar and implemented using the same principle.  

```
bacalhau node approve [nodeid] <-m message>

bacalhau node reject [nodeid] <-m message>
```

it is expected that this same approach can be used for `node pause` and `node drain` etc.  
Currently the system does nothing to react to the change of state, a future PR will add pubsub for node events so that jobs can be scheduled etc based on approval/rejection/join/leave events. 

Additionally this PR resolves a small issue in AuthenticatingClient which was proxying calls with the wrong methods, and adds a basenats test suite for testing against a nats enabled devstack.

TODO: 

* [ ] Clarify who is allowed to do this.

### DIFF
--- a/cmd/cli/node/action.go
+++ b/cmd/cli/node/action.go
@@ -1,0 +1,54 @@
+package node
+
+import (
+	"fmt"
+
+	"github.com/bacalhau-project/bacalhau/cmd/util"
+	"github.com/bacalhau-project/bacalhau/pkg/publicapi/apimodels"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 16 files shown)


aronchick (approved):
LGTM

simonwo (approved):
> TODO:
>  Clarify who is allowed to do this.

You can probably leave that to the purview of the authn/z system.

PR #3602, opened 2024-03-12
Manually cleanup repo directory to work around flaky window test.

rossjones:
On windows we've been seeing failing tests with directories already existing, presumably from previous/parallel runs. This PR switches to using a tempdir provided by the test framework so that it is cleaned up when the test completes.

Edit: It appears this is a common problem on windows when removing directories under certain conditions.  We've not moved to manually cleaning up the folder.

### DIFF
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -2,7 +2,6 @@ package setup
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


aronchick (approved):
LGTM

rossjones:
> `The process cannot access the file because it is being used by another process.` Likely because some of out testing cleanup logic is buggy.

Ah, it might be parallel tests then. The complaint I was seeing was "the plugins folder already exists".

PR #3618, opened 2024-03-13
Updates timeout for TestNack_TimeoutReset

rossjones:
We often see this test failing on Windows, but always be a reasonably small amount. This may be just performance on Windows, but as it is an arbitrary value and we're unlikely to optimise _just_ for windows, I've increased the timeout within this test.


aronchick (approved):
LGTM

PR #3620, opened 2024-03-13
[circleci] Build swagger reference - this is an automatic commit [skip ci]

simonwo:


rossjones (approved):
Approving in the hope that future diffs will show it not worth generating a PR.

PR #3634, opened 2024-03-14
Fix documentation not being built or deployed.

simonwo:
Docs haven't been deployed for over a month. The build takes all of 1 minute, and happens in parallel, so there's no need to rate limit this.


rossjones:
There have been a lot of changes to the documentation site recently, and some to structure which unfortunately seems a bit random - https://github.com/bacalhau-project/expanso-planning/issues/522.  Are we sure we're happy with the current docs before we deploy them? Might be best to check with @MichaelHoepler what the plan is?

simonwo:
I think we need the content to be up to date in time for the 1.3.0 release – or even, in advance, so we can put links to relevant documentation into release notes and blog posts. So I'd be inclined to choose "correct content, incorrect structure" over "incorrect content, correct structure". At least the search will work?

rossjones (approved):

PR #3638, opened 2024-03-14
Clarifies error message when bacalhau encounters a newer repo version

rossjones:
When using an older version of bacalhau, in an environment with a repo version that is higher than that supported by that version (e.g. launching bacalhau 1.2.1 where the repo version has been upgrade to 3) the error message is:

```
failed to initialize bacalhau repo at '/Users/ross/.bacalhau':
failed to check if repo exists: unknown repo version 3
```

The error message is now:

```
failed to initialize bacalhau repo at '/Users/ross/.bacalhau':
Unsupported repository version 3.
Bacalhau does not know how to read this configuration format and
you will need to upgrade to a newer version to upgrade this repository.
```

### DIFF
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -35,12 +35,11 @@ func SetupBacalhauRepo(repoDir string) (*repo.FsRepo, error) {
 		return nil, fmt.Errorf("failed to create repo: %w", err)
 	}
 	if exists, err := fsRepo.Exists(); err != nil {
-		switch err.(type) {
-		case repo.ErrUnknownRepoVersion:
+		if repo.IsUnknownVersion(err) {
 			return nil, err
-		default:
-			return nil, fmt.Errorf("failed to check if repo exists: %w", err)
 		}
+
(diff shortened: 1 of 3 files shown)


pkg/setup/setup.go
```diff
@@ -35,7 +35,12 @@ func SetupBacalhauRepo(repoDir string) (*repo.FsRepo, error) {
 		return nil, fmt.Errorf("failed to create repo: %w", err)
 	}
 	if exists, err := fsRepo.Exists(); err != nil {
-		return nil, fmt.Errorf("failed to check if repo exists: %w", err)
+		switch err.(type) {
+		case repo.ErrUnknownRepoVersion:
```
frrist, line 39: Would it be worth following a similar pattern to how `os.IsNotExist` works? i.e. maybe something like: `repo.IsUnknownVersion`?

rossjones: Have implemented in 0e7f06cfe
Didn't think I was going to like it, but it saved on the type switch and so yeah, much better.

PR #3639, opened 2024-03-14
Remove custom yaml tag on TLS ServerKey

### LINKED ISSUE #3633: Setting TLS server key doesn't work
The [TLS docs](https://docs.bacalhau.org/running-node/configuring-tls/#self-signed-certificates) say:
> If you are using a configuration file, you can set the values inNode.ServerAPI.TLS.ServerCertificate and Node.ServerAPI.TLS.ServerTLSKey instead.

And if we look at the defaults, we see:
```
% bacalhau config default | grep 'ServerAPI' -A11
    ServerAPI:
        Host: 0.0.0.0
        Port: 1234
        ClientTLS:
            UseTLS: false
            CACert: ""
            Insecure: false
        TLS:
            AutoCert: ""
            AutoCertCachePath: ""
            ServerCertificate: ""
            ServerTLSKey: ""
```

But:
```
% bacalhau config set node.serverapi.tls.servertlskey ...
Error: invalid configuration key "node.serverapi.tls.servertlskey": not found
```
And this does not have the desired effect i.e. TLS is not enabled:
```
echo 'Node.ServerAPI.TLS.ServerTLSKey: ...' >> $BACALHAU_DIR/config.yaml
```

But this works:
```
% bacalhau config set node.serverapi.tls.serverkey ...
% cat $BACALHAU_DIR/config.yaml
node:
    serverapi:
        tls:
            serverkey: ...
```

So it looks like whilst the `default` command will respect the tag name defined on the config structure, the rest of the config system only pays attention to the Go field name and not the defined YAML tag.

rossjones:
Previously we had a customer tag on the TLSConfiguration struct marking ServerKey as ServerTLSKey when generating YAML.  This was confusing and didn't work.

This resets the yaml key (and documentation) to ServerKey instead.

Fixes #3633

### DIFF
--- a/docs/docs/setting-up/running-node/configuring-tls.md
+++ b/docs/docs/setting-up/running-node/configuring-tls.md
@@ -1,14 +1,14 @@
 ---
-sidebar_label: 'Configuring TLS'
+sidebar_label: "Configuring TLS"
 sidebar_position: 185
-title: 'Configuring Transport Level Security'
+title: "Configuring Transport Level Security"
 description: How to configure TLS for the requester node APIs
 ---
 
 By default, the requester node APIs used by the Bacalhau CLI are accessible over HTTP, but it is possible to configure it to use Transport Level Security (TLS) so that they are accessible over HTTPS instead. There are several ways to obtain the necessary certificates and keys, and Bacalhau supports obtaining them via ACME and Certificate Authorities or even self-signing them.
 
 :::info
-Once configured, you must ensure that instead of using __http__://IP:PORT you use __https__://IP:PORT
+Once configured, you must ensure that instead of using **http**://IP:PORT you use **https**://IP:PORT
 to access the Bacalhau API
 :::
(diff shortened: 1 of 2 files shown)


simonwo (approved):
Do we need to worry about migrations or anything?

rossjones:
> Do we need to worry about migrations or anything?

@frrist any opinion on this?  Happy to do the manual migration, but it might be better if we have a few things to update?  From my understanding of the migration stuff, we've not really done a change in field name (even if only the one generated for `config default`).

simonwo (approved):

frrist:
> @frrist any opinion on this? Happy to do the manual migration, but it might be better if we have a few things to update? From my understanding of the migration stuff, we've not really done a change in field name (even if only the one generated for config default).

I think this should be fine without a migration.

frrist (approved):

PR #3640, opened 2024-03-14
Upgrades the codebase to Go 1.21

rossjones:
We've previously been blocked from keeping up with more recent Go versions, but this has been resolved by the dependencies that were blocking.

This PR upgrades to Go version 1.21.8 after also upgrading IPFS/Boxo. There were some breaking changes but they are resolved in this commit. Unfortunately these changes mean we need to be on 1.21 as these dependencies won't build with earlier versions.

When ready to merge we might want to coordinate to ensure everyone's dev machines are up to date.


frrist (approved):
:tada:

simonwo (approved):

PR #3656, opened 2024-03-19
Publisher processing for exec command

rossjones:
Publishers are no longer required for submitting jobs, but the original processing of the legacy -o flag always presents a resultpath to the exec command.  

This PR now checks the default outputs, and 

* if there is no publisher specified, will ignore the default. 
* If there is a publisher then it will continue to process the default (or any -o flag)
* If there is no publisher but the user specifies -o, then it will correctly fail validation.

### DIFF
--- a/cmd/cli/exec/exec.go
+++ b/cmd/cli/exec/exec.go
@@ -222,6 +222,11 @@ func PrepareJob(cmd *cobra.Command, cmdArgs []string, unknownArgs []string, opti
 			Type:   publisherSpec.Type.String(),
 			Params: publisherSpec.Params,
 		}
+	} else {
+		job.Tasks[0].Publisher = &models.SpecConfig{
+			Type:   "ipfs",
+			Params: map[string]interface{}{},
+		}
 	}
 
 	// Handle ResultPaths by using the legacy parser and converting.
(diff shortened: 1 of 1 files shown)


cmd/cli/exec/exec.go
```diff
@@ -222,6 +222,11 @@ func PrepareJob(cmd *cobra.Command, cmdArgs []string, unknownArgs []string, opti
 			Type:   publisherSpec.Type.String(),
 			Params: publisherSpec.Params,
 		}
+	} else {
+		job.Tasks[0].Publisher = &models.SpecConfig{
+			Type:   "ipfs",
+			Params: map[string]interface{}{},
```
wdbaruni, line 228: why default to ipfs instead of local publisher?

rossjones: Because the local publisher is also currently borked and in need of testing in the RC during the 1.3 release.  So I wanted to play it safe and use IPFS which we know works (mostly).

rossjones: Going to take another look because we shouldn't _need_ to have a default publisher.  I suspect we're getting some defined outputs from somewhere (that requires a publisher) that might be unnecessary.

rossjones: redone.

PR #3660, opened 2024-03-19
Local publisher address tf

rossjones:
Set the address to use for the local publisher explicitly as a command line parameter, and this is used for the link that is generated in the response.  The actual publisher server will bind to all interfaces as we don't know which (if any) are actually the public address.

### DIFF
--- a/ops/terraform/main.tf
+++ b/ops/terraform/main.tf
@@ -75,6 +75,8 @@ export BACALHAU_NODE_NETWORK_TYPE=${var.network_type}
 export BACALHAU_NODE_NETWORK_ORCHESTRATORS="${var.internal_ip_addresses[0]}:4222"
 export BACALHAU_NODE_NETWORK_ADVERTISEDADDRESS="${var.public_ip_addresses[count.index]}:4222"
 export BACALHAU_NODE_NETWORK_CLUSTER_PEERS=""
+export BACALHAU_NODE_COMPUTE_LOCAL_PUBLISHER_ADDRESS="${var.public_ip_addresses[count.index]}"
+
 
 ### secrets are installed in the install-node.sh script
 export SECRETS_GRAFANA_CLOUD_PROMETHEUS_API_KEY="${var.grafana_cloud_prometheus_api_key}"
@@ -204,8 +206,8 @@ EOF
 
 resource "google_compute_address" "ipv4_address" {
   region = var.region
-  name  = "bacalhau-ipv4-address-${terraform.workspace}-${count.index}"
-  count = var.protect_resources ? var.instance_count : 0
+  name   = "bacalhau-ipv4-address-${terraform.workspace}-${count.index}"
+  count  = var.protect_resources ? var.instance_count : 0
   lifecycle {
     prevent_destroy = true
   }
@@ -308,8 +310,8 @@ resource "google_compute_firewall" "bacalhau_ingress_firewall" {
   allow {
     protocol = "udp"
     ports = [
-      "4001",  // ipfs swarm
-      "1235",  // bacalhau swarm
+      "4001", // ipfs swarm
+      "1235", // bacalhau swarm
     ]
   }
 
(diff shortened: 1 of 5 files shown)


ops/terraform/remote_files/scripts/start-bacalhau.sh
```diff
@@ -97,5 +98,6 @@ else
     --web-ui-port 80 \
     --labels owner=bacalhau \
     --requester-job-translation-enabled \
-    --default-publisher ipfs
+    --default-publisher local
+    --local-publisher-address "${BACALHAU_NODE_COMPUTE_LOCAL_PUBLISHER_ADDRESS}"
```
wdbaruni, line 102: missing `\` after local. Also you don't have to set cli flag is you used the correct env variable, which is `BACALHAU_LOCAL_PUBLISHER_ADDRESS `

ops/terraform/main.tf
```diff
@@ -75,6 +75,8 @@ export BACALHAU_NODE_NETWORK_TYPE=${var.network_type}
 export BACALHAU_NODE_NETWORK_ORCHESTRATORS="${var.internal_ip_addresses[0]}:4222"
 export BACALHAU_NODE_NETWORK_ADVERTISEDADDRESS="${var.public_ip_addresses[count.index]}:4222"
 export BACALHAU_NODE_NETWORK_CLUSTER_PEERS=""
+export BACALHAU_NODE_COMPUTE_LOCAL_PUBLISHER_ADDRESS="${var.public_ip_addresses[count.index]}"
+
```
wdbaruni, line 79: You won't have to set cli flag if you used the right env variable `BACALHAU_LOCAL_PUBLISHER_ADDRESS="${var.public_ip_addresses[count.index]}"`

ops/terraform/remote_files/scripts/start-bacalhau.sh
```diff
@@ -43,7 +43,8 @@ if [[ "${BACALHAU_NODE_NETWORK_TYPE}" == "nats" ]]; then
     --web-ui-port 80 \
     --labels owner=bacalhau \
     --requester-job-translation-enabled \
-    --default-publisher ipfs
+    --default-publisher local \
+    --local-publisher-address "${BACALHAU_LOCAL_PUBLISHER_ADDRESS}"
```
wdbaruni, line 47: nit: redundant to pass the cli flag as `BACALHAU_LOCAL_PUBLISHER_ADDRESS` env variable translates to the same config

frrist: I don't think this flag actually expects an IP address, it expects one of these values: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/lib/network/network.go#L111

rossjones: No it can be an IP address as well.  The problem here is that 'public' won't resolve to anything for google cloud as it doesn't give out public addresses to the VMs at runtime.

ops/terraform/remote_files/scripts/start-bacalhau.sh
rossjones: I think this, and patience waiting for the build were the answer.

ops/terraform/remote_files/scripts/start-bacalhau.sh
rossjones: Have pushed a code change, will redeploy to staging this evening.

pkg/publisher/local/publisher.go
```diff
@@ -93,20 +93,17 @@ func (p *Publisher) PublishResult(
...
+			return addrs[0]
+		} else {
+			log.Ctx(ctx).Error().Err(err).Msg("failed to resolve network address by type, using 127.0.0.1")
+			return "127.0.0.1"
+		}
 	}
 
-	log.Ctx(ctx).Error().Err(err).Stringer("AddressType", addressType).Msgf("unable to find address for type, using 127.0.0.1")
-	return "127.0.0.1"
+	return address
```
rossjones, line 108: This just tries to resolve a name (public etc) to an address, and leaves IP addresses untouched.  We might want to re-assess whether the name (public, local, etc) lookup is worthwhile as it doesn't solve the actual problem of potentially not having a public address.

wdbaruni (approved):

PR #3667, opened 2024-03-21
 Use local node in Bash tests when starting hybrid node

simonwo:
Previously tests accidentally used demo network nodes if the node that was started was a hybrid.


js-ts (approved):
Thanks

PR #3672, opened 2024-03-21
Allow nodes to call register on startup, even if previously registered

rossjones:
More gracefully handles registration flow for compute nodes that are already registered.  Allows for re-registration of nodes by removing the sentinel file (if any)

### DIFF
--- a/pkg/node/manager/node_manager.go
+++ b/pkg/node/manager/node_manager.go
@@ -57,8 +57,10 @@ func (n *NodeManager) Register(ctx context.Context, request requests.RegisterReq
 			}, nil
 		}
 
+		// Otherwise we'll allow the registration, but let the compute node
+		// that it has already been registered on a previous occasion.
 		return &requests.RegisterResponse{
-			Accepted: false,
+			Accepted: true,
 			Reason:   "node already registered",
 		}, nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/node/manager/node_manager.go
```diff
@@ -57,8 +57,10 @@ func (n *NodeManager) Register(ctx context.Context, request requests.RegisterReq
 			}, nil
 		}
 
+		// Otherwise we'll allow the registration, but let the compute node
+		// that it has already been registered on a previous occasion.
 		return &requests.RegisterResponse{
-			Accepted: false,
+			Accepted: true,
 			Reason:   "node already registered",
 		}, nil
```
wdbaruni, line 65: I agree that the requester node should handle duplicate registration gracefully and not fail the request. Though I am not sure if the compute node should always regsiter at startup. Our goal is to allow the compute nodes to start even when they are disconnected, but this will prevent that from happening. 

I think we discussed about having a separate path for compute nodes to periodically check their registration status with the requester node and to revoke their own registration status if it sees inconsistencies.

rossjones: So is it reasonable to keep the registration file for the compute node, so it knows it doesn't need to register - but in cases where the state is incorrect, we can remove the registration file and have the node re-register?

wdbaruni: yes, and that can be post 1.3. For now we should just let the requester node handle duplicate registeration requestes gracefully

wdbaruni: if that still causes pain, the we can let the compute node always register at startup and fix that post 1.3

wdbaruni (approved):

PR #3673, opened 2024-03-21
Single requester node

wdbaruni:


frrist (approved):
Has this change been deployed? If it has let's merge, if it hasn't deploy then merge please.

wdbaruni:
> Has this change been deployed? If it has let's merge, if it hasn't deploy then merge please.

Yes deployed on Monday. In general I only send PRs about infrastructure changes after they have been deployed and verified they are working as expected

PR #3681, opened 2024-03-22
Use GetByPrefix to find nodes when approving/rejecting them

rossjones:
Rather than requiring an exact node id, we can accept a shortened version of it and use getbyprefix.  This will complain appropriately if there is more than one node returned by the prefix.


aronchick (approved):
LGTM

PR #3685, opened 2024-03-22
Allow `node list` to sort and filter by approval status

### LINKED ISSUE #3682: Need: bacalhau approve --list
List all nodes that are awaiting approving

rossjones:
When calling `node list`, it is possible to sort by various fields, but it is not possible to sort by approval status.
Supports 'approval' (as an alias) and 'status' for `--order-by` when calling `node list`. This also work when specifying `--order-reversed`.

```
bacalhau node list --order-by status 
bacalhau node list --order-by status --order-reversed 
```

In addition to sorting, it is also possible to only return nodes with a specific status. This is also applied when filtering by labels.

```
bacalhau node list --filter-status approved 
bacalhau node list --filter-status approved  --labels env=devstack
```

Resolves #3682

### DIFF
--- a/cmd/cli/node/list.go
+++ b/cmd/cli/node/list.go
@@ -14,7 +14,7 @@ import (
 )
 
 var defaultColumnGroups = []string{"labels", "capacity"}
-var orderByFields = []string{"id", "type", "available_cpu", "available_memory", "available_disk", "available_gpu"}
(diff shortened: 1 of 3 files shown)


aronchick (approved):
LGTM - well done, great fast turn around

PR #3695, opened 2024-03-22
Update python dependencies

rossjones:
The python dependencies are out of date, and this is reflected in the integrations that are showing old (insecure) versions of libraries. Attempts to update the dependencies in the integrations themselves will fail as the requirement for the old versions is here.


aronchick (approved):
Lgtm

PR #3697, opened 2024-03-23
Update Legacy Tests to V2

js-ts:
Fixes https://github.com/bacalhau-project/expanso-planning/issues/535

This will Break Existing Tests in Create, Validate Test Suites

Which should be deprecated.

The tests have been moved to /cmd/cli/job/run_test.go


wdbaruni:
as discussed previously, we should keep the tests that are specifically targeting legacy cli. The goal is migrate devstack and scenario tests that testing our system end to end to no longer use the legacy API, and start using the new APIs for job submission and listing. In terms of CLI testing, we already have tests that test the new CLIs.

PR #3700, opened 2024-03-25
[Documentation] add declarative example to Custom containers

AlekseyKorzik:


simonwo (approved):
The declarative version runs the job 3 times whereas the non-declarative version only runs it once.

AlekseyKorzik:
fixed

PR #3701, opened 2024-03-25
[Documentation] add declarative example to Pandas

AlekseyKorzik:

### DIFF
--- a/docs/docs/setting-up/workload-onboarding/python-pandas/index.md
+++ b/docs/docs/setting-up/workload-onboarding/python-pandas/index.md
@@ -129,13 +129,11 @@ tasks:
           - -c
           - python read_csv.py
     InputSources:
-      - Target: "/files"
-        Source:
-          Type: "s3"
+      - Source:
+          Type: "ipfs"
           Params:
-            Bucket: "bacallhau-python-pandas"
-            Key: "*"
-            Region: "us-east-1"
+            CID: "QmfKJT13h5k1b23ja3ZCVg5nFL9oKz2bVXc8oXgtwiwhjz"
+        Target: "/files"
 ```
 
 The job description should be saved in `.yaml` format, e.g. `pandas.yaml`, and then run with the command:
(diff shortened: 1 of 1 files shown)


simonwo (commented):
- The imperative example is pulling data from IPFS whereas the declarative example is using S3. Should these be consistent?
- Is the S3 bucket public?
- The above text says "Our working directory is /files. This is the folder where we will save the model as it will automatically get uploaded to IPFS as outputs". I don't tihnk this is true in either case. The default output directrory for imperative jobs is /outputs. I don't think declarative jobs receive any output directories automatically, so the job outputs in the declarative case will go nowhere. Does there need to be some `ResultPaths`?
- When I run the declarative example I get the output "python: can't open file '/files/read_csv.py': [Errno 2] No such file or directory" which is presumably not a success.

AlekseyKorzik:
Replaced source back to IPFS, now the job passes with

```
ID            = j-6b5633a4-7052-4ed8-a703-3b9835c9f573
Name          = Running Pandas
Namespace     = default
Type          = batch
State         = Completed
Count         = 1
Created Time  = 2024-03-27 11:20:45
Modified Time = 2024-03-27 11:20:47
Version       = 0

Summary
Completed = 1

Executions
 ID          NODE ID   STATE      DESIRED  REV.  CREATED  MODIFIED  COMMENT 
 e-682a4ea4  QmPLPUUj  Completed  Stopped  6     11s ago  9s ago            

Standard Output
                                                hash  ...  transaction_type
0  0x04cbcb236043d8fb7839e07bbc7f5eed692fb2ca55d8...  ...                 0
1  0xcea6f89720cc1d2f46cc7a935463ae0b99dd5fad9c91...  ...                 0
2  0x463d53f0ad57677a3b430a007c1c31d15d62c37fab5e...  ...                 0
3  0x05287a561f218418892ab053adfb3d919860988b1945...  ...                 0

[4 rows x 15 columns]
```

PR #3704, opened 2024-03-25
[Documentation] add declarative example to R Helloworld page

AlekseyKorzik:


simonwo (approved):
This one works :) But again, the imperative example mounts from IPFS. Should these be consistent?

PR #3707, opened 2024-03-25
Generate certificates requester three

olgibbons:
If TLS certificates are not provided, the requester node will generate self-signed certificates.


cmd/cli/serve/serve.go
```diff
@@ -274,34 +275,30 @@ func serve(cmd *cobra.Command) error {
...
 		nodeConfig.RequesterAutoCert = config.ServerAutoCertDomain()
 		nodeConfig.RequesterAutoCertCache = config.GetAutoCertCachePath()
-
 		cert, key := config.GetRequesterCertificateSettings()
-		nodeConfig.RequesterTLSCertificateFile = cert
-		nodeConfig.RequesterTLSKeyFile = key
+		nodeConfig.RequesterTLSCertificateFile, nodeConfig.RequesterTLSKeyFile, err = GetTLSCertificate(&nodeConfig, cert, key)
+		if err != nil {
+			return err
+		}
```
rossjones, line 287: How about including the previous line in the GetTLSCertificate function, so it becomes

```
cert, key, err := GetTLSCertificate(nodeConfig.RequesterSelfSigned)
if err != nil { ... } 

nodeConfig.RequesterTLSCert = cert
nodeConfig.RequesterTLSKey = key
```

the parameter passed to GetTLSCertificate can then be `shouldSelfSign bool`

cmd/cli/serve/serve.go
```diff
@@ -571,3 +562,51 @@ func pickP2pAddress(addresses []multiaddr.Multiaddr) multiaddr.Multiaddr {
 
 	return addresses[0]
 }
+func GetTLSCertificate(nodeConfig *node.NodeConfig, cert string, key string) (string, string, error) {
+
+	if nodeConfig.RequesterAutoCert != "" {
+		return "", "", nil
+	}
```
rossjones, line 569: Worth a comment to explain this,.
```suggestion
        // If there are configuration values for autocert we should return an let autocert 
        // do what it does later on in the setup.
	if nodeConfig.RequesterAutoCert != "" {
		return "", "", nil
	}
```

cmd/cli/serve/serve.go
```diff
@@ -571,3 +562,51 @@ func pickP2pAddress(addresses []multiaddr.Multiaddr) multiaddr.Multiaddr {
...
+		if privKey, err := crypto.LoadPKCS1KeyFile(key); err != nil {
+			return "", "", err
+		} else if caCert, err := crypto.NewSelfSignedCertificate(privKey, false, ips); err != nil {
+			return "", "", errors.Wrap(err, "failed to generate server certificate")
+		} else if err = caCert.MarshalCertficate(certFile); err != nil {
+			return "", "", errors.Wrap(err, "failed to write server certificate")
+		}
+		cert = certFile.Name()
+		return cert, key, nil
+	}
```
rossjones, line 609: Instead of this huge indented block, why not return early with 

```
if !nodeConfig.RequesterSelfSigned {
    return "", "", nil 
}
```

and then carry on with the generating of the cert?

cmd/util/flags/configflags/api.go
```diff
@@ -83,4 +83,11 @@ Using this option results in the API serving over HTTPS`,
 		Description:          `Specifies a TLS key file matching the certificate to be used by the requester node`,
 		EnvironmentVariables: []string{"BACALHAU_TLS_KEY"},
 	},
+	{
+		FlagName:             "selfsigned",
```
rossjones, line 87: self-signed ?

cmd/cli/serve/serve_test.go
```diff
@@ -196,8 +198,12 @@ func (s *ServeSuite) TestAPINotPrintedForRequesterNode() {
 
 func (s *ServeSuite) TestCanSubmitJob() {
 	docker.MustHaveDocker(s.T())
-	port, _ := s.serve("--node-type", "requester", "--node-type", "compute")
+	port, err := s.serve("--node-type", "requester,compute")
+	s.Require().NoError(err)
 	client := client.NewAPIClient(client.NoTLS, "localhost", port)
+
+	// Create a v2 client, that does not verify the certificate but does use TLS. This
+	// is only used to check the nodes are up.
```
rossjones, line 206: This comment is incorrect.

frrist:
We can close this in-favor of: https://github.com/bacalhau-project/bacalhau/pull/3721
Additionally I don't think we should be generating TLS certs in the requester node if one isn't provided. That ought to be left to the node operator. In almost every case, to properly generate a cert, the requester will need to be aware of its public IP. And in almost every case that will not be possible.

rossjones:
> In almost every case, to properly generate a cert, the requester will need to be aware of its public IP

For a self-signed certificate, we can use internal addresses for the SAN in the certificate.  The aim of this I believe is to make it easier to use a self-signed cert without having to find/generate one.

pkg/jobstore/mocks.go
simonwo, line 1: This shouldn't have changed.

pkg/config/types/storagetype_string.go
simonwo, line 1: Did you change this?

pkg/compute/mocks.go
simonwo, line 1: This shouldn't have changed.

ops/aws/canary/lambda/pkg/scenarios/utils.go
```diff
@@ -149,7 +149,7 @@ func getClient() *client.APIClient {
 
 func getClientV2() clientv2.API {
 	host, port := getClientHostAndPort()
-	return clientv2.New(fmt.Sprintf("http://%s:%d", host, port))
+	return clientv2.New(fmt.Sprintf("http://%s:%d", host, port), clientv2.WithTLS(true), clientv2.WithInsecureTLS(true))
```
simonwo, line 152: Is this right? The protocol is still specified as HTTP? The default should still be non-TLS which is what the canary should expect. Is this a left over from when you were doing TLS-by-default?

cmd/testing/basetls.go
```diff
@@ -66,7 +66,7 @@ func (s *BaseTLSSuite) SetupTest() {
 	s.Host = s.Node.APIServer.Address //NOTE: 0.0.0.0 will not work because we're testing TLS validation
 	s.Port = s.Node.APIServer.Port
 	s.Client = client.NewAPIClient(client.LegacyTLSSupport{UseTLS: true, Insecure: false}, s.Host, s.Port)
-	s.ClientV2 = clientv2.New(fmt.Sprintf("http://%s:%d", s.Host, s.Port), clientv2.WithTLS(true))
+	s.ClientV2 = clientv2.New(fmt.Sprintf("http://%s:%d", s.Host, s.Port))
```
simonwo, line 69: This should be consistent with the line above – why is the v1 client expecting TLS but the v2 client is not?

cmd/cli/serve/serve_test.go
```diff
@@ -37,7 +38,7 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/util/closer"
 )
 
-const maxServeTime = 15 * time.Second
+const maxServeTime = 10 * time.Second
```
simonwo, line 41: Why does this need to change?

cmd/cli/serve/serve.go
```diff
@@ -267,39 +270,44 @@ func serve(cmd *cobra.Command) error {
...
 		Labels:                config.GetStringMapString(types.NodeLabels),
 		AllowListedLocalPaths: allowedListLocalPaths,
 		NodeInfoStoreTTL:      nodeInfoStoreTTL,
 		NetworkConfig:         networkConfig,
 	}
-
 	if isRequesterNode {
+		var cert string
+		var key string
+		var err error
```
simonwo, line 282: Don't think you need to delcare these separately, just use `cert, key, err := ...`

simonwo (commented):
There are a few things to clarify and/or clean up but then it looks good to go.

cmd/testing/basetls.go
```diff
@@ -66,7 +66,7 @@ func (s *BaseTLSSuite) SetupTest() {
 	s.Host = s.Node.APIServer.Address //NOTE: 0.0.0.0 will not work because we're testing TLS validation
 	s.Port = s.Node.APIServer.Port
 	s.Client = client.NewAPIClient(client.LegacyTLSSupport{UseTLS: true, Insecure: false}, s.Host, s.Port)
-	s.ClientV2 = clientv2.New(fmt.Sprintf("http://%s:%d", s.Host, s.Port), clientv2.WithTLS(true))
+	s.ClientV2 = clientv2.New(fmt.Sprintf("http://%s:%d", s.Host, s.Port), clientv2.WithTLS(true), clientv2.WithInsecureTLS(true))
```
rossjones, line 69:
```suggestion
	s.ClientV2 = clientv2.New(fmt.Sprintf("http://%s:%d", s.Host, s.Port))
```

still haven't removed the TLS stuff here...

simonwo: Is this the right change? The v1 client on the line above has `UseTLS: true`

rossjones: Oops, totally missed that this was the basetls suite. My bad.

rossjones (approved):

PR #3709, opened 2024-03-26
Node heartbeats

rossjones:
Implements heartbeats for compute nodes, sending heartbeat messages to the requester node over NATS PubSub.  The server, upon receiving a heartbeat updates the map of nodes to include the current server-side timestamp.  

Compute nodes using the heartbeat client, will continuously send heartbeat messages every n seconds.  

The heartbeat server receiving these heartbeats maintains a priority queue, which dequeues oldest items (lowest timestamp) first.  Every 5 seconds any item older than a specific timestamp is dequeued, and its state either set to unhealthy (if it is the first missed heartbeat) or unknown if it is the second.  The default for timestamps is 

* 30s since heartbeat - unhealthy
* 60s since heartbeat - unknown (node may be live but disconnected)

The next heartbeat sent by a unhealthy of unknown node will make it healthy again and ready to receive work.

The current state of the node is added to the nodeinfo during a Get/GetByPrefix/List call to the node info store.  This means that the liveness is dynamic and not persisted to the kvstore for node info.


pkg/compute/management_client.go
```diff
@@ -128,10 +133,23 @@ func (m *ManagementClient) updateResources(ctx context.Context) {
...
+	// The heartbeat ticker will fire twice as often as the configured, to ensure that
+	// we don't slip outside the window.  If we only ever sent on the configured
+	// frequency we are at risk of the node's liveness flapping between good and bad.
+	heartbeatTicker := time.NewTicker((heartbeatFrequencySeconds / 2) * time.Second)
+
+	var heartbeatSequence uint64 = 0
+
 	loop := true
 	for loop {
 		select {
```
frrist, line 155: we can remove the boolean and return when ctx is done or the closeChannel is closed (maybe just call this channel `done`?) the calls to stop at the bottom of this method can then be handled via a deferred function.

pkg/node/heartbeat/client.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (h *HeartbeatClient) SendHeartbeat(ctx context.Context, sequence uint64) error {
+	log.Ctx(ctx).Trace().Msgf("sending heartbeat seq: %d", sequence)
+	return h.Publish(ctx, Heartbeat{NodeID: h.nodeID, Sequence: sequence})
+}
+
+func (h *HeartbeatClient) Publish(ctx context.Context, message Heartbeat) error {
+	return h.publisher.Publish(ctx, message)
+}
```
frrist, line 45: if this isn't being called externally it should be private.

pkg/node/heartbeat/server.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := h.subscription.Subscribe(ctx, h); err != nil {
+		return err
+	}
+
+	go func(ctx context.Context) {
+		log.Ctx(ctx).Info().Msg("Heartbeat server started")
+		<-ctx.Done()
+		_ = h.subscription.Close(ctx)
+		log.Ctx(ctx).Info().Msg("Heartbeat server shutdown")
+	}(ctx)
```
frrist, line 50: remove the go routine and call `h.subscription.Close(ctx)` in a deferred function.

frrist (commented):
few drive by comments.

pkg/node/heartbeat/client.go
rossjones: Part of the interface

pkg/node/heartbeat/server.go
rossjones: But then it'll close it when the function returns, and I don't want to close it until the context is cancelled.  Deliberately didn't run this Start method in it's own goroutine so that I can return the error if setup fails.

frrist: I mean a deferred function in the go routine below this one that also handles context cancellation. Wouldn't that work?
e.g.
```go
func (h *HeartbeatServer) Start(ctx context.Context) error {
	if err := h.subscription.Subscribe(ctx, h); err != nil {
		return err
	}
       	log.Ctx(ctx).Info().Msg("Heartbeat server started")
       	
       	go func(ctx context.Context) {
       	        defer func() {
       	                h.subscription.Close(ctx) // TODO and maybe handle error with log message
       	                log.Ctx(ctx).Info().Msg("Heartbeat server shutdown")
       	        }
		ticker := time.NewTicker(heartbeatQueueCheckFrequency)

		for {
			select {
			case <-ctx.Done():
				return
			case <-ticker.C:
				// We'll iterate over the priority queue and check if the last heartbeat
				// received from each node is older than 10 seconds. If it is, we'll
				// consider the node as dead.
			}
		}
	}(ctx)

	return nil
}

Comment (rossjones) [reply pkg/node/heartbeat/server.go:50]: Oh yeah, got it. I think I refactored the other goroutine to no longer being necessary but forgot to remove it.

File: pkg/node/heartbeat/client.go [hunk 4]
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return &HeartbeatClient{publisher: publisher, nodeID: nodeID}, nil
+}
+
+func (h *HeartbeatClient) Start(ctx context.Context) error {
+	// Waits until we are cancelled and then closes the publisher
+	<-ctx.Done()
+	return h.publisher.Close(ctx)
+}
```
frrist, line 35: `Start` usually implies a non-blocking operation, e.g.  [HeartbeatServer Start method](https://github.com/bacalhau-project/bacalhau/pull/3709/files#diff-901570ddabfcccd3c9bf01ede24a06556fe7cef0d696a10428b74887ba9c7d6bR52). This method is blocking until the context passed to it is canceled. 

I think the change here is to implement this `Start` method in a similar manner to the `HeartbeatServer` stop method - e.g. spawn a go-routine that waits on the context and closes the publisher when its canceled.

pkg/node/heartbeat/types.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "time"
+
+const (
+	heartbeatTopic               = "heartbeat"
+	heartbeatQueueCheckFrequency = 5 * time.Second
+
+	unhealthyAfter = 30 * time.Second
+	unknownAfter   = 60 * time.Second
+)
```
frrist, line 11: In a similar vein to my comment on the ManagementClient, I think these should be configurable, with defaults that don't require the ManagementClient to heartbeat twice as frequently as the configured heartbeat interval.

pkg/node/node.go
```diff
@@ -283,11 +292,20 @@ func NewNode(
...
+			Heartbeats: heartbeatSvr,
 		})
 
+		// Start the nodemanager, ensuring it doesn't block the main thread and
+		// that any errors are logged. If we are unable to start the manager
+		// then we should not start the node.
+		if err := nodeManager.Start(ctx); err != nil {
+			return nil, errors.Wrap(err, "failed to start node manager")
+		}
+
```
frrist, line 308: Any reason not to perform this construction in the `NewRequesterNode` method?

pkg/node/node.go
```diff
@@ -348,6 +366,24 @@ func NewNode(
...
+			natsClient, err := nats_transport.CreateClient(ctx, natsConfig)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create NATS client for node info store")
+			}
+
+			hbClient, err = heartbeat.NewClient(natsClient.Client, config.NodeID)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create heartbeat client")
+			}
+		}
```
frrist, line 385: Similarly, could this be done in `NewComputeNode`?

frrist (commented):
Good stuff! 

Left a few comment's I'd like addressed/discussed before merging.

pkg/models/node_state.go
```diff
@@ -0,0 +1,94 @@
+package models
```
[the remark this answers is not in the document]
rossjones: It was, but ran up against everything being in the same package, so for instance `unknown` was already defined :|

pkg/node/node.go
rossjones: natsConfig doesn't live any longer than NewNode(), would make more sense to do this as part of the cleanup/refactor because there's a lot of overlapping checks in NewNode() and NewCompute() doesn't really know anything about NATS (currently).

pkg/node/node.go
rossjones: heartbeatSvr is constructed from a client that is optionally created in NewNode().  It should probably move as part of rework/refactor how these nodes are constructed as at present there's been attempt to isolate all the libp2p/nats stuff in NewNode()

pkg/compute/management_client.go
```diff
@@ -11,45 +11,50 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/compute/capacity"
 	"github.com/bacalhau-project/bacalhau/pkg/models"
 	"github.com/bacalhau-project/bacalhau/pkg/models/requests"
+	"github.com/bacalhau-project/bacalhau/pkg/node/heartbeat"
 )
 
 const (
 	infoUpdateFrequencyMinutes     = 5
 	resourceUpdateFrequencySeconds = 30
+	heartbeatFrequencySeconds      = 30
```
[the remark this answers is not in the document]
rossjones: Configuration done in 08b9ab2cd, but will add documentation in a following commit.

pkg/compute/management_client.go
```diff
@@ -124,36 +129,57 @@ func (m *ManagementClient) updateResources(ctx context.Context) {
...
 func (m *ManagementClient) Start(ctx context.Context) {
 	infoTicker := time.NewTicker(infoUpdateFrequencyMinutes * time.Minute)
 	resourceTicker := time.NewTicker(resourceUpdateFrequencySeconds * time.Second)
 
-	loop := true
-	for loop {
+	// The heartbeat ticker will fire twice as often as the configured, to ensure that
+	// we don't slip outside the window.  If we only ever sent on the configured
+	// frequency we are at risk of the node's liveness flapping between good and bad.
+	heartbeatTicker := time.NewTicker((heartbeatFrequencySeconds / 2) * time.Second)
```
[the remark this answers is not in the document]
rossjones: Yeah, that's reasonable.  In 08b9ab2cd I updated the comment to make it clear that it is the users responsibility to make sure it is lower than the requester node's expectations.  I'll add more in docs.

pkg/compute/management_client.go
```diff
@@ -124,36 +129,57 @@ func (m *ManagementClient) updateResources(ctx context.Context) {
...
 		case <-infoTicker.C:
 			// Send the latest node info to the requester node
 			m.deliverInfo(ctx)
 		case <-resourceTicker.C:
 			// Send the latest resource info
 			m.updateResources(ctx)
+		case <-heartbeatTicker.C:
+			// Send a heartbeat to the requester node
+			heartbeatSequence += 1
+			m.heartbeat(ctx, heartbeatSequence)
```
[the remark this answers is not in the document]
rossjones: Messages are already logged in nats.PubSub

pkg/models/node_state.go
```diff
@@ -0,0 +1,94 @@
...
+		unhealthy:    "UNHEALTHY",
+		healthy:      "HEALTHY",
+	}
+
+	typeLivenessMap = map[string]liveness{
+		"UNKNOWN":   unknownState,
+		"UNHEALTHY": unhealthy,
+		"HEALTHY":   healthy,
+	}
+)
```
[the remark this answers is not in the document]
rossjones: Added in 92fbc1f41

pkg/node/heartbeat/client.go
rossjones: Actually was able to move this to a Close() method and call it when the containing component shuts down instead. This is the only place that uses it so it feels safe.

cmd/cli/node/list.go
```diff
@@ -16,15 +16,17 @@ import (
 
 var defaultColumnGroups = []string{"labels", "capacity"}
 var orderByFields = []string{"id", "type", "available_cpu", "available_memory", "available_disk", "available_gpu", "status"}
-var filterStatusValues = []string{"approved", "pending", "rejected"}
+var filterApprovalValues = []string{"approved", "pending", "rejected"}
+var filterStatusValues = []string{"healthy", "unhealthy", "unknown"}
```
aronchick, line 20: i think it should be binary - "connected" or "unconnected" (we could add something more sophisticated like "unhealthy" when we have more health checks

docs/docs/dev/cli-reference/cli/node/list/index.md
```diff
@@ -1,52 +1,74 @@
...
 
 - `--hide-header`:
+
   - Do not display the column headers in the output.
 
+- `--filter-approval`:
+
+  - Only show nodes with the specified approval status. Valid values are: `approved`, `pending`, `rejected`.
+
+- `--filter-approval`:
```
aronchick, line 33: Dupe?

aronchick (approved):
LGTM  but i think we should go with just a binary classification

docs/docs/dev/cli-reference/cli/node/list/index.md
rossjones: Fixed in 63336e12f

cmd/cli/node/list.go
rossjones: Just for the filters, or for the`node list` as well?

aronchick: i think the node list as well - "healthy" and "unhealthy" imply something i don't think we know. We only know if it's connected or disconnected.

cmd/cli/node/list.go
```diff
@@ -42,22 +44,24 @@ func NewListCmd() *cobra.Command {
...
+		fmt.Sprintf("Filter nodes by approval. One of: %q", filterApprovalValues))
 	nodeCmd.Flags().StringVar(&o.FilterByStatus, "filter-status", o.FilterByStatus,
 		fmt.Sprintf("Filter nodes by status. One of: %q", filterStatusValues))
 
 	return nodeCmd
 }
 
 // Run executes node command
-func (o *ListOptions) run(cmd *cobra.Command, _ []string) {
+func (o *ListOptions) run(cmd *cobra.Command, _ []string) error {
```
frrist, line 64: if we are going to return an error from this method can we replace all the `util.Fatal` calls while we are at it? (be wary  this might break some tests that (annoyingly) override the `util.Fatal` field :disappointed:)

cmd/cli/node/list.go
```diff
@@ -69,15 +73,22 @@ func (o *ListOptions) run(cmd *cobra.Command, _ []string) {
...
+	if o.FilterByApproval != "" {
+		if !slices.Contains(filterApprovalValues, o.FilterByApproval) {
+			return fmt.Errorf("cannot use '%s' as filter approval value, should be one of: %q", o.FilterByApproval, filterApprovalValues)
+		}
+	}
+
 	if o.FilterByStatus != "" {
 		if !slices.Contains(filterStatusValues, o.FilterByStatus) {
-			util.Fatal(cmd, fmt.Errorf("cannot use '%s' as filter status value, should be one of: %q", o.FilterByStatus, filterStatusValues), 1)
+			return fmt.Errorf("cannot use '%s' as filter status value, should be one of: %q", o.FilterByStatus, filterStatusValues)
```
frrist, line 84: consider printing the specific flag name in the error message e.g. `filter-status` and `filter-approval`

docs/docs/dev/cli-reference/cli/node/approve/index.md
```diff
@@ -0,0 +1,67 @@
+---
+sidebar_label: approve
+---
+
+# Command: `node approve`
+
+The `bacalhau node approve` command offers administrations the ability to approve the cluster membership for a node using its unique identifier.
```
frrist, line 7: command offers ~administrations~ administrators?
unique identifier being synonymous with name, right?

docs/docs/dev/cli-reference/cli/node/approve/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+  - The unique identifier of the node you wish to describe.
+
+- `-h`, `--help`:
+
+  - Displays the help documentation for the `describe` command.
+
+- `-m message`:
+
+  - A message to be attached to the approval action.
```
frrist, line 31: makes me wonder if we should also include the ClientID of the user issuing the approval, but that could always be included later.

docs/docs/dev/cli-reference/cli/node/delete/index.md
```diff
@@ -0,0 +1,67 @@
+---
+sidebar_label: delete
+---
+
+# Command: `node approve`
```
frrist, line 5: node delete

docs/docs/dev/cli-reference/cli/node/delete/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+The `bacalhau node delete` command offers administrations the ability to remove a node from the cluster using its unique identifier.
```
frrist, line 7: similar comment about administrations

docs/docs/dev/cli-reference/cli/node/delete/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Description:
+
+Using the `delete` sub-command, administrators can remove a node from the list of available compute nodes in the cluster. This feature is necessary for the management of the infrastructure.
```
frrist, line 11: for consistency with above:  Using the `delete` sub-command under the `bacalhau node` umbrella

docs/docs/dev/cli-reference/cli/node/delete/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+bacalhau node delete [id] [flags]
+```
+
+## Flags:
+
+- `[id]`:
+
+  - The unique identifier of the node you wish to describe.
```
frrist, line 23: ~describe~ delete

docs/docs/dev/cli-reference/cli/node/approve/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+bacalhau node approve [id] [flags]
+```
+
+## Flags:
+
+- `[id]`:
+
+  - The unique identifier of the node you wish to describe.
```
frrist, line 23: ~describe~ approve

docs/docs/dev/cli-reference/cli/node/approve/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `-h`, `--help`:
+
+  - Displays the help documentation for the `describe` command.
```
frrist, line 27: ~describe~ node approve

docs/docs/dev/cli-reference/cli/node/approve/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `--log-mode logging-mode`:
+
+  - Determines the log format preference.
+  - Options: `'default','station','json','combined','event'`
+  - Default: `'default'`
+
+- `--repo string`:
+  - Points to the bacalhau repository's path.
+  - Default: `"`$HOME/.bacalhau"`
```
frrist, line 53: I rather not have these repeated here and just document them one, optionally providing a link, otherwise it draw attention away from what this page is documenting imo.

docs/docs/dev/cli-reference/cli/node/delete/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `-h`, `--help`:
+
+  - Displays the help documentation for the `describe` command.
```
frrist, line 27: replace describe

docs/docs/dev/cli-reference/cli/node/delete/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `--log-mode logging-mode`:
+
+  - Determines the log format preference.
+  - Options: `'default','station','json','combined','event'`
+  - Default: `'default'`
+
+- `--repo string`:
+  - Points to the bacalhau repository's path.
+  - Default: `"`$HOME/.bacalhau"`
```
frrist, line 53: I rather not have these repeated here and just document them one, optionally providing a link, otherwise it draw attention away from what this page is documenting imo.

docs/docs/dev/cli-reference/cli/node/index.md
```diff
@@ -12,48 +12,75 @@ bacalhau node [command]
...
 bacalhau node [command] --help
 ```
 
 ## Flags
 
 - `-h`, `--help`:
-    - Description: Shows the help information for the `node` command.
+  - Description: Shows the help information for the `node` command.
 
 ## Global Flags
```
frrist, line 67: similar sentiment around not repeating global flags

docs/docs/dev/cli-reference/cli/node/index.md
```diff
@@ -12,48 +12,75 @@ bacalhau node [command]
...
+- `--log-mode logging-mode`:
 
----
+  - Description: Chooses the preferred log format. Available choices are: `default`, `station`, `json`, `combined`, and `event`.
+  - Default: `default`
 
-This should provide an organized and structured overview of the `node` command and its functionalities!
+- `--repo string`:
+  - Description: Specifies the path to the bacalhau repository.
+  - Default: `/Users/walid/.bacalhau`
```
frrist, line 86: who's walid? :trollface:

docs/docs/dev/cli-reference/cli/node/reject/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+bacalhau node rejected [id] [flags]
+```
+
+## Flags:
+
+- `[id]`:
+
+  - The unique identifier of the node you wish to describe.
```
frrist, line 23: not describe

docs/docs/dev/cli-reference/cli/node/reject/index.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `-h`, `--help`:
+
+  - Displays the help documentation for the `describe` command.
```
frrist, line 27: not describe

docs/docs/dev/node_management.md
```diff
@@ -0,0 +1,96 @@
+# Node Management
+
+Bacalhau clusters are composed of requester nodes, and compute nodes. The requester nodes are responsible for managing the compute nodes that make up the cluster. This functionality is only currently available when using NATS for the network transport.
+
+The two main areas of functionality for the requester nodes are, managing the membership of compute nodes that require approval to take part in the cluster, and monitoring the health of the compute nodes. They are also responsible for collecting information provided by the compute nodes on a regular schedule.
+
+## Compute node membership
+
+As compute nodes start, they register their existence with the requester nodes. Once registered, they will maintain a sentinel file to note that they are already registered, this avoids unnecessary registration attempts.
```
frrist, line 9: When authentication is enabled, in order to even connect to a requester node, compute nodes must have the correct secret/auth values, right?

Asking as I am curious about the behavior when two compute nodes with the same ID/Name attempt to register, if one is approved will the second node with the same ID also be considered approved? The ID is just a field users provide in their config file so its very possible for two nodes to have the same ID/Name

docs/docs/dev/node_management.md
```diff
@@ -0,0 +1,96 @@
...
+
+Once registered, the requester node will need to approve the compute node before it can take part in the cluster. This is to ensure that the requester node is aware of all the compute nodes that are part of the cluster. In future, we may provide mechanisms for auto-approval of nodes joining the cluster, but currently all compute nodes registering default to the PENDING state.
```
frrist, line 11: We should make sure this isn't thought of as a security feature.

pkg/config/types/compute.go
```diff
@@ -62,3 +63,20 @@ type LocalPublisherConfig struct {
 	Port      int    `yaml:"Port"`
 	Directory string `yaml:"Directory"`
 }
+
+type ComputeControlPlaneConfig struct {
```
frrist, line 67: Naming is hard. Think `ComputeHeartbeatConfig` might make more sense? or `ManagmentClientConfig` since that's the constructor that accepts this?

pkg/config/types/requester.go
```diff
@@ -53,3 +55,20 @@ type S3StorageProviderConfig struct {
 type JobDefaults struct {
 	ExecutionTimeout Duration `yaml:"ExecutionTimeout"`
 }
+
+type RequesterControlPlaneConfig struct {
```
frrist, line 59: Again, naming is hard. Maybe `HeartbeatServerConfig`? based on the constructor that accepts this?

pkg/models/node_state.go
```diff
@@ -0,0 +1,99 @@
...
+	liveness
+}
+
+type liveness int
+
+// To add a new state,
+// * add it to the end of the list in the const below
+// * add it to strLivenessArray and typeLivenessMap
+// * add it to the livenessContainer and corresponding NodeStates var.
+// * add it to the All() method in the livenessContainer
```
frrist, line 17: <3

pkg/compute/management_client.go
```diff
@@ -124,36 +127,62 @@ func (m *ManagementClient) updateResources(ctx context.Context) {
 		Resources: resources,
 	})
 	if err != nil {
-		log.Ctx(ctx).Error().Err(err).Msg("failed to send resource update to requester node")
```
frrist, line 127: While I am thinking about this. Above on line 103 where we call `getNodeInfo` we do:
```go
nodeInfo := m.getNodeInfo(ctx)
	response, err := m.managementProxy.UpdateInfo(ctx, requests.UpdateInfoRequest{
		Info: nodeInfo,
	})
	if err != nil {
		log.Ctx(ctx).Error().Err(err).Msg("failed to send update info to requester node")
	}
// TODO response will be nil when err != nil and this code will panic (has panicked for me)
	if response.Accepted {
		log.Ctx(ctx).Debug().Msg("update info accepted")
	} else {
		log.Ctx(ctx).Error().Msgf("update info rejected: %s", response.Reason)
	}
```
Can we address the panic here?

pkg/node/heartbeat/server.go
```diff
@@ -0,0 +1,221 @@
...
+		return err
+	}
+
+	log.Ctx(ctx).Info().Msg("Heartbeat server started")
+
+	tickerStartCh := make(chan struct{})
+
+	go func(ctx context.Context) {
+		defer func() {
+			_ = h.subscription.Close(ctx) // We're closing down, not much we can do with an error
```
frrist, line 87: we can log it on the line below

pkg/node/manager/node_manager.go
```diff
@@ -37,9 +40,24 @@ func NewNodeManager(params NodeManagerParams) *NodeManager {
 	return &NodeManager{
 		resourceMap: concurrency.NewStripedMap[models.Resources](resourceMapLockCount),
 		nodeInfo:    params.NodeInfo,
+		heartbeats:  params.Heartbeats,
 	}
 }
 
+func (n *NodeManager) Start(ctx context.Context) error {
+	log.Ctx(ctx).Info().Msg("Node manager started")
```
frrist, line 48: log this after the thing starts, it technically hasn't started yet, and it my not start if `n.heartbeats` is `nil`

pkg/node/manager/node_manager.go
```diff
@@ -37,9 +40,24 @@ func NewNodeManager(params NodeManagerParams) *NodeManager {
...
+
+	if n.heartbeats != nil {
```
frrist, line 50: when can this condition evaluate to false?

frrist (approved):
Looks good, once comments are address :shipit:

cmd/cli/node/list.go
rossjones: Can definitely add a ticket to address this, yeah. There's a mix of approaches atm.

rossjones: Added https://github.com/bacalhau-project/bacalhau/issues/3764

pkg/compute/management_client.go
rossjones: I think I intended this to return after logging the error.  Fixed in a96c1b00137a442f20a1ae8b920c9b632c6f7daa

pkg/node/manager/node_manager.go
rossjones: When we're running some tests.

cmd/cli/node/list.go
rossjones: Went with Connected/Disconnected for now, discussed with Walid and will add another duration in future to mark the point beyond which we don't think the node is coming back.

PR #3711, opened 2024-03-27
Add authentication to the Web UI

### LINKED ISSUE #3641: WebUI cannot access authenticated bacalhau node
When a bacalhau cluster is created that requires authentication the WebUI is unable to communicate with its API as the WebUI does not provide the required authentication headers.

### LINKED ISSUE #3679: Web UI: TypeError: node.PeerInfo is undefined
The nodes dashboard does not display any nodes and complains in the console with `TypeError: node.PeerInfo is undefined`.

The code appears to be expecting the structure returned from `/api/v1/node` to contain a `PeerInfo` object but this API does not return this.

simonwo:
This commit hooks up the Web UI to use our authentication APIs. It can now query for a list of acceptable authentication methods and show the user a web form for filling out the required credentials.

The UI will monitor requests and if it detects an Unauthorized status will immediately open the auth flow. So as with the CLI, auth happens automagically when required and not otherwise.

With default authorization policy in place, which doesn't require authz to use read-only APIs, the user won't see any authentication because it is not required.

But, for example, with a username/password and non-anon policy in place, the user will be redirected to a page to provide credentials:

image

The system remains entirely flexible, and the required credentials is still controlled by the auth policy. So with the "shared secret" auth policy used by the Marketplace deployment, the user will see:

image

Resolves #3679
Resolves #3641
Resolves https://github.com/bacalhau-project/expanso-planning/issues/668
Resolves https://github.com/bacalhau-project/expanso-planning/issues/669

### DIFF
--- a/pkg/publicapi/endpoint/orchestrator/endpoint.go
+++ b/pkg/publicapi/endpoint/orchestrator/endpoint.go
@@ -6,7 +6,6 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/orchestrator"
 	"github.com/bacalhau-project/bacalhau/pkg/publicapi/middleware"
 	"github.com/labstack/echo/v4"
-	echo_middleware "github.com/labstack/echo/v4/middleware"
 )
 
 type EndpointParams struct {
@@ -34,7 +33,6 @@ func NewEndpoint(params EndpointParams) *Endpoint {
 	// JSON group
 	g := e.router.Group("/api/v1/orchestrator")
 	g.Use(middleware.SetContentType(echo.MIMEApplicationJSON))
-	g.Use(echo_middleware.CORS())
 	g.PUT("/jobs", e.putJob)
 	g.POST("/jobs", e.putJob)
 	g.GET("/jobs", e.listJobs)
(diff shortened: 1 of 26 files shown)


rossjones (commented):
LGTM but can't currently build it to try it out because of  missing 'react-toastify' and a mention of PeerInfo in the node mock.

pkg/publicapi/endpoint/orchestrator/endpoint.go
```diff
@@ -34,7 +33,6 @@ func NewEndpoint(params EndpointParams) *Endpoint {
 	// JSON group
 	g := e.router.Group("/api/v1/orchestrator")
 	g.Use(middleware.SetContentType(echo.MIMEApplicationJSON))
-	g.Use(echo_middleware.CORS())
```
frrist, line 37: why did we need to remove this? I don't have a problem with the change - just curious

rossjones: It's added at https://github.com/bacalhau-project/bacalhau/pull/3711/files#diff-19907c38cadb733516c5d0d2d452e8ef1dde593253b301251fec43f1a8a8d315 instead

simonwo: https://github.com/bacalhau-project/bacalhau/pull/3711/commits/134db3f2c8a082744a4b8232ba30a7cc54fe18db

simonwo:
It builds now.

rossjones (approved):
LGTM.

PR #3716, opened 2024-03-27
Adds a `node delete` command

### LINKED ISSUE #3715: Add node command to delete nodes from the info store
Nodes are currently always kept in the node info store once they have been seen.  There are likely use-cases where we want to manually delete nodes from the store.  We should add `node delete`

rossjones:
Allows for nodes to be delete from the node info store. If the node is found in the store, it is deleted, and then will not re-appear when the requester node is re-started.  If however the compute node restarts, it will re-register so this command is only useful if you know that nodes will not be coming back.

In future if we want to block the node-id from ever being an entry, we should add tombstone entries to the node info store that blog it's registration.

Closes #3715

### DIFF
--- a/pkg/publicapi/apimodels/node.go
+++ b/pkg/publicapi/apimodels/node.go
@@ -69,7 +69,7 @@ func (n NodeAction) Description() string {
 	case NodeActionReject:
 		return "Reject a node whose membership is pending"
 	case NodeActionDelete:
-		return "Delete a node from the cluster. Note this does not stop them re-connecting."
+		return "Delete a node from the cluster."
 	}
 	return ""
 }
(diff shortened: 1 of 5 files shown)


pkg/publicapi/apimodels/node.go
```diff
@@ -67,10 +68,12 @@ func (n NodeAction) Description() string {
 		return "Approve a node whose membership is pending"
 	case NodeActionReject:
 		return "Reject a node whose membership is pending"
+	case NodeActionDelete:
+		return "Delete a node from the cluster. Note this does not stop them re-connecting."
```
aronchick, line 72: I don't think we need this warning. If a node disconnects, does it have to go pending again? (it should)

aronchick (approved):
LGTM other than the one question about reconnecting

pkg/publicapi/apimodels/node.go
rossjones: Yes, would go back to Pending, unless whatever criteria we're using for auto-approve (tbd) puts them in that state.

PR #3721, opened 2024-03-27
Support MP deployments with self-generated TLS certificate

frrist:
Generates a TLS certificate for communication between clients and requester node.


ops/marketplace-tf/vars.tfvars
```diff
@@ -17,9 +17,9 @@ bacalhau_accept_networked_jobs = true
 
 bacalhau_otel_collector_endpoint = "http://analytics.bacalhau.tech:4318"
 
-bacalhau_requester_api_token = "token_for_requester_api"
-bacalhau_compute_api_token = "token_for_compute_api"
+bacalhau_requester_api_token = "password"
```
aronchick, line 20: did you mean to check in these as defaults? Terraform offers a feature to generate a random string (or this should be left blank and required)

aronchick (approved):
LGTM other than the password/defaults.

ops/marketplace-tf/vars.tfvars
frrist: If these values are blank a random token will automatically be generated and used. I commit this since typing `password` was easier than the former.

ops/marketplace-tf/modules/cloud-init/compute-cloud-init.yml
```diff
@@ -0,0 +1,107 @@
+# create the bacalhau user that will run the bacalhau process
+packages:
+  - vim
```
rossjones, line 3: :heart:

ops/marketplace-tf/modules/cloud-init/compute-cloud-init.yml
```diff
@@ -0,0 +1,107 @@
...
+  # bacalhau service file to run the bacalhau process
+  - path: /etc/systemd/system/bacalhau.service
+    encoding: b64
+    owner: root:root
+    permissions: "0600"
+    content: |
+      ${bacalhau_service_file}
+
+  # bacalhau install-script to install a custom bacalhau version
+  - path: /etc/install-bacalhau.sh
```
rossjones, line 55: /etc probably not the best place for this. /usr/local/sbin's the best place for binaries that need su,or /usr/local/bin if it is intended for wider use.  If it's only used by the installer, then probably less of an issue.

ops/marketplace-tf/modules/cloud-init/requester-cloud-init.yml
```diff
@@ -0,0 +1,123 @@
...
+  # bacalhau service file to run the bacalhau process
+  - path: /etc/systemd/system/bacalhau.service
+    encoding: b64
+    owner: root:root
+    permissions: "0600"
+    content: |
+      ${bacalhau_service_file}
+
+  # bacalhau install-script to install a custom bacalhau version
+  - path: /etc/install-bacalhau.sh
```
rossjones, line 82: _if_ you change the other one (not compulsory) then change this one too.

ops/marketplace-tf/modules/gcp/compute_instances/compute/main.tf
```diff
@@ -74,7 +74,7 @@ locals {
   bacalhau_env_vars = {
     LOG_LEVEL                   = "debug"
     BACALHAU_NODE_LOGGINGMODE   = "default"
-    BACALHAU_DIR                = "/data"
+    BACALHAU_DIR                = "/bacalhau_repo"
```
rossjones, line 77: This is okay, but wonder if it might be better in /opt/bacalhau/* eventually ?
 
c.f. https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html

rossjones (approved):
LGTM. Couple suggestions for potential directory changes, but nothing urgent.

ops/marketplace-tf/vars.tfvars
wdbaruni: I hate that we auto-generate tokens and ask the users to go through the logs or to a local file on desk to get whatever token was generated. We should either allow unsecure access by default, or fail if no token was provided. but we shouldn't auto generate tokens

aronchick: i think i'm ok as well if the token goes into a defined file somewhere. Not logs. This feels like best practice to me.

PR #3723, opened 2024-03-28
fix: attempt to return an error if ListenAndServe fails

frrist:


pkg/publicapi/server.go
```diff
@@ -211,27 +212,54 @@ func (apiServer *Server) ListenAndServe(ctx context.Context) error {
...
+	// one of which is when there are issues reading the TLS configuration.
+	// In most cases, if the server starts successfully, http.Serve* will block and not return an error.
+	// We use a select statement to wait for either an error from errChan or a timeout of 2 seconds.
+	// If an error is received within the timeout, it means the server encountered an issue during startup.
+	// If no error is received within the timeout, we assume the server started successfully.
+	// TODO: call os.Exit(1) or log.Fatal if the server stops running for a reason other than ErrServerClosed
+	select {
+	case startErr := <-errChan:
+		return startErr
+	case <-time.After(time.Second * 2):
```
rossjones, line 258: It seems a shame to introduce a 2 second pause to startup just for the log message.
How about logging this message at the start of the goroutine instead?
Then we can replace the time.After with a ctx.Done() check

rossjones (approved):
LGTM but seems to be failing tests because shutdown isn't clean and it doesn't get a chance to close the NATS server.

pkg/publicapi/server.go
frrist: Yeah, on second look I don't think this is the right change. I wrote it as when deploying bacalhau with terraform. The user running bacalhau didn't have permissions to read the TLS config. Definitely a rare case, but annoying to debug since the Bacalhau process kept running but didn't respond to any requests. Further I failed to see `failed to start API server` in the log messages.

How would we feel about calling `log.Fatal` in the event the server returns an error? If we did that the user would see:
`15:03:12.041 | FTL pkg/publicapi/server.go:231 > Api server can't run. Cannot serve client requests! error="failed to start server" [NodeID:n-fde9a353]` and the process would exit. I suggest this as I don't see any value in running bacalhau if the API server isn't active. On the other hand this might leave some datastores in a weird state.

Alternatively I am happy to just close this PR and discount my experience as a one-off issue that probably won't happen again.

rossjones: > How would we feel about calling log.Fatal in the event the server returns an error

I am 100% in favour of it visibily crashing rather than quietly failing :D

PR #3735, opened 2024-03-29
wip: first pass at writing a job schema in cue

frrist:
1. Download the latest version of `cue`
  - https://github.com/cue-lang/cue/releases/tag/v0.8.0 
2. vet your job specs with this via the command:
  - `cue vet -d "#Job" testdata/jobs/docker-output.yaml pkg/models/job-schema.cue`

Couple notes:
- This is a work in progress, not perfect
- At present the schema requires that keys in your yaml be upper case

### DIFF
--- a/pkg/models/job-schema.cue
+++ b/pkg/models/job-schema.cue
@@ -1,0 +1,161 @@
+import "strings"
+
+#Job: {
+    Name:       string
+    Namespace:  string
+    Type:       "daemon" | "ops" | "batch" | "service"
+    Priority:   int & >=0 | *1
+    Count:      int & >0
+    Constraints?: [...#LabelSelectorRequirement]
+    Labels?: [string]: string
+    Tasks: [...#Task]
+}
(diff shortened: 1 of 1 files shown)


frrist:
@coderabbitai /review

rossjones:
Interestingly, if we used https://cuelang.org/docs/integration/go/#processing-cue-in-go, could we remove all of our individual `Validatable`/`Normalizable` implementations?   We also get validation of the SpecConfigs that we don't have at present, which seems useful.  

What other benefits are there to using this to implement `bacalhau job validate` as opposed to importing the models and using the `Validatable` interface?  It looks like it might be possible to have a client generate the job spec entirely from a cue file (perhaps even versions) without the client having to juggle all of the existing models - downside being we'd miss out on the openapi specification of the endpoint methods/objects etc.

frrist:
Will revisit when job validation becomes higher priority.

PR #3738, opened 2024-03-29
Improve bashtub coverage

js-ts:


aronchick (changes requested):
Can you break this up into many pull requests? Start with just the smallest necessary to verify that 'version' works properly and nothing else.

aronchick:
I don't understand thing - we asked you to do one, and you did 27. Please do one so that we can actually review it.

test/jobs/permutations.json
```diff
@@ -0,0 +1,49 @@
+[
+{"Engine":"docker", "StorageSources":"", "Publishers":"", "Count":1}
+,
+{"Engine":"docker", "StorageSources":"", "Publishers":"ipfs", "Count":1}
```
js-ts, line 4: @aronchick 1 test

PR #3744, opened 2024-04-01
Replace use of multierror libraries with now standardised errors.Join

simonwo:
`errors.Join` preserves the tree of errors by producing structures with `Unwrap() []error` calls available whereas multierr{or} only produces a linear chain of errors using `Unwrap() error` calls. This is a largely mechanical change using find/replace.

This is necessary for subsequent work to returned properly structured errors from the API and show them in the CLI.

### DIFF
--- a/cmd/cli/serve/util.go
+++ b/cmd/cli/serve/util.go
@@ -2,6 +2,7 @@ package serve
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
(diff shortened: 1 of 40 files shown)


simonwo:
Yes, it's true. In the next commits I'll introduce a better mechanism for printing errors  in the CLI, which will look more like the latter example. So if you want me to close this PR and merge it in with that work, I can do that.

PR #3749, opened 2024-04-02
Improve error reporting when job resources exceed capacity

simonwo:
These messages are now:
- specific about which resource(s) is/are over the limit
- do not include resources that are not problematic
- print numbers to a sensible level of precision
- also fix capacity string output

Resolves #3351.
Resolves https://github.com/bacalhau-project/expanso-planning/issues/692.


aronchick (approved):
LGTM

PR #3756, opened 2024-04-04
Return structured errors from API calls and improve job error display

simonwo:
This PR adds special handling for a number of common error cases:
* Don't show the full list of nodes if they all output the same error, and if the job error matches one of the execution errors (common case) don't show that separately either.
* Show the number of nodes that were available even if unsuitable, because users often misinterpret "0 available" as meaning there is a connection error with their compute nodes
* Summarise nodes when they all print the same output rather than listing all of them

An example of how execution errors looked previously:
```
% bin/darwin/arm64/bacalhau job run testdata/jobs/docker.yaml 
Job successfully submitted. Job ID: j-6207c87b-d2ce-4458-b55e-1b2efb9d2a91

	   Creating job for submission  ................  err  ❌  13.5s

Error submitting job: not enough nodes to run job. requested: 1, available: 0. 
	Node node-1: job rejected
	Node node-2: job rejected
	Node node-3: job rejected
	Node node-4: job rejected
Job Results By Node:
• Node node-3, node-2, node-1, node-4: 
	rejected bid: this node does not Could not inspect image "not-existyyyy:latest" - could be due to repo/image not existing, or registry needing authorization: Error response from daemon: errors:
	denied: requested access to the resource is denied
	unauthorized: authentication required

	bacalhau job describe j-6207c87b-d2ce-4458-b55e-1b2efb9d2a91

	bacalhau job executions j-6207c87b-d2ce-4458-b55e-1b2efb9d2a91
```

An example of how execution errors look now:
carbon

So this doesn't fix all of the long ugly go errors which need to be dealt with on a case-by-case basis, but it at least simplifies the output and allows users to focus on the information that is pertinent.

Resolves https://github.com/bacalhau-project/expanso-planning/issues/693.

### DIFF
--- a/pkg/lib/bad/error.go
+++ b/pkg/lib/bad/error.go
@@ -1,0 +1,184 @@
+// Package bad helps you deal with errors.
+package bad
+
+import (
+	"strings"
+)
+
+// ErrorSubject identifies where the error arose according to the component that
+// emitted it.
+type ErrorSubject string
+
+const (
+	// The error arose for an unspecified reason.
+	ErrSubjectNone ErrorSubject = ""
+	// The error arose because user input did not conform to the required
+	// specification. The user will need to modify their request before
+	// submitting it again.
+	ErrSubjectInput ErrorSubject = "input"
+	// The error arose because of an internal logic problem with the component
+	// that created the error, normally signaling a programming issue. The
+	// system developer will need to modify their code to meet this request.
+	ErrSubjectInternal ErrorSubject = "internal"
+	// The error arose because some other service or component that the
+	// component that created the error depends on also returned an error. The
+	// error may be transient, especially if this error is a leaf, or may
+	// require operator intervention (e.g. out of disk space).
+	ErrSubjectDependency ErrorSubject = "dependency"
+)
+
+// Error represents a node in a tree of errors. Error (and thus the whole tree)
+// is JSON marshallable so that a tree of Errors can be sent across HTTP.
+type Error struct {
+	// The type of the error. For native errors, this is just the error text,
+	// but more structured errors this is a unique identifier that clients can
+	// use to display certain types of important errors in specific ways.
+	// Essentially, the "type" identifies the type of thing in the "data" field,
+	// and allows the client to unmarshall the data field appropriately.
+	//
+	// For example, if a client knows how to represent errors about resource
+	// limits being unmet, the type may be "Unmet resource limits" and the data
+	// field may include structured information about what resource limits are
+	// being unmet that the client can choose to display in a number of ways.
+	Type string `json:"type"`
+	Data any    `json:"data"`
+
+	// The subject of the error, which identifies where the problem occurred and
+	// is used to select appropriate status codes and retry information.
+	//
+	// Not every Error needs to have a Subject specified – instead subjects can
+	// be added to tag any error from a certain place. For example, a Validate()
+	// method will almost always be returning an error due to user input, so any
+	// error emitted from this place can be wrapped with a ErrSubjectInput.
+	Subject ErrorSubject `json:"subject"`
+
+	// Any child errors that this error wraps.
+	Errs []Error `json:"errs"`
+}
+
+// Error() is necessary for Error to implement the error interface, which makes
+// it much easier to pass back from functions. We shouldn't actually use this
+// except in internal logging, but it implements a simple printer that will
+// output the entire tree of errors, skipping any blank levels.
+func (err *Error) Error() string {
+	var out strings.Builder
+	if err == nil {
+		return ""
+	}
+
+	indent := ""
+	if len(err.Type) > 0 {
+		out.WriteString("* ")
+		out.WriteString(err.Type)
+		indent = "\t"
+		if len(err.Errs) > 0 {
+			out.WriteString(": ")
+		}
+		out.WriteRune('\n')
+	}
+	for _, subError := range err.Errs {
+		for _, line := range strings.SplitAfter(subError.Error(), "\n") {
+			if line == "" {
+				continue
+			}
+			out.WriteString(indent)
+			out.WriteString(line)
+		}
+	}
+	return out.String()
+}
+
+var _ error = (*Error)(nil)
+
+// Leaves returns just the Errors from the tree that do not have any children.
+// These Errors are normally the ones with the most specific error information.
+// The order in which the leaves are returned is unspecified.
+func (err *Error) Leaves() []Error {
+	if err == nil {
+		return nil
+	}
+
+	if len(err.Errs) == 0 {
+		return []Error{*err}
+	}
+
+	leaves := make([]Error, 0, len(err.Errs))
+	for _, subErr := range err.Errs {
+		leaves = append(leaves, subErr.Leaves()...)
+	}
+	return leaves
+}
+
+// Input transforms the error to report that it was the result of user input not
+// meeting requirements. Data is an optional parameter that should be specified
+// 0 or 1 times and will replace any data in the error if specified.
+func Input(err error, data ...any) error {
+	e := ToError(err)
+	if e == nil {
+		return nil
+	}
+	e.Subject = ErrSubjectInput
+	if len(data) > 1 {
+		panic("too much error data passed to Input()")
+	}
+	if len(data) > 0 {
+		e.Data = data[0]
+	}
+	return e
+}
+
+type singleWrapper interface {
+	error
+	Unwrap() error
+}
+
+type multiWrapper interface {
+	error
+	Unwrap() []error
+}
+
+// ToError converts a Go native error in an Error. If the error is already an
+// Error, it is returned unchanged. If the error is nil, nil is returned. If the
+// error wraps other errors, they are also converted to Errors and the full tree
+// of errors returned.
+func ToError(err error) *Error {
+	if err == nil {
+		return nil
+	}
+
+	if apiError, ok := (err).(*Error); ok {
+		return apiError
+	}
+
+	// If this error could wrap another error, make an error and add its single
+	// wrapped error as an Errs.
+	var subErrs []error
+	if singleErr, ok := (err).(singleWrapper); ok {
+		subErrs = append(subErrs, singleErr.Unwrap())
+	} else if multiErr, ok := (err).(multiWrapper); ok {
+		subErrs = multiErr.Unwrap()
+	}
+
+	apiError := &Error{
+		Type: err.Error(),
+		Errs: make([]Error, 0, len(subErrs)),
+	}
+
+	// Discard this error.Error() if its repeating the wrapped error messages,
+	// but obviously not if it is a leaf error with no sub-errors. The logic
+	// here is that calling Error() on errors that wrap other errors often just
+	// returns the concatenation of the wrapped Error() calls, which is not
+	// useful to have in the Type field.
+	for _, subErr := range subErrs {
+		subAPIErr := ToError(subErr)
+		if subAPIErr == nil {
+			continue
+		}
+
+		apiError.Errs = append(apiError.Errs, *subAPIErr)
+		apiError.Type = strings.Replace(apiError.Type, subErr.Error(), "", 1)
+	}
+	apiError.Type = strings.TrimRight(strings.TrimSpace(apiError.Type), ":")
+
+	return apiError
+}
--- a/pkg/lib/bad/error_test.go
+++ b/pkg/lib/bad/error_test.go
@@ -1,0 +1,108 @@
+//go:build unit || !integration
+
+package bad
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/samber/lo"
+	"github.com/stretchr/testify/require"
+)
+
+var errorTree = errors.Join(
+	errors.Join(
+		errors.New("1"),
+		fmt.Errorf("test: %w", errors.New("2")),
+	),
+	errors.New("3"),
+)
+
+var errorTreeOutput = strings.Join([]string{
+	"* 1",
+	"* test: ",
+	"\t* 2",
+	"* 3",
+	"",
+}, "\n")
+
+// Leaves should return only the errors in the tree with no children.
+func TestLeaves(t *testing.T) {
+	leaves := ToError(errorTree).Leaves()
+	types := lo.Map(leaves, func(e Error, _ int) string { return e.Type })
+	require.ElementsMatch(t, types, []string{"1", "2", "3"})
+
+	require.Nil(t, (*Error)(nil).Leaves())
+}
+
+type testSingleUnwrap struct{}
+
+func (testSingleUnwrap) Error() string { return "single" }
+func (testSingleUnwrap) Unwrap() error { return nil }
+
+type testMultiUnwrap struct{}
+
+func (testMultiUnwrap) Error() string   { return "multi" }
+func (testMultiUnwrap) Unwrap() []error { return []error{nil} }
+
+// ToError should return nil for nil errors, the same object for Errors, and a
+// new Error for native errors.
+func TestToError(t *testing.T) {
+	require.Nil(t, ToError(nil))
+
+	x := Error{Type: "x"}
+	require.Same(t, &x, ToError(&x))
+
+	y := errors.New("y")
+	require.Equal(t, "y", ToError(y).Type)
+
+	z := ToError(testSingleUnwrap{})
+	require.Equal(t, "single", z.Type)
+	require.Empty(t, z.Errs)
+
+	a := ToError(testMultiUnwrap{})
+	require.Equal(t, "multi", a.Type)
+	require.Empty(t, a.Errs)
+}
+
+// An error that just outputs the messages of its wrapped errors should not be
+// converted into an Error with a Type.
+func TestConcatenation(t *testing.T) {
+	err := errors.Join(
+		errors.New("test 1"),
+		errors.New("test 2"),
+	)
+	require.Equal(t, "test 1\ntest 2", err.Error())
+
+	badErr := ToError(err)
+	require.Empty(t, badErr.Type)
+	require.Len(t, badErr.Errs, 2)
+}
+
+// Output should display a tree of errors in a simple text view.
+func TestErrorOutput(t *testing.T) {
+	require.Empty(t, ToError(nil).Error())
+	require.Equal(t, errorTreeOutput, ToError(errorTree).Error())
+}
+
+// Input should tag an error as being from an input source, should return nil on
+// nil input, and should panic if more than one data item is passed.
+func TestInput(t *testing.T) {
+	require.Nil(t, Input(nil))
+
+	err := Input(errors.New("err"))
+	badErr, ok := (err).(*Error)
+	require.True(t, ok)
+	require.Equal(t, ErrSubjectInput, badErr.Subject)
+
+	err = Input(errors.New("err"), "things")
+	badErr, ok = (err).(*Error)
+	require.True(t, ok)
+	require.Equal(t, "things", badErr.Data)
+
+	require.Panics(t, func() {
+		Input(errors.New("err"), "1", "2")
+	})
+}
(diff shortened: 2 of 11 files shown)


wdbaruni (commented):
Can you please share the API output using `curl 0.0.0.01234/api/v1/orchestrator/jobs/<job_id>`? Also will there be similar improvements to `bacalhau job describe <job_id>` output?

wdbaruni (commented):
My general comment about the approach is that we are doing a lot of implicit things on behalf of the developers (us) to avoid changing the existing errors we are reporting. This can result in unexpected behaviours, still noisy messages, and difficulty to structure human readable messages upstream at the API handler or even scheduler, where downstream components have better understanding of the error.

My recommendation is to define few types of structured errors that implement a common interface and follow a hierarchy, and expect each component to intercept errors and expliclity return an instance of the structured error. 

For example, our engines should always return `ExecutionError` with a human readable message along with optional debugging details in a separate field, and maybe an indication if it is retryable or not. It should work as follows:
1. The docker engine implementation should intercept all docker daemon errors and convert them to an `ExecutionError` instance along with selecting one of our documented error codes for execution engines. 
2. The `BaseExecutor` will intercept the error thrown by the engine. If it is an instance of `ExecutionError`, then it will return it as is, and maybe add some debugging details, but without changing or adding a prefix to the error message if that doesn't add value. If it is not an `ExecutionError`, then it will return a new `ExecutionError` with an Unknown code and a general message, and can encode the actual error under the details section. 

Slowly this will force us in reducing the unknown and undocumented errors by our engines, and make sure we have good coverage of explicit structured errors returned.

So the goal here is to define the framework for error propagation and reporting, and not to have a good coverage or improve the visibility of existing errors. With that in mind, it would be better to have more breath than depth where we return few structured errors in 2 or 3 places to make sure the framework is working as expected, compared to taking docker engine and improve its error coverage.

My last question is have we looked into how kubernetes or nomad do their error reporting and propagation?

frrist:
> My general comment about the approach is that we are doing a lot of implicit things on behalf of the developers (us) to avoid changing the existing errors we are reporting. This can result in unexpected behaviours, still noisy messages, and difficulty to structure human readable messages upstream at the API handler or even scheduler, where downstream components have better understanding of the error.

+1.

My preference is for our packages to return error types similar to Walid's description of `ExecutionError` - although with more granularity regarding the type of error (not found, already running, stopped, etc.). i.e. https://github.com/bacalhau-project/bacalhau/blob/main/pkg/executor/types.go#L77
Then we provide methods to assert on the various error types similar to the go standard library: https://github.com/golang/go/blob/ebf7747dbe0204a930815200b3cbc63a09967bf9/src/os/error.go#L90.

pkg/lib/bad/error.go
```diff
@@ -0,0 +1,184 @@
...
+	// The error arose because some other service or component that the
+	// component that created the error depends on also returned an error. The
+	// error may be transient, especially if this error is a leaf, or may
+	// require operator intervention (e.g. out of disk space).
+	ErrSubjectDependency ErrorSubject = "dependency"
+)
+
+// Error represents a node in a tree of errors. Error (and thus the whole tree)
+// is JSON marshallable so that a tree of Errors can be sent across HTTP.
+type Error struct {
```
frrist, line 32: Can this type be used with the builtin methods `errors.Is` and `errors.As`? If so how?

pkg/lib/bad/error.go
```diff
@@ -0,0 +1,184 @@
...
+	for _, subErr := range err.Errs {
+		leaves = append(leaves, subErr.Leaves()...)
+	}
+	return leaves
+}
+
+// Input transforms the error to report that it was the result of user input not
+// meeting requirements. Data is an optional parameter that should be specified
+// 0 or 1 times and will replace any data in the error if specified.
+func Input(err error, data ...any) error {
```
frrist, line 115: (I think) This is variadic to allow for optional values - we can pass nothing or something. In the event that more than one value is provided, which readers of the method declaration will believe to be acceptable at first glance, the method will panic. We can't have this type of behavior in our code.

pkg/lib/bad/error_test.go
```diff
@@ -0,0 +1,108 @@
...
+	require.Equal(t, errorTreeOutput, ToError(errorTree).Error())
+}
+
+// Input should tag an error as being from an input source, should return nil on
+// nil input, and should panic if more than one data item is passed.
+func TestInput(t *testing.T) {
+	require.Nil(t, Input(nil))
+
+	err := Input(errors.New("err"))
+	badErr, ok := (err).(*Error)
```
frrist, line 96: Is this how we're supposed to use/assert the error type, will `errors.Is` work - I assume it won't?

pkg/publicapi/validator.go
```diff
@@ -21,12 +19,12 @@ type CustomValidator struct {
...
 func (cv *CustomValidator) Validate(i interface{}) error {
 	if err := cv.validator.Struct(i); err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+		return bad.Input(err)
 	}
 	if v, ok := i.(validatable); ok {
 		err := v.Validate()
 		if err != nil {
-			return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+			return bad.Input(err)
```
frrist, line 27: From a readability perspective this is a step backwards, how do I know/specify which status code is being returned for other types of errors in the API?

pkg/orchestrator/errors.go
```diff
@@ -33,16 +35,22 @@ func NewErrNotEnoughNodes(requestedNodes int, availableNodes []NodeRank) ErrNotE
...
+		if len(nodes) > 1 {
+			fmt.Fprintf(&message, "%d of %d nodes", len(nodes), len(e.AvailableNodes))
 		} else {
-			nodeErrors += fmt.Sprintf("\n\tNode %s: %s", idgen.ShortNodeID(rank.NodeInfo.ID()), rank.Reason)
+			fmt.Fprintf(&message, "Node %s", idgen.ShortNodeID(nodes[0].NodeInfo.ID()))
 		}
+		fmt.Fprintf(&message, ": %s", reason)
 	}
-	return fmt.Sprintf("not enough nodes to run job. requested: %d, available: %d. %s", e.RequestedNodes, available, nodeErrors)
+	return message.String()
```
frrist, line 53: generally a fan of this, but we shouldn't be including newlines in our errors.

pkg/lib/bad/error.go
simonwo: This is the pattern used by other packages such as echo for handling optional parameters: https://github.com/labstack/echo/blob/447c92d842e2f91c07eb032f619bea212beeee60/echo.go#L892-L899 Yes, it does require the user to actually read the comment or to test their code. 

In any case, it sounds like we maybe don't want this function and/or an option fn may be more useful.

pkg/lib/bad/error.go
simonwo: It is a valid `error` so will work fine with `Is` and `As`. 

Do you mean "will `errors.Is(typeOfWrappedError)` return true if I previously called `bad.ToError(typeOfWrappedError)"? If so, no, that would require this class to implement its own Is or As methods. It's not how I envisage this being used.

pkg/lib/bad/error_test.go
simonwo: `error.Is` is used to detect if an error "ican be treated as equivalent to an existing error". Is a `bad.Error` equivalent to any other `bad.Error` with a different (or empty) type? My take is that it is not – a `bad.Error` that wraps e.g. `io.EOF` is not equivalent to `os.PathError`. But what we do want to know is if the error we have recevied can be unmarhalled as a `bad.Error` – so I think doing a type assertion is the correct way to detect this.

pkg/publicapi/validator.go
simonwo: I would contend that it's a step forwards – what status code to return is a concern only of the API. The idea is to define metadata on the error so that the API is able to make such a distinction in a single place, rather than having to wrap every error with a status code and have that logic spread out thinly across the whole codebase.

pkg/orchestrator/errors.go
simonwo: It's currently the only way to get newlines to appear in the CLI, which enables the screenshot above. We _do_ want to use newlines there, and for ranking errors this PR does not include a way to structurally store errors in the job or execution state, which is where this error will end up. Assuming we make this change in a future PR, we can remove the newlines here (and most of this method).

pkg/lib/bad/error.go
frrist: The only behavior I take issue with is panicking.

 I would rather this method drop any extra fields, as echo does, or we could add a new metho - `InputWiffMsg` - that accepts a single field. functional options are, err, an option as well - but I worry they will be a bit verbose to write.

pkg/publicapi/client/v2/errors.go
```diff
@@ -131,13 +134,18 @@ func fromHTTPResponse(resp *http.Response) unexpectedResponseErrorSource {
...
 			// Body has been tested as safe to close more than once
-			_ = resp.Body.Close()
-			body := strings.TrimSpace(buf.String())
+			closer.CloseWithLogOnError(resp.Request.RequestURI, resp.Body)
+
+			var apiErr bad.Error
+			if err := json.NewDecoder(&buf).Decode(&apiErr); err == nil {
+				u.err = &apiErr
+			} else {
+				u.body = strings.TrimSpace(buf.String())
```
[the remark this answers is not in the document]
simonwo: If we're in this method, we have already checked that the status code is not what we expect and are preparing to return an `UnexpectedResponseError` – so we are assuming that the payload is an error payload.

But to be fair, this is a more general method and could be used to assert that we are _not_ receiving a 2xx status code, although we don't use it like that currently. But if you'd like me to check the status code again I can do that, although I think this probably works pragmatically.

pkg/lib/bad/error.go
```diff
@@ -0,0 +1,184 @@
...
+	if e == nil {
+		return nil
+	}
+	e.Subject = ErrSubjectInput
+	if len(data) > 1 {
+		panic("too much error data passed to Input()")
+	}
+	if len(data) > 0 {
+		e.Data = data[0]
+	}
```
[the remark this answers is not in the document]
simonwo: Yes, that's probably a better choice. Depending on what we decide more generally, we may not need this anyway.

pkg/lib/bad/error.go
```diff
@@ -0,0 +1,184 @@
...
+
+	indent := ""
+	if len(err.Type) > 0 {
+		out.WriteString("* ")
+		out.WriteString(err.Type)
+		indent = "\t"
+		if len(err.Errs) > 0 {
+			out.WriteString(": ")
+		}
+		out.WriteRune('\n')
```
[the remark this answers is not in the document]
simonwo: Yes, this method is really only intended as fallback that will appear in logs anyway – so maybe I can take the newlines out of this and just concatenate it to a whole string.

pkg/lib/bad/error.go
```diff
@@ -0,0 +1,184 @@
...
+	// component that created the error depends on also returned an error. The
+	// error may be transient, especially if this error is a leaf, or may
+	// require operator intervention (e.g. out of disk space).
+	ErrSubjectDependency ErrorSubject = "dependency"
+)
+
+// Error represents a node in a tree of errors. Error (and thus the whole tree)
+// is JSON marshallable so that a tree of Errors can be sent across HTTP.
+type Error struct {
+	// The type of the error. For native errors, this is just the error text,
```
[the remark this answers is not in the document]
simonwo: Ok so:

The `Type` field is intended to uniquely identify a type of error, and the type of what is in the `Data` field, so that the API Client can switch on the type field to work out how it wants to present the error – either by using the data in the `Data` field to work out/build up what it wants to display or just to replace the error message with it's own interpretation of what would be more helpful to the user. 

A good example of where this might happen is if there is some future browser-based UI that allows users to submit jobs, a validation error that comes back with a `Type` of e.g. `"Job did not pass validation"` and `Data` of 
```json
[
    {"field": "Name", "error": "contains null bytes"}, 
    {"field": "Docker.Image", "error": "image does not have a recognised platform", "validValues": ["linux/amd64", "linux/arm64", ...]}
]
``` 
In this case, the browser-based UI could recognise the `Type` and handle the error by highlighting all the form fields that are mentioned in the `Data` and can put the individual error messages next to those fields for the user to review in context.

Similarly, the CLI could do the same interpretation of the `Type` and show hints as I showed examples of in the design doc by working out where each field has taken it's value from and suggesting to the user how to modify these. So the UX is quite different for each client and the structured data allows each client to pick the best UX. 

So the requirement is that the `Type` should identify a way to **handle** this error uniquely. Different modules may want to emit the same `Type` if the way a client should handle this error is the same, or may want to emit different `Types` if the handling should be different. That's ultimately a case-by-case thing, and there's only so abstract that this can get – the granularity in `Types` needs to be driven by the actual clients people are developing and what logic people want to apply there. 

I deliberately chose against making `Type` an enum, to meet your requirement of modules being able to implement their own errors with little required visiblity at a higher level, which is sovling the existing problem with the bacerrors package requiring all error definitions to be centralised.

I am not a fan of error codes. It's just a level of indirection that I don't see as necessary – so my design is to just use a human-readable string as the type. With a code, ultimately anyone interpreting the code will need to look up its meaning in some dictionary, which is either a document we'll need to maintain or they'll have to go and read the code. At least with a human readable type string, the user can see the error for the first time coming from the API and immediately be able to understand it. The expectation is that `Type` field is a human-readable string that identifies the type of error, so doesn't contain specifics – think "job requested more resources than are available" than "job requested more memory (200G) than is available (14G)". 

Hopefully the above example also demonstrates why I didn't feel a `map[string]string` is appropriate – the level of detail that provides is not sufficient here, and [the example I show in the design doc](https://www.notion.so/expanso/Errors-b22b6b3ab35c4e12a92a4636b17d3c16?pvs=4#83072fa075644ac9b9e8a0df154648ed) is not a particularly complicated case but uses a lot of structure. Even the pre-existing errors [already require a level of detail beyond that](https://github.com/bacalhau-project/bacalhau/blob/9530b2bae9388465c679818af2ca359eeb8a9185/pkg/orchestrator/errors.go#L24-L28). I think the "future" you are talking about is already here – I make the case in the design doc as for why errors are not strings, and clients already need more detail to do any better interpretation of errors.

I picked `any` to allow modules to define their own data types that fit the specific situation, and there doesn't seem to be a more general interface. Can you say more about your concern with a more structured field? What is the risk with an arbitrary structure that makes it less maintainable? Yes, it will require us to document/define what things are going to come across the API and apply a level of rigour to not changing this interface arbitrarily... but we need to do that anyway, and even for a `map[string]string` it needs to have clear and slowly-changing keys with documentation on what format they take and what they mean, which is broadly the same information as for an arbitrary structure. 

I was assuming that `bad.Error` would at least be required to wrap native errors, because otherwise we'd be trying to send two different types as an error response (string or bad.Error), depending on which part of the code has been uplifted? In the design doc I talk about how for native errors [only outputting the final layer of the error message](https://www.notion.so/expanso/Errors-b22b6b3ab35c4e12a92a4636b17d3c16?pvs=4#b02ae231d2bc42f3a2ad148429631e1a) already beings immediate value without needing to visit every error site, so that's what I'm trying to achieve here.

simonwo:
> My general comment about the approach is that we are doing a lot of implicit things on behalf of the developers (us) to avoid changing the existing errors we are reporting. This can result in unexpected behaviours, still noisy messages, and difficulty to structure human readable messages upstream at the API handler or even scheduler, where downstream components have better understanding of the error.

Yes, the current proposal was to provide as much value _as we can_ to existing errors, rather than predicate any improvement at all on visiting every error site. In the design doc I talk about how just being able to unwrap errors to their lowest or penultimate level of detail already has a lot of value today.

The plan for this work was broadly:

1. Create a structured error type, and return that through the API and use it to store structure anywhere errors get stored as strings currently (i.e. job state, execution state).
2. Wrap existing errors and only show the final level of detail in the CLI by default, with all levels of detail show in `describe`.
3. Visit error sites and replace the existing errors with  a set of error types and data structures that can be returned from each one, and teach the client to use the information in these errors to provide a better UX

So the work to be done in 3 is to a) define the UX that would be better for the user in each case, b) get the erroring code to emit a structured error type that contains that information, c) teach the clients to interpret that structured information into their own UX that is better than a string, with an example for invalid job fields in the other comment I just wrote.

Do you think that step 2 is unnecessary? Do you think I should give up on trying to do anything with existing erorrs at all? 

> My recommendation is to define few types of structured errors that implement a common interface and follow a hierarchy, and expect each component to intercept errors and expliclity return an instance of the structured error.
> 
> For example, our engines should always return ExecutionError with a human readable message along with optional debugging details in a separate field, and maybe an indication if it is retryable or not. It should work as follows:
> 
> - The docker engine implementation should intercept all docker daemon errors and convert them to an ExecutionError instance along with selecting one of our documented error codes for execution engines.
> - The BaseExecutor will intercept the error thrown by the engine. If it is an instance of ExecutionError, then it will return it as is, and maybe add some debugging details, but without changing or adding a prefix to the error message if that doesn't add value. If it is not an ExecutionError, then it will return a new ExecutionError with an Unknown code and a general message, and can encode the actual error under the details section.

So my analysis in the design doc was that _only_ defining types to cover general classes of error is not very useful – what is needed is to:

1. interrogte the hierarchy (so we can differentiate e.g. an S3 error coming from input from an S3 error coming from a publisher)
2. allow structured data at the _lowest_ level of detail, because that is where the information useful to a user actually lives, to give clients the opportunity to provide a better UX than just a single error string (e.g. hints in the CLI, and visual elements in a browser-based UI). 

It sounds from the above like you don't want to do 2, and just want to categorise errors under specific error codes? I'm not quite sure what clients will do with this type code without also having access to specific structured fields, other than just continue to print out the lowest level of error detail? If that's all we need, I'm not sure we need any more than is already in this PR, and we can just go and visit each error site and transform the errors at the lowest level.

Where you say:
> it will return a new ExecutionError with an Unknown code and a general message, and can encode the actual error under the details section
...that sounds like what the CLI actually wants to print is the error under the details section and ignore the rest, because that's the part the user will need to actually resolve the error. Simply printing "there was an execution error for execution e-123456" and leaving it at that doesn't seem to add any clarity, and is exactly the sort of detail that I've been removing from the CLI output in this PR, to focus on the lower down errors that actually tell the user what the problem is. Although I can see it might be helpful to see that in the extended error trace, so it would still be in the error tree for debugging.

Ultimately it sounds like the design you have in mind is not the design I am proposing in the design doc – I am talking about having highly structured errors for each very specific error case, with fields that allow the client to apply it's own UX. E.g. for node ranking, each ranking strategy would return its own type of structured error, with different fields that cover each strategy's specific case, (e.g. the max resource strategy would have an struct containing what was requested and the limits applied, the min version strategy would have a struct containing the node's version and the server version, etc.) 

So to sum up what I think you are asking for, it looks a bit more like just a structure that can marshal an error tree intro JSON:
```go
type Error struct {
    // A human readable message that can be displayed to the user
    Error string
    // Zero or more errors that caused this aggregate error
    Causes []Error
}
```

And then the clients can just decide on what level of detail they want to print – in the CLI, the lowest level by default, and the whole tree as part of `describe`. Does that match up more with what you want? As I say, I'm not sure what the value of the error code is or what the client will do with it – in my design, it was to allow the client to unmarshal a data field and apply handling based on the type, but if we don't want to do that I don't see what a code does for us.

Whatever structure you choose, we will still need to store/return a marshallable tree structure in all the places we store error strings currently, which is at least:

1. From the API directly (done in this PR)
2. In the `RunCommandResult`, as execution errors end up in here
3. In the `Job.State` instead of stringly-typed `Message` field, as ranking errors end up in here
4. In the `Execution.State` instead of the stringly-typed `Message` field, as compute-side errors pre- and post-execution end up in here
... and the various other places that lead to these, such as the `Plan` and `BidResult`.

So I'll work on getting those places to store a `bad.Error` in the meantime, whilst we work out what the type contains, because at the very least we seem to agree it will be a tree and be marshallable.

PR #3760, opened 2024-04-06
Just a bunch of spelling fixes

aronchick:


frrist (approved):
Do we know why CI isn't catching these in review?

aronchick:
@frrist it is catching them - but then we're not blocking the approval when they fail. https://github.com/bacalhau-project/bacalhau/actions/runs/8570277736

Then it blocks docs deployments (but everything else works)

PR #3761, opened 2024-04-07
compiling many dependabot warnings into a single PR

aronchick:


cmd/cli/agent/alive.go
```diff
@@ -42,7 +42,7 @@ func (o *AliveOptions) runAlive(cmd *cobra.Command, _ []string) {
 		util.Fatal(cmd, fmt.Errorf("could not get server alive: %w", err), 1)
 	}
 
-	writeErr := output.OutputOneNonTabular(cmd, o.OutputOpts, response)
+	writeErr := output.OutputOneNonTabular(cmd, o.OutputOpts, response) //nolint
```
wdbaruni, line 45: what is lint complaining about here?

integration/airflow/requirements.txt
```diff
@@ -213,12 +207,12 @@ idna==3.6
     #   httpx
     #   requests
     #   yarl
-importlib-metadata==6.11.0
+importlib-metadata==7.0.0
```
wdbaruni, line 210: multiple dependencies are being upgraded to new major versions. how do we know that this won't break things? can we skip dependabot on integrations?

pkg/nats/proxy/callback_handler.go
```diff
@@ -54,13 +54,13 @@ func (h *CallbackHandler) handle(msg *nats.Msg) {
...
+		processCallback(ctx, msg, h.callback.OnBidComplete) //nolint
 	case OnRunComplete:
-		processCallback(ctx, msg, h.callback.OnRunComplete)
+		processCallback(ctx, msg, h.callback.OnRunComplete) //nolint
 	case OnCancelComplete:
-		processCallback(ctx, msg, h.callback.OnCancelComplete)
+		processCallback(ctx, msg, h.callback.OnCancelComplete) //nolint
 	case OnComputeFailure:
-		processCallback(ctx, msg, h.callback.OnComputeFailure)
+		processCallback(ctx, msg, h.callback.OnComputeFailure) //nolint
```
wdbaruni, line 63: what are skipping here? should we override lint config?

PR #3763, opened 2024-04-08
Improve ranking error output

simonwo:
This is a commit that was included in #3756 and seemed uncontroversial. It didn't actually make any use of the structured error output, and just improves the node ranking error mesages by:

- Shows the number of nodes that were available even if unsuitable, because users often misinterpret "0 available" as meaning there is a connection error between nodes in the network
- Summarises nodes when they all print the same output rather than listing all of them

As an example, failures to rank now look like:
```
% bin/darwin/arm64/bacalhau job run testdata/jobs/docker.yaml
Job successfully submitted. Job ID: j-2baef550-3023-4211-822d-eb2c88c4dfe1

	   Creating job for submission  ................  err  ❌  0.5s

Error submitting job: not enough nodes to run job. requested: 1, available: 4, suitable: 0.
• 4 of 4 nodes: job requires more memory (200 GB) than the maximum available (14 GB)

	bacalhau job describe j-2baef550-3023-4211-822d-eb2c88c4dfe1

	bacalhau job executions j-2baef550-3023-4211-822d-eb2c88c4dfe1
```

### DIFF
--- a/pkg/orchestrator/errors.go
+++ b/pkg/orchestrator/errors.go
@@ -2,8 +2,10 @@ package orchestrator
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/bacalhau-project/bacalhau/pkg/util/idgen"
+	"github.com/samber/lo"
 )
 
 // ErrSchedulerNotFound is returned when the scheduler is not found for a given evaluation type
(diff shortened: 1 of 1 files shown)


rossjones (approved):
LGTM!

PR #3769, opened 2024-04-09
Ensure compute node labels are published over LibP2P connections

simonwo:
A previous commit accidentally removed the labels used by the compute node from the NodeInfoProvider used by LibP2P. This commit restores those labels to that provider and adds a test that labels are available via the CLI in both LibP2P and NATS networking.


aronchick (approved):
LGTM - nice turn around

PR #3771, opened 2024-04-10
Structured events in place of comment strings

simonwo:
This PR implements the structure proposed in [Improve Error Reporting](https://www.notion.so/expanso/Improve-Error-Reporting-c19f5516822b47de980d76ff43ff4bbe) as a first step towards providing richer progress reporting during job execution. 

The "tl;dr;" is that we will move to using an event stream for reporting progress on jobs. The event stream will help users understand the progress of their job and give them extra context about any failures that occur. This will allow us to show a richer view in the UI, e.g. the user will be able to see "downloading Docker image" instead of just "job running". 

To achieve this vision, we need to build this infrastructure for generating events, recording them in the job history, and displaying them (done), replace the orchestrator/compute callbacks mechanism (later PR), and then give lower level components the ability to push events (later PR). 

This PR also includes some facility for structured error reporting. This allows low-level components to throw structured errors that provide a richer event than the ones generated automatically. This is used in e.g. the ErrNotEnoughNodes case and docker ImageUnavailable case so far.

This gives us the ability to output hints as part of our messages back to the user:

carbon

The output of `describe` now shows a split history between the overall job and its executions:
```
% ./bin/darwin/arm64/bacalhau job describe j-66081fef-8dd2-48de-9997-bbe23a62f0be

ID            = j-66081fef-8dd2-48de-9997-bbe23a62f0be
Name          = Docker Job
Namespace     = default
Type          = batch
State         = Completed
Count         = 1
Created Time  = 2024-04-10 06:56:26
Modified Time = 2024-04-10 06:56:29
Version       = 0

Summary
Completed = 1

Job History
 TIME       REV.  STATE      TOPIC       EVENT          DETAILS 
 0s         1     Pending    Submission  Job submitted          
 2.618376s  2     Running                                       
 2.840423s  3     Completed                                     

Executions
 ID          NODE ID     STATE      DESIRED  REV.  CREATED  MODIFIED  COMMENT      
 e-5886e01f  n-ffc3e455  Completed  Stopped  6     8s ago   5s ago    Accepted job 

Execution e-5886e01f History
 TIME       REV.  STATE              TOPIC            EVENT         DETAILS               
 0s         1     New                                                                     
 8.165ms    2     AskForBid                                                               
 2.569966s  3     AskForBidAccepted  Requesting node  Accepted job  FailsExecution: false 
                                                                    IsError: false        
                                                                    Retryable: false      
 2.590902s  4     AskForBidAccepted                                                       
 2.613668s  5     BidAccepted                                                             
 2.803923s  6     Completed                                                               

Standard Output
15
```

Resolves https://github.com/bacalhau-project/expanso-planning/issues/693.
Resolves https://github.com/bacalhau-project/expanso-planning/issues/694.

### TODO in this PR
- [x] Add more documentation
- [x] Sort execution histories by time DESC so that most relevant execution is first
- [x] Do some more examples of using structured errors from compute node components


simonwo:
@coderabbitai review

pkg/models/error.go
```diff
@@ -0,0 +1,85 @@
...
+// This method allows errors to implement extra interfaces (above) to do
+// "attribute-based error reporting". The design principle is that errors can
+// report a set of extra flags that have well defined semantics which the system
+// can then respond to with specific behavior. This allows introducing or
+// refactoring error types without higher-level components needing to be
+// modified – they simply continue to respond to the presence of attributes.
+//
+// This is instead of the sysetm having a centralized set of known error types
+// and programming in specific behavior in response to them, which is brittle
+// and requires updating all of the error responses when the types change.
```
rossjones, line 61:
```suggestion
// This is instead of the system having a centralized set of known error types
// and programming in specific behavior in response to them, which is brittle
// and requires updating all of the error responses when the types change.
```

pkg/models/error.go
```diff
@@ -0,0 +1,85 @@
...
+func EventFromError(topic EventTopic, err error) Event {
+	event := Event{
+		Message:   err.Error(),
+		Timestamp: time.Now(),
+		Topic:     topic,
+		Details:   make(map[string]string, 4), //nolint:gomnd // number of inserts below
```
rossjones, line 67: This feels kinda brittle, and I wonder whether the same purpose might be achieved with a concrete type with sensible defaults?   I'm assuming these defaults are applied/infered elsewhere, but it isn't clear to me whether the defaults would be different based on where it was used (or not).

pkg/models/job_history.go
```diff
@@ -49,5 +49,55 @@ type JobHistory struct {
...
+// unique.
+type EventTopic string
+
+// Event represents a progress report made by the system in its attempt to run a
+// job. Events are generated by the orchestrator and also passed back to the
+// orchestrator from the compute node.
+//
+// Events may be delivered in an async fashion – i.e, they may arrive much later
+// than the moment they occurred.
+type Event struct {
```
rossjones, line 88: Not sure I'd expect to find this in job_history.go as opposed to event.go - even started to add a comment that I couldn't find it.

pkg/models/error.go
rossjones: Found a use where it was adding none hint/retryable/fails-execution details, which makes sense. I think I'm just suggesting that hint/retryable/fails-execution have their own fields.

rossjones (commented):
LGTM, although I have added a couple of minor comments about impl/structure of the event struct.

pkg/models/job_history.go
simonwo: Yeah fair enough. I will move it.

pkg/models/error.go
simonwo: Yeah... in the design doc there was a comment about "it's not clear whether these things in Details should be first-class or not". @wdbaruni has a compelling reason for them not to be from watching Nomad evolve their similar structure away from having 30+ unused members.

rossjones (approved):

cmd/cli/job/describe.go
```diff
@@ -156,6 +188,33 @@ func (o *DescribeOptions) printExecutions(cmd *cobra.Command, executions []*mode
...
+		NoStyle: true,
+	}
+	jobHistoryCols := []output.TableColumn[*models.JobHistory]{
+		timeSinceCol,
+		historyRevisionCol,
+		historyStateCol,
+		historyTopicCol,
+		historyEventCol,
+		historyDetailsCol,
+	}
```
wdbaruni, line 213: to avoid having a sparse column, it might make sense to show the details under the the even message

cmd/cli/job/describe.go
```diff
@@ -156,6 +188,33 @@ func (o *DescribeOptions) printExecutions(cmd *cobra.Command, executions []*mode
...
 
+func (o *DescribeOptions) printHistory(cmd *cobra.Command, label string, history []*models.JobHistory) error {
+	if len(history) < 1 {
+		return nil
+	}
+
+	firstTime := history[0].Occurred()
+	timeSinceCol := output.TableColumn[*models.JobHistory]{
+		ColumnConfig: table.ColumnConfig{Name: historyTimeCol.ColumnConfig.Name, WidthMax: 10, WidthMaxEnforcer: text.WrapText},
+		Value:        func(h *models.JobHistory) string { return h.Occurred().Sub(firstTime).String() },
```
wdbaruni, line 199: showing event elapsed time since first even (t0) is interesting and can give quick insight on how long an even took since the job creation. Though it might be confusing with long running jobs, and when we add job updates. Will change this to elapsed since current time and will re-evaluation t0 in the future

PR #3772, opened 2024-04-10
Generate events when nodes are approved/rejected

rossjones:
When nodes are approved or rejected,  or connected/disconnected, the system may want to respond to these events by creating new evaluations, or cancelling previous tasks.

This PR adds an event emitter which allows implementations of `NodeEventHandler` to receive events that are one of:

```go 
NodeEventApproved
NodeEventRejected
NodeEventDeleted
NodeEventConnected
NodeEventDisconnected
```

all events are delivered in a goroutine, and so it is the recipients responsibility to handle the event quickly, or to create a new goroutine and return immediately.  

Currently this PR delivers events to an event listener in the `manager` package which currently logs that it has received and event.  In future, we will want to modify this behaviour to create new evaluations, or stop current executions based on the node info that is delivered with the event, the job types currently deployed,  and the type of event.  To enable this, the NodeEventListener is provided with a reference to the evaluation broker and the jobstore.


pkg/node/manager/events.go
```diff
@@ -0,0 +1,119 @@
...
+	}()
+
+	waitDuration := 1 * time.Second
+
+	for {
+		select {
+		case <-completedChan:
+			return nil
+		case <-ctx.Done():
+			return nil
```
frrist, line 114: Probably worth returning `ctx.Err()` here

pkg/node/manager/events.go
```diff
@@ -0,0 +1,119 @@
...
+		case <-e.clock.After(waitDuration):
+			return fmt.Errorf("timed out waiting for %s event callbacks to complete", event.String())
+		}
+	}
```
frrist, line 118: we can remove this for loop since `select` will block until one of its cases can run.

pkg/node/manager/events.go
```diff
@@ -0,0 +1,119 @@
...
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	completedChan := make(chan struct{})
+	wg := sync.WaitGroup{}
+
+	for _, hlr := range e.callbacks {
+		wg.Add(1)
+		go func(handler NodeEventHandler) {
+			handler.HandleNodeEvent(ctx, info, event)
```
frrist, line 95: I am thinking it's worth passing a sub-context with a timeout to these methods instead of the parent context.  This way we ensure that all the handlers receive the cancellation signal and can stop their execution, and (try to) prevent goroutine leaks on timeout. What do you think?

pkg/node/manager/events.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// Wait for the waitgroup and then close the channel to signal completion. This allows
+	// us to select on the completed channel as well as the timeout
+	go func() {
+		defer close(completedChan)
+		wg.Wait()
+	}()
+
+	waitDuration := 1 * time.Second
```
frrist, line 107: Can we make this a field on the `NodeEventEmitter`? It will be easier to change in the event we need to adjust this value later.

pkg/node/manager/interfaces.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+)
+
+type NodeEventHandler interface {
+	HandleNodeEvent(ctx context.Context, info models.NodeInfo, event NodeEvent)
+}
```
frrist, line 12: I would find the code slightly more readable if we declared this interface where it's expected/used before the `RegisterHandler` method on the `NodeEventEmitter` instead of in a separate file.

pkg/orchestrator/evaluation/inmemory_broker.go
```diff
@@ -815,3 +816,10 @@ func (b *InMemoryBroker) registerMetrics() (metric.Registration, error) {
 	}, orchestrator.EvalBrokerReady, orchestrator.EvalBrokerInflight, orchestrator.EvalBrokerPending,
 		orchestrator.EvalBrokerWaiting, orchestrator.EvalBrokerCancelable)
 }
+
+func (b *InMemoryBroker) HandleNodeEvent(ctx context.Context, info models.NodeInfo, evt manager.NodeEvent) {
```
frrist, line 820: For readability add a compile time assertion to the top of this file:
`var _ manager.NodeEventHandler = &InMemoryBroker{}`

pkg/models/node_state.go
```diff
@@ -69,7 +69,6 @@ func (t liveness) IsValid() bool {
 type livenessContainer struct {
 	CONNECTED    NodeState
 	DISCONNECTED NodeState
-	HEALTHY      NodeState
```
frrist, line 72: `CONNECTED` now implies healthy I assume? or was this just never used?

pkg/node/manager/events.go
```diff
@@ -0,0 +1,119 @@
...
+
+type NodeEventEmitterOption func(emitter *NodeEventEmitter)
+
+// WithClock is an option that can be used to set the clock for the NodeEventEmitter. This is useful
+// for testing purposes.
+func WithClock(clock clock.Clock) NodeEventEmitterOption {
+	return func(emitter *NodeEventEmitter) {
+		emitter.clock = clock
+	}
+}
```
frrist, line 49: Not sure this is used anywhere, was it meant to be used in a test?

pkg/node/manager/events.go
rossjones: Done in ff7828f3f

pkg/node/manager/events.go
rossjones: Done in ff7828f3f, not sure why it was there in the first place, probably force of habit.

pkg/node/manager/events.go
rossjones: Done in ff7828f3f

pkg/node/manager/interfaces.go
rossjones: Reasonble, done in ff7828f3f.  I think I put it in an interfaces.go as there's been a bit of a pattern with the eval stuff that there's an interfaces.go and a types.go.

pkg/orchestrator/evaluation/inmemory_broker.go
rossjones: Done in ff7828f3f

pkg/node/manager/events.go
rossjones: It was, and is with b9a69535f

pkg/node/manager/events.go
rossjones: Good idea, will give it a try.

frrist: > The problem might be that the timeout will not be using the mock clock, and so is only really useful for cleaning up unfinished callbacks rather than being useful to replace the timeout on each callback

oo yeah, good point. Change looks good here - thanks.

pkg/node/manager/events_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (s *EventEmitterSuite) TestEmitWithSlowCallback() {
+	e := manager.NewNodeEventEmitter(manager.WithClock(s.clock))
+	s.NotNil(e)
+
+	e.RegisterHandler(testSleepyHandler{s.clock})
+
+	go func() {
+		s.clock.Add(10 * time.Second)
+	}()
```
frrist, line 89: why does this need a go routine? I'd assume adding 10 seconds to the clock is non-blocking?

frrist (approved):

pkg/node/manager/events_test.go
rossjones: Some weirdness with the mock clock where it won't work when it's inline.  I've been trying to work out if I'm just holding it wrong but I suspect it's something to do with the work in EmitEvent happening in a goroutine and requiring a yield (but I'm not 100% sure)

frrist:
@rossjones I have implemented a fix for the NodeInfo overwrite bug here: https://github.com/bacalhau-project/bacalhau/pull/3785
It was a very large mechanical change to make and sadly touched a lot of files. If it looks good to you, could we land your change  (this PR) after it to avoid conflicts? I suspect it will be easier to resolved conflicts from your change here than the other way around. However if you spot any significant issues with https://github.com/bacalhau-project/bacalhau/pull/3785 go ahead and merge this and I'll sort the conflicts on my end.

wdbaruni:
@frrist are you owning this PR now? what do we need to do to unblock it and merge it?

Issue #3773, opened 2024-04-10
bacalhau authenication via token CLI borken


frrist:
I have deployed a cluster, but not authenticated to it yet via the CLI. e.g.
```
cat ~/.bacalhau/tokens.json 
{}
```

Attempting to run a docker job fails due to invalid token and no prompt is given (as was previous behavior) to enter a token:
```
bacalhau docker run ubuntu:latest echo hello world
token: Error: Invalid token. 
Invalid token. 
```

Using the V2 API does prompt for a token, but it does so twice:
```
bacalhau job run testdata/jobs/docker-hello.yaml
token: token: 
```
I am unable to enter a token successfully here

frrist:
On the surface this appears to be related to the update service, e.g.
This fails:
```bash
frrist@cypress ~/.bacalhau> pwd
/home/frrist/.bacalhau
frrist@cypress ~/.bacalhau> ls
cert.crt  compute_store  config.yaml  executor_storages  libp2p_private_key  orchestrator_store  plugins  repo.version  tokens.json  user_id.pem
frrist@cypress ~/.bacalhau> bacalhau node list
token: token: 

This works because there is an update.json file in the repo node:
```bash
frrist@cypress ~/.bacalhau> echo '{"LastCheck":"2024-04-10T15:37:38.341767311-07:00"}' > update.json
frrist@cypress ~/.bacalhau> bacalhau node list
token: 
```

PR #3775, opened 2024-04-10
fix: update checker doesn't require server connection to work

frrist:
- and liberally decorate the code with TODOs so we can make this better in the future where someone has some spare cycles
- fixes https://github.com/bacalhau-project/bacalhau/issues/3773

### DIFF
--- a/cmd/util/hook/version.go
+++ b/cmd/util/hook/version.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/bacalhau-project/bacalhau/cmd/util"
+	"github.com/spf13/cobra"
+
 	"github.com/bacalhau-project/bacalhau/pkg/models"
 	"github.com/bacalhau-project/bacalhau/pkg/version"
-	"github.com/spf13/cobra"
 )
 
 var printMessage *string = nil
@@ -19,13 +19,7 @@ func StartUpdateCheck(cmd *cobra.Command, args []string) {
 	version.RunUpdateChecker(
 		cmd.Context(),
 		func(ctx context.Context) (*models.BuildVersionInfo, error) {
-			if response, err := util.GetAPIClientV2(cmd).Agent().Version(ctx); err != nil {
-				return nil, err
-			} else if response != nil {
-				return response.BuildVersionInfo, nil
-			} else {
-				return nil, nil
-			}
+			return version.Get(), nil
 		},
 		func(_ context.Context, ucr *version.UpdateCheckResponse) { printMessage = &ucr.Message },
 	)
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -484,6 +484,7 @@ func NewNode(
 	updateCheckCtx, stopUpdateChecks := context.WithCancel(ctx)
 	version.RunUpdateChecker(
 		updateCheckCtx,
+		// TODO(forrest) [correctness]: this code is literally the server, why are we returning nil???!!!
 		func(ctx context.Context) (*models.BuildVersionInfo, error) { return nil, nil },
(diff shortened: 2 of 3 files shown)


cmd/util/hook/version.go
```diff
@@ -19,13 +19,7 @@ func StartUpdateCheck(cmd *cobra.Command, args []string) {
...
 		cmd.Context(),
 		func(ctx context.Context) (*models.BuildVersionInfo, error) {
-			if response, err := util.GetAPIClientV2(cmd).Agent().Version(ctx); err != nil {
-				return nil, err
-			} else if response != nil {
-				return response.BuildVersionInfo, nil
-			} else {
-				return nil, nil
-			}
+			return version.Get(), nil
```
frrist, line 22: This is the fix for: https://github.com/bacalhau-project/bacalhau/issues/3773:

This is because the update check runs async on every command and incorrectly required a client, which prompted for a token to ask the server what version is was**. The actual command being run (e.g. `job run`) also created a client and asked for a token which is why the CLI asked for a token twice until and `update.json` file existed. The fix here is to simply return the version of the binary running the update checker.

** no idea why we do this, update checker should tell you the binary you are using is out of date, not the server you are talking to. Server operators will become aware their server is out of date via log message the update server prints when its update checker runs

aronchick (approved):
LGTM

frrist:
@coderabbitai /review

cmd/util/hook/version.go
wdbaruni: maybe we need to understand the purpose of the update checker better. Is it to tell that the client is outdated compared to the orchestrator's version? or outdated compared to latest releases from bacalhau?

As a user, I _might_ expect the following:
1. If I am running a compute node, I would like the cli to compare my version with the orchestrator's version. I don't want the cli to ask me to update to a version that is beyond the orchestrator's
2. If I am running client commands, such as `bacalhau job run`, I would like the cli to also compare my version with the orchestrator's
3. If I a m running an orchestrator node, then I would like the cli to compare my version with the latest release from bacalhau

@aronchick @MichaelHoepler what do you think the right behaviour here?

frrist: I'd propose we reduce the responsibility of the UpdateChecker to only retrieve the latest version of bacalhau from the update server. This will remove the dependency on a bacalhau client - and thus authentication - from the UpdateChecker service - which will fix this bug.

The UpdateChecker may then persist the value it retrieves from the update server to some "system configuration field" (will be very easy to do when [#3959](https://github.com/bacalhau-project/bacalhau/pull/3959/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R71) lands). From there we can make the following changes:
1. Clients and Compute nodes can compare their version, the orchestrator's version, and the latest version notifying accordingly in stderr or log messages.
2. Orchestrator nodes can compare their version with the latest version notifying accordingly in log messages.

----------------------------------------------------------

Alternatively, and this would be my preference: 
1. Only Servers (Compute and Requester Nodes) use the update checker (as we already do now) to retrieve the latest version. If they are behind it they produce a log message stating so.
2. Clients are only notified when the server they are communicating with is on a version different than theirs. Clients commands do not run the UpdateChecked.

In this flow, clients are only notified they need to update when their server operator updates.

wdbaruni: Yeah those are valid points. I've approved the PR as figuring out the right behaviour of the update checker is a bigger task, and you are not changing the behaviour with this PR
https://github.com/bacalhau-project/bacalhau/issues/3980

wdbaruni (approved):

PR #3777, opened 2024-04-11
Make it easier to get in progress jobs of a certain type

rossjones:
Currently the inprogress index is just the ID of the job that is inprogress.  This PR switches the inprogress index to a composite key containing both the ID and the Type of the job.  This allows for filtering of active jobs of a certain type without having to serde each entire job referenced in the index.

We may at a future date want to introduce CompositeIndex type alongside the current Index type.


pkg/jobstore/boltdb/store_test.go
```diff
@@ -45,6 +45,7 @@ func (s *BoltJobstoreTestSuite) SetupTest() {
 
 	jobFixtures := []struct {
 		id              string
+		typ             string
```
aronchick, line 48: Stupid question - why not e the field "jobType" instead of "typ". Is this something idiomatic about go?

aronchick (approved):
Lgtm - just naming comment

pkg/jobstore/boltdb/store_test.go
rossjones: type is a [keyword](https://go.dev/ref/spec#Keywords), but I guess jobType should be okay.

PR #3778, opened 2024-04-11
Deprecates some IPFS flags

rossjones:
In order to smooth the migration from an embedded IPFS node to a self-hosted IPFS node, we will deprecate the flags used for the internal node, rather than removing the functionality in one release.

This will allow for the functionality to be removed over two or more minor version releases.


aronchick (approved):
Lgtm

PR #3779, opened 2024-04-11
Fixes a panic in the wasm executor when entrypoint is not known

### LINKED ISSUE #3753: Failure to run a WASM job does not produce a job error
If you submit a WASM job that fails to load, e.g. because the entrypoint specified in the Task does not exist as a function in the WASM module, the WASM does not get run but it still reported as successfully completed.

It is not even possible to see that there was a problem:
```
% gsed -i 's/_start/not_start/g' testdata/jobs/wasm.yaml
% bacalhau job run testdata/jobs/wasm.yaml 
Job successfully submitted. Job ID: j-6decc4ab-74a0-4415-9200-b31f31374def

	   Creating job for submission  ................  done ✅  0.5s

	bacalhau job describe j-6decc4ab-74a0-4415-9200-b31f31374def

	bacalhau job executions j-6decc4ab-74a0-4415-9200-b31f31374def

%  bacalhau job describe j-6decc4ab-74a0-4415-9200-b31f31374def

ID            = j-6decc4ab-74a0-4415-9200-b31f31374def
Name          = Wasm Job
Namespace     = default
Type          = batch
State         = Completed
Count         = 1
Created Time  = 2024-04-03 07:55:44
Modified Time = 2024-04-03 07:55:45
Version       = 0

Summary
Completed = 1

Executions
 ID          NODE ID     STATE      DESIRED  REV.  CREATED  MODIFIED  COMMENT 
 e-e824fd26  n-ba653f9f  Completed  Stopped  6     26s ago  26s ago           

% bacalhau job logs j-6decc4ab-74a0-4415-9200-b31f31374def    


% bacalhau job executions j-6decc4ab-74a0-4415-9200-b31f31374def

 CREATED   MODIFIED  ID          NODE ID     REV.  STATE      DESIRED 
 07:55:44  07:55:45  e-e824fd26  n-ba653f9f  6     Completed  Stopped 
```

rossjones:
Currently we try to reference the `ExportedFunction` which is documented as returning nil when the name is not found. Instead it returns a non-nil object with an inner nil field which will cause a panic when `Call`-ed.

This commit instead looks up the function before referencing it to ensure it exists, returning an error early if not. This allows us to remove the call to recover which was handling the panic() for us.

Fixes #3753

### DIFF
--- a/pkg/executor/wasm/handler.go
+++ b/pkg/executor/wasm/handler.go
@@ -64,6 +64,19 @@ type executionHandler struct {
 func (h *executionHandler) run(ctx context.Context) {
 	ActiveExecutions.Inc(ctx)
 	defer func() {
+		if r := recover(); r != nil {
+			h.logger.Error().
+				Str("recover", fmt.Sprintf("%v", r)).
+				Msg("execution recovered from panic")
+
+			// The recover was originally here for a bug we think is now fixed, but given
+			// the propensity for panics in this area, we're being extra-cautious and
+			// ensuring we can handle any future panics that arise.
+			h.result = executor.NewFailedResult(
+				fmt.Sprintf("WASM executor failed with an internal error: %v", r),
+			)
+		}
+
 		ActiveExecutions.Dec(ctx)
 	}()
 
--- a/pkg/executor/wasm/handler.go
+++ b/pkg/executor/wasm/handler.go
@@ -64,13 +64,6 @@ type executionHandler struct {
 func (h *executionHandler) run(ctx context.Context) {
 	ActiveExecutions.Inc(ctx)
 	defer func() {
-		if r := recover(); r != nil {
-			h.logger.Error().
-				Str("recover", fmt.Sprintf("%v", r)).
-				Msg("execution recovered from panic")
-			// TODO don't do this.
-			h.result = &models.RunCommandResult{}
-		}
 		ActiveExecutions.Dec(ctx)
 	}()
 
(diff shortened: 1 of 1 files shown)


pkg/executor/wasm/handler.go
```diff
@@ -64,13 +64,6 @@ type executionHandler struct {
 func (h *executionHandler) run(ctx context.Context) {
 	ActiveExecutions.Inc(ctx)
 	defer func() {
-		if r := recover(); r != nil {
-			h.logger.Error().
-				Str("recover", fmt.Sprintf("%v", r)).
-				Msg("execution recovered from panic")
-			// TODO don't do this.
-			h.result = &models.RunCommandResult{}
-		}
```
frrist, line 73: If we are 100% sure the code called below here won't panic then I am for removing this defer statement, otherwise I'd find it safer to leave it here.

frrist (approved):

pkg/executor/wasm/handler.go
rossjones: If we leave it, then I'd suggest we change it to returning an error instead as at the moment it makes it look like the execution succeeded.

frrist: That sounds good to me. Basically do what was done here:
```
	h.result = executor.NewFailedResult(
			fmt.Sprintf("unable to find the entrypoint '%s' in the WASM module", h.arguments.EntryPoint),
		)
		return
```
but change the error message to state a panic occurred?

simonwo (approved):

pkg/executor/wasm/handler.go
rossjones: Done in 94b9c8e2dab677fd61d21b86fc8328ed6a488d20

PR #3782, opened 2024-04-11
Stop approval states being overwritten

rossjones:
When computes nodes send their info updates, they were sending a version of the NodeInfo struct that contains an empty approval value.  We shouldn't accept them at all, and so this commit will overwrite whatever the client sends with the current approval state.

We should probably change what the compute node sends so that the NodeInfo on the requester node is an enhanced version of the info the compute node has.

Possible fix for: https://github.com/bacalhau-project/bacalhau/issues/3783


rossjones:
This is unnecessary given the existence of #3785 which is a better long-term fix.

PR #3785, opened 2024-04-11
fix: remove Liveness and Acceptance from NodeInfo

### LINKED ISSUE #3783: Compute Nodes broadcast `NodeInfo` with `unknown` approval status which overrides previous approvals/rejections
Auto-approved/pending nodes receive unknown status since compute nodes broadcast empty approval state to requester node. There are several solutions to address this:
- shoe-horn in a fix to the requester that ignores the status from compute nodes
- remove the approval state and connected state from the `NodeInfo` structure as details are unknown to compute node. It's only relevant in the context of a requester node.

frrist:
- fixes #3783
- Introduces NodeState type used to track NodeInfo, Liveness, and Acceptance. Removes the idea of Livenss and Acceptance from data sent by compute nodes to the Requester(s) since compute nodes should not influence their Liveness or Acceptance. Those are values related to the Requesters view of the network.


frrist:
@coderabbitai /review

cmd/cli/job/describe.go
```diff
@@ -102,7 +102,7 @@ func (o *DescribeOptions) printHeaderData(cmd *cobra.Command, job *models.Job) {
 		{Left: "Name", Right: job.Name},
 		{Left: "Namespace", Right: job.Namespace},
 		{Left: "Type", Right: job.Type},
-		{Left: "State", Right: job.State.StateType},
+		{Left: "Liveness", Right: job.State.StateType},
```
rossjones, line 105: This is the state of the job, rather than the node, is Liveness the correct word here?

cmd/cli/job/executions.go
```diff
@@ -97,7 +97,7 @@ var (
 		Value:        func(e *models.Execution) string { return strconv.FormatUint(e.Revision, 10) },
 	}
 	executionColumnState = output.TableColumn[*models.Execution]{
-		ColumnConfig: table.ColumnConfig{Name: "State", WidthMax: 10, WidthMaxEnforcer: text.WrapText},
+		ColumnConfig: table.ColumnConfig{Name: "Liveness", WidthMax: 10, WidthMaxEnforcer: text.WrapText},
```
rossjones, line 100: Same question, is Liveness the word we want to use for the execution?

cmd/cli/job/history.go
```diff
@@ -108,7 +108,7 @@ var historyColumns = []output.TableColumn[*models.JobHistory]{
 		},
 	},
 	{
-		ColumnConfig: table.ColumnConfig{Name: "New State", WidthMax: 20, WidthMaxEnforcer: text.WrapText},
+		ColumnConfig: table.ColumnConfig{Name: "New Liveness", WidthMax: 20, WidthMaxEnforcer: text.WrapText},
```
rossjones, line 111: New Liveness sounds a bit weird.

pkg/compute/management_client.go
```diff
@@ -101,7 +102,7 @@ func (m *ManagementClient) RegisterNode(ctx context.Context) error {
 func (m *ManagementClient) deliverInfo(ctx context.Context) {
 	// We _could_ avoid attempting an update if we are not registered, but
 	// by doing so we will get frequent errors that the node is not
-	// registered.
+	// registered.c
```
rossjones, line 105:
```suggestion
	// registered.
```

pkg/node/manager/node_manager.go
```diff
@@ -86,10 +87,13 @@ func (n *NodeManager) Register(ctx context.Context, request requests.RegisterReq
...
 	}
 
-	request.Info.Approval = n.defaultApprovalState
-
-	if err := n.nodeInfo.Add(ctx, request.Info); err != nil {
-		return nil, errors.Wrap(err, "failed to save nodeinfo during node registration")
+	if err := n.store.Add(ctx, models.NodeState{
+		Info:     request.Info,
+		Approval: n.defaultApprovalState,
+		// NB(forrest): by virtue of a compute node calling this endpoint we can consider it connected
```
rossjones, line 93: This _should_ be okay, but using other signals than the heartbeat might lead to flapping, so if that happens, this might be a reasonable place to look.

pkg/node/manager/node_manager.go
```diff
@@ -121,8 +125,14 @@ func (n *NodeManager) UpdateInfo(ctx context.Context, request requests.UpdateInf
 	}
 
 	// TODO: Add a Put endpoint that takes the revision into account?
-	if err := n.nodeInfo.Add(ctx, request.Info); err != nil {
-		return nil, errors.Wrap(err, "failed to save nodeinfo during node registration")
+	if err := n.store.Add(ctx, models.NodeState{
+		Info: request.Info,
+		// the nodes approval state is assumed to be approved here, but re-use existing state
+		Approval: existing.Approval,
+		// TODO can we assume the node is connected here?
```
rossjones, line 132: Probably. It might be best to inject a heartbeat as the liveness is held in memory rather than the store and we probably want to avoid situations where the node is not in the liveness map and we accidentally send Connected rather than the default of Disconnected.

pkg/orchestrator/selection/discovery/chained.go
rossjones, line 1: These aren't still needed for libp2p are they?

rossjones (approved):
Looks great. Definitely an improvement to break into different types rather than jamming more fields into the existing one.

Couple of comments about the word liveness, as it feels a bit weird to use it in CLI output, but it might just be me.

Slight concern about the deletions and whether they affect libp2p usage, as we no longer test with the libp2p transport, it might be worth testing manually (if you haven't already) to ensure it continues to work.

cmd/cli/job/describe.go
frrist: whoops, my editor was a little overzealous with the name change, will revert this.

cmd/cli/job/executions.go
frrist: continuing to blame the enthusiasm of my editor

cmd/cli/job/history.go
frrist: will revert

frrist:
Based on conversation with @wdbaruni and @seanmtracey we need to evaluate if this change breaks backwards compatibility and if it does handle it gracefully by adjusting marshaling or adjusting the topic we exchange these messages on to include a version

frrist:
This change has been validated with a v1.3.0 Requester communicating with a v1.3.1-rc Compute node and v1.3.0 Compute node as well as a v1.3.1-rc Requester node communicating with a v1.3.1-rc Compute node and v1.3.0 Compute node. Meaning it’s backwards and forwards compatible.

pkg/models/node_approval.go
```diff
@@ -4,93 +4,95 @@ import (
 	"fmt"
 )
 
-type NodeApproval struct {
-	approval
+// TODO if we ever pass a pointer to this type and use `==` comparison on it we're gonna have a bad time
+// implement an `Equal()` method for this type and default to it.
```
wdbaruni, line 8: this sounds bad. can you open an issue to track this TODO?

pkg/models/node_connection.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (t connection) IsValid() bool {
+	return t >= connection(1) && t <= connection(len(strConnectionArray))
+}
+
+type livenessContainer struct {
+	CONNECTED    NodeConnectionState
+	DISCONNECTED NodeConnectionState
+	HEALTHY      NodeConnectionState
```
wdbaruni, line 74: What is HEALTHY?

pkg/nats/transport/nats.go
```diff
@@ -85,7 +86,7 @@ type NATSTransport struct {
 	natsClient        *nats_helper.ClientManager
 	computeProxy      compute.Endpoint
 	callbackProxy     compute.Callback
-	nodeInfoPubSub    pubsub.PubSub[models.NodeInfo]
+	nodeInfoPubSub    pubsub.PubSub[models.NodeState]
```
wdbaruni, line 89: shouldn't PubSub be about NodeInfo and not NodeState?

pkg/node/manager/node_manager.go
```diff
@@ -24,24 +25,24 @@ const (
...
-	nodeInfo             routing.NodeInfoStore
+	store                routing.NodeInfoStore
 	resourceMap          *concurrency.StripedMap[models.Resources]
 	heartbeats           *heartbeat.HeartbeatServer
-	defaultApprovalState models.NodeApproval
+	defaultApprovalState models.NodeMembershipState
 }
 
 type NodeManagerParams struct {
 	NodeInfo             routing.NodeInfoStore
```
wdbaruni, line 35: should be `Store`

pkg/orchestrator/selection/discovery/chained.go
```diff
@@ -1,63 +0,0 @@
...
-	"github.com/bacalhau-project/bacalhau/pkg/orchestrator"
-	pkgerrors "github.com/pkg/errors"
-	"github.com/rs/zerolog/log"
-	"golang.org/x/exp/maps"
-)
-
-type Chain struct {
-	discoverers  []orchestrator.NodeDiscoverer
-	ignoreErrors bool
-}
```
wdbaruni, line 17: I remember you saying at one point that you are removing all chain implementations and replacing them with simple arrays. While the Chain looks like a simple array, it implements `NodeDiscoverer` interface which allows different components to just accept the interface rather than an array of interfaces. This simplifies testing, error handling (e.g. `ignoreErrors`), and can allow chain implementations with `AND` or `OR` logic.

wdbaruni (approved):
Few minor comments

wdbaruni:
Ahh .. I didn't see auto-merge was enabled. Anyways, there are few comments that can be addressed if a follow-up PR if the comments are valid

PR #3799, opened 2024-04-18
Revert Self Signed Certificate Generation

frrist:


pkg/config/types/generated_constants.go
```diff
@@ -30,7 +29,6 @@ const NodeServerAPITLSAutoCert = "Node.ServerAPI.TLS.AutoCert"
 const NodeServerAPITLSAutoCertCachePath = "Node.ServerAPI.TLS.AutoCertCachePath"
```
wdbaruni, line 29: Are the AutoCert configs still valid?

wdbaruni (approved):
Minor comment

pkg/config/types/generated_constants.go
frrist: Yeah, it's still a config value: https://github.com/bacalhau-project/bacalhau/pull/3799/files#diff-d9e3882850ed4375c2a194c445781f20b764080f60e62cf48721ce4b88aac55dL90

PR #3957, opened 2024-04-25
fix: require connected and approved nodes for scheduling

### LINKED ISSUE #3784: The requester node attempts to schedule work on disconnected nodes resulting in the job never running
Due to changes here: 
- https://github.com/bacalhau-project/bacalhau/pull/3768/files#diff-16bad1556b6ffdd42e5c57a235f78726a613fc46759a999fb23703ce93f92ecbR161-R163
- https://github.com/bacalhau-project/bacalhau/pull/3768/files#diff-3034b567889a6a7f1ea37dcf47d446e1fbf1a2928894a4fb7d9d2cfbafc857ecR93
- https://github.com/bacalhau-project/bacalhau/pull/3768/files#diff-6552869d10da40039d76eca490cf89bb20a2e4eb860c892ac1569b7a0eb88bebR110

Current proposal is to:
- make auto-approve work via a fix for: https://github.com/bacalhau-project/bacalhau/issues/3783
- modify the aforementioned code to only schedule on connected nodes.

cc @rossjones & @wdbaruni to weigh in on how the new event system introduced in https://github.com/bacalhau-project/bacalhau/pull/3772 can be used to force scheduling of executions when offline compute nodes come online again.

frrist:
- This change modifies the Requester nodes scheduling constraints s.t. jobs will only be scheduled on nodes that are online and approved. Disconnected nodes and nodes that are rejected or pending will not be eligible to run jobs.
- Additionally, this change cleans up some code by making constraints an parameter to the node selector - which simplify various parts of dependency construction.
- Lastly, this change removes some *Param types to avoid the possibility of NPD.
- fixes #3784

## Summary by CodeRabbit

- **Refactor**
  - Improved the initialization and configuration process for job schedulers and node selectors for better performance and maintainability.
  - Streamlined node selection logic by removing redundant parameters and simplifying method signatures.
  - Renamed several methods and fields to enhance clarity and consistency across the application.

- **New Features**
  - Enhanced node selection capabilities with additional constraints to better tailor node selection to specific job requirements.

### DIFF
--- a/pkg/node/requester.go
+++ b/pkg/node/requester.go
@@ -145,41 +145,24 @@ func NewRequesterNode(
 		retryStrategy = retryStrategyChain
 	}
 
-	// TODO(forrest): [refactor] the selector constraints ought to be a parameter to the node selector.
 	// node selector
-	nodeSelector := selector.NewNodeSelector(selector.NodeSelectorParams{
-		NodeDiscoverer: nodeInfoStore,
-		NodeRanker:     nodeRankerChain,
-	})
-	// selector constraints: require nodes be online and approved to schedule
-	selectorConstraints := orchestrator.NodeSelectionConstraints{
-		RequireConnected: true,
-		RequireApproval:  true,
-	}
+	nodeSelector := selector.NewNodeSelector(
+		nodeInfoStore,
+		nodeRankerChain,
+		// selector constraints: require nodes be online and approved to schedule
+		orchestrator.NodeSelectionConstraints{
+			RequireConnected: true,
+			RequireApproval:  true,
+		},
+	)
 
 	// scheduler provider
-	batchServiceJobScheduler := scheduler.NewBatchServiceJobScheduler(
-		jobStore,
-		planners,
-		nodeSelector,
-		retryStrategy,
-		selectorConstraints,
-	)
+	batchServiceJobScheduler := scheduler.NewBatchServiceJobScheduler(jobStore, planners, nodeSelector, retryStrategy)
 	schedulerProvider := orchestrator.NewMappedSchedulerProvider(map[string]orchestrator.Scheduler{
 		models.JobTypeBatch:   batchServiceJobScheduler,
 		models.JobTypeService: batchServiceJobScheduler,
-		models.JobTypeOps: scheduler.NewOpsJobScheduler(
-			jobStore,
-			planners,
-			nodeSelector,
-			selectorConstraints,
-		),
-		models.JobTypeDaemon: scheduler.NewDaemonJobScheduler(
-			jobStore,
-			planners,
-			nodeSelector,
-			selectorConstraints,
-		),
+		models.JobTypeOps:     scheduler.NewOpsJobScheduler(jobStore, planners, nodeSelector),
+		models.JobTypeDaemon:  scheduler.NewDaemonJobScheduler(jobStore, planners, nodeSelector),
 	})
 
 	workers := make([]*orchestrator.Worker, 0, requesterConfig.WorkerCount)
--- a/pkg/node/requester.go
+++ b/pkg/node/requester.go
@@ -99,12 +99,6 @@ func NewRequesterNode(
 		}),
 	)
 
-	// node selector
-	nodeSelector := selector.NewNodeSelector(selector.NodeSelectorParams{
-		NodeDiscoverer: nodeInfoStore,
-		NodeRanker:     nodeRankerChain,
-	})
-
 	// evaluation broker
 	evalBroker, err := evaluation.NewInMemoryBroker(evaluation.InMemoryBrokerParams{
 		VisibilityTimeout:    requesterConfig.EvalBrokerVisibilityTimeout,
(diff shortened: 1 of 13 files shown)


pkg/test/requester/node_selection_test.go
```diff
@@ -63,13 +63,10 @@ func (s *NodeSelectionSuite) SetupSuite() {
 			},
 		},
 	}
-	requesterConfig, err := node.NewRequesterConfigWith(
-		node.RequesterConfigParams{
-			NodeRankRandomnessRange: 0,
-			OverAskForBidsFactor:    1,
-		},
-	)
```
frrist, line 71: No ides how this ever worked before `NodeRankRandomnessRange` cannot be `0` or the code that uses it panics, somehow it never panicked, meaning the setting of these values were likely ineffectual

pkg/test/requester/retries_test.go
```diff
@@ -132,13 +132,10 @@ func (s *RetriesSuite) SetupSuite() {
 	}
 	ctx := context.Background()
 
-	requesterConfig, err := node.NewRequesterConfigWith(
-		node.RequesterConfigParams{
-			NodeRankRandomnessRange: 0,
-			OverAskForBidsFactor:    1,
-		},
-	)
```
frrist, line 140: No ides how this ever worked before `NodeRankRandomnessRange` cannot be `0` or the code that uses it panics, somehow it never panicked, meaning the setting of these values were likely ineffectual

pkg/node/requester.go
```diff
@@ -151,26 +145,24 @@ func NewRequesterNode(
...
+	)
+
 	// scheduler provider
-	batchServiceJobScheduler := scheduler.NewBatchServiceJobScheduler(scheduler.BatchServiceJobSchedulerParams{
-		JobStore:      jobStore,
-		Planner:       planners,
-		NodeSelector:  nodeSelector,
-		RetryStrategy: retryStrategy,
-	})
+	batchServiceJobScheduler := scheduler.NewBatchServiceJobScheduler(jobStore, planners, nodeSelector, retryStrategy)
```
wdbaruni, line 160: Keeping params in the constructor helps maintain code stability and makes the overall code more maintainable. This ensures that constructor signatures and tests do not require changes when new dependencies are introduced. This also makes the code more readable even if we fail to provide clear field names

wdbaruni (changes requested):
To make reviewing the PRs easier, avoid coupling refactoring with actual changes the PR is mainly about, such as fixing node selection in this case

pkg/node/requester.go
frrist: I generally get where you are coming from - maintainable code is great! But please hear me out here:

> Keeping params in the constructor helps maintain code stability and makes the overall code more maintainable.

I'd argue it does the opposite: the params pattern doesn't provide compile-time safety. Additionally, whenever a new dependency is introduced, we need to modify the the constructor, the call sites, _and_ the params type. 

Comparing this with declaring dependencies in the function signature: It provides compile-time safety, and reduces the number of modifications we need to make when introducing new dependencies. We only need to update the constructor and the call sites. (And the compiler will complain very loudly in places we don't, instead of panicking later when the application runs - this is what soured my additional refactor here.)

> This ensures that constructor signatures and tests do not require changes when new dependencies are introduced

I don't follow. Constructors and tests still need to change when new dependencies are introduce - all fields are _required_ and therefore must be provided.

> This also makes the code more readable even if we fail to provide clear field names

I'll acknowledge we have a different in opinions here. Personally I want to know the type of something rather than its name in a structure.

-------------------------------


Given the points regarding of compile time safety and reduced maintenance overhead, would you be alright moving forward with this code as is?

wdbaruni: Having methods or constructors with long parameters is by standard a bad code design in the industry. First time I hear a discussion favouring long parameters :) 

I hear your point about wanting to fail early at compile-time. It is all about tradeoffs. An example, you have picked fx as our dependency injection framework, and that will move a lot of compile-time failures to runtime. You have prioritized code readability, maintainability and testability in favour of compile-time checks, and that is okay as the gains are worth the cost.

Cons of long params include:
1. Increased error potential as more params increase the likelihood of passing arguments in the wrong order
2. Poor readability as methods with many params are harder to read and understand at a glance, without having to open the method implementation to understand what param does. The IDEs might help, but not always
3. Optional fields still need to be passed with their zero values
4. Harder to maintain as adding or modifying parameters requires changes in all your method calls. This is where you have a different opinion that you prefer the compiler to fail and force the developer to update all calls. Yes, initialization oversight is a true problem with using a single param object. Though in _practice_, it still makes the code more maintainable for the follow reasons:
    1. Compile time are not sufficient to validate the correct initialization of types. They are great to fail fast, but we   shouldn't assume if the compiler is not complaining that we don't need to do runtime validation or that we won't get NPE. This is [an example](https://github.com/bacalhau-project/bacalhau/blob/cad894a69cf196f6b2b62a93a8dd50f608ecd712/pkg/nats/transport/nats.go#L54-L55) where param object can help with runtime validation. I am still saying that failing at compile-time is great, but it is not the only criteria we should evaluate tradeoffs at. Runtime validation during node startup can be as good.
    2. Usually constructors/methods are called once or very few times in production code, but multiple times in tests to test different branches and combinations. In practice, and with good tests, I see the parameter object is initialized once with default testing values, and then different tests override certain values that they want to test. This makes tests more maintainable and readable compared to having to modify tens of method signatures and that is before even adding any new tests
    3. A lot of fields and configurations are exposed in the constructor mainly for testing purposes with fallback default values if not configured. e.g. allow to inject a clock to have more predictable tests. Having a param object enables this without complicating the type initialization

A bonus is fx seem to work well with param objects out of the box https://uber-go.github.io/fx/parameter-objects.html#using-parameter-objects

Overall I feel silly blocking the PR because of this code style, specially that the number of parameters (4) is still manageable. My concerns are I feel you are planning to change this code design in more places, I expect the scheduler to have more params soon as I work on the queueing, and our code is still in its early stages and more parameters might be added in places other than the scheduler.

frrist: There is certainly a limit to consider when declaring function parameters, and when encapsulating them starts to make sense. I completely agree that this isn't a black and white decision - nuance is required depending on the specific context. 

> A lot of fields and configurations are exposed in the constructor mainly for testing purposes with fallback default values if not configured

In this case I'd prefer to use functional options (we do a good job with this in some places already) as it makes it clearer when parameters are optional vs required. It's also easy to extend later with more options. 

> My concerns are I feel you are planning to change this code design in more places

I'll probably make a change like this again in the future if it makes sense while refactoring code. But I don't intend to remove param arguments everywhere! In this specific case 4 felt manageable enough to make the change.

frrist: Since it seems we are in agreement about 4 arguments being manageable here, can this get a ✅? 🙏

wdbaruni (approved):

PR #3959, opened 2024-04-26
refactor: define config instance. config is no longer global

frrist:
# Motivation
Previously we had a global config store. This is a bad pattern and complicates many aspects of creating a node. We should instead have a config instance that we pass to methods requiring config.

# Background

Bacalhau uses cobra to accept CLI flags and viper to manage configuration state. Viper and Cobra may be used together by binding viper to CLI flag definitions defined via cobra. Binding in this way allows configuration to be specified using a mixture of defaults, config files, environment variables, and flags. The priority viper treats these different options is as follows with highest precedence first:

1. Flag
2. Env Var
3. Config File
4. Defaults

This mean that when no config file, env vars, or flags are provided, the default settings are used. When a config file is provided, the values present in the file override default values. When env vars are provided they will override the defaults and/or the config file. When flags are passed they will override any other medium providing configuration.

# Implementation

The first point of interaction with cobra and viper is in `cmd/cli/root.go` where we access the global viper instance and bind it to all persistent flags present on the root command. This configures the viper instance with values related to the repo path, API, and logging. From there, child commands (all other commands) may bind additional flags to viper. For example, `cmd/cli/serve/serve.go` binds many different flags that configure how the server operates. When a child command is executed the following steps are taken to fully configure viper and produce a final configuration state.

1. An instance of `config` is created with the global viper instance
2. The repo path set in `cmd/cli/root.go` is retrieved from viper
3. A repo is initialized or opened using the repo path and configuration instance
   1. When opening or initializing a repo, it first checks if a config file is present under its path. If one is found it loads the values from the file into the `config` instance.
   2. Next the repo ensures paths to various files and directories are set on the config instance. The **repo will modify the config to include these paths**. These paths include:
      1. Authentication Tokens
      2. User private key
      3. executor plugin path
      4. compute storage (`executor_storages`) path
      5. update checker file path
      6. TLS certificate cache
      7. Libp2p private key
      8. orchestrator nats store
      9. orchestrator jobs database (`jobs.db`)
      10. compute store database (`executions.db`)
   3. Once the repo has modified the config to contain the various path, joined on its own path, it will ensure the paths are present on the filesytem by creating them or stat-ing them.
4. The initialized repo and reconciled configuration file are then used by the command to execute. The reconciled config (`types.BacalhauConfig`) contains values specified by defaults, config file, env vars, and/or flags merged in the order described above.

For the rest of the commands implementation services accept `types.BacalhauConfig` or a subset of it to configure themselves. They no longer depend on the previously global `config` package.

## Questions?

Various fields mentioned above contain paths to locations on disk where values may be read/stored. At present these paths can be anywhere on the filesystem - they need not live under the repo path. Since it is not possible to define defaults for these fields ahead of time, we set them based on the repo path when it is initialized.

In future iterations of this work, as described in the [Rethinking Configuration](https://www.notion.so/expanso/Rethinking-Configuration-435fbe87419148b4bbc5119d413786eb#5e137f97d2f24078b0c13853bed0462c) document, we intend to decouple the config from the repo. This can be achieved by removing the ability to configure the paths mentioned above from the config system and instead make them all relative to the repo path with hard-coded values. In this way, the repo path can become a field of the config that is provided to the repo when it initializes.


aronchick:
We should be really thoughtful as we fix tests. Every fix is an incompatibility with the way it worked previously - not that it's bad just let's be thoughtful about the user experience

frrist:
@coderabbitai review

cmd/cli/root.go
```diff
@@ -45,43 +42,21 @@ func NewRootCmd() *cobra.Command {
...
-				// this error indicates `defaultRepo` was unable to find a default location and the user
-				// didn't provide on.
-				util.Fatal(cmd, fmt.Errorf("bacalhau repo not set, please use BACALHAU_DIR or --repo"), 1)
-			}
-			if _, err := setup.SetupBacalhauRepo(repoDir); err != nil {
-				util.Fatal(cmd, fmt.Errorf("failed to initialize bacalhau repo at '%s': %w", repoDir, err), 1)
-			}
-
-			if err := configflags.BindFlags(cmd, rootFlags); err != nil {
+			if err := configflags.BindFlags(cmd, cfg, rootFlags); err != nil {
```
frrist, line 51: Review: We no longer create a config when the bacalhau binary runs. We instead create or open a repo in the places it's needed.

cmd/cli/root.go
```diff
@@ -45,43 +42,21 @@ func NewRootCmd() *cobra.Command {
...
 				util.Fatal(cmd, err, 1)
 			}
 
-			// If a CA certificate was provided, it must be a file that exists. If it does not
-			// exist we should not continue.
-			if caCert, err := config.Get[string](types.NodeClientAPIClientTLSCACert); err == nil && caCert != "" {
-				if _, err := os.Stat(caCert); os.IsNotExist(err) {
-					util.Fatal(cmd, fmt.Errorf("CA certificate file '%s' does not exist", caCert), 1)
-				}
-			}
```
frrist, line 76: Review: I can't imagine a more out-of-place location to perform this validation :face_with_spiral_eyes: I have removed this and add 1468f6d0968057049f184d03ffca12ee5b7ff970 to prevent the possibility of a panic in the client code.

cmd/cli/root.go
```diff
@@ -110,10 +85,12 @@ func NewRootCmd() *cobra.Command {
...
 	}
 	RootCmd.PersistentFlags().String("repo", defaultRepo, "path to bacalhau repo")
-	if err := viper.BindPFlag("repo", RootCmd.PersistentFlags().Lookup("repo")); err != nil {
+
+	// Bind the repo flag to the system configuration
+	if err := cfg.System().BindPFlag("repo", RootCmd.PersistentFlags().Lookup("repo")); err != nil {
 		util.Fatal(RootCmd, err, 1)
 	}
-	if err := viper.BindEnv("repo", "BACALHAU_DIR"); err != nil {
+	if err := cfg.System().BindEnv("repo", "BACALHAU_DIR"); err != nil {
```
frrist, line 93: The repo path is part of the 'System' config. And currently is the only value in that instance. Users are still free to specify a path for their repo.

cmd/cli/serve/serve_test.go
```diff
@@ -262,53 +264,56 @@ func (s *ServeSuite) TestDefaultServeOptionsHavePrivateLocalIpfs() {
...
-		// skip checking environments other than test, because
-		// system.GetEnvironment() in getPeers() always returns "test" while testing
-		if envName.String() != "test" {
-			continue
-		}
-
-		// this is required for the below line to succeed as environment is being deprecated.
-		config.Set(configenv.Testing)
-		peers, err = serve.GetPeers("env")
+	s.T().Skip("we are deprecating libp2p, skipping")
```
frrist, line 267: Review: we are deprecating Libp2p very shortly and I didn't want to bother trying to fix the compilation issues with this test because of that. Any objections?

cmd/util/api.go
```diff
@@ -1,35 +1,107 @@
...
+	repoDir, err := c.RepoPath()
+	if err != nil {
+		return nil, err
+	}
+	// TODO(forrest) [refactor]: a repo is required here as it contains the clients
+	// private key and a pk is needed in order to perform the authentication flow
+	// iff the configured authentication mode is challenge. We could decide to only
+	// require a repo here if the client _wants_ to communicate with a requester
+	// that uses authorization. In the event a requester doesn't need authorization
+	// we could use a un-authenticated client and not open/init a repo here.
```
frrist, line 30: I would prefer to make the repo responsible for managing client keys. Sensitive items should not be stored in the config. This TODO lays out the intention for this change.

cmd/util/auth/ask.go
```diff
@@ -1,95 +0,0 @@
...
-	"github.com/santhosh-tekuri/jsonschema/v5"
-	"github.com/spf13/cobra"
-	"golang.org/x/term"
-)
-
-// Returns a responder that responds to authentication requirements of type
-// `authn.MethodTypeAsk`. Reads the JSON Schema returned by the `ask` endpoint
-// and uses it to ask appropriate questions to the user on their terminal, and
-// then returns their response as serialized JSON.
-func askResponder(cmd *cobra.Command) responder {
```
frrist, line 19: this wasn't deleted, rather moved to a `Responder` interface.

cmd/util/flags/configflags/register.go
```diff
@@ -43,14 +44,22 @@ type Definition struct {
...
+func BindFlags(cmd *cobra.Command, c config.Context, register map[string][]Definition) error {
+	seen := make(map[string]Definition)
 	for _, defs := range register {
 		for _, def := range defs {
+			// sanity check to ensure we are not binding a config key on more than one flag.
+			if dup, ok := seen[def.ConfigPath]; ok {
+				return fmt.Errorf("DEVELOPER ERROR: duplicate regsistration of config key %s for flag %s"+
+					" previously registered on on flag %s", def.ConfigPath, def.FlagName, dup.FlagName)
+			}
+			seen[def.ConfigPath] = def
```
frrist, line 56: added to prevent bugs like the one in [cmd/util/flags/configflags/timeouts.go](https://github.com/bacalhau-project/bacalhau/pull/3959/files#diff-ddb4476b86c89712ae5ba6c7f256d7b0afcab61918271f36f3fc46e00c2bd25c) from going unnoticed.

cmd/util/hook/hooks.go
```diff
@@ -68,7 +68,9 @@ func AfterParentPostRunHook(hook runHookE) runHookE {
 // should have applied.
 var ClientPreRunHooks runHookE = Chain(
 	Adapt(ApplyPorcelainLogLevel),
-	Adapt(StartUpdateCheck),
+	// TODO(forrest) [fixme/are-you-fucking-kidding-me]: need to un-wang this from here
+	// gut check idea is to perform this in a single place at the root level
+	// Adapt(StartUpdateCheck),
```
frrist, line 73: This whole thing is really gross, especially now that we have moved away from a global config structure. In order to run the update checker we need an API client so we can ask the server what it's version is and then notify the user that the server they are talking to is out of date. I don't think we should notify users their server is out of date - I beleive the utility of this serve should be to tell users their _binary_ is out of date. Further, doing all this in a pre-run hook is really complicated since it requires a client , TLS config, and api token to connect to the server and authenticate.

We still run an update checker that notifies users when they run `serve` but with this commented out we won't notify users when they run client commands. 

I think a the follow on change here ought to be:
1. Dial get.bacalhau.org to get the latest version
2. if the version is newer than the current binary notify, else do nothing.

cmd/util/tokens.go
frrist, line 1: Review: logic here is unchanged, difference is rather than lazily rely on the global config we instead provide a path to read and write token from as an argument.

cmd/util/version.go
```diff
@@ -24,44 +20,25 @@ type Versions struct {
 	UpdateMessage string                   `json:"updateMessage,omitempty"`
 }
 
-func CheckVersion(cmd *cobra.Command, args []string) error {
```
frrist, line 27: method was unused

pkg/config/config.go
```diff
@@ -40,112 +59,112 @@ const (
...
 	}
+	return repoPath, nil
+}
 
-	for _, opt := range opts {
-		opt(params)
+func (c *config) Load(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		// if the config file doesn't exist then we obviously cannot load it
+		return fmt.Errorf("config file not found at at path: %q", path)
```
frrist, line 112: Load now requires a config to be present at the path, if no file is there is returns an error.

pkg/config/config.go
```diff
@@ -40,112 +59,112 @@ const (
...
 	environmentVariableReplace = strings.NewReplacer(".", "_")
 	DecoderHook                = viper.DecodeHook(mapstructure.TextUnmarshallerHookFunc())
 )
 
-const (
-	ConfigFileName = "config.yaml"
-	ConfigFileMode = 0666
-)
-
-func Init(path string) (types.BacalhauConfig, error) {
```
frrist, line 55: Removed as calling `config.New()` initializes a config

pkg/config/config.go
```diff
@@ -40,112 +59,112 @@ const (
...
-	defaultConfig.Auth.TokensPath = filepath.Join(path, TokensPath)
-
-	// We default to the folder which contains the job store, and add
-	// a subfolder for the network store.
-	defaultConfig.Node.Network.StoreDir = filepath.Join(
-		filepath.Dir(defaultConfig.Node.Requester.JobStore.Path),
-		NetworkTransportStore,
-	)
-
-	return defaultConfig
```
frrist, line 90: this logic is now the responsibility of the repo.

pkg/repo/migrations/v2_3.go
frrist, line 1: @wdbaruni the behavior remains unmodified here but would be good to review closely

pkg/config/config.go
```diff
@@ -4,17 +4,36 @@ import (
...
 	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/viper"
+
+	"github.com/bacalhau-project/bacalhau/pkg/config/configenv"
+	"github.com/bacalhau-project/bacalhau/pkg/config/types"
 )
 
+var _ Context = (*config)(nil)
+
+type Context interface {
```
wdbaruni, line 19: the name `Context` is confusing to me. I keep relating it to the context we pass around methods, and I thought this interface extends `context.Context`

pkg/config/config.go
```diff
@@ -4,17 +4,36 @@ import (
...
+	Load(path string) error
+	Current() (types.BacalhauConfig, error)
+
+	Set(key string, value interface{})
+	SetIfAbsent(key string, value interface{})
+	ForKey(key string, cfg interface{}) error
+
+	User() *viper.Viper
+	System() *viper.Viper
+}
```
wdbaruni, line 30: I am not sure what value this interface is providing. It is tightly coupled to viper and strictly file based config. Also how should we access config? Through `Current()` or `User()`?

pkg/config/config.go
```diff
@@ -40,112 +59,112 @@ const (
...
-	// initialize the configuration with default values.
-	return initConfig(path,
-		WithDefaultConfig(getDefaultConfig(path)),
-	)
+type config struct {
+	// viper instance for holding user provided configuration
+	user *viper.Viper
+	// viper instance for holding system specific configuration
+	system *viper.Viper
+	// the default configuration values to initialize with
```
wdbaruni, line 72: I have to say this is very confusing for me to follow and understand. What I think we need to do in general is just use and pass `BacalhauConfig` around instead of `config.Context`. You can have different implementations to construct `BacalhauConfig` with viper being our only option for now. This allows you to test things without having to setup viper or have a file based config. Just construct `BacalhauConfig` instance, set some values, and pass it to your test.

What I think will simplify things is making the repo path part of `BacalhauConfig` instead of something independent. The [Rethinking Configuration](https://www.notion.so/expanso/Rethinking-Configuration-435fbe87419148b4bbc5119d413786eb) doc is suggesting that as well

wdbaruni: What this might look like is moving `types.BacalhauConfig` to `config.Config`, and moving the current viper implementation to `config/viper/...` as an implementation to setup our config

pkg/config/helpers.go
```diff
@@ -0,0 +1,426 @@
...
+	"github.com/spf13/viper"
+
+	"github.com/bacalhau-project/bacalhau/pkg/config/types"
+	baccrypto "github.com/bacalhau-project/bacalhau/pkg/lib/crypto"
+	"github.com/bacalhau-project/bacalhau/pkg/lib/network"
+	"github.com/bacalhau-project/bacalhau/pkg/logger"
+)
+
+func Get[T any](c Context, key string) (T, error) {
+	raw := c.User().Get(key)
```
wdbaruni, line 27: These global getters and methods should be part of the `config.Context`

pkg/config/config.go
wdbaruni: The challenge that you might face is config setters. We need to decouple that from `BacalhauConfig` and we can have a separate interface for writing configs, with viper being an implementation.

pkg/config/util.go
```diff
@@ -0,0 +1,117 @@
...
+	IsSet(string) bool
+	Get(string) interface{}
+	AllKeys() []string
+}
+
+// unmarshalCompositeKey takes a key and an output structure to unmarshal into. It gets the
+// composite value associated with the given key and decodes it into the provided output structure.
+// It's especially useful when the desired value is not directly associated with the key, but
+// instead is spread across various nested sub-keys within the configuration.
+func unmarshalCompositeKey(v unmarshalContext, key string, output interface{}) error {
```
wdbaruni, line 21: not sure I understand this?

pkg/executor/docker/bidstrategy/semantic/image_platform.go
```diff
@@ -22,13 +22,13 @@ var _ bidstrategy.SemanticBidStrategy = (*ImagePlatformBidStrategy)(nil)
 var ManifestCache cache.Cache[docker.ImageManifest]
 var mu sync.Mutex
 
-func NewImagePlatformBidStrategy(client *docker.Client) *ImagePlatformBidStrategy {
+func NewImagePlatformBidStrategy(client *docker.Client, cfg config.Context) *ImagePlatformBidStrategy {
```
wdbaruni, line 25: we shouldn't be passing bacalhau config all the way down to these internal components, including executor, tracer, s3 storage provider and so on. The boundary for the config and `types.BacalhauConfig` should be the `node` package.

pkg/node/compute.go
```diff
@@ -197,9 +199,9 @@ func NewComputeNode(
 	if managementProxy != nil {
 		// TODO: Make the registration lock folder a config option so that we have it
 		// available and don't have to depend on getting the repo folder.
-		repo, _ := pkgconfig.Get[string]("repo")
+		computeDataPath, _ := pkgconfig.Get[string](cfg, types.NodeComputeDataPath)
 		regFilename := fmt.Sprintf("%s.registration.lock", nodeID)
-		regFilename = filepath.Join(repo, pkgconfig.ComputeStorePath, regFilename)
+		regFilename = filepath.Join(computeDataPath, regFilename)
```
wdbaruni, line 204: we want to minimize number of paths the users have to define in the config, and follow more convtions over configurations. The idea is to allow users to only define the repo path, and let each component create its sub-directories and files under the repo instead of allowing users to configure those as well

wdbaruni (changes requested):
this has been a very long and difficult PR to review, and I don't think I was able to cover all the details. There are major changes suggested, but still we need to break this PR before we can merge it even after addressing those proposals. 

As a start, you can create a PR where global configs are not used beyond `node` package and that `node` package is responsible in preparing and injecting all configs required by other components, such as publishers, executors and downloaders

Second PR can be my suggestion to decouple configs from viper and pass `BacalhauConfig` around and to the `node` package

Another is passing the client to commands as suggested, and so many ways to break into more PRs

pkg/config/helpers.go
frrist: They cannot be because of generics

cmd/cli/config/auto.go
```diff
@@ -89,7 +91,8 @@ func (a *autoSettings) validate() error {
 	return nil
 }
 
-func newAutoResourceCmd() *cobra.Command {
+// TODO(forrest) [fixme]
+func newAutoResourceCmd(cfg config.Context) *cobra.Command {
```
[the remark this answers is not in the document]
frrist: It's a TODO I left to remind myself to fix something and I forgot to remove the comment before pushing. Will remove.

pkg/config/util.go
frrist: I am only moving this code to a different file. This isn't anything new in the PR.

cmd/cli/config/auto.go
```diff
@@ -101,7 +104,16 @@ func newAutoResourceCmd() *cobra.Command {
...
-			return autoConfig(cmd.Context(), settings)
+			// create or open a repo
+			repoPath, err := cfg.RepoPath()
+			if err != nil {
+				return err
+			}
+			_, err = setup.SetupBacalhauRepo(repoPath, cfg)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
frrist: > why do you need to initialize a repo

Since we no longer create a repo in the root command, we must create on here if it's not present. This is because this CLI method modifies the config file in a repo, creating a config file when one isn't present.

> You are already passing a config to newAutoResourceCmd. How was that initialized?

As was done prior to this change - [the config instance is created with default values](https://github.com/bacalhau-project/bacalhau/pull/3959/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R88)
see `pkg/config/config.go`
```go
func New(opts ...Option) *config {
	c := &config{
		user:       viper.New(),
		system:     viper.New(),
		defaultCfg: configenv.Production,
	}
	for _, opt := range opts {
		opt(c)
	}
	c.user.SetEnvPrefix(environmentVariablePrefix)
	c.user.SetTypeByDefaultValue(inferConfigTypes)
	c.user.AutomaticEnv()
	c.user.SetEnvKeyReplacer(environmentVariableReplace)
	c.setDefault(c.defaultCfg)
	return c
}
```

> and how initializing the repo will change the config?

This is slightly different than before:
Prior to this change, the config instance [set paths on itself](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/config.go#L69) that the [repo would be configured based on](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/repo/util.go#L21).
Now when initializing the repo, it modifies the config by setting paths on the config instance via the method [`EnsureRepoPathsConfigured`](https://github.com/bacalhau-project/bacalhau/pull/3959/files#diff-51b4dbd635889500fad1dac99dd1fbd3ecff53680cd5bad57f1f05a09fa9afe5R242). 
This change was implemented in this manner because we discussed removing the option for users to set all paths on the configurations themselves that fell within the repo. Instead, the repository now manages the paths it sets, with the top-level path being the repository path, which users are still allowed to set.

cmd/cli/config/default.go
```diff
@@ -1,28 +1,28 @@
...
 		Use:   "default",
 		Short: "Show the default bacalhau config.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return defaultConfig(cmd)
 		},
 	}
-	showCmd.PersistentFlags().String("path", viper.GetString("repo"), "sets path dependent config fields")
+	showCmd.PersistentFlags().String("path", cfg.System().GetString("repo"), "sets path dependent config fields")
 	return showCmd
 }
```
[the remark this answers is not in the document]
frrist: We could just delete this command in favor of clearer documentation around what the default config structure is since lack of said docs are why it existed in the first place.

cmd/cli/create/create.go
```diff
@@ -78,7 +79,7 @@ func NewCmd() *cobra.Command {
 	return createCmd
 }
 
-func create(cmd *cobra.Command, cmdArgs []string, OC *CreateOptions) error { //nolint:funlen,gocyclo
+func create(cmd *cobra.Command, cfg config.Context, cmdArgs []string, OC *CreateOptions) error { //nolint:funlen,gocyclo
 	ctx := cmd.Context()
```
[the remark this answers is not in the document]
frrist: If we pass the client instead that will mean the repo needs to be created in the root command which is something I feel we should avoid.
However I do agree that we should be passing a client to methods called below this instead of the whole config. Opted not to do that here as it would balloon the size of this change, which is already quite large.

cmd/cli/devstack/devstack.go
```diff
@@ -166,17 +166,17 @@ func runDevstack(cmd *cobra.Command, ODs *devstack.DevStackOptions, IsNoop bool)
...
 
-	fsRepo, err := setup.SetupBacalhauRepo(repoPath)
+	fsRepo, err := setup.SetupBacalhauRepo(repoPath, cfg)
 	if err != nil {
 		return err
 	}
 
 	// make sure we don't run devstack with a custom IPFS path - that must be used only with serve
-	if path, err := config.Get[string](types.NodeIPFSServePath); err == nil && path != "" {
+	if path, err := config.Get[string](cfg, types.NodeIPFSServePath); err == nil && path != "" {
```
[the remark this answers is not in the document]
frrist: Because methods cannot have type parameters (generics): https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#no-parameterized-methods

cmd/cli/job/root.go
```diff
@@ -2,23 +2,25 @@ package job
...
-	cmd.AddCommand(NewLogCmd())
-	cmd.AddCommand(NewRunCmd())
-	cmd.AddCommand(NewStopCmd())
+	cmd.AddCommand(NewDescribeCmd(cfg))
+	cmd.AddCommand(NewExecutionCmd(cfg))
+	cmd.AddCommand(NewHistoryCmd(cfg))
+	cmd.AddCommand(NewListCmd(cfg))
+	cmd.AddCommand(NewLogCmd(cfg))
+	cmd.AddCommand(NewRunCmd(cfg))
+	cmd.AddCommand(NewStopCmd(cfg))
```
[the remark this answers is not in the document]
frrist: > recommendation is to create the client during the root's PersistentPreRunE and make it available to the command

This would require the repo to be created in the root command, which is something I feel we should avoid.

> Since not all commands need a client, there are two ways we can do that

I think these are worth exploring in a follow on, especially when it's no longer the case that different commands need different versions the client (v1 v v2) to operate.

cmd/cli/create/create.go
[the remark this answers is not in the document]
wdbaruni: No we don't have to create the repo in root. Check my comment here on how we can do this https://github.com/bacalhau-project/bacalhau/pull/3959#discussion_r1596903257

cmd/cli/config/list.go
```diff
@@ -29,7 +29,16 @@ func newListCmd() *cobra.Command {
...
-			return list(cmd, o)
+			// create or open a repo
+			repoPath, err := cfg.RepoPath()
+			if err != nil {
+				return err
+			}
+			_, err = setup.SetupBacalhauRepo(repoPath, cfg)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
frrist: https://github.com/bacalhau-project/bacalhau/pull/3959#discussion_r1600396698

cmd/cli/config/show.go
```diff
@@ -1,27 +1,38 @@
...
+			// create or open a repo
+			repoPath, err := cfg.RepoPath()
+			if err != nil {
+				return err
+			}
+			_, err = setup.SetupBacalhauRepo(repoPath, cfg)
+			if err != nil {
+				return err
+			}
+
```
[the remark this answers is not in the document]
frrist: https://github.com/bacalhau-project/bacalhau/pull/3959#discussion_r1600396698

cmd/cli/job/root.go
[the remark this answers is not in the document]
wdbaruni: > This would require the repo to be created in the root command, which is something I feel we should avoid.

No we don't have to. Both options I explained don't require you to create a repo for all commands, but only for the commands with certain annotation of specific type

> I think these are worth exploring in a follow on

What we are doing now is not a good practice where we create a cfg, pass it to commands, and then modify it in preRun, where we can do most things in preRun. It is not right to push this now and then try to fix it later, and having two clients is not complicating things significantly

cmd/cli/root.go
```diff
@@ -45,43 +42,21 @@ func NewRootCmd() *cobra.Command {
 		"api":     configflags.ClientAPIFlags,
 		"logging": configflags.LogFlags,
 	}
+	cfg := config.New()
```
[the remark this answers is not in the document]
frrist: Its created with default values and bound to flags as was done before.

pkg/config/helpers.go
wdbaruni: I see. It would be nice if we just pass BacalhauConfig and avoid having these getters in general.

cmd/cli/create/create.go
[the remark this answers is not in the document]
frrist: If we don't create or open the repo in the root command then:
- how will values be read from the config file if one is present?
- how will the clientID be used to sign requests?

cmd/util/api.go
```diff
@@ -1,35 +1,130 @@
...
+	}
+	var apiPort uint16
+	if err := c.ForKey(types.NodeClientAPIPort, &apiPort); err != nil {
+		return nil, err
+	}
+	apiSheme := "http"
+	if tlsCfg.UseTLS {
+		apiSheme = "https"
+	}
+	base := fmt.Sprintf("%s://%s:%d", apiSheme, apiHost, apiPort)
```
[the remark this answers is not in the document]
frrist: This does that
```go
var tlsCfg types.ClientTLSConfig
	if err := c.ForKey(types.NodeClientAPITLS, &tlsCfg); err != nil {
		return nil, err
	}
```

Are you saying you would like this logic encapsulated into a single method?

pkg/config/config.go
frrist: I can rename this type to something different, suggestions welcome.
I picked this name because `config.Context` felt descriptive (it's the context of the current configuration) and didn't stutter like `config.Config`

pkg/executor/docker/bidstrategy/semantic/image_platform.go
frrist: I agree. Implementing that change as a part of this PR would have made this already large change even larger so I opted not to do it. I would prefer to do that in a follow on.

pkg/node/compute.go
frrist: I am in agreement so long as we make it the responsibility of the repo to create and manage these paths on disk, or whatever medium is backing the repo.

cmd/cli/create/create.go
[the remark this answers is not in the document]
wdbaruni: What I meant is the repo will be created in the PersistentPreRunE in the root, and a client will be created only for the commands that needs one. There shouldn't be any difference on when the repo will be create compared to your current implementation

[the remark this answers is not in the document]
frrist: > No we don't have to create the repo in root
>> What I meant is the repo will be created in the PersistentPreRunE in the root

Sorry, but which is it?

pkg/repo/migrations/helpers.go
```diff
@@ -1,35 +1,33 @@
...
 		return nil, err
 	}
 
-	configFile := filepath.Join(repoPath, config.ConfigFileName)
-	v := viper.New()
-	v.SetTypeByDefaultValue(true)
-	v.SetConfigFile(configFile)
-
+	configFile := filepath.Join(repoPath, config.FileName)
+	c := config.New()
```
wdbaruni, line 22: this will load config with default values and env variables. The intention here is to just get the configs from the config file

pkg/repo/migrations/v1_2.go
```diff
@@ -46,16 +46,16 @@ var V1Migration = repo.NewMigration(
 		// if no incorrect values are present they are left as is.
 		doWrite := false
 		if haveSameElements(oldSwarmPeers, cfg.Node.IPFS.SwarmAddresses) {
-			v.Set(types.NodeIPFSSwarmAddresses, []string{})
+			currentCtx.Set(types.NodeIPFSSwarmAddresses, []string{})
```
wdbaruni, line 49: you should keep using viper directly for the migrations as we want to work directly with config files and not with fully loaded configs with defaults and binded env variables. Also I don't think it is making things any simpler by working with config.Context

pkg/repo/migrations/v2_3.go
```diff
@@ -32,24 +32,31 @@ var V2Migration = repo.NewMigration(
...
 		// reading the correct libp2p key in case the user is overriding the default value.
-		resolvedCfg, err := config.Load(repoPath)
+		newCtx := config.New()
+		if _, err := os.Stat(filepath.Join(repoPath, config.FileName)); err == nil {
+			if err := newCtx.Load(filepath.Join(repoPath, config.FileName)); err != nil {
+				return err
+			}
+		}
+		r.EnsureRepoPathsConfigured(newCtx)
+		resolvedCfg, err := newCtx.Current()
```
wdbaruni, line 42: I think all looks good, except ensuring repo paths are configured. Will this create the repo paths if they don't exist, or just populate the config with the paths?

pkg/repo/migrations/v2_3.go
```diff
@@ -21,7 +21,7 @@ var V2Migration = repo.NewMigration(
 	repo.RepoVersion2,
 	repo.RepoVersion3,
 	func(r repo.FsRepo) error {
-		v, fileCfg, err := readConfig(r)
+		currentCtx, currentCfg, err := readConfig(r)
```
wdbaruni, line 24: we only want to read the configs in the file. Otherwise you will be writing all default configs to the file instead of just the new configs that we want to persist

pkg/repo/fs.go
```diff
@@ -210,3 +237,22 @@ func (fsr *FsRepo) WriteRunInfo(ctx context.Context, summaryShellVariablesString
...
+	c.SetIfAbsent(types.NodeComputeStoragePath, fsr.join(config.ComputeStorePath))
+
+	c.SetIfAbsent(types.UpdateCheckStatePath, fsr.join(config.UpdateCheckStatePath))
+	c.SetIfAbsent(types.NodeClientAPITLSAutoCertCachePath, fsr.join(config.AutoCertCachePath))
+	c.SetIfAbsent(types.UserLibp2pKeyPath, fsr.join(config.Libp2pPrivateKeyFileName))
+	c.SetIfAbsent(types.NodeNetworkStoreDir, fsr.join(config.OrchestratorStorePath, config.NetworkTransportStore))
+
+	c.SetIfAbsent(types.NodeRequesterJobStorePath, fsr.join(config.OrchestratorStorePath, "jobs.db"))
+	c.SetIfAbsent(types.NodeComputeExecutionStorePath, fsr.join(config.ComputeStorePath, "executions.db"))
+}
```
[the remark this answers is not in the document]
frrist: Yeah, this is a bit of :chicken: :egg: problem.
> Config should use the repo as base for the different paths it holds, such as execution and job store paths

This sounds interesting - would it imply the paths in the config are relative paths the repo uses rather than complete paths?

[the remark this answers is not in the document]
wdbaruni: default paths are relative to repo path, which is also the case today. Based on the proposal here https://www.notion.so/expanso/Rethinking-Configuration-435fbe87419148b4bbc5119d413786eb, we will only have a repo path and will no longer expose configurations for each individual path

wdbaruni (approved):
Minor comments. Looks good

pkg/config/config.go
```diff
@@ -4,17 +4,48 @@ import (
...
+}
+
+type Reader interface {
+	Current() (types.BacalhauConfig, error)
+	Get(string) interface{}
+	ForKey(key string, cfg interface{}) error
+}
+
+type Writer interface {
+	Load(path string) error
```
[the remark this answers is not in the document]
frrist: The Load method loads the config file at the provided path into config. Quick refresher on how this all plays out:

when no config file, env vars, or flags are provided, the default settings are used. **When a config file is provided, the values present in the file override default values**. When env vars are provided they will override the defaults and/or the config file. When flags are passed they will override any other medium providing configuration.

So calling load will modify the config by loading whatever data is in the corresponding file in to the config.

PR #3964, opened 2024-04-29
fix: allow timeouts in the client methods that make API call

frrist:
Allows `bacalhau version` to fail quickly if the server is offline.

### DIFF
--- a/cmd/cli/version/version.go
+++ b/cmd/cli/version/version.go
@@ -18,6 +18,7 @@ package version
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/rs/zerolog/log"
@@ -100,8 +101,12 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
 	if oV.ClientOnly {
 		versions.ClientVersion = version.Get()
 	} else {
+		// NB(forrest): since `GetAllVersions` is an API call - in the event the server is un-reachable
+		// we timeout after 3 seconds to avoid waiting on an unavailable server to return its version information.
+		vctx, cancel := context.WithTimeout(ctx, time.Second*3)
+		defer cancel()
 		var err error
-		versions, err = util.GetAllVersions(ctx)
+		versions, err = util.GetAllVersions(vctx)
 		if err != nil {
 			// No error on fail of version check. Just print as much as we can.
 			log.Ctx(ctx).Warn().Err(err).Msg("failed to get updated versions")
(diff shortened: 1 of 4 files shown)


cmd/cli/version/version.go
```diff
@@ -100,8 +101,12 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
 	if oV.ClientOnly {
 		versions.ClientVersion = version.Get()
 	} else {
+		// NB(forrest): since `GetAllVersions` is an API call - in the event the server is un-reachable
+		// we timeout after 3 seconds to avoid waiting on an unavailable server to return its version information.
+		vctx, cancel := context.WithTimeout(ctx, time.Second*3)
+		defer cancel()
 		var err error
-		versions, err = util.GetAllVersions(ctx)
+		versions, err = util.GetAllVersions(vctx)
```
frrist, line 109: this change will cause the `bacalhau version` command to return quickly if the server (requester) is unreachable.

cmd/util/api.go
```diff
@@ -59,8 +59,8 @@ func GetAPIClientV2(cmd *cobra.Command) clientv2.API {
 			PersistCredential: func(cred *apimodels.HTTPCredential) error {
 				return WriteToken(base, cred)
 			},
-			Authenticate: func(a *clientv2.Auth) (*apimodels.HTTPCredential, error) {
-				return auth.RunAuthenticationFlow(cmd, a)
+			Authenticate: func(ctx context.Context, a *clientv2.Auth) (*apimodels.HTTPCredential, error) {
+				return auth.RunAuthenticationFlow(ctx, cmd, a)
```
frrist, line 63: This change allows calls to the authentication API to respect context cancellation in the event the server (requester) serving the `/api/v1/auth` endpoint is unreachable. Previously we used the less explicit command context for this method

cmd/cli/version/version.go
```diff
@@ -100,8 +101,12 @@ func (oV *VersionOptions) Run(ctx context.Context, cmd *cobra.Command) error {
 	if oV.ClientOnly {
 		versions.ClientVersion = version.Get()
 	} else {
+		// NB(forrest): since `GetAllVersions` is an API call - in the event the server is un-reachable
+		// we timeout after 3 seconds to avoid waiting on an unavailable server to return its version information.
+		vctx, cancel := context.WithTimeout(ctx, time.Second*3)
```
wdbaruni, line 106: what do we log or return to the user after the 3 seconds timeout?

frrist: We only return the client version in that case:
```bash
> export BACALHAU_API_HOST=1.1.1.1
> bacalhau version
 CLIENT                UPDATE MESSAGE 
 v1.3.0-43-g9debb92e9                 

> export BACALHAU_API_HOST=bootstrap.production.bacalhau.org
> bacalhau version
 CLIENT                SERVER  LATEST  UPDATE MESSAGE 
 v1.3.0-43-g9debb92e9  v1.3.0  v1.3.0                 
```

The log message below this comment on line 112 isn't returned to the user on the CLI since we only log `Fatal` for CLI calls.

wdbaruni (approved):

PR #3965, opened 2024-04-30
Make workload-onboarding examples testable

js-ts:
This is ran along side the bashtub tests but it isn't a bashtub test itself, since the bashtub tests execute sequentially

Taking a subset of examples and making them testable
All Examples are executed in parallel
Test Passes when the Job is complete and the exit code is 0

when test passes this badge is displayed in the Example markdown
Static Badge

when It fails this batch is disaplayed
Static Badge


js-ts:
All of the V2 Declarative jobspecs aren't still merged into main and are in this branch https://github.com/bacalhau-project/bacalhau/tree/docupdating-a1qa

wdbaruni:
@MichaelHoepler is this still useful for the docs?

wdbaruni:
@frrist do we need to do this work in the docs repo or not needed at all?

wdbaruni:
@MichaelHoepler I am closing this as docs have been migrated to their own repo. Please track this somewhere

PR #3967, opened 2024-05-01
feat: premit update check configuration w/ envvar

frrist:
- allows update checker endpoint to be specified via the environment variable BACALHAU_UPDATE_CHECKER_ENDPOINT. If unset the default endpoint is used.


aronchick (approved):
Perfect!

PR #3968, opened 2024-05-01
refactor: return client version by default

### LINKED ISSUE #3308: The version command never fails
When running `bacalhau version` it should always print the client version.
It currently also attempts to get the server version, but if it is unable to do so does not explain why.
This surfaced when testing with TLS and when the version command failed with a TLS error, the error was silently swallowed.

Whilst it is fine that we allow the command to run without access to the network, 
we might consider any other error (e.g. there is a network but no TLS specified)
as an error and inform the user as such.

frrist:
- replaces the '--client' flag on the `version` command with '--server'. The default value of '--server' is false. This results in the `version` command always succeeding.
- closes #3308
- see https://github.com/bacalhau-project/get.bacalhau.org/pull/28

New Behavior:
```bash
> bacalhau version
 CLIENT                UPDATE MESSAGE 
 v1.3.0-44-g2b688b95b                 

> bacalhau version --client=false
Flag --client has been deprecated, use --server
 CLIENT                SERVER  LATEST  UPDATE MESSAGE 
 v1.3.0-44-g2b688b95b  v1.3.0  v1.3.0                 

> bacalhau version --client=true
Flag --client has been deprecated, use --server
 CLIENT                UPDATE MESSAGE 
 v1.3.0-44-g2b688b95b                 

> .bacalhau version --server
 CLIENT                SERVER  LATEST  UPDATE MESSAGE 
 v1.3.0-44-g2b688b95b  v1.3.0  v1.3.0                 

> bacalhau version --server=false
 CLIENT                UPDATE MESSAGE 
 v1.3.0-44-g2b688b95b

### DIFF
--- a/cmd/cli/version/version.go
+++ b/cmd/cli/version/version.go
@@ -20,8 +20,8 @@ import (
 	"fmt"
 
 	"github.com/jedib0t/go-pretty/v6/table"
-	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"github.com/bacalhau-project/bacalhau/pkg/version"
 
@@ -32,12 +32,15 @@ import (
 )
 
 type VersionOptions struct {
-	ClientOnly bool
+	Client     bool
+	Server     bool
 	OutputOpts output.OutputOptions
 }
 
 func NewVersionOptions() *VersionOptions {
 	return &VersionOptions{
+		Client:     true,
+		Server:     false,
 		OutputOpts: output.OutputOptions{Format: output.TableFormat},
 	}
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by aronchick — APPROVED** [review]

**Review by wdbaruni — CHANGES_REQUESTED** [review]
I don't think we need this change given you introduced a short timeout when querying server version here https://github.com/bacalhau-project/bacalhau/pull/3964

It might make sense to give it sometime and see if that changes hits the sweet spot of improved user experience where they can easily know if their clients are outdated, and  not hang if client is disconnected. Feel free to reduce the 3 seconds timeout to 2 seconds if you believe it is not short enough

**Comment by aronchick** [comment]
> I don't think we need this change given you introduced a short timeout when querying server version here #3964
> 
> It might make sense to give it sometime and see if that changes hits the sweet spot of improved user experience where they can easily know if their clients are outdated, and not hang if client is disconnected. Feel free to reduce the 3 seconds timeout to 2 seconds if you believe it is not short enough
> 
> @aronchick what do you think?

I'll be honest, I can see both sides. 

Is there a scenario where someone should be able to get the version of a binary without connecting to a server? yes - i just downloaded it.

Is the function of our client app ONLY to connect to the server? Yes, so it makes sense we do that too.

I _guess_ the ideal user experience, on first download would for this to come back in 1 second.

```
 CLIENT  SERVER  LATEST  UPDATE MESSAGE
 v1.3.0  Disconnected  v1.3.0
```
 
And the moment API_Server (or whatever) is set - 

```
 CLIENT  SERVER  LATEST  UPDATE MESSAGE
 v1.3.0  v1.3.0  v1.3.0
 ```
 
 We should also have a `bacalhau connect <IP_ADDRESS>` command....

wdbaruni:
I like the suggestion of giving feedback to the user that we couldn't figure out the server's version because it was disconnected. So reducing the timeout from 3 seconds to possibly 1 second should do the trick instead of only retrieving the client version by default

Issue #3832, opened 2024-02-19
Update test suites to use new CLIs and APIs


wdbaruni:
All our test suites are using legacy `/pkg/model` package and legacy APIs to submit jobs. We should switch to using new APIs and V2 client
```[tasklist]
### Tasks
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4140
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4141
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4142
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4143
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4144
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4145
- [ ] https://github.com/bacalhau-project/bacalhau/issues/4159
```

frrist:
being addressed at an individual CLI level via https://github.com/bacalhau-project/bacalhau/issues/3838

frrist:
This issue also includes the devstack, and anything else that is using the old API methods. When this issue closes we must be able to delete the model and v1 api packages.

Issue #3816, opened 2024-03-27
Remove embedded IPFS node


rossjones:
Remove embedded IPFS node from bacalhau and associated tests. The end result should be that bacalhau communicates with IPFS solely via it's API and not to an embedded IPFS node.  This will require documentation explaining how to set up an IPFS node (covered in other tickets).

After taking an initial look at this work in https://github.com/bacalhau-project/bacalhau/pull/3712 it appears that this is quite a large change to implement, which is likely to take quite a while. Given the potential impact on people currently running their own cluster, and the length of time the change is likely to take, it would be better to take a staged approach to this work.  This staged approach will allow users to change over to the new and preferred approach at their own pace and without the frustration of breaking changes. 

### Stage 1 - Deprecate CLI flags ✅

**Proposed release:** 1.3.1

Deprecate any IPFS command line flags that will not be needed once this change happens.  The deprecation will only show a message alongside the CLI help, and will explain that these flags will not be supported in future. This was completed in https://github.com/bacalhau-project/bacalhau/pull/3778

### Stage 1.5 - Explanation on a web page

**Proposed release:** 1.3.1

Optional: Write a web-page (probably not a blog) outlining the intent and the approach being taken.  This doesn't need to be a novel, but should include enough information so that interested parties can understand the change, and what they will need to plan for over upcoming releases.  This will serve as a URL you can share for people looking for more information.

### Stage 2 - Deprecation warning 

**Proposed release:** 1.3.2

Any time the constructors in https://github.com/bacalhau-project/bacalhau/blob/main/pkg/ipfs/node.go are called, they should log a warning that the embedded IPFS node is going away in a future version. If stage 1.5 happened, it should link to that.  Ideally this would not bother logging in tests. 

### Stage 3 - Inform users of upcoming change

**Proposed release:** 1.4

Clearly include in the release notes the link from Stage 1.5 (if it exists) and a warning that it will happen in 1.5 
Be clear that IPFS will still be supported, but how we support it will change.

### Stage 4 - Disable embedded node in devstack/serve.

**Proposed release:** 1.5 

Fully disable the use of the embedded node for non-test use by only supporting --ipfs-connect (with swarm key option).  e.g. only allow ipfs.NewNodeWithConfig to be called from tests.  For non-test uses change any relevant Source or Publisher to use a remote client.  You can create a client with 

```
client, err := ipfs.NewClientUsingRemoteHandler(ctx, connectionString)
```

[...]

PR #3969, opened 2024-05-01
refactor: remove embedded ipfs

frrist:
- see https://github.com/bacalhau-project/bacalhau/pull/3712 for prior art/an alternative approach.
- This PR is more of a "bulldozer" approach than #3712. I am removing all references to IPFS as a storage provider and publish, along with any tests making assertions on that functionality. Many tests assume an IPFS client is available since that was always the case prior to removing the embedded IPFS node. And I suspect it will be far simpler to remove IPFS all together and then add it back in a follow on.
- we have a heap of tests that still use the "legacy" models and said tests are tightly coupled to IPFS. These are known as Scenario tests. I would like to remove these tests since they are using models (`model`) we plan to remove very soon as we move to the new API and `models` package. We can always add similar testing back in a follow on.


wdbaruni:
I am sorry but I don't agree with the approach. You are removing almost all of our test scenarios, and pushing a commit with no support for IPFS users that can block our release pipeline until it is added back in. We still need to give notice to our users and inform them to spin up their own IPFS node instead of the embedded one

Instead of deleting the tests, the right approach is to migrate them to the new models as planned in https://github.com/bacalhau-project/bacalhau/issues/3832, and to local publisher instead of IPFS as planned in https://github.com/bacalhau-project/bacalhau/issues/3816. Now we can speed things and remove embedded ipfs from tests before 1.5 release, but we shouldn't do more intrusive things or fully remove embedded ipfs without communicating that early to users.

frrist:
@wdbaruni agreed, which is why this remains in draft.

Issue #3831, opened 2024-02-19
Fully remove existing existing CLIs, APIs and models


wdbaruni:
Delete `pkg/model` package, delete `publicapi/endpoints/requester` APIs and rename V2 client to just client!

frrist:
being addressed by separate issues in https://github.com/bacalhau-project/bacalhau/issues/3838

Issue #3719, opened 2024-03-27
Better errors on job version


aronchick:
```
❯ bacalhau job run job.yaml
the job provided is invalid: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Job.Version of type uint64
```

This error is very hard to understand. What should the version be? And why would we tell someone what unmarshalling is?

wdbaruni:
the job version is auto generated and updated by the requester node and should be ignored if it was provided by the user. This is a bug and should be handled better

frrist:
> cannot unmarshal string into Go struct field Job.Version of type uint64

@aronchick can you share the `job.yaml` file you used? And in the future for errors like this please always share the corresponding file :smile: 

> the job version is auto generated and updated by the requester node and should be ignored if it was provided by the user. This is a bug and should be handled better

Hummm, on the other hand I think there is value to alerting users they are providing a field that has no affect/shouldn't be specified - else they may continue to do it and not realize its a noop, basically a waste of their time and increased cogitative  overhead. 
Better, we should exclude fields used by the protocol/internally from the data users submit

wdbaruni:
fair enough. I think the benefit I was looking for is allowing users to submit a similar version of a job by doing something like `bacalhau job describe --output yaml | bacalhau job run`

aronchick:
Agreed - but I think it's perfectly valid to just ignore and alert the user as a warning "hey we ignored this field"

wdbaruni:
I like the warning approach instead of failing or silently ignoring the fields

frrist:
I've created an issue to track separating user submission from system representation here: https://github.com/bacalhau-project/bacalhau/issues/3983

wdbaruni:
@frrist lets just ignore system defined fields such as `version` for now

frrist:
One way to reproduce this error is with the following job spec:
```yaml
Name: A Simple Docker Job
Type: batch
Version: "10"
Count: 1
Tasks:
  - Name: My main task
    Engine:
      Type: docker
      Params:
        Image: ubuntu:latest
        Entrypoint:
          - /bin/bash
        Parameters:
          - -c
          - echo Hello Bacalhau!
```
The key thing here is that `Version` is of type `string` which doesn't decode into the expected `int64` type of the job structure. However, the same spec with `Version` as an int runs just fine; the field is ignored. Similar behavior applies to `State`, `Revision`, `CreateTime` and `ModifyTime`. 
[...]

PR #3970, opened 2024-05-02
refactor: migrate docker & wasm run to v2 api

frrist:
This change removes the `pkg/model` dependency from the `docker run` and `wasm run` commands and migrates both of them to the new API methods. The flags on each command remain unchanged.

makes progress towards #3831 and #3832 

### TODO:
- fix some todo's I'll call out in review


cmd/cli/docker/docker_run.go
```diff
@@ -139,101 +134,127 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, opts *DockerRunOptions) err
...
+	// Normalize and validate the job spec
+	job.Normalize()
+	if err := job.ValidateSubmission(); err != nil {
+		return fmt.Errorf("%s: %w", userstrings.JobSpecBad, err)
 	}
 
+	// TODO(forrest) [refactor]: this options is _almost_ useful. At present it marshals the entire
+	// job spec to yaml, said spec cannot be used with `bacalhau job run` since it contains fields that
+	// users are not permitted to set, like ID, Version, ModifyTime, State, etc.
+	// The solution here is to have a "JobSubmission" type that is different from the actual job spec.
```
frrist, line 151: Looking for opinions on this. We don't need to land the proposed change as part of this PR, but it would be good to do soon

cmd/cli/docker/docker_run.go
```diff
@@ -139,101 +134,127 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, opts *DockerRunOptions) err
...
+		WithEntrypoint(opts.Entrypoint...).
+		WithEnvironmentVariables(opts.SpecSettings.EnvVar...).Build()
 	if err != nil {
 		return nil, err
 	}
 
-	labels, err := parse.Labels(ctx, opts.SpecSettings.Labels)
-	if err != nil {
-		return nil, err
+	// TODO(forrest) [refactor]: this logic is duplicated in wasm_run
```
frrist, line 206: I should extract the common logic here before merging.

cmd/cli/docker/docker_run.go
```diff
@@ -139,101 +134,127 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, opts *DockerRunOptions) err
...
-		legacy_job.WithOutputs(outputs...),
-		legacy_job.WithAnnotations(labels...),
-		legacy_job.WithNodeSelector(nodeSelectorRequirements),
-		legacy_job.WithDeal(
-			opts.DealSettings.TargetingMode,
-			opts.DealSettings.Concurrency,
-		),
-	)
+	task, err := models.NewTaskBuilder().
+		Name("TODO").
```
frrist, line 216: We currently don't have a means for users to specify a task name with the current flags on `wasm run` or `docker run` I could stub in a UUID here and add a `--task-name` flag later, thoughts?

cmd/cli/docker/docker_run.go
```diff
@@ -139,101 +134,127 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, opts *DockerRunOptions) err
...
 	}
 
+	constraints, err := parse.NodeSelector(opts.SpecSettings.Selector)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("parseing job contstrints: %w", err)
+	}
+	job := &models.Job{
+		Name:        "TODO",
+		Namespace:   "TODO",
```
frrist, line 248: Similar point to task name in my comment above - these could be UUIDs, values from flags we add, or something else?

cmd/cli/docker/docker_run.go
```diff
@@ -139,101 +134,127 @@ func dockerRun(cmd *cobra.Command, cmdArgs []string, opts *DockerRunOptions) err
...
+		Type:        models.JobTypeBatch,
+		Priority:    0,
```
frrist, line 250: I believe it is assumed that docker and wasm jobs are batch when run over the CLI, but we could easily change that. Additionally priority doesn't have a flag. Gut says we need flags for these - ideas on good default values appreciated.

cmd/cli/exec/exec.go
frrist, line 1: Unfortunately when we implement the `exec` feature we chose to use the v1 APIs. So we will need to migrate this command over to use the v2 APIs. This can be done in a follow on, or in this PR. Currently (I think?) `exec` is a 'experimental' command so we should be clear to make changes to its behavior.

cmd/cli/wasm/wasm_run.go
```diff
@@ -139,111 +135,146 @@ func newRunCmd() *cobra.Command {
...
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("wasm env vars invalid: %w", err)
 	}
-
-	nodeSelectorRequirements, err := parse.NodeSelector(opts.SpecSettings.Selector)
+	engineSpec, err := models.WasmSpecBuilder(&models.InputSource{
+		Source: entryModule,
+		Alias:  "TODO",
+		Target: "TODO",
```
frrist, line 204: Unsure what `Alias` and `Target` actually do - obviously need to fix this, just not sure what the right answer is. cc @wdbaruni

cmd/util/flags/cliflags/spec.go
```diff
@@ -35,7 +35,7 @@ func NewSpecFlagDefaultSettings() *SpecFlagSettings {
 type SpecFlagSettings struct {
 	Publisher     opts.PublisherOpt // Publisher - publisher.Publisher
 	Inputs        opts.StorageOpt   // Array of inputs
-	OutputVolumes []string          // Array of output volumes in 'name:mount point' form
+	OutputVolumes map[string]string // Map of output volumes in 'name:mount point' form
```
frrist, line 38: This maps to a slice of `models.ResultPath`. Using a map felt more intuitive and idiomatic, as it allows for direct enforcement of the name-to-mount mapping through the flags library, rather than performing the validation after the fact.

cmd/util/flags/cliflags/spec.go
```diff
@@ -57,13 +57,13 @@ func SpecFlags(settings *SpecFlagSettings) *pflag.FlagSet {
...
-	flags.StringSliceVarP(
+	flags.StringToStringVarP(
 		&settings.OutputVolumes,
 		"output",
 		"o",
 		settings.OutputVolumes,
-		`name:path of the output data volumes. `+
-			`'outputs:/outputs' is always added unless '/outputs' is mapped to a different name.`,
+		`name=path of the output data volumes. `+
+			`'outputs=/outputs' is always added unless '/outputs' is mapped to a different name.`,
```
frrist, line 66: I don't think this is true anymore. Will need to address this before merging.

cmd/util/opts/storage.go
```diff
@@ -70,7 +70,9 @@ func (o *StorageOpt) Set(value string) error {
 			return fmt.Errorf("unpexted key %s in field %s", key, field)
 		}
 	}
-	storageSpec, err := legacy_job.ParseStorageString(sourceURI, destination, options)
+	// TODO(forrest) [correctness]: need to allow aliases to be provided over CLI
+	alias := "TODO"
```
frrist, line 74: Need to address, not sure if the alias should be passed by the client or picked by the CLI.

pkg/models/docker_spec_config.go
frrist, line 1: This was ported from the model package. We will be deleting the `model` package at which point the duplicated code will be remove.

pkg/models/publisher_spec_config.go
frrist, line 1: also ported from `model`.

pkg/models/storage_source_config.go
frrist, line 1: also ported from `model`.

frrist (commented):
left some comments to guide review

cmd/cli/docker/docker_run.go
aronchick: Generally speaking, intermediate representation are not necessarily a bad thing. Forcing the user or user layer to understand the capabilities of what is required at the submission level is not going to give a great UX.

However, this is going to be an area we end up coming back to often, because this translation will change when we decide to change internal functions.

frrist: Is this for or against a separate type for job submissions? We could even embed said type in the current Job model.

frrist: Conversation here related to this: https://github.com/bacalhau-project/bacalhau/issues/3719#issuecomment-2061238679

wdbaruni: I hear your point. My preference is to delay any decision to change our API specs and have different ways to submit jobs until we collect more feedback about current APIs, and have more dedicated time to explore alternatives and align on the right option. Options might include splitting the job type into spec and state top level fields, where state hold the system defined fields, and spec is what the user submits. Again, I truly don't recommend figuring this out just yet.

A low hanging fruit is to introduce SanitizeForSubmission() method under models.Job that will clear out all fields that are reserved by the system so that something like bacalhau job describe <job_id> --output yaml | bacalhau job run would work

cmd/cli/docker/docker_run.go
wdbaruni: we can use "main" as the default task name. Task names are only useful when we support multiple tasks per job, and that won't be doable through `docker run` and `wasm run` as they will most likely submit a single task per job. So calling it "main" or any other standard name is acceptable

cmd/cli/docker/docker_run.go
wdbaruni: we have a default namespace for `bacalhau job run` https://github.com/bacalhau-project/bacalhau/blob/6f00c8a89e23b28a882e467885d1b6074f81f63e/pkg/models/constants.go#L5

though we are still using clientID for `docker run`. I think we should use the default namespace, but expose a flag to set the desired namespace. This means users won't be able to filter their own jobs in the public network, but that might be acceptable based on our longer term plans. Thoughts?

cmd/cli/docker/docker_run.go
wdbaruni: there is a `--target` flag that controls this:
```
      --target all|any                          Whether to target the minimum number of matching nodes ("any") (default) or all matching nodes ("all") (default any)
```

any is a batch job, where all is an ops job

cmd/util/flags/cliflags/spec.go
wdbaruni: why changed from `name:path` to `name=path`? this will break users. Also we will need to define a default output path `outputs:/outputs` for both `docker run` and `wasm run` to avoid breaking existing behavour and breaking all our examples and docs. I don't think it is a bad design anyways for those commands, but worth exploring the right long term option with @aronchick

cmd/cli/docker/docker_run.go
frrist: The "current APIs" are the ones serving `bacalhau create/list/describe/etc.` - AKA the V1 APIs.
The APIs we are moving to in this PR are still considered "experimental" - AKA the V2 APIs. They are not the canonical bacalhau APIs we expected user to be interacting with. We should be striving to get them right before moving to them. 

> My preference is to delay any decision to change our API specs and have different ways to submit jobs until we collect more feedback about current APIs, and have more dedicated time to explore alternatives and align on the right option.

Given the above, I strongly disagree with this point. We shouldn't be putting off improvements to the new APIs we are moving to. The whole point of creating these new API was to make improvements to it before they become canonical. As you have already alluded to, and outlined in greater detail via https://github.com/bacalhau-project/bacalhau/issues/3983, separating the system representation of a job from the user submission is an excellent improvement - both from a development perspective and user perspective.
We have made it a habit of encouraging the development team to make improvements to the V2 APIs before we make them the default, let's ensure we allow that before moving forward.

> Again, I truly don't recommend figuring this out just yet.

Just the opposite: I would recommend we figure this out now, since if we don't, we'll have to introduce yet another API - V3 - as changing the payload and responses expected by the V2 API will be a breaking change. 

> A low hanging fruit is to introduce SanitizeForSubmission() method under models.Job that will clear out all fields that are reserved by the system so that something like bacalhau job describe <job_id> --output yaml | bacalhau job run would work

I believe this confuses low hanging fruit "accept technical debt", and as I've stated above, given these are still experimental APIs we should be making the effort to implement them correctly, rather than postpone valuable improvements. 

> have more dedicated time to explore alternatives and align on the right option

The change here is easier than we think. Initially we can keep it small by simply changing the request payload expected by the APIs to a subset of the Job spec as described in https://github.com/bacalhau-project/bacalhau/issues/3983. We don't need to fully separate the spec from the state across the system as a whole until later. And said change can be made without affecting the API schema later.

cmd/cli/docker/docker_run.go
frrist: My comment here was with respect to the priority fields. Are we okay adding a flag called `--priority` that sets their field?

With respect to the job type, currently controlled via `--target`:
Would we be willing to drop the target flag in favor of a `--type` flag whose default value is `batch` and other possible values are `service`, `ops`, and `daemon`? This would bring the CLI more inline with the fields of the job spec.

cmd/util/flags/cliflags/spec.go
frrist: For context here @aronchick: This comment is about a change in behavior of the output flag from expecting `:` as a separator to `=` as a separator.

When using a flag that requires pairs of strings (key,value) - for example environment variables - and in the case here ResultPaths I find a `StringToStringVar` flag type (a map) to provide clearer validation compared to accepting a `StringArrayVar` flag (a slice). As an example, consider:

**Valid**:
`bacalhau docker run -o outputs=/outputs -o inputs=/inputs`

**Invalid**:
`bacalhau docker run -o outputs=/outputs -o inputs`

Cobra, our CLI library, will immediately catch the invalid case and return an error since `inputs` is missing a value field.

Comparatively, the behavior prior to this change accepted both options, and it was on us (developers) to write the validation logic to catch the invalid case of a key missing a value.

If we feel this is too large of a change for our users to understand - replacing `:` with `=` - then I can write a bespoke implementation of a map flag parser that uses `:` as the separator instead of `=` to preserve existing behavior.

cmd/cli/docker/docker_run.go
```diff
@@ -93,156 +84,123 @@ func NewCmd() *cobra.Command {
...
-			return fmt.Errorf("converting job to yaml: %w", err)
+			return err
 		}
-		cmd.Print(string(yamlBytes))
+		cmd.Print(out)
 		return nil
 	}
 
-	executingJob, err := util.ExecuteJob(ctx, j, opts.RunTimeSettings)
+	api := util.GetAPIClientV2(cmd)
```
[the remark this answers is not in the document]
frrist: Yes I can unify the method calls across the docker, wasm, exec, and run commands in this PR. Great way to de-dupe some of this code!

pkg/models/selector.go
```diff
@@ -5,9 +5,10 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/bacalhau-project/bacalhau/pkg/lib/validate"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+
+	"github.com/bacalhau-project/bacalhau/pkg/lib/validate"
 )
```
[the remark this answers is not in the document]
frrist: Great point, here is my current setup in order to import: group stdlib together, group all 3rd party imports together, group all `bacalhau-project` packages together
image

pkg/util/idgen/idgen.go
```diff
@@ -14,6 +14,12 @@ const (
 
 	// NodeIDPrefix is the prefix of node ID.
 	NodeIDPrefix = "n-"
+
+	// TaskNamePrefix is the prefix of a system generated task name.
+	TaskNamePrefix = "t-name-"
+
+	// JobNamePrefix is the prefix of a system generated job name.
+	JobNamePrefix = "j-name-"
```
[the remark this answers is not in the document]
frrist: Yeah, me either, I just needed something to generate a name if one wasn't provided. What is the expected behavior if the user doesn't provide a name? The `bacalhau job run` command will fail, should `docker run` and `wasm run` do the same? Note, this would be a change in their existing behavior.

cmd/cli/docker/docker_run.go
wdbaruni: This was discussed somewhere else and agreed not to change things for now

cmd/cli/docker/docker_run.go
wdbaruni: lets do the migration one step at a time. Lets avoid exposing or renaming flags as this PR's main focus should be internal and to just switch to new APIs.

To answer your questions though, `--priority` is not fully implemented as the compute nodes for example don't respect priority when queueing executions locally in their buffer. Will make sense to expose priority, blog about it and make sure it is documented when it is fully implemented. We should use whatever default value `bacalhau job run` has.

As for `--target`, we don't and should not support service or daemon jobs using imperative job submissions (i.e. `docker run` and `wasm run`). We should allow users to easily update their long running jobs, compare what is deployed in the network and what their local job spec says, and utilize gitops to deploy and update jobs based on committed specs. All that is not feasible with imperative job submissions, and we shouldn't make it easy for users to do bad things to their network.

cmd/util/flags/cliflags/spec.go
wdbaruni: Can we please stop changing behaviour or adding more complexity to the migration story? The migration of the APIs and CLIs is a huge and complex story as is, and we've discussed that no change in behaviour to be done while we migrate to avoid as many issues as possible.

While I appreciate you identifying areas of improvements as you work on this task, the right way to handle this is by cutting issues to track them.

pkg/util/idgen/idgen.go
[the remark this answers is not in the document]
wdbaruni: You are right about job name. This PR should help https://github.com/bacalhau-project/bacalhau/pull/4005
For task, use `main`

frrist:
Scope of this change is larger than expected and there are some breaking changes in the new job spec. Descoping this to a smaller change via https://github.com/bacalhau-project/bacalhau/pull/4020.

PR #3974, opened 2024-05-06
ExecutionTimeout to fail executions instead of jobs

wdbaruni:
This work is a pre-requirement for job queues as the orchestrator is currently failing jobs that take longer than ExecutionTimeout from the time they were submitted. ExecutionTimeout should only fail executions that have been running for too long, and not fail jobs. If a job haven't started and has been in the queue for too long, then this timeout should take effect as there will be other config to timeout of fail a job has been in the queue for too long or has been retried too many times.

A side effect of this PR is that timed out executions will be retried by the scheduler on other nodes until one succeed or no more nodes to retry. This will be handled in next job queue related PRs


pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		// filter terminal executions, and enrich executions with job information
+		for j := range executions {
+			if executions[j].IsTerminalState() {
+				continue
+			}
+			executions[j].Job = job
+			activeExecutions = append(activeExecutions, &executions[j])
+		}
+	}
```
frrist, line 164: While I don't think there is anything wrong here, I am dubious of taking the address of variables in loops, especially nested loops. See https://go.dev/blog/loopvar-preview for details.

Blocking:
 Let's make sure there is test coverage for cases when there are multiple jobs with multiple executions - at present this scenario has no coverage.

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		log.Ctx(h.ctx).Err(err).Msgf("failed to create evaluation %+v", eval)
+		return
+	}
+
+	err = h.evaluationBroker.Enqueue(eval)
+	if err != nil {
+		log.Ctx(h.ctx).Err(err).Msgf("failed to enqueue evaluation %+v", eval)
+		return
+	}
```
frrist, line 216: I would prefer we push these error up the stack to the root caller `runHousekeepingTasks` and deal with them consistently in a single location.

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+	waitGroupDone := make(chan struct{})
+	go func() {
+		h.waitGroup.Wait()
+		close(waitGroupDone)
+	}()
+
+	select {
+	case <-waitGroupDone:
+	case <-ctx.Done():
+		h.waitGroup.Done()
```
frrist, line 105: Blocking:
This isn't guaranteed to cause the above go routine to exit since several active go routines in `timeoutExecutions` may have added to it; `Done` only decrements the counter by 1. 
Further, this code may panic if the current count of the wait group is 0 and `Done` is called since the wait group panics if the count goes below zero: https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/sync/waitgroup.go;l=32

pkg/orchestrator/housekeeping_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+const (
+	mockJobTimeout                 = 60              // 1 minute
+	timeoutBuffer                  = 1 * time.Minute // another minute. only executions older than 2 minutes are expired
+	notExpiredModifyTime           = -30 * time.Second
+	expiredWithinBufferModifyTime  = -90 * time.Second
+	expiredOutsideBufferModifyTime = -120 * time.Second
+)
+
+type HousekeepingTestSuite struct {
```
frrist, line 25: Strongly prefer we test things involving time with github.com/benbjohnson/clock as we have done elsewhere in the code base. This way we can avoid depending on time in tests, preventing flakes, and also improve how tests behave when using a debugger: Currently these tests will fail if a debugger is used to inspect them as the timeouts are out of our control.

pkg/orchestrator/scheduler/batch_job_test.go
```diff
@@ -303,6 +304,41 @@ func (s *BatchJobSchedulerTestSuite) TestProcess_ShouldMarkJobAsFailed_NoRetry()
...
 }
 
+func (s *BatchJobSchedulerTestSuite) TestProcess_ShouldStopExpiredExecutions() {
+	ctx := context.Background()
+	job, executions, evaluation := mockJob()
+	// Set the job type to batch and the timeout to 60 minutes
+	job.Task().Timeouts.ExecutionTimeout = int64((60 * time.Minute).Seconds())
+	// Set the start time of the executions to exceed the timeout
+	for i := range executions {
+		executions[i].ModifyTime = time.Now().Add(-90 * time.Minute).UnixNano()
```
frrist, line 314: further motivation to adopt github.com/benbjohnson/clock in our testing.

sample_config.yaml
frrist, line 1: delete

pkg/orchestrator/housekeeping_test.go
```diff
@@ -0,0 +1,343 @@
...
+	var tests = []struct {
+		name                 string
+		ModifyTimes          []time.Time
+		expectedEnqueueCount int
+		jobCount             int
+		jobType              string
+		executionState       models.ExecutionStateType
+	}{
+		{
+			name: "TestExpiredExecutionOutsideBuffer",
```
frrist, line 64: nit: including 'Test' in the name is redundant.

frrist (changes requested):
Three major points of concern:
- Parallelism in Housekeeping has several issues, one of which the potential for panicking. Suggest adopting the workerpool package I mentioned instead of implementing our own.
- Don't embed `context.Context` in the housekeeping structure.
- We need to be mocking time in our tests, especially as this area of the code base grows.

pkg/models/event.go
```diff
@@ -57,7 +57,12 @@ func (e *Event) WithMessage(message string) *Event {
...
 func (e *Event) WithError(err error) *Event {
-	e.Message = err.Error()
+	return e.WithErrorMessage(err.Error())
+}
+
+// WithErrorMessage returns a new Event with the given error.
+func (e *Event) WithErrorMessage(message string) *Event {
+	e.Message = message
 	return e.WithDetail(DetailsKeyIsError, "true")
 }
```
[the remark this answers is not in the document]
wdbaruni: Not enough. `WithMessage` accepts a string and doesn't' flag the event as `DetailsKeyIsError`. `WithError` accepts an error type. The alternatives of not adding `WithErrorMessage` are:
- Changing `WithError` to accept a string instead of error type. This doesn't sound right
- Construct an error type of a string to pass it to `WithError`. Doesn't sound right for callers to construct errors only to pass them to the event builder and not return the error!
- Calling `WithError` and then manually setting `DetailsKeyIsError` in the caller. We will have more similar cases and I prefer to simplify the creation of events and have one place for common setup of events

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+
+type Housekeeping struct {
+	evaluationBroker EvaluationBroker
+	jobStore         jobstore.Store
+	interval         time.Duration
+	timeoutBuffer    time.Duration
+
+	workersSem chan struct{}
+	waitGroup  sync.WaitGroup
+	ctx        context.Context
```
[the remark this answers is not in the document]
wdbaruni: Done in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/851b0690b9aba4c6cc577d2baf612c932fab40a0

pkg/orchestrator/events.go
```diff
@@ -1,14 +1,16 @@
...
 
 	"github.com/bacalhau-project/bacalhau/pkg/models"
 )
 
 const (
-	EventTopicJobSubmission models.EventTopic = "Submission"
-	EventTopicJobScheduling models.EventTopic = "Scheduling"
+	EventTopicJobSubmission    models.EventTopic = "Submission"
+	EventTopicJobScheduling    models.EventTopic = "Scheduling"
+	EventTopicExecutionTimeout models.EventTopic = "Exec Timeout"
```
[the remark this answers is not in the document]
wdbaruni: Topics main use is improving readability of events and let users understand which component or lifecycle or job orchestration the event is related. Not an internal detail and not intended to be consumed to other applications. 

As the goal is improve human readability, allowing spaces can be improve the user experience and will improve how we display them on the CLI and UI without having to truncate them or wrap them at weird palces

pkg/orchestrator/events.go
```diff
@@ -63,6 +68,14 @@ func ExecStoppedByNodeUnhealthyEvent() models.Event {
 	return event(EventTopicJobScheduling, execStoppedByNodeUnhealthyMessage, map[string]string{})
 }
 
+func ExecStoppedByExecutionTimeoutEvent(timeout time.Duration) models.Event {
+	e := models.NewEvent(EventTopicExecutionTimeout).
+		WithErrorMessage(fmt.Sprintf("%s. Execution took longer than %s", executionTimeoutMessage, timeout)).
```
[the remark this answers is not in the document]
wdbaruni: answered above

pkg/models/execution.go
```diff
@@ -135,6 +135,12 @@ func (e *Execution) GetModifyTime() time.Time {
 	return time.Unix(0, e.ModifyTime).UTC()
 }
 
+// HasExecutionExpired returns true if the execution has been running longer than
+// the provided timeout duration.
+func (e *Execution) HasExecutionExpired(timeout time.Duration) bool {
```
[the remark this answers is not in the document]
wdbaruni: Doen in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/cfb11f8f4bc0cddf7ee4227e6a9d250de00084ee

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+	// validate params
+	err := errors.Join(
+		validate.IsNotNil(params.EvaluationBroker, "evaluation broker cannot be nil"),
+		validate.IsNotNil(params.JobStore, "job store cannot be nil"),
+		validate.IsGreaterThanZero(params.Interval, "interval must be greater than zero"),
+		validate.IsGreaterThanZero(params.Workers, "workers must be greater than zero"),
+		validate.IsGreaterThanZero(params.TimeoutBuffer, "timeout buffer must be greater than zero"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error validating housekeeping params. %w", err)
```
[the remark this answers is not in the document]
wdbaruni: You seem to be using the term "Blocking" very lightly 😅. Maybe worth explaining why

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+	}()
+
+	select {
+	case <-waitGroupDone:
+	case <-ctx.Done():
+		h.waitGroup.Done()
+	}
+}
+
+func (h *Housekeeping) runHousekeepingTasks() {
```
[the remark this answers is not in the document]
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/851b0690b9aba4c6cc577d2baf612c932fab40a0

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+			h.timeoutExecutions(activeExecutions)
+		case <-h.ctx.Done():
+			log.Ctx(h.ctx).Debug().Msg("stopped housekeeping task")
+			return
+		}
+	}
+}
+
+// fetchActiveExecutions fetches all active executions
+func (h *Housekeeping) fetchActiveExecutions() []*models.Execution {
```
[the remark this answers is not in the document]
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/851b0690b9aba4c6cc577d2baf612c932fab40a0

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+	h.startOnce.Do(func() {
+		h.ctx, h.cancel = context.WithCancel(ctx)
+		go h.runHousekeepingTasks()
+	})
+}
+
+func (h *Housekeeping) Stop(ctx context.Context) {
+	if h.cancel != nil {
+		h.cancel()
+	}
```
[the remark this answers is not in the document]
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/851b0690b9aba4c6cc577d2baf612c932fab40a0

pkg/orchestrator/housekeeping.go
```diff
@@ -0,0 +1,219 @@
...
+	return true
+}
+
+// Start starts the housekeeping task
+func (h *Housekeeping) Start(ctx context.Context) {
+	h.startOnce.Do(func() {
+		h.ctx, h.cancel = context.WithCancel(ctx)
+		go h.runHousekeepingTasks()
+	})
+}
```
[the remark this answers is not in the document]
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/851b0690b9aba4c6cc577d2baf612c932fab40a0

pkg/orchestrator/housekeeping.go
wdbaruni: I am happy they are fixing this soon. In my case, I am looping over indexes and fetching the reference from the original array to avoid failing into the issue of referencing the wrong pointer

Added test in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/4fff4d01760ea9be8c1a03e8ef318a2b792af16b

pkg/orchestrator/housekeeping.go
wdbaruni: Good catch. That shouldn't be there. Removed in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/4527fcdac6e33cd4e08cee9f926e03318dc36ade

pkg/orchestrator/housekeeping_test.go
wdbaruni: renamed in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/7a0b02e08cfa7910311c7db16b763fd962d1c68f

pkg/orchestrator/housekeeping.go
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/f380e99ba72b60e64aef96166efff56967bc73b1

pkg/orchestrator/housekeeping_test.go
wdbaruni: I am a big fan of controlling the clock for testing purposes. Though in this case it will require that we pass the clock to `Execution.IsExpired(...)` which I am not a fan of. Still, the biggest advantage of mocking the clock is to have more predictable test scenarios and to avoid having to sleep here and there to mimic some scenarios. This is not the case here as I am explicitly setting the executions time. I am only using the wall clock to check if those times have expired or not. 

I agree that if you put a break point in very specific places the tests will fail, but I don't see this as a problem. 

In summary, I like clocks, but for this specific test suite I don't see they add significant value and justify  passing the clock to `Execution.IsExpired(...)`

pkg/orchestrator/scheduler/batch_job_test.go
wdbaruni: answered above

pkg/models/event.go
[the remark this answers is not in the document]
frrist: My concern is DevX as this change obscures usages of this package.  Pass WithError an error as I suggested below. We are returning the event that encapsulates the error to the caller so I'll good there.

wdbaruni:
> Parallelism in Housekeeping has several issues, one of which the potential for panicking. Suggest adopting the workerpool package I mentioned instead of implementing our own.

Not sure about the several issues. You pointed to one issue with a wrong call to `waitGroup.Done()` and that was fixed. You forgot to mention the workerpool package, but the semaphore seem to be working fine. The other requirement I have is I don't want the housekeeper to start another iteration before all tasks are completed, and thats why I am using the `waitGroup`

> Don't embed context.Context in the housekeeping structure.

Fixed

> We need to be mocking time in our tests, especially as this area of the code base grows.

Answered inline

pkg/orchestrator/housekeeping_test.go
frrist: > Though in this case it will require that we pass the clock to Execution.IsExpired(...) which I am not a fan of

Or we could pass the time at which the execution ought to expire instead of a duration as we do now. Actually I might like that better as it removes time from the model. Wdyt?

frrist:
Here's the worker pool package: https://github.com/gammazero/workerpool

The several issues were embedded context in structure, and the potential for panic with the wait group. The look to be addressed now will give another review in a sec.

pkg/orchestrator/events.go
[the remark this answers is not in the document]
wdbaruni: removed in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/918ce4e965be0318f11c1558335d2852d15aff40

pkg/orchestrator/scheduler/batch_job_test.go
wdbaruni: using clocks in https://github.com/bacalhau-project/bacalhau/pull/3974/commits/da4a3ae81130b423b822fcbdf67ed7d97d1e10cf

wdbaruni:
I've previously looked into the worker pool library you mentioned, and it doesn't enable waiting for housekeeping tasks before calling the iteration as complete and starting another one. Again our requirements are pretty simple and I don't see a need for such a library at this point. The semaphore is limiting the number of concurrent tasks, which is what the workerpool library does, and the waitGroup is preventing iterations from overlapping with each other, which the workerpool library is missing.

frrist (approved):

Issue #3983, opened 2024-05-08
Separate Job spec from system state


frrist:
At present the system representation of a job and user submitted data are combined together into a single structure, the Job type.

## Problem

It feels like we are making this easy for ourselves, instead of ours users. Or what's more likely - making this hard for BOTH ourselves and users. At present we are munging together user submitted data and system data into multiple structures - the job and its tasks. Further we are making modifications to the data submitted by users depending on various edge cases. Lastly, we are inefficiently storing job state when persisting to disk in the jobstore. 

### Ways we are making it harder for ourselves:
- There are several different methods on the job type that must be called to validate and sanitize it. Some of these methods must only be called on the client side, while others must only be called on the server side. This is poor DevEx.
  - Some of these methods are only needed in the first place because we are munging together system representation and user data. e.g. [`SanitizeSubmission`](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/job.go#L274) only exists because we allow users to submit system fields in the first place.
- The meta field of a job has various "reserved" keys that cannot be set by users, if they are we delete them: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/job.go#L307. This is both poor DevEx and UX.
- We only disallow user submissions with the JobID set if the job type is batch or ops. system and daemon jobs may provide their own jobID on submission: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/job.go#L295-L300. 
  - If this is the intended behavior, and I bet its not, that's really confusing UX. If it's not the intended behavior, its a bug serving as evidence we are making this hard for ourselves: We should prevent this field from even being an option for users to provide.
- We inefficiency store the entire job structure, and its tasks(!!), repeatedly to disk when only modifying a small portion of the structure: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/jobstore/boltdb/store.go#L848-L852. The State of a job doesn't need to be in the same structure as the users submission. It would be significantly more efficient to store the user submission _once_ and make updates to system state representation separately.
- Whenever the time comes to make changes to the system state representation of the job or task (e.g. add a new field, remove a field, modify a fields type, etc.) we will need to modify everything from the client to the server to accommodate it. Ohh and let's not forget the various SDKs!

### Ways we are making it hard for our users
[...]

PR #3985, opened 2024-05-08
refactor: implement JobSpec type used for submitting job to V2 api

frrist:
This change implements the `JobSpec` type which contains only the fields of a Job users are permitted to set when submitting a job to run. This change modifies the `PutJobRequest` to use `JobSpec` as the new payload. 
- progress towards #3983


frrist:
Based on conversation in https://github.com/bacalhau-project/bacalhau/issues/3983 this is an extremely controversial change and therefore cannot be accepted.

Issue #3993, opened 2024-05-13
Atomic job store transactions


wdbaruni:
Today we run job store operations in their isolation transaction, which prevents us from doing atomic transactions that span multiple operations, such as handling a single scheduler plan that can involve updating the job state, updating the state of multiple executions, enqueueing delayed evaluations and adding history events.

We need way to start a single transaction and have all these operations to be committed atomically or rolled-back

PR #3996, opened 2024-05-13
add atomic job store transactions

wdbaruni:
This PR allows executing multiple job store write operations in a single atomic transaction. This done by optionally providing a transactional context to the operation, which will prevent automatic commit after each operation and delegates that control to the caller.

#### A transaction per operation
This example is for the same behaviour we have today where each operation is committed automatically, which will continue to work
```
	ctx := context.Background()
	var err error

	// The job will be updated immediately if no error was returned
	if err = s.store.UpdateJobState(ctx, jobstore.UpdateJobStateRequest{}); err != nil {
		return err
	}

	// The execution will be updated immediately if no error was returned
	if err = s.store.UpdateExecution(ctx, jobstore.UpdateExecutionRequest{}); err != nil {
		return err
	}
```

#### A single transaction for both operations
This is an example of the new behaviour where we can group multiple operations in a single atomic operation 
```
	var err error
	txCtx, err := s.store.BeginTx(context.Background())
	if err != nil {
		return err
	}

	// make sure we rollback the transaction if there is an error
	defer func() {
		if err != nil {
			_ = txCtx.Rollback()
		}
	}()

	// updating the job won't be committed yet
	// Note we are passing the transaction context txCtx instead of a regular context
	if err = s.store.UpdateJobState(txCtx, jobstore.UpdateJobStateRequest{}); err != nil {
		return err
	}

	// updating the execution won't be committed yet
	if err = s.store.UpdateExecution(txCtx, jobstore.UpdateExecutionRequest{}); err != nil {
		return err
	}

	// commit both operations
	return txCtx.Commit()
```

Closes [3993](https://github.com/bacalhau-project/bacalhau/issues/3993)


pkg/orchestrator/planner/state_updater.go
```diff
@@ -25,20 +26,36 @@ func NewStateUpdater(store jobstore.Store) *StateUpdater {
...
+	txContext, err := s.store.BeginTx(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+
+	defer func() {
+		if err != nil {
+			_ = txContext.Rollback()
+		}
+	}()
```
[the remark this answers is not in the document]
wdbaruni: I hear your concerns about the general recommendation of not passing values in context, but we need to think about the bigger picture, understand why it is an anti-pattern, whether it applies here, and the tradeoffs of insisting not using transactional contexts

First, the transactional context is telling operations that they are within a scope of a transaction and not to create their own. They are well scoped, short lived and things won't panic if we call operations with a regular context, as it means they are within their own transaction scope. They are not a way to pass optional parameters as you pointed, but to tell the scope we are in.

Second, it is the least intrusive change to our code base to introduce transactions only when needed. All existing reads and writes are still functional. This is a priority to keep in mind when we do changes given our team size, and higher priority items we need to tackle. We need to think how can we find simple solutions to complex problems

Third, the alternatives are not as simple as you might think and will require major refactoring:
- A closure means all our jobstore operations will require a transaction as parameter. This includes read operations as well to enable reads inside a write transaction by reusing the same transaction (https://github.com/bacalhau-project/bacalhau/pull/3996/commits/1ff510fc6ef37c13c4a88c079d16f24970cecb8c) . This means that even the most common `GetJob` API will require a transaction that must be manually started and rolled-back, or use a closure, which won't make much of difference for single operations. 
- A transaction type means all our operations, both writes and reads, will need to move from jobstore interface to this new transaction type. Again we would have to create this type first, and manually commit/rollback even for simple `GetJob` operations

Fourth, not only the suggestions require major refactoring today, I am not sure if they are on the right path as they  will impact our developer experience 

Fifth, jobstore is an internal type and not something that external users will need to interact with. We don't need to be as worried as how people will abuse it.

Lastly, if in the future we see the need to set explicit transactions in all our operations, then we can always do that change when needed. We are not taking a one way door decision now. I am not doing major refactors today, and will be smarter to delay major refactors when they are actually needed and justifies the impact on developer experience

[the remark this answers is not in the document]
frrist: I would prefer a functional options pattern like the below using the option `jobstore.WithBoltTransaction`
```go
func (s *StateUpdater) Process(ctx context.Context, plan *models.Plan) (err error) {
	// ...

	txContext, err := s.store.BeginTx()
	if err != nil {
		return fmt.Errorf("failed to begin transaction: %w", err)
	}

	defer func() {
		if err != nil {
			err = txContext.Rollback()
		}
	}()

	// Create new executions
	for _, exec := range plan.NewExecutions {
		err = s.store.CreateExecution(ctx, *exec, plan.Event, jobstore.WithBoltTransaction(txContext))
		if err != nil {
			return err
		}
	}

	// Update existing executions
	for _, u := range plan.UpdatedExecutions {
		err = s.store.UpdateExecution(ctx, jobstore.UpdateExecutionRequest{
			ExecutionID: u.Execution.ID,
			NewValues: models.Execution{
				DesiredState: models.State[models.ExecutionDesiredStateType]{
					StateType: u.DesiredState,
					Message:   u.Event.Message,
				},
			},
			Condition: jobstore.UpdateExecutionCondition{
				ExpectedRevision: u.Execution.Revision,
			},
			Event: u.Event,
		}, jobstore.WithBoltTransaction(txContext))
		if err != nil {
			return err
		}
	}

	// Update job state if necessary
	if !plan.DesiredJobState.IsUndefined() {
		err = s.store.UpdateJobState(ctx, jobstore.UpdateJobStateRequest{
			JobID:    plan.Job.ID,
			NewState: plan.DesiredJobState,
			Event:    plan.Event,
			Condition: jobstore.UpdateJobCondition{
				ExpectedRevision: plan.Job.Revision,
			},
		}, jobstore.WithBoltTransaction(txContext))
		if err != nil {
			return err
		}
	}
	return txContext.Commit()
}
```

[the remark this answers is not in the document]
wdbaruni: I am still insisting that utilizing context for this purpose is acceptable and very understandable for all the reasons I've mentioned previously.

Using functional options for every database operation is an overkill for this purpose, confusing, and I don't see how they are improving things more than a scoped context value:
1. Multiple database operations already accept sort of options today, such as `JobQuery` to set limits and offsets, `GetExecutionsOptions` to include job, `JobHistoryFilterOptions` to set start time and filter execution or job level events, and `UpdateJobCondition` to set conditions. We are using struct types to define those parameters as they map easily to our HTTP APIs, clearer what each API expects, and we don't need a great level of flexibility for those calls. It will be very confusing to provide additional types of options to configure the calls and optionally allowing passing transactions
2. Functional patterns are best when wanting a great level of flexibility, but not as readable or obvious what options are available. We don't need this level of flexibility here and I don't foresee the need for more options in the future other than just the transaction. Again we are not taking one way door decision here as the jobstore is an internal library, and we can add functional options in the future _if_ there is a need for more options


Lets not get stuck at anti-patterns without thinking about the bigger picture or the impact, which I've shared in my previous answer. This PR is blocking my progress for job queues. If you don't have any other comment, I would like us to move forward and we can always reevaluate when we need more flexible options or realize most of our calls span a single transaction

pkg/jobstore/boltdb/context.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/bacalhau-project/bacalhau/pkg/jobstore"
+	bolt "go.etcd.io/bbolt"
+)
+
+// contextKey is a custom type to avoid key collisions in context values.
+type contextKey int
+
+// txContextKey is the key used to store the transaction context in the context.
+const txContextKey contextKey = 0
```
frrist, line 14:
```suggestion
// contextKey is a custom type to avoid key collisions in context values.
type contextKey struct{}

// txContextKey is the key used to store the transaction context in the context.
var txContextKey = contextKey{}
```
> WithValue returns a copy of parent in which the value associated with key is val.
> Use context Values only for request-scoped data that transits processes and APIs, **not for passing optional parameters to functions**.
> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys. To avoid allocating when assigning to an interface{}, context keys often have concrete type struct{}. Alternatively, exported context key variables' static type should be a pointer or interface. 

https://pkg.go.dev/context#WithValue

pkg/jobstore/boltdb/context.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	// Start a goroutine that listens for the context's Done channel.
+	go func() {
+		<-innerCtx.Done()
+		// If the transaction has not been committed when the context is done, rollback the transaction.
+		// In boltdb transaction, db is nil if the transaction is already committed or rolled back.
+		if txCtx.tx.DB() != nil {
+			_ = txCtx.tx.Rollback()
+		}
+	}()
```
frrist, line 40: Few concerns:
1. Don't ignore the error if the rollback fails, at least log it.
2. When `txContext.Commit()/Rollback()` is called we commit/rollback the transaction, then cancel the inner context, if the DB is not nil we then rollback the transaction. Since boltdb's transaction implementation closes the connection on commit or rollback this is fine - But we've tightly coupled the logic here to the internals of boltdb, which is unfortunate and unavoidable given this approach. 
3. When the parent context is canceled we rollback the transaction, but depending on how this is used we may call `Commit` or `Rollback` again later which will result in `ErrTxClosed` errors. Again unfortunate but unavoidable.

Lets address 1 with a code change, and make 2-3 explicitly clear in comments

pkg/jobstore/boltdb/context_test.go
```diff
@@ -0,0 +1,97 @@
+//go:build unit || !integration
```
frrist, line 1: Is it worth testing the case where a context contains more than a single transaction? Or will that never happen?

pkg/jobstore/boltdb/store.go
```diff
@@ -690,11 +699,94 @@ func (b *BoltJobStore) CreateJob(ctx context.Context, job models.Job, event mode
...
+		tx, err = b.database.Begin(true)
+		if err != nil {
+			return err
+		}
+	}
+
+	// always rollback the transaction if we did create it and we're returning an error
+	defer func() {
+		if !managed && err != nil {
+			_ = tx.Rollback()
```
frrist, line 734: should we return this error to the caller?

frrist, line 734: If we rollback the transaction here, then also rollback again it callers like `state_updater.go` is that alright?

pkg/jobstore/boltdb/store.go
```diff
@@ -690,11 +699,94 @@ func (b *BoltJobStore) CreateJob(ctx context.Context, job models.Job, event mode
...
+
+	err = view(tx)
+	if err != nil {
+		return err
+	}
+
+	// if we created the transaction, then we need to commit it
+	if !managed {
+		err = tx.Rollback()
+	}
```
frrist, line 786: Is this comment wrong, or is the code incorrect?

pkg/jobstore/context.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	return t.TxContext.Commit()
+}
+
+func (t TracingContext) Rollback() error {
+	if time.Since(t.startTime) > maxTracingDuration {
+		log.Debug().Msgf("transaction took longer than %s to rollback", maxTracingDuration.String())
+	}
+	return t.TxContext.Rollback()
+}
```
frrist, line 38: If I see these log lines, how will I know which transaction is the culprit? I think all I will see is:
`DEBUG transaction took longer than 10ms to rollback` which isn't explicit enough to be helpful.
Please remove this wrapper and instead opt for actionable log line in the job store on its respective methods.

pkg/jobstore/boltdb/context.go
wdbaruni: Yeah I am aware of the recommendation to not use built-in types to avoid collisions, and thats why I introduced a custom `contextKey` type. It doesn't have to be a struct
https://go.dev/blog/context#TOC_3.2.:~:text=Print(err)%0A%20%20%20%20%20%20%20%20return%0A%20%20%20%20%7D-,Package%20userip,-The%20userip%20package

Playground verification
https://go.dev/play/p/py6iZ6uL1BK

pkg/jobstore/boltdb/context.go
wdbaruni: improved things here https://github.com/bacalhau-project/bacalhau/pull/3996/commits/2aa9913b48bd13efd934d5af2fe311119b45f5ce, but I've also realized boltdb transactions are not thread safe and fixed it here https://github.com/bacalhau-project/bacalhau/pull/3996/commits/99eec412c1fc12a7481514ebc488a6413f602981

pkg/jobstore/boltdb/context_test.go
wdbaruni: boltdb doesn't allow concurrent active transactions. If you attempt to begin a transaction while another one is still active, whether in the same context or note, it will block. I've added additional tests in https://github.com/bacalhau-project/bacalhau/pull/3996/commits/b0c7060a3e822fbf2cc59eabfaa42fa286882c4d for botldb jobstore to verify we are handling this gracefully

pkg/jobstore/boltdb/store.go
wdbaruni: we should only return the original error that caused the transaction to fail, not the rollback error. Think of rollback as a best effort here as the issue that failed the transaction can also fail the rollback. Even logging rollback errors here can be noisy and wouldn't added value

pkg/jobstore/boltdb/store.go
wdbaruni: The comment is wrong as we don't commit readonly transactions, just rollback them to release the lock. Fixed in https://github.com/bacalhau-project/bacalhau/pull/3996/commits/f2180af41a362b9bcc94050facc9838a48f25651

pkg/jobstore/context.go
wdbaruni: The log lines are useful in knowing something is wrong that requires investigation. They are not as great as metrics or actual traces, but this is a cheap and quick solution to call for investigation until we have better ways for tracing. These logs are specially useful now as we've just introduced multi-op transaction and can be removed after some time.

Log lines in in the job store methods are not useful as the they are not the ones holding the transaction. The `StateUpdaterPlanner` for example is a transaction holder and would be a bug if it is doing many slow operations while holding the transaction.

For now I've added the context in the log statement, and we can always add more logging contexts in the future to identify the source operation https://github.com/bacalhau-project/bacalhau/pull/3996/commits/33c2a51fd072486c37813f19388fa04b7793d07c

pkg/jobstore/boltdb/store.go
wdbaruni: similar to https://github.com/bacalhau-project/bacalhau/pull/3996#discussion_r1604832916

pkg/jobstore/boltdb/context_test.go
frrist: I believe boltdb supports concurrent read-only transactions: https://github.com/boltdb/bolt?tab=readme-ov-file#transactions

pkg/jobstore/boltdb/context.go
frrist: fair enough. It may be worth maintaining consistency with other usages of contextKeys ([1](https://github.com/bacalhau-project/bacalhau/blob/fb9a5461b016e6016e4c31069bf08631b398b59f/cmd/util/cleanup.go#L9), [2](https://github.com/bacalhau-project/bacalhau/blob/fb9a5461b016e6016e4c31069bf08631b398b59f/cmd/cli/root.go#L192)) in the codebase, but will leave it to you.

frrist (approved):
Will let you decide if https://github.com/bacalhau-project/bacalhau/pull/3996/files#r1605288596 is worth addressing.

PR #3997, opened 2024-05-14
adding update checker test env variable

aronchick:


pkg/version/update.go
```diff
@@ -61,6 +61,10 @@ func CheckForUpdate(
 	}
 	q.Set("clientID", clientID)
 	q.Set("InstallationID", InstallationID)
+	if os.Getenv("BACALHAU_UPDATE_CHECKER_TEST") != "" {
+		q.Set("bacalhau_update_checker_test", "true")
+	}
+
```
wdbaruni, line 67: What will this do? Might be worth adding a comment here

aronchick: will do

pkg/version/update.go
```diff
@@ -61,6 +61,14 @@ func CheckForUpdate(
 	}
 	q.Set("clientID", clientID)
 	q.Set("InstallationID", InstallationID)
+
+	// The BACALHAU_UPDATE_CHECKER_TEST is an env variable a user can set so that we can track
+	// when the binary is being run by a non-user, to enable easier filtering of queries
+	// to their update server for internal/CI.
```
frrist, line 67: Can I assume this has some relation to: https://github.com/bacalhau-project/getbacalhauorg/pull/29 ?

> when the binary is being run by a non-user

What does this mean? Like a linux user or a person who uses bacalhau?

aronchick: person who uses / deploys / etc bacalhau.

Basically just let users set (if they want) a key indicating this run is "internal".

PR #4002, opened 2024-05-15
compute nodes no longer reject bids based on queued capacity

### LINKED ISSUE #3992: No bid rejection for transient reasons. E.g. no available capacity
To simplify our scheduling logic and minimize the edge cases and decision branches, I would like us to move to a state where bid rejections by compute nodes are only for terminal reasons and not for transient reasons, such as no available capacity. Compute nodes already queue jobs locally if they don't have immediate capacity, they can share both the available and queued capacity with the orchestrator, and the orchestrator can make a decision on ranking nodes based on those capacities and can filter compute nodes with a large queue and retry later. This enables higher priority jobs to be scheduled in compute nodes with large queue of pending jobs, but get the chance to run sooner if we use priority queues in the executor_buffer

Even if the Orchestrator's view of the network's utilization is stale, we have solutions that can mitigation that including:
1. Use Sparrow scheduling where the orchestrator will select multiple compute nodes, and they only accept when the job reaches the head of the of the queue and ready to be executed https://github.com/bacalhau-project/bacalhau/issues/3856
2. Add a Housekeeping task that tracks executions taking too long in negotiation phase and enqueue an evaluation. The scheduler might decide to pick another node if there are better ranked ones or keep the same node otherwise

wdbaruni:
This PR not only closes #3992, but also adds a node ranker based on available and queued capacity, which we didn't have before. I had to introduce the ranker along with this PR to make sure some tests are passing that were relying on bid rejection due to exceeding resource limits, specifically `TestParallelGPU`


pkg/orchestrator/selection/ranking/available_capacity.go
```diff
@@ -0,0 +1,151 @@
...
+		ranks[i] = orchestrator.NodeRank{
+			NodeInfo: node,
+			Rank:     int(rank),
+			Reason: fmt.Sprintf(
+				"Ranked based on available capacity %s and queue capacity %s",
+				node.ComputeNodeInfo.AvailableCapacity.String(), node.ComputeNodeInfo.QueueUsedCapacity.String()),
+			Retryable: true,
+		}
+		log.Ctx(ctx).Trace().Object("Rank", ranks[i]).Msg("Ranked node")
+	}
```
frrist, line 148: For readability these two loops could be separate functions, but I don't feel strongly about it. your choice.

pkg/orchestrator/selection/ranking/available_capacity.go
```diff
@@ -0,0 +1,151 @@
...
+		// Assign rank and reason to the node
+		ranks[i] = orchestrator.NodeRank{
+			NodeInfo: node,
+			Rank:     int(rank),
+			Reason: fmt.Sprintf(
+				"Ranked based on available capacity %s and queue capacity %s",
+				node.ComputeNodeInfo.AvailableCapacity.String(), node.ComputeNodeInfo.QueueUsedCapacity.String()),
+			Retryable: true,
+		}
+		log.Ctx(ctx).Trace().Object("Rank", ranks[i]).Msg("Ranked node")
```
frrist, line 147: Can we make this a debug log? Or add a log outside of this loop that prints the nodes in ranked order?

pkg/orchestrator/selection/ranking/available_capacity_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+type rankNodesTestCase struct {
+	name         string
+	nodes        []nodeScenario
+	jobResources models.ResourcesConfig
+	expected     []string
+	equalCheck   bool
+}
+
+func (suite *AvailableCapacityNodeRankerSuite) TestRankNodes() {
```
frrist, line 38: I'd find this test _slightly_ more readable if the node resources were scaled from a common value, e.g.
But don't feel strongly enough to push for change
```go
bc := models.Resources{CPU: 4, Memory: 16000, Disk: 200000, GPU: 1}
bq :=  models.Resources{CPU: 1, Memory: 4000, Disk: 100000, GPU: 0},
nodes: []nodeScenario{
	{
		nodeID:            "node1",
		availableCapacity: bc,
		queueUsedCapacity: bq,
	},
	{
		nodeID:            "node2",
		availableCapacity: models.Resources{CPU: bc.CPU*2, Memory: bc.Memory*2, Disk: bc.Disk/2, GPU: bc.GPU*2},
		queueUsedCapacity: models.Resources{CPU: bq.CPU/2, Memory: bq.Memory/2, Disk: bc.Disk/2, GPU: bc.GPU},
	},
	{
		nodeID:            "node3",
		availableCapacity: models.Resources{CPU: bc.CPU/2, Memory: bc.Memory, Disk: bc.Disk-5000, GPU: bc.GPU*4},
		queueUsedCapacity: models.Resources{CPU: bq.CPU/2, Memory: bq.Memory/4, Disk: bq.Disk/10, GPU: bq.GPU*2},
	},
},
```

pkg/orchestrator/selection/ranking/available_capacity.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return ranks, nil
```
frrist, line 150: Let's return the ranks in sorted order, will remove one step from the tests and make it easier for humans to parse log lines.

pkg/orchestrator/selection/ranking/available_capacity.go
```diff
@@ -0,0 +1,151 @@
...
+// AvailableCapacityNodeRanker ranks nodes based on their available capacity and queue used capacity.
+type AvailableCapacityNodeRanker struct{}
+
+// NewAvailableCapacityNodeRanker creates a new instance of AvailableCapacityNodeRanker.
+func NewAvailableCapacityNodeRanker() *AvailableCapacityNodeRanker {
+	return &AvailableCapacityNodeRanker{}
+}
+
+// dynamicWeights calculates the weights for resources based on the job requirements.
+func dynamicWeights(jobRequirements *models.Resources) (float64, float64, float64, float64) {
```
frrist, line 43: probably worth using named returns here, or encapsulate the result in a struct.

pkg/orchestrator/selection/ranking/available_capacity.go
```diff
@@ -0,0 +1,151 @@
...
+	"fmt"
+
+	"github.com/bacalhau-project/bacalhau/pkg/lib/math"
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+	"github.com/bacalhau-project/bacalhau/pkg/orchestrator"
+	"github.com/rs/zerolog/log"
+)
+
+// Constants to normalize resource values to a comparable scale.
+const (
```
frrist, line 14: passing thought: These constants are great candidates for system configuration values.

frrist (approved):
LGTM. Consider addressing a few of the comments.

pkg/compute/bidder.go
```diff
@@ -274,14 +274,16 @@ func (b Bidder) runSemanticBidding(
...
+		if err != nil || !resp.ShouldBid {
+			log.Ctx(ctx).WithLevel(logger.ErrOrDebug(err)).
+				Err(err).
+				Str("Job", request.Job.ID).
+				Str("Strategy", strategyType).
+				Bool("Bid", resp.ShouldBid).
+				Bool("Wait", resp.ShouldWait).
+				Str("Reason", resp.Reason).
+				Send()
+		}
```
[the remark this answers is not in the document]
wdbaruni: The logs were too noisy. I've changed to only log if the bid was rejected or if an error was raised. Happy cases such as when bids were accepted were too many and too noisy when trying to troubleshoot an issue

pkg/compute/bidder.go
```diff
@@ -335,14 +337,16 @@ func (b Bidder) runResourceBidding(
...
```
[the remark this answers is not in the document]
wdbaruni: same as above

pkg/node/manager/node_manager.go
```diff
@@ -154,11 +159,14 @@ func (n *NodeManager) UpdateResources(ctx context.Context,
 		return &requests.UpdateResourcesResponse{}, nil
 	}
 
-	log.Ctx(ctx).Debug().Msg("updating resources availability for node")
+	log.Ctx(ctx).Debug().Msgf("updating node resources availability: %+v", request)
```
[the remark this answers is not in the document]
wdbaruni: I am also interested in the queued capacity reported in the request

pkg/orchestrator/selection/ranking/available_capacity.go
```diff
@@ -0,0 +1,151 @@
...
+	// Scaling factors for different resource types to normalize them to a comparable scale.
+	memoryScale = 1e6 // Scale memory to megabytes
+	diskScale   = 1e9 // Scale disk to gigabytes
+	gpuScale    = 1   // No scaling needed for GPU as it's already in units
+	cpuScale    = 1   // No scaling needed for CPU as it's already in cores
```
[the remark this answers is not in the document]
wdbaruni: for consistency with other scales, and to make it clear how we are handling gpu and cpu units

pkg/orchestrator/selection/ranking/available_capacity.go
wdbaruni: Done

pkg/orchestrator/selection/ranking/available_capacity.go
wdbaruni: sorting node ranks is not cheap. We are already sorting in the selector after each ranker provided their own rank, and will be waste to sort again here for debug logging purpose

pkg/orchestrator/selection/ranking/available_capacity.go
wdbaruni: answered in https://github.com/bacalhau-project/bacalhau/pull/4002#discussion_r1605968664

pkg/orchestrator/selection/ranking/available_capacity.go
wdbaruni: done in https://github.com/bacalhau-project/bacalhau/pull/4002/commits/c5fbcb9a37ab61cd9bb19f9e02591fec71ebde1d

pkg/orchestrator/selection/ranking/available_capacity.go
wdbaruni: I would like us to think more of conventions over configurations, and reduce the number of available configurations that are not expected to change, whether we call them user or system configurations. If a value is not expected to change by users or system operators, not shared between packages and only expected to be used by current package, then it should live as a constant in the package itself. This improves the code's readability and make it clear to developers/operators what the values are, and that they are not configurable in different environments.

Even when it comes to testing, we should minimize the need to override those values for testing purposes to make sure we are are testing similar paths to production. But when there is a need to override those values, then we should still keep the default values as constants in the same package.

Issue #3971, opened 2024-05-04
Move `bacalhau get` to `bacalhau job get`


wdbaruni:
today we have `bacalhau get` to fetch results. We v2 CLIs we started to introduce the resource type the command relates to, such as `bacalhau job describe` instead of `bacalhau describe`, and `bacalhau job list` instead of `bacalhau list`

This issue is about introducing `bacalhau job get` that should act similarly to `bacalhau get`

If `bacalhau get` is used print the error message:
```
This command has moved! Please use `job get` to download job results.
```

PR #4008, opened 2024-05-17
added get to job type command

leostelon:
PR contains
1. Add bacalhau job get to download results. #3971 
2. Removed unused parameters from legacy get_test.go.
3. Modified description to download jobs using the "job get" command.


cmd/cli/job/get.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func NewGetOptions() *GetOptions {
+	return &GetOptions{
+		DownloadSettings: cliflags.NewDefaultDownloaderSettings(),
+	}
+}
+
+func NewGetCmd() *cobra.Command {
+	OG := NewGetOptions()
```
wdbaruni, line 44: What would be great is if we can have the existing `get` command call this new implementation to avoid duplicating code. Similar to how we are handling `bacalhau logs` and `bacalhau job logs` https://github.com/bacalhau-project/bacalhau/blob/main/cmd/cli/logs/logs.go

I would also move the tests to this package as our goal is to eventually deprecate `bacalhau get` and only keep `bacalhau job get`

cmd/util/printer/print_legacy.go
```diff
@@ -129,11 +129,11 @@ func PrintJobExecutionLegacy(
 
 	hasResults := slices.ContainsFunc(js.Executions, func(e model.ExecutionState) bool { return e.RunOutput != nil })
 	if !quiet && hasResults {
-		cmd.Printf("\nTo download the results, execute:\n\t"+os.Args[0]+" get %s\n", j.ID())
+		cmd.Printf("\nTo download the results, execute:\n\t"+os.Args[0]+" job get %s\n", j.ID())
 	}
 
 	if !quiet {
-		cmd.Printf("\nTo get more details about the run, execute:\n\t"+os.Args[0]+" describe %s\n", j.ID())
+		cmd.Printf("\nTo get more details about the run, execute:\n\t"+os.Args[0]+" job describe %s\n", j.ID())
```
wdbaruni, line 136: we still want the legacy commands to refer to the legacy `get` and `describe` commands as `bacalhau describe`'s output is different than `bacalhau job describe`

wdbaruni (changes requested):
Thanks Leo for your awesome contribution. Just some minor comments

cmd/cli/job/get.go
leostelon: Can you check now, I have removed the duplicate code.

leostelon:
Can you check now, I have removed the duplicate code.

wdbaruni:
Thanks @leostelon for your contribution. Your changes were merged using a different PR as we have an issue with cricleci that we are trying to fix
https://github.com/bacalhau-project/bacalhau/commit/36a859e1a362fe2fa5c38b7fe49d854c07043415

PR #4009, opened 2024-05-20
run demo network on nats

### LINKED ISSUE #3958: Redeploy Demo network with NATs as the transport layer
Currently the production network is deployed using libp2p as the transport layer. It should instead be deployed using the default, NATS.
Additionally the default config used by bacalhau will need to be modified to include the production networks orchestrator IP address as the default orchestrator compute nodes use when they start up.

wdbaruni:
Closes #3958


frrist (approved):
assuming the deployment wen as expected LGTM. If you didn't deploy from this change then let's not merge until that is done.

wdbaruni:
> assuming the deployment wen as expected LGTM. If you didn't deploy from this change then let's not merge until that is done.

already deployed on Monday

PR #4012, opened 2024-05-20
fix txContext auto-cancellation of context

wdbaruni:
Currently we auto cancel the context when the transaction is committed, which means any operation using that context can fail if it depends on the context's done state. This is causing flakiness in `TestBeginMultipleTransactions_Sequential` test in windows as the second transaction is created from the previous context, which will also be cancelled and rolled-back when the first one is committed and then cancelled.

This PR fixes the issue by no longer cancelling the context of tx commit or rollback

### DIFF
--- a/pkg/jobstore/boltdb/context.go
+++ b/pkg/jobstore/boltdb/context.go
@@ -26,22 +26,20 @@ const txContextKey contextKey = 0
 // TODO: Evaluate the trade-offs and consider delegating the handling of context cancellation to the caller.
 type txContext struct {
 	context.Context
-	tx         *bolt.Tx
-	cancelFunc context.CancelFunc
-	closed     bool
-	closeCh    chan struct{}
-	mu         sync.Mutex
+	tx      *bolt.Tx
+	closed  bool
+	closeCh chan struct{}
+	mu      sync.Mutex
 }
 
 // newTxContext creates a new transactional context for a BoltDB transaction.
 // It embeds a standard context and manages transaction commit/rollback based on the context's lifecycle.
 func newTxContext(ctx context.Context, tx *bolt.Tx) *txContext {
-	innerCtx, cancelFunc := context.WithCancel(context.WithValue(ctx, txContextKey, tx))
+	innerCtx := context.WithValue(ctx, txContextKey, tx)
 	txCtx := &txContext{
-		Context:    innerCtx,
-		tx:         tx,
-		cancelFunc: cancelFunc,
-		closeCh:    make(chan struct{}),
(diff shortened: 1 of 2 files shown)


pkg/jobstore/boltdb/context.go
```diff
@@ -39,14 +41,19 @@ func newTxContext(ctx context.Context, tx *bolt.Tx) *txContext {
...
+		defer func() {
+			// Attempt to rollback the transaction, which is a no-op if already committed or rolled back.
+			if err := txCtx.doRollback(); err != nil {
+				log.Ctx(txCtx).Error().Err(err).Msg("failed to rollback transaction on tx cleanup")
+			}
+		}()
+		select {
+		case <-innerCtx.Done():
+		case <-txCtx.closeCh:
 		}
```
frrist, line 57: We can simplify this by extracting the logic of this deferred function to below the select statement and removing the defer statement.

pkg/jobstore/boltdb/context.go
```diff
@@ -28,6 +28,8 @@ type txContext struct {
 	context.Context
 	tx         *bolt.Tx
 	cancelFunc context.CancelFunc
```
frrist, line 30: should delete if no longer being used.

pkg/jobstore/boltdb/context.go
```diff
@@ -62,28 +69,37 @@ func txFromContext(ctx context.Context) (*bolt.Tx, bool) {
...
 
+// close closes the transactional context.
+// already called with the mutex held.
+func (b *txContext) close() {
+	if !b.closed {
+		close(b.closeCh)
+		b.closed = true
+	}
+}
+
```
frrist, line 103: since this logic should not be called without locking my preference would be to inline the logic of this function into the various places it's used and drop the function.

pkg/jobstore/boltdb/context.go
```diff
@@ -39,14 +41,19 @@ func newTxContext(ctx context.Context, tx *bolt.Tx) *txContext {
 		Context:    innerCtx,
 		tx:         tx,
 		cancelFunc: cancelFunc,
+		closeCh:    make(chan struct{}),
 	}
 	// Start a goroutine that listens for the context's Done channel.
 	go func() {
-		<-innerCtx.Done()
```
frrist, line 45: I think we want to remove the `conext.WithCancel` line above this, right?

frrist (commented):
Can you provide a link to the CI failure that motivates this change?

frrist (approved):
please consider addressing remaining comments. then lgtm

PR #4020, opened 2024-05-22
Migrate docker CLI to API v2

frrist:
# What
- Migrates `docker run` to use the model**s** package and the V2 client for job submission.
- adds several new flags to the docker commands which may be found in `cmd/util/flags/cliflags/job.go` 
- marks some existing flags (`--concurrency`, and `--selector`) as deprecated. 

The deprecation process for flags we plan to remove is as follows:
- Hide the old flag - it no longer appears in help, but can still be used. This keeps things working for existing users, and prevents new users from adopting patterns we are moving away from.
- Mark old flag as deprecated - a warning will be displayed if users use it.
- Implement the new flag definition and wire up to command
- Accept a value from either the new or old flag values when command runs. If the old one was used we convert its value to align with new expectations. Otherwise proceed as normal.
- Ensure a deprecated flag and its replacement cannot be passed simultaneously. 
  - For example the command `bacalhau docker run --count=2 --concurrency=2 ubuntu:latest echo "hello world"` returns the error: `if any flags in the group [count concurrency] are set none of the others can be; [count concurrency] were all set` (this is a feature of Cobra, our CLI package)


wdbaruni:
Why are you creating a new PR and not continuing with https://github.com/bacalhau-project/bacalhau/pull/3970?

cmd/cli/docker/docker_run.go
```diff
@@ -106,143 +95,88 @@ func newDockerRunCmd() *cobra.Command { //nolint:funlen
...
-func dockerImageContainsTag(image string) bool {
-	if strings.Contains(image, ":") {
-		return true
-	}
-	if strings.Contains(image, "@") {
-		return true
+	// Normalize and validate the job spec
+	job.Normalize()
+	if err := job.ValidateSubmission(); err != nil {
+		return nil, fmt.Errorf("%s: %w", userstrings.JobSpecBad, err)
```
wdbaruni, line 178: is there a way to put job normalization, validation, submission, and printing of progress in one place shared between `docker run`, `wasm run` and `job run`?

cmd/cli/get/get_test.go
```diff
@@ -125,6 +126,7 @@ func (s *GetSuite) getDockerRunArgs(extraArgs []string) []string {
 // it makes it's own folder to put the results in and does not splat results
 // all over the current directory
 func (s *GetSuite) TestDockerRunWriteToJobFolderAutoDownload() {
+	s.T().Skip("--download not supported in v2")
```
wdbaruni, line 129: why is `--download` not supported?

cmd/cli/helpers/todo.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+
+	"github.com/bacalhau-project/bacalhau/cmd/util/flags/cliflags"
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+)
+
+func JobToYaml(job *models.Job) (string, error) {
+	yamlBytes, err := yaml.Marshal(job)
+	if err != nil {
```
wdbaruni, line 15: what is the purpose of this file and why is it called `todo.go`?

cmd/util/flags/cliflags/job.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (j *JobSettings) Type() string {
+	if j.cmd.Flags().Changed("target") {
+		jobType := models.JobTypeBatch
+		if j.legacy.targetingMode == model.TargetAll {
+			jobType = models.JobTypeOps
+		}
+		return jobType
+	}
+	return j.jobType
```
wdbaruni, line 48: Lets discuss with product about this flag and moving from `--target` to `--job-type`. As I mentioned before, I am questioning enabling users to submit long running jobs using declarative approach

cmd/util/flags/cliflags/job.go
```diff
@@ -0,0 +1,232 @@
...
+
+const RegexString = "A-Za-z0-9._~!:@,;+-"
+
+func IsSafeAnnotation(s string) bool {
+	matches := SafeAnnotationRegex().FindString(s)
+	return matches == ""
+}
+
+func returnAllEmojiString() string {
+	//nolint:lll,stylecheck // exception for one time string
```
wdbaruni, line 86: is this the best way to do this? we are using label selector library from k8s, and I am sure we can use something from that library for label validation

cmd/util/flags/cliflags/job.go
```diff
@@ -0,0 +1,232 @@
...
+	labels []string
+	// Deprecated: use 'JobSettings.constraints'
+	selectors string
+	//Deprecated: use `JobSettings.count`
+	concurrency int
+}
+
+func DefaultJobSettings() *JobSettings {
+	return &JobSettings{
+		name:        "main",
```
wdbaruni, line 131: "main" should be the name of the task and not the job

cmd/util/flags/cliflags/job.go
```diff
@@ -0,0 +1,232 @@
...
+		namespace:   "default",
```
wdbaruni, line 132: use `models.DefaultNamespace`

cmd/util/flags/cliflags/job.go
```diff
@@ -0,0 +1,232 @@
...
+		panic(err)
+	}
+	if err := fs.MarkDeprecated("target", "use --type"); err != nil {
+		panic(err)
+	}
+
+	// NB(forrest): the `label` flag is replacing `labels` flag. Hide the `labels` flag and add deprecation notice.
+	fs.StringToStringVar(&s.labels, "label", s.labels,
+		`List of labels for the job. Enter multiple in the format '-label a -label 2'.
+All characters not matching /a-zA-Z0-9_:|-/ and all emojis will be stripped.`)
```
wdbaruni, line 193: we are using plural format for similar flags, including `--labels` with `bacalhau serve`

wdbaruni (commented):
Minor comments, but overall looks great. Please don't force push future commits to help review changes faster ;)

cmd/cli/docker/docker_run.go
```diff
@@ -106,143 +95,88 @@ func newDockerRunCmd() *cobra.Command { //nolint:funlen
...
+}
+
+func build(args []string, opts *DockerRunOptions) (*models.Job, error) {
+	image := args[0]
+	parameters := args[1:]
+	engineSpec, err := engine_docker.NewDockerEngineBuilder(image).
+		WithParameters(parameters...).
+		WithWorkingDirectory(opts.WorkingDirectory).
+		WithEntrypoint(opts.Entrypoint...).
+		WithEntrypoint(opts.EnvironmentVariables...).
```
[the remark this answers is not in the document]
frrist: Indeed, good catch.

cmd/cli/docker/docker_run.go
frrist: Yes, although I'd like to wait on unification across the the CLI's until we migrate wasm to the new API spec. That alright with you?

cmd/cli/get/get_test.go
frrist: The printer logic used by V2 CLIs does no have support for downloads - e.g. the job run command doesn't support these auto download flags. We can either address this in a follow on where we unify auto download flags across all CLIs. Or I can attempt to address this now and preserve the existing flag. Which would you prefer?

cmd/cli/helpers/todo.go
frrist: It contains helper methods to be used by various CLIs, just not sure what to name the file - will go with `helpers.go` unless an alternative is proposed here.

cmd/util/flags/cliflags/job.go
frrist: This code is ported over from the [legacy job CLI package](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/legacyjob/emojiarray.go) in an attempt to preserve backwards compatibility. I think the better way would be using the validation provided by the K8s selectors, but that would mean a breaking change to the `labels` flag. 

Generally speaking there isn't a straight forward way maintain compatibility across this flag since
- Previously labels were a string slice: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/model/job.go#L230
- Now they are a map of strings: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/job.go#L99

To accommodate this change we must accept and parse the flags differently on the CLI and include a [plural flag name](https://github.com/bacalhau-project/bacalhau/pull/4020#discussion_r1615175328) for the map case. When we accept a slice instead of a map we migrate it to a map with empty keys: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/migration/legacy/from.go#L34. But empty keys are considered invalid by the k8s label package.

My suggestion here, if you're willing to move forward with it, would be to drop the flag that accepts a slice and make this a breaking change only in the context of this flag. what do you think?

cmd/util/flags/cliflags/job.go
frrist: see https://github.com/bacalhau-project/bacalhau/pull/4020#discussion_r1619426801

cmd/util/flags/cliflags/job.go
frrist: What should be the default name of the job? Seems we still need to set one..

Now that I have merged this with main and compared with the changes made in https://github.com/bacalhau-project/bacalhau/pull/4005 I see we are setting the job name to the ID: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/job.go#L166 this Normalization will work on the server side after the job has been submitted and an ID has been created, but the change is causing failures in this branch as the ID isn't set on the client side and Validate thus returns an error: https://github.com/bacalhau-project/bacalhau/blob/main/pkg/models/job.go#L214

How would you like to address this issue?

cmd/util/flags/cliflags/job.go
frrist: Disregard previous comment, I was using the wrong validate method.

Afshinam1489 (commented):
Azadam

cmd/cli/docker/docker_run_cli_test.go
```diff
@@ -0,0 +1,333 @@
...
+			flags: []string{"image:tag", "--workdir=/dir"},
+			assertJob: func(t *testing.T, j *models.Job) {
+				defaultJobAssertions(t, j)
+				task := j.Task()
+				defaultTaskAssertions(t, task)
+
+				ds, err := dm.DecodeSpec(j.Task().Engine)
+				require.NoError(t, err)
+				assert.Equal(t, "image:tag", ds.Image)
+				assert.Equal(t, "/dir", ds.WorkingDirectory)
```
Afshinam1489, line 84: Azadam

cmd/cli/docker/docker_run.go
wdbaruni: yeah sound good

cmd/cli/get/get_test.go
wdbaruni: I see. My preference to preserve the existing flag and address this as part of this change. I don't want merging this PR to block the release pipeline, and we shouldn't release without it

cmd/util/flags/cliflags/job.go
wdbaruni: We don't need separate flags. What I am trying to say is if k8s selectors enable emoji's, then we don't need to prevent them, and if they have additional validations, then we should enforce them. We shouldn't categorize this as a breaking change as most likely the additional validations are not common cases that will break our limited set of existing users.

As for existing labels being a slice, and new ones being key/value, we've discussed this before that we can still accept a slice of strings and parse them as keys with empty values. We don't need to document this so we can stop the bleeding, but we don't need to break all existing submissions and we don't need to continue supporting two different flags

cmd/util/flags/cliflags/job.go
```diff
@@ -229,11 +182,10 @@ Matching objects must satisfy all of the specified label constraints.`)
...
 	// NB(forrest): don't allow the legacy flag name to be used together with the new flag name.
 	cmd.MarkFlagsMutuallyExclusive("count", "concurrency")
-	cmd.MarkFlagsMutuallyExclusive("labels", "label")
 	cmd.MarkFlagsMutuallyExclusive("selector", "constraints")
 	cmd.MarkFlagsMutuallyExclusive("target", "type")
 	//NB(forrest): we require the name with the type.
 	// TODO (forrest): FOR REVIEW this ais a question for PRODUCT cc Aronchick
 	// - do we want to support type? do we want a dtype flag for each kind of job? --type-ops, --type-daemon, etc?
-	cmd.MarkFlagsRequiredTogether("type", "name")
+	//cmd.MarkFlagsRequiredTogether("type", "name")
```
wdbaruni, line 190: I think this is the only item left to merge the PR. My preference is to decouple the introduction of `--type` flag and enabling different submitting long running job types through `docker run` from this PR as it might take time to convey the problem to product and align on a decision, specially that no one asked for this yet.

Afshinam1489 (changes requested):
Azaawd

Afshinam1489 (changes requested):
Azawd

cmd/cli/docker/docker_run.go
```diff
@@ -1,23 +1,22 @@
...
 	"github.com/bacalhau-project/bacalhau/cmd/util/flags/configflags"
 	"github.com/bacalhau-project/bacalhau/cmd/util/hook"
-	"github.com/bacalhau-project/bacalhau/cmd/util/parse"
 	"github.com/bacalhau-project/bacalhau/cmd/util/printer"
-	"github.com/bacalhau-project/bacalhau/pkg/bacerrors"
-	legacy_job "github.com/bacalhau-project/bacalhau/pkg/legacyjob"
-	"github.com/bacalhau-project/bacalhau/pkg/model"
+	engine_docker "github.com/bacalhau-project/bacalhau/pkg/executor/docker/models"
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+	"github.com/bacalhau-project/bacalhau/pkg/publicapi/apimodels"
```
Afshinam1489, line 18: Azaawde

cmd/util/flags/cliflags/job.go
frrist: From our sync on this: We will maintain one flag for labels that accepts either keys or key values and leverage the k8s validation.

cmd/util/flags/cliflags/job.go
frrist: I have removed the type flag from the command. `target` remains on the command with the same behavior as before.

cmd/util/flags/cliflags/job.go
frrist: addressed

wdbaruni (approved):

cmd/cli/get/get_test.go
frrist: Tracking for this here. Will implement once the wasm CLI is migrated: https://github.com/bacalhau-project/bacalhau/issues/4058

Issue #4024, opened 2024-05-22
`bacalhau node list` returns error `failed request: invalid node type: nodeTypeUndefined`


frrist:
### Bug Description
See title

### Expected Behavior
It lists the nodes

### Steps to Reproduce
1. install main
2. run a server
3. list nodes
4. see error

### Bacalhau Versions
- **Agent Version**:  `v1.3.1-rc1
- **CLI Client Version**: `v1.3.1-rc1`

### Host Environment
*Provide details about the environment where the bug occurred:*
- Operating System: linux
- CPU Architecture: x86

frrist:
I am working from the staging cluster which contains 4 nodes.. It appears there is an extra node (a 5th) in the store somewhere that is of type undefined causing these errors:
```
curl http://bootstrap.staging.bacalhau.org:1234/api/v1/orchestrator/nodes
{
  "NextToken": "",
  "Nodes": [
    {
      "Info": {
        "NodeID": "",
        "NodeType": "nodeTypeUndefined",
        "Labels": null,
        "BacalhauVersion": {
          "GitVersion": "",
          "GitCommit": "",
          "BuildDate": "0001-01-01T00:00:00Z",
          "GOOS": "",
          "GOARCH": ""
        }
      },
      "Membership": "",
      "Connection": "DISCONNECTED"
    },
    {
      "Info": {
        "NodeID": "QmRr9qPTe4mU7aS9faKnWgvn1NtXt36FT8YUULRPCn2f3K",
        "NodeType": "Compute",
        "Labels": {
          "Architecture": "amd64",
          "Operating-System": "linux",
          "git-lfs": "false",
          "owner": "bacalhau"
        },
        "ComputeNodeInfo": {
          "ExecutionEngines": [
            "docker",
            "wasm"
          ],
          "Publishers": [
            "s3",
            "local",
            "noop",
            "ipfs"
          ],
          "StorageSources": [
            "inline",
            "repoclone",
            "repoclonelfs",
            "s3",
            "ipfs",
            "urldownload"
          ],
          "MaxCapacity": {
            "CPU": 3.2,
            "Memory": 13406204723,
            "Disk": 83047314227
          },
          "QueueCapacity": {},
          "AvailableCapacity": {
            "CPU": 3.2,
            "Memory": 13406204723,
            "Disk": 83047314227
          },
          "MaxJobRequirements": {
            "CPU": 3.2,
            "Memory": 13406204723,
            "Disk": 83047314227
          },
[...]

PR #4030, opened 2024-05-23
Implement repo version 4 with migration

frrist:
A simpler alternative  of https://github.com/bacalhau-project/bacalhau/pull/4029  
  - fixes https://github.com/bacalhau-project/bacalhau/issues/4024

### DIFF
--- a/pkg/compute/management_client.go
+++ b/pkg/compute/management_client.go
@@ -22,7 +22,6 @@ type ManagementClientParams struct {
 	NodeInfoDecorator        models.NodeInfoDecorator
 	AvailableCapacityTracker capacity.Tracker
 	QueueUsageTracker        capacity.UsageTracker
-	RegistrationFilePath     string
 	HeartbeatClient          *heartbeat.HeartbeatClient
 	ControlPlaneSettings     types.ComputeControlPlaneConfig
 }
@@ -39,7 +38,6 @@ type ManagementClient struct {
 	nodeInfoDecorator        models.NodeInfoDecorator
 	availableCapacityTracker capacity.Tracker
 	queueUsageTracker        capacity.UsageTracker
-	registrationFile         *RegistrationFile
 	heartbeatClient          *heartbeat.HeartbeatClient
 	settings                 types.ComputeControlPlaneConfig
 }
@@ -51,7 +49,6 @@ func NewManagementClient(params *ManagementClientParams) *ManagementClient {
 		managementProxy:          params.ManagementProxy,
 		nodeID:                   params.NodeID,
 		nodeInfoDecorator:        params.NodeInfoDecorator,
-		registrationFile:         NewRegistrationFile(params.RegistrationFilePath),
 		availableCapacityTracker: params.AvailableCapacityTracker,
 		queueUsageTracker:        params.QueueUsageTracker,
 		heartbeatClient:          params.HeartbeatClient,
@@ -68,17 +65,8 @@ func (m *ManagementClient) getNodeInfo(ctx context.Context) models.NodeInfo {
 	return ni
 }
 
-// RegisterNode sends a registration request to the requester node. If we successfully
-// register, a sentinel file is created to indicate that we are registered. If present
-// the requester node will know it is already registered.  If not present, it will
-// attempt to register again, expecting the requester node to gracefully handle any
-// previous registrations.
+// RegisterNode sends a registration request to the requester node.
 func (m *ManagementClient) RegisterNode(ctx context.Context) error {
-	if m.registrationFile.Exists() {
-		log.Ctx(ctx).Debug().Msg("not registering with requester, already registered")
-		return nil
-	}
-
 	nodeInfo := m.getNodeInfo(ctx)
 	response, err := m.managementProxy.Register(ctx, requests.RegisterRequest{
 		Info: nodeInfo,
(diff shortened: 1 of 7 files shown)


pkg/compute/management_client.go
```diff
@@ -68,17 +65,8 @@ func (m *ManagementClient) getNodeInfo(ctx context.Context) models.NodeInfo {
...
-// the requester node will know it is already registered.  If not present, it will
-// attempt to register again, expecting the requester node to gracefully handle any
-// previous registrations.
+// RegisterNode sends a registration request to the requester node.
 func (m *ManagementClient) RegisterNode(ctx context.Context) error {
-	if m.registrationFile.Exists() {
-		log.Ctx(ctx).Debug().Msg("not registering with requester, already registered")
-		return nil
-	}
-
```
wdbaruni, line 81: Why are we removing the registration state? This means the compute node will have to re-register every time at startup and the node will fail to start if it is offline or disconnected from the requester.

Overall, I think we need to do better for node registration and allow the compute node to periodically re-check with the requester if the registration info match, and re-register if needed. So not only during startup, and should be best effort if the compute node is offline. Meaning the compute node should continue to run the locally allocated jobs even if it fails to register or re-register. I have a NodeManagerV2 placeholder for 1.5 that should track and fix most issues related to node management.

Since we don't have full offline capabilities today, it might make sense to force compute nodes to re-register at startup if we know what we are doing, and we can fix things with NodeManagerV2

pkg/repo/migrations/v3_4.go
```diff
@@ -0,0 +1,28 @@
...
+	func(r repo.FsRepo) error {
+		repoPath, err := r.Path()
+		if err != nil {
+			return err
+		}
+		resolvedCfg, err := config.Load(repoPath)
+		if err != nil {
+			return err
+		}
+		return os.RemoveAll(resolvedCfg.Node.Network.StoreDir)
```
wdbaruni, line 26: Actually `Node.Network.StoreDir` gives the parent
`~/.bacalhau/orchestrator_store` which holds both job store and nats store. I don't recall the motivation for doing it this way.

I am also concerned that we delete something not intended if the user misconfigured this path to be something outside of the bacalhau dir and something we shouldn't delete.

If we were to go with the repo migration route instead of migrating entries, then we need to validate this path   , and only delete specific sub-directories that are related to node kv store and not the whole path

pkg/repo/migrations/v3_4.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/bacalhau-project/bacalhau/pkg/config"
+	"github.com/bacalhau-project/bacalhau/pkg/repo"
+)
+
+// V3Migration deletes the network store (NATS KV store) so that it may recreated once the node starts.
+// It deletes the store since in v1.3.0 the store contains models.NodeInfo and in v1.3.1 the store
+// contains models.NodeState and these types are incompatible with each other.
+// Full details may be found in https://github.com/bacalhau-project/bacalhau/issues/4024
+var V3Migration = repo.NewMigration(
+	repo.RepoVersion3,
```
wdbaruni, line 15: please add a test scenario for this migration. We already have a test suite for repo migrations and should be trivial to add this one

wdbaruni: also you might want to validate that migration will be graceful if the store doesn't exist or the path is not configured properly

wdbaruni (changes requested):

pkg/repo/migrations/v3_4.go
frrist: > Actually Node.Network.StoreDir gives the parent
~/.bacalhau/orchestrator_store which holds both job store and nats store. I don't recall the motivation for doing it this way.

I don't think that is the case see https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/config.go#L87 and attached debugger screenshot.
image

We can check if the path is outside the repo, and if that is the case abort the migration, but I think the odds of that happening are next to zero.

pkg/repo/migrations/v3_4.go
frrist: Was waiting for feedback on this vs https://github.com/bacalhau-project/bacalhau/pull/4029 before doing any more work here. But if we decide we like this PR better I will add tests before merging.

pkg/compute/management_client.go
frrist: > Why are we removing the registration state?

Its a requirement for the migration added here to work since we are deleting the requester nodes nodestore. Removing this lock ensures each time a compute node is started it attempts to register itself wit the requester. This is important since in the event a requester loses it nodestate store (i.e. the migration) compute nodes will re-register themselves with it. If they have already registered with a requester node registering again idempotent.

> the node will fail to start if it is offline or disconnected from the requester

True, although the compute nodes will also fail to start if starting fresh and they are unable to connect to the requester. I don't think a failure here should result in an aborted start, rather the node should continue attempting to register itself until successful (e.g. loop until success)

wdbaruni: Yeah I understand. As I mentioned it is fine to remove the registration state for now

frrist:
closing in favor of alternative  https://github.com/bacalhau-project/bacalhau/pull/4029

PR #4051, opened 2024-05-30
filter out nodes with high queue capacity

wdbaruni:
This change is related to https://github.com/bacalhau-project/bacalhau/pull/4049 where instead of queueing locally in each compute node, we try to queue in the requester instead so that jobs are scheduled to new nodes that join, or to the first node that frees up its resources.

The current state is we don't filter out nodes if they don't have immediate available capacity or if their queue is growing large. We rank nodes with more capacity higher, but we don't filter out nodes with no capacity. This change allows operators to define `NodeOverSubscriptionFactor` in the requester node to allow it to filter out any compute node with total active and queue capacity beyond the factor. The default is `1.5` which means the compute node can queue locally half of its total capacity in addition to the running capacity.

## Testing
This change has been tested with https://github.com/bacalhau-project/bacalhau/pull/4049 in dev stack as documented in that issue


aronchick (approved):
Very cool feature!

pkg/node/config_requester.go
```diff
@@ -73,10 +76,17 @@ func NewRequesterConfigWithDefaults() (RequesterConfig, error) {
...
 //nolint:gosimple
 func NewRequesterConfigWith(params RequesterConfigParams) (RequesterConfig, error) {
-	if err := mergo.Merge(&params, getRequesterConfigParams()); err != nil {
+	defaults := getRequesterConfigParams()
+	if err := mergo.Merge(&params, defaults); err != nil {
 		return RequesterConfig{}, fmt.Errorf("creating requester config: %w", err)
 	}
 
+	// TODO: move away from how we define approval states as they don't have clear
+	//  zero value and don't play nicely with merge
```
frrist, line 85: Feels like `PENDING` ought to be the default/zero value.

frrist (approved):
lgtm

PR #4055, opened 2024-06-04
fail jobs exceeding TotalTimeout

wdbaruni:
`TotalTimeout` is a new field that was introduced in https://github.com/bacalhau-project/bacalhau/pull/4049#discussion_r1625020598 that covers both ExecutionTimeout and QueueTimeout, but not taking affect and not processed by Housekeeper or Scheduler

This PR fixes that and maps `--timeout` flag in `docker run` to `TotalTimeout` instead of `ExecutionTimeout`


pkg/orchestrator/housekeeping.go
```diff
@@ -191,42 +210,38 @@ func (h *Housekeeping) timeoutExecutions(ctx context.Context, activeExecutions [
...
-		return fmt.Errorf("failed to create evaluation %+v: %w", eval, err)
-	}
-
-	log.Ctx(ctx).Debug().Msgf("enqueued evaluation for timed-out execution %+v", eval)
-	return nil
+func (h *Housekeeping) enqueueTimeoutTask(ctx context.Context, job *models.Job, trigger, comment string) {
+	h.workersSem <- struct{}{}
+	h.waitGroup.Add(1)
+
+	go func() {
```
frrist, line 232: why is it necessary to perform this operation in a go routine?

frrist (approved):

pkg/bidstrategy/semantic/timeout_strategy.go
```diff
@@ -38,17 +37,9 @@ const (
...
-
-	// Timeout will be multiplied by 1000000000 (time.Second) when it gets
-	// converted to a time.Duration (which is an int64 underneath), so make sure
-	// that it can fit into it.
-	var maxTimeout = int64(model.NoJobTimeout.Seconds())
-	if request.Job.Task().Timeouts.ExecutionTimeout > maxTimeout {
-		return bidstrategy.NewBidResponse(false, maxReason, timeoutSeconds, maxTimeout), nil
+	timeout := request.Job.Task().Timeouts.GetExecutionTimeout()
+	if timeout <= 0 {
+		return bidstrategy.NewBidResponse(true, minReason, timeout.String(), 0), nil
```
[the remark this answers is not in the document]
wdbaruni: I hear your point. What do have in mind to make it more readable?

pkg/orchestrator/housekeeping.go
wdbaruni: today our datastore is local and may not be that beneficial, but when we use a remote datastore or when we have to replicate datastore operations across multiple requesters, we can benefit from parallelizing unrelated operations or else the housekeeper may not be able to keep up

frrist: Since we aren't at that point yet this seems unnecessary/premature. But given this PR already merged I suppose it doesn't matter.

pkg/bidstrategy/semantic/timeout_strategy.go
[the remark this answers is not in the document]
frrist: Return the bid structure rather than using this construct.

PR #4060, opened 2024-06-06
refactor: migrate wasm CLI to v2

### LINKED ISSUE #4021: Use new APIs for `wasm run`
Use new APIs when handling `bacalhau wasm run`. There shouldn't be any user visible change with this PR, and the CLI flags should not be changed.

We should continue to support old flag names to avoid breaking existing jobs and examples in a single release. Though only new flag names should be printed when calling `bacalhau wasm run --help`

frrist:
- closes #4021


frrist:
@coderabbitai /review

wdbaruni (approved):

PR #4061, opened 2024-06-08
fully remove embedded ipfs

wdbaruni:
This PR fully removes embedded IPFS from our code base, including tests where we are now using `local` publisher and `localDirectory` storage source instead.

## Testing
IPFS related tests are skipped unless IPFS connect endpoint is provided, such as through `BACALHAU_NODE_IPFS_CONNECT` environment variable.

These tests are running in circleci test workflow for liux-amd64-integration where we also run an IPFS daemon in that environment and set IPFS connect configuration.

## Other Changes

Few other issues were surfaced when switching to local publisher that had to be fixed, including:
- local publisher used a different compression logic than s3 publisher, but both used the same downloader. The issue with the previous compression is that it used absolute paths when setting file headers and that resulted in failures with windows during decompression as it resulted in invalid file names.
- repo clone is a weird storage source that copies repos from git to ipfs, index to kv.bacalhau.org (no longer running), and tries to load repos from ipfs if they are cached instead of git. It is a whole mess and have been removed
- s3 integration tests didn't run in circelci due to a change in determining if they should be skipped or not where we tried to write a fake file to s3. The problem is circelci credentials only had read permissions to a defined prefix. Changed the logic to try to read a non-existing file, and run the test if we get 404 NoSuchKey, otherwise we skip the test such as with 403 AccessDenied.


cmd/testing/base.go
```diff
@@ -29,18 +29,27 @@ import (
...
 // before each test
 func (s *BaseSuite) SetupTest() {
 	logger.ConfigureTestLogging(s.T())
 
 	fsr, cfg := setup.SetupBacalhauRepoForTesting(s.T())
+	s.Config = cfg
+
+	// TODO: Update checker is configured with production default configs
+	//  and not respecting the test environment. This is a temporary fix
+	os.Setenv("BACALHAU_UPDATE_SKIPCHECKS", "true")
```
frrist, line 50: So I hit this issue as well during some refactor work. I think the issue (and you may already be aware of this) is the config we provide to the devstack for BaseSuite is used by the _server_ but not the client - meaning the server isn't checking for updates, but the client is. This results in flaky tests since when output from a client command is inspected the update message is occasionally present.

I implemented a small mitigation for this via https://github.com/bacalhau-project/bacalhau/blob/7ecfa81ff4494b3f07e501111a56881950111ba7/cmd/testing/base.go#L86 which separates the stdout and stderr from the command result returned by the client. This allows tests to inspect stdout (where results are written) while ignoring stderr (where update messages are written)

frrist, line 50: We can use `config.KeyAsEnvVar` here instead of typing out the environment variable

cmd/util/printer/print_legacy.go
```diff
@@ -48,7 +48,7 @@ func PrintJobExecutionLegacy(
 	runtimeSettings *cliflags.RunTimeSettingsWithDownload,
 	apiLegacy *client.APIClient,
 	apiV2 clientv2.API,
-	cfg types.IpfsConfig,
+	cfg types.BacalhauConfig,
```
frrist, line 51: I don't think we need the config here anymore.

pkg/downloader/download_test.go
```diff
@@ -164,8 +153,11 @@ func publishToS3Unsigned(ds *DownloaderSuite, dir string) *models.SpecConfig {
 // output directory.
 func requireFileExists(ds *DownloaderSuite, path ...string) string {
 	testPath := filepath.Join(ds.downloadSettings.OutputDir, filepath.Join(path...))
-	require.FileExistsf(ds.T(), testPath, "File %s not present", testPath)
-
+	_, err := os.Stat(testPath)
+	if err != nil {
+		ds.T().Logf("Could not find file at %s", testPath)
+	}
+	require.NoError(ds.T(), err)
```
frrist, line 160: can we just add this log message to the require statement?

pkg/downloader/ipfs/downloader_test.go
```diff
@@ -46,25 +27,18 @@ func randomText(t *testing.T) []byte {
 
 func TestIPFSDownload(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
-	cm := system.NewCleanupManager()
-	defer cm.Cleanup(ctx)
 	defer cancel()
 
-	server, err := ipfs.NewNodeWithConfig(ctx, cm, types.IpfsConfig{PrivateInternal: true})
+	t.Skip("Enable test only when IPFS is running")
```
frrist, line 32: maybe just remove this test as from the looks of things it's always skipped? Or do we have plans to unskip later (issue number)?

pkg/lib/gzip/gzip.go
```diff
@@ -8,29 +8,112 @@ import (
...
 // Decompress takes the path to a .tar.gz file and decompresses it into the specified directory.
 func Decompress(tarGzPath, destDir string) error {
+	log.Debug().Msgf("Decompressing %s to %s", tarGzPath, destDir)
 	return DecompressWithMaxBytes(tarGzPath, destDir, DefaultMaxDecompressSize)
 }
 
 // DecompressWithMaxBytes takes the path to a .tar.gz file and decompresses it into the specified directory.
 // It enforces a maximum decompressed file size (per file) to prevent decompression bombs.
 func DecompressWithMaxBytes(tarGzPath, destDir string, maxDecompressSize int64) error {
+	fmt.Printf("Decompressing %s to %s\n", tarGzPath, destDir)
```
frrist, line 104: Shouldn't this be a log line?

frrist (approved):
Please look at addressing a few of the comments before merging. Generally looks great :+1:

frrist:
:man_facepalming: well there goes auto-merge. Might need to address several of these comments in a follow on.

.circleci/config.yml
```diff
@@ -266,6 +266,28 @@ commands:
...
+            cd go-ipfs
+            sudo bash install.sh
+            ipfs --version
+      - run:
+          name: Initialize IPFS
+          command: |
+            ipfs init 
+            ipfs config Addresses.API /ip4/127.0.0.1/tcp/5001
+            ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
+      - run:
```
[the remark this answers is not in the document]
wdbaruni: These are separate steps to install ipfs, init/configure ipfs and another for running the daemon in the background. I didn't know about circleci caching. Good point

cmd/cli/create/create_test.go
```diff
@@ -48,14 +48,6 @@ func (s *CreateSuite) TestCreateGenericSubmitBetter() {
...
 			Fixture: testdata.YamlJobNoopUrl,
 		},
-		{
-			Name:    "docker task json",
-			Fixture: testdata.IPVMTaskDocker,
-		},
-		{
-			Name:    "task with config json",
-			Fixture: testdata.IPVMTaskWithConfig,
-		},
```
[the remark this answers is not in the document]
wdbaruni: I didn't remove them yet. I am planning a separate PR to remove all IPVM stuff

cmd/cli/create/create.go
```diff
@@ -254,5 +254,5 @@ func create(
 		return fmt.Errorf("submitting job for execution: %w", err)
 	}
 
-	return printer.PrintJobExecutionLegacy(ctx, executingJob, cmd, OC.DownloadFlags, OC.RunTimeSettings, apiV1, apiV2, cfg.Node.IPFS)
+	return printer.PrintJobExecutionLegacy(ctx, executingJob, cmd, OC.DownloadFlags, OC.RunTimeSettings, apiV1, apiV2, cfg)
```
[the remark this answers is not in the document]
wdbaruni: it is needed today to call the DownloadResultsHandler. The name `PrintJobExecutionLegacy` is misleading as it also triggers the downloader and not just print info

cmd/cli/get/get.go
```diff
@@ -104,7 +104,7 @@ func get(cmd *cobra.Command, cmdArgs []string, api client.API, cfg types.Bacalha
 	if err := util.DownloadResultsHandler(
 		ctx,
 		cmd,
-		cfg.Node.IPFS,
+		cfg,
```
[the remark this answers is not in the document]
wdbaruni: we still use it as we need it to instantiate NewStandardDownloaders
```
downloaderProvider, err := util.NewStandardDownloaders(ctx, cfg.Node.IPFS.Connect)
	if err != nil {
		return err
	}
```

cmd/util/printer/print_legacy.go
wdbaruni: answered above

pkg/downloader/ipfs/downloader_test.go
wdbaruni: good catch. Fixing in a follow up PR

pkg/lib/gzip/gzip.go
wdbaruni: ah I forgot to remove these debug lines

PR #4068, opened 2024-06-10
Optimize locking in producer and consumer NATS streaming client

udsamani:
## What are we doing here ?

- With this PR, we are making sure we do locking correctly. 
- In addition, we come up with a good way to cancel the stream. We store a `cancel` function as part of stream.  This cancel function is attached to the context which gets passed on to the engine, which serves the stream (currently only the log stream).  If a heartbeat response is received from the consumer stating that it is no longer interested in this stream, then we invoke this `cancel` function. There might be scenarios where few logs will make it through after context cancelation due to race conditions. This should be fine and won't result in `nil` pointers or writing to a closed writer as we only close the writer after source channel is closed.

### DIFF
--- a/pkg/nats/proxy/compute_handler.go
+++ b/pkg/nats/proxy/compute_handler.go
@@ -158,11 +158,13 @@ func processAndStream[Request, Response any](ctx context.Context, streamingClien
 		return
 	}
 
+	childCtx, cancel := context.WithCancel(ctx)
 	err = streamingClient.AddStream(
 		streamRequest.ConsumerID,
 		streamRequest.StreamID,
 		msg.Subject,
 		streamRequest.HeartBeatRequestSub,
+		cancel,
 	)
 
 	defer streamingClient.RemoveStream(streamRequest.ConsumerID, streamRequest.ConsumerID)
@@ -172,17 +174,20 @@ func processAndStream[Request, Response any](ctx context.Context, streamingClien
 		return
 	}
 
-	ch, err := f(ctx, *request)
+	ch, err := f(childCtx, *request)
 	if err != nil {
-		_ = writer.CloseWithCode(stream.CloseInternalServerErr,
+		closeError := writer.CloseWithCode(stream.CloseInternalServerErr,
 			fmt.Sprintf("error in handler %s: %s", reflect.TypeOf(request).Name(), err))
+		if closeError != nil {
+			log.Err(closeError).Msg("error while closing NATS Stream")
+		}
 		return
 	}
 
 	for res := range ch {
-		_, err := streamingClient.WriteResponse(streamRequest.ConsumerID, streamRequest.StreamID, res, writer)
+		_, err := writer.WriteObject(res)
 		if err != nil {
-			log.Ctx(ctx).Error().Msgf("error writing response to stream: %s", err)
+			log.Err(err).Msg("error writing log to NATS subject")
 			break
 		}
 	}
--- a/pkg/nats/proxy/compute_handler.go
+++ b/pkg/nats/proxy/compute_handler.go
@@ -158,6 +158,12 @@ func processAndStream[Request, Response any](ctx context.Context, streamingClien
 		return
 	}
 
+	// This context is passed down to particular engine serving the stream. The cancel function is stored as part of
+	// the StreamInfo. When the consumer client informs the producer client via heartBeat that it is no longer interested
+	// in the stream, we call this cancel function. This also informs the engine that we are no longer interested in this
+	// stream and hence close it. There might be scenarios where few logs will make it through after context cancelation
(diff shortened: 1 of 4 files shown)


pkg/nats/proxy/compute_handler.go
```diff
@@ -172,17 +174,20 @@ func processAndStream[Request, Response any](ctx context.Context, streamingClien
...
 
-	ch, err := f(ctx, *request)
+	ch, err := f(childCtx, *request)
 	if err != nil {
-		_ = writer.CloseWithCode(stream.CloseInternalServerErr,
+		closeError := writer.CloseWithCode(stream.CloseInternalServerErr,
 			fmt.Sprintf("error in handler %s: %s", reflect.TypeOf(request).Name(), err))
+		if closeError != nil {
+			log.Err(closeError).Msg("error while closing NATS Stream")
+		}
```
wdbaruni, line 183: this is great, and let me explain how I understand this works, and please confirm we are not missing other edge cases.

This proxy handler is the one that is creating the stream writer, and it is the only one that is writing to and eventually closing the writer. We don't have any concurrent access to the writer, and that is how we don't need to synchronize its access. 

As we are using the context to close the downstream channel and source of generated logs, there might be scenarios where few logs will make it through after context cancelation due to race conditions. This should be find and won't result in `nil` pointers or writing to a closed writer as we only close the writer _after_ source channel is closed, right?

Maybe wroth adding comments on how this works ;)

pkg/nats/stream/consumer_client.go
```diff
@@ -71,8 +71,9 @@ type ConsumerClientParams struct {
 
 // ConsumerClient represents a NATS streaming client.
 type ConsumerClient struct {
-	Conn *nats.Conn
-	mu   sync.RWMutex // Protects access to the response map.
+	Conn        *nats.Conn
+	mu          sync.RWMutex // Protects access to the response map.
+	muReqSubMap sync.RWMutex // Protects access to reqSubMap
```
wdbaruni, line 76: I think having two locks is an over-optimization here not worth the added complexity. Just using `mu` should be enough, as long as we use read locks when processing heartbeats? What do you think?

pkg/nats/stream/consumer_client.go
```diff
@@ -266,21 +267,21 @@ func (nc *ConsumerClient) heartBeatRespHandler(msg *nats.Msg) {
 	request := new(HeartBeatRequest)
 	err := json.Unmarshal(msg.Data, request)
 	if err != nil {
-		log.Err(err)
+		log.Err(err).Msg("failured to parse heart beat request for NATs based consumer client")
```
wdbaruni, line 270: typo

pkg/nats/stream/producer_client.go
```diff
@@ -130,50 +133,30 @@ func (pc *ProducerClient) heartBeat(ctx context.Context) {
...
 				}
 
 				nonActiveStreamIds[c] = getStringList(heartBeatResponse.NonActiveStreamIds)
 			}
 
-			pc.updateActiveStreamInfo(nonActiveStreamIds)
+			pc.mu.RUnlock()
+			if len(nonActiveStreamIds) != 0 {
+				pc.updateActiveStreamInfo(nonActiveStreamIds)
+			}
```
wdbaruni, line 152: how about adding all this in a `defer` statement right after acquiring the lock?

pkg/nats/stream/types.go
```diff
@@ -47,6 +48,8 @@ type StreamInfo struct {
 	RequestSub string
 	// CreatedAt represents the time the stream was created.
 	CreatedAt time.Time
+	// Function to cancel the stream
+	Cancel context.CancelFunc
```
wdbaruni, line 52: maybe more comments will be helpful

udsamani: Done.

pkg/nats/stream/producer_client.go
udsamani: This unlock is happening in an infinite for loop. I don't think using `defer` here would be a good idea ! But let me know if you think otherwise.

pkg/nats/proxy/compute_handler.go
udsamani: Indeed. Well put what I am trying to do here :) Let me put a comment explaining it.

pkg/nats/stream/consumer_client.go
udsamani: Fixed  :)

pkg/nats/stream/producer_client.go
wdbaruni: You are right. I missed the loop

wdbaruni (approved):

frrist (approved):
LGTM

PR #4069, opened 2024-06-11
circleci cache ipfs + previous PR fixes

wdbaruni:
follow up fixes for https://github.com/bacalhau-project/bacalhau/pull/4061


pkg/lib/gzip/gzip.go
```diff
@@ -127,56 +122,44 @@ func DecompressWithMaxBytes(tarGzPath, destDir string, maxDecompressSize int64)
 			break // End of archive
 		}
 		if err != nil {
-			return fmt.Errorf("failed to read tar header: %w", err)
+			return err
```
frrist, line 125: Any reason for making these error messages less verbose?

.circleci/config.yml
```diff
@@ -267,16 +267,35 @@ commands:
...
+              cd go-ipfs
+              sudo bash install.sh
+              ipfs --version
+            else
+              echo "IPFS is already installed, skipping download."
+            fi
+      - save_cache:
+          key: ipfs-<< parameters.cache_version >>-{{ checksum ".circleci/config.yml" }}
+          paths:
+            - /usr/local/bin/ipfs
```
frrist, line 298: How about something similar to this. The suggested change will enable us to modify the IPFS version and invalidate the cache together
```suggestion
install_ipfs:
  parameters:
    ipfs_version:
      type: string
      default: v0.18.0
  description: Install IPFS for linux-amd64
  steps:
    - restore_cache:
        keys:
          - ipfs-<< parameters.ipfs_version >>
    - run:
        name: Install IPFS
        command: |
          if [ ! -f /usr/local/bin/ipfs ]; then
            wget https://dist.ipfs.tech/go-ipfs/<< parameters.ipfs_version >>/go-ipfs_<< parameters.ipfs_version >>_linux-amd64.tar.gz -O ipfs.tar.gz
            tar -xzf ipfs.tar.gz
            cd go-ipfs
            sudo bash install.sh
            ipfs --version
          else
            echo "IPFS is already installed, skipping download."
    - save_cache:
        key: ipfs-<< parameters.ipfs_version >>
        paths:
          - /usr/local/bin/ipfs
    - run:
        name: Initialize IPFS
        command: |
          ipfs init 
          ipfs config Addresses.API /ip4/127.0.0.1/tcp/5001
          ipfs config Addresses.Gateway /ip4/0.0.0.0/tcp/8080
    - run:
        name: Start IPFS Daemon
        command: |
          ipfs daemon
        background: true
```

frrist (commented):
I'd like to adjust the cache setting in CI before merging.

.circleci/config.yml
wdbaruni: I am using a similar pattern as in other existing dependencies where we include the checksum of the whole config file. IPFS version shouldn't be enough as we should invalidate the cache if we change how we are installing ipfs. I understand that any change in the config will invalidate the cache, but that shouldn't be a problem.

I've added the ipfs version in the cache key and fixed a hardcoded v0.18.0 here https://github.com/bacalhau-project/bacalhau/pull/4069/commits/d1106e5ca2e8260c42e977785f9f74213769f2ec

pkg/lib/gzip/gzip.go
wdbaruni: brought them back in https://github.com/bacalhau-project/bacalhau/pull/4069/commits/d1106e5ca2e8260c42e977785f9f74213769f2ec

frrist (approved):

PR #4073, opened 2024-06-12
remove libp2p

wdbaruni:


frrist (approved):
This is doing a bit more than deprecating libp2p :wink:  but structurally looks good as I assume this is mostly mechanical. If there is anything specific you'd like review on let me know - otherwise looks good.

wdbaruni:
> This is doing a bit more than deprecating libp2p 😉 but structurally looks good as I assume this is mostly mechanical. If there is anything specific you'd like review on let me know - otherwise looks good.

It is removing libp2p and things that were specific to libp2p transport. The main one that _looks_ unrelated is NodeInfoStore and NodeInfoPublisher, which were replaced by NodeManager for NATS transport. Also NodeInfo are only specific to requester node for nats compared to compute node as well in libp2p to find the peerInfo of the requester node

PR #4086, opened 2024-06-14
fix ipfs storage test flakiness

wdbaruni:
we check public dht for the cid providers, and can take time for the cid to be discoverable. The fix is to retry up to 10 seconds before failing


frrist (approved):
LGTM, as a follow on to this I've filed https://github.com/bacalhau-project/bacalhau/pull/4088 which removes the need to check the DHT for local content all together.

PR #4087, opened 2024-06-14
Debugging GH Secrets

aronchick:
Fixing GH to allow for bots to manage PRs.


.circleci/trigger_circleci.py
```diff
@@ -35,6 +35,8 @@ def main():
     }
     data.update(target)
 
+    print(f"Full data: {data}")
```
frrist, line 38: any chance of this leaking secrets?

frrist (approved):
One comment regarding what we print.

.circleci/trigger_circleci.py
aronchick: no - this is from github event and github auto hides.

but will remove when this is debugged.

PR #4107, opened 2024-06-18
Removing flyte

aronchick:
Removing flyte - it's out of date, and needs rewriting. Non blocking for 1.4.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,10 @@ parameters:
     default: "v0.0.0"
     type: string
 
+  full_name:
+    default: "bacalhau-project/bacalhau"
+    type: string
+
 executors:
   # Available resource classes - https://circleci.com/product/features/resource-classes/
   linux-amd64:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -256,36 +279,16 @@ commands:
...
-              cd go-ipfs
-              sudo bash install.sh
-              ipfs --version
-            else
-              echo "IPFS is already installed, skipping download."
-            fi
-      - save_cache:
-          key: ipfs-<< parameters.cache_version >>-<< parameters.ipfs_version >>-{{ checksum ".circleci/config.yml" }}
-          paths:
-            - /usr/local/bin/ipfs
```
frrist, line 288: Is there a reason this caching is being removed?

frrist (approved):
Please add back the ipfs download cache and provide details on buildkite/bacalhau when you can.

.circleci/config.yml
aronchick: fixed

PR #4111, opened 2024-06-19
refactor: migrate List*Response names to Items

### LINKED ISSUE #3839: Rename ListX responses to `Items`
Instead of having ListJob, ListJobExecutions,... return an array of `Jobs`, `Executions`,... , unify this to always return `Items` array along with `NextToken`

Preferably we should do this before announcing V1 API deprecation. Also would be nice to do this while maintaining backward compatibility with existing clients. Maybe we should explore adding bacalhau version as an http header, and return `Items` only for new versions? Up for discussion

frrist:
- maintains backwards compatibility with old clients
- closes #3839

### DIFF
--- a/pkg/publicapi/endpoint/orchestrator/job.go
+++ b/pkg/publicapi/endpoint/orchestrator/job.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/gorilla/websocket"
 	"github.com/labstack/echo/v4"
 	"github.com/rs/zerolog/log"
@@ -103,6 +104,11 @@ func (e *Endpoint) getJob(c echo.Context) error { //nolint: gocyclo
 			for i := range history {
 				response.History.Items[i] = &history[i]
 			}
+			// migrate to old response if required
+			if listResponseRequiresMigration(apimodels.GetClientVersion(c.Request())) {
+				response.History.History = response.History.Items //nolint: staticcheck
+				response.History.Items = make([]*models.JobHistory, 0)
+			}
 		case "executions":
 			// ignore if user requested executions twice
 			if response.Executions != nil {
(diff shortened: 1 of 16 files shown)


pkg/publicapi/endpoint/orchestrator/job.go
```diff
@@ -114,13 +115,26 @@ func (e *Endpoint) getJob(c echo.Context) error {
...
 			for i := range executions {
-				response.Executions.Executions[i] = &executions[i]
+				response.Executions.Items[i] = &executions[i]
 			}
 		}
 	}
+
+	clientVersion := apimodels.GetClientVersion(c.Request())
+	if !(clientVersion.GreaterThan(version.V1_3_2) || clientVersion.Equal(version.Development) || clientVersion.Prerelease() != "") {
+		if response.History != nil && len(response.History.History) > 0 { //nolint: staticcheck
```
wdbaruni, line 128: Why do we need this check?

pkg/publicapi/endpoint/orchestrator/job.go
```diff
@@ -114,13 +115,26 @@ func (e *Endpoint) getJob(c echo.Context) error {
...
 			}
 			for i := range executions {
-				response.Executions.Executions[i] = &executions[i]
+				response.Executions.Items[i] = &executions[i]
 			}
 		}
 	}
+
+	clientVersion := apimodels.GetClientVersion(c.Request())
+	if !(clientVersion.GreaterThan(version.V1_3_2) || clientVersion.Equal(version.Development) || clientVersion.Prerelease() != "") {
```
wdbaruni, line 127: Since we have been encoding the client version for a while, we shouldn't only return old format if the version is known and it is older than 1.3.2. For development and unknown versions we should return the latest format

wdbaruni (approved):
minor comments

pkg/publicapi/endpoint/orchestrator/job.go
frrist: So, we do return the new version if the client's version is Development:
```shell
curl -s -H "X-Bacalhau-Git-Version: v0.0.0-xxxxxxx" 0.0.0.0:1234/api/v1/orchestrator/jobs | jq '{JobsCount: (if .Jobs == null then 0 else .Jobs | length end), ItemsCount: .Items | length}'
{
  "JobsCount": 0,
  "ItemsCount": 4
}
```
But don't if the client's version is unknown:
(with header)
```shell
curl -s -H "X-Bacalhau-Git-Version: v0.0.0" 0.0.0.0:1234/api/v1/orchestrator/jobs | jq '{JobsCount: (if .Jobs == null then 0 else .Jobs | length end), ItemsCount: .Items | length}'
{
  "JobsCount": 4,
  "ItemsCount": 0
}
```
(without header)
```
curl -s  0.0.0.0:1234/api/v1/orchestrator/jobs | jq '{JobsCount: (if .Jobs == null then 0 else .Jobs | length end), ItemsCount: .Items | length}'
{
  "JobsCount": 4,
  "ItemsCount": 0
}
```
Meaning the requested change here is to use the new format if the client version is not set/unknown?

pkg/publicapi/endpoint/orchestrator/job.go
frrist: we don't I've adjusted in latest commit.

PR #4115, opened 2024-06-21
add Queued state and history event

wdbaruni:
This PR improves the visibility of queued jobs by adding a new history event, dedicated job state, and enables updating the job state from running back to queued if necessary.

### This change includes:
- Decoupling adding history events from updating job and executions states by exposing first class methods to add events
- Make sure we are using transactions in many places we missed using them
- Refactor the structure of `models.JobHistory` so that events are not coupled with state updates where we can add events even if the job/execution state remains the same. This plays nicely with our intention to add more granular events for job execution
- Drop unhelpful and too granular history events

### Issues
To maintain backward compatibility with v1.4 and to avoid `nil` exceptions due to missing `JobState` fields, we are still keeping those fields and adding undefined value for them at the API layer. Older clients will still see `Undefined` as the job state, but that is better than `nil` exception. When `v1.5` is released, we will set that version as the min version and all requests from older clients will be rejected with a message asking them to update.

#### Old Format
```
→       bacalhau job describe j-ae5403b8-0934-4d86-a6e0-31e2f981f021
ID            = j-ae5403b8-0934-4d86-a6e0-31e2f981f021
Name          = j-ae5403b8-0934-4d86-a6e0-31e2f981f021
Namespace     = default
Type          = batch
State         = Completed
Count         = 1
Created Time  = 2024-06-21 09:18:10
Modified Time = 2024-06-21 09:18:10
Version       = 0

Summary
Completed = 1

Job History
 TIME                 REV.  STATE      TOPIC       EVENT
 2024-06-21 09:18:10  1     Pending    Submission  Job submitted
 2024-06-21 09:18:10  2     Running
 2024-06-21 09:18:10  3     Completed

Executions
 ID          NODE ID   STATE      DESIRED  REV.  CREATED    MODIFIED   COMMENT
 e-6403a485  QmPLPUUj  Completed  Stopped  6     4h37m ago  4h37m ago  Accepted job

Execution e-6403a485 History
 TIME                 REV.  STATE              TOPIC            EVENT
 2024-06-21 09:18:10  1     New
 2024-06-21 09:18:10  2     AskForBid
 2024-06-21 09:18:10  3     AskForBidAccepted  Requesting Node  Accepted job
 2024-06-21 09:18:10  4     AskForBidAccepted
 2024-06-21 09:18:10  5     BidAccepted
 2024-06-21 09:18:10  6     Completed

Standard Output
heelo
```

#### New Format
```
[...]

PR #4117, opened 2024-06-21
set minimum client version on http server

wdbaruni:
Some changes in the upcoming release are not compatible with existing clients, such as https://github.com/bacalhau-project/bacalhau/pull/4115

This PR prints a human readable message asking users to update their client


pkg/publicapi/middleware/version.go
```diff
@@ -80,3 +82,37 @@ func VersionNotifyLogger(logger *zerolog.Logger, serverVersion semver.Version) e
...
+					"Error": "Invalid client version " + clientVersionStr,
+				})
+			}
+
+			if clientVersion.LessThan(&minVersion) {
+				// Client version is less than the minimum required version
+				return c.JSON(http.StatusForbidden, map[string]string{
+					"Error":          "Client version is outdated. Update your client",
+					"ServerVersion":  serverVersion.String(),
+					"MinimumVersion": minVersion.String(),
```
frrist, line 110: Currently this will return the minimum version is https://github.com/bacalhau-project/bacalhau/pull/4117/files#diff-19907c38cadb733516c5d0d2d452e8ef1dde593253b301251fec43f1a8a8d315R30 (1.3.3) which isn't actually a version. Should we bump the min version to 1.4.0?

frrist (approved):

PR #4132, opened 2024-06-24
Add Streaming Client interaction test

udsamani:
With this change, we are adding tests for interaction between the producer and consumer client.
The test scenarion being added is where we kill the consumer client and make sure that producer client kills the stream as well

### DIFF
--- a/pkg/nats/stream/producer_client.go
+++ b/pkg/nats/stream/producer_client.go
@@ -91,17 +91,17 @@ func (pc *ProducerClient) AddStream(
 	return nil
 }
 
-func (pc *ProducerClient) RemoveStream(consumerID string, streamID string) {
+func (pc *ProducerClient) RemoveStream(consumerID string, streamID string) error {
 	pc.mu.Lock()
 	defer pc.mu.Unlock()
 
 	activeStreamIdsForConn := pc.activeConsumers[consumerID].ActiveStreamInfo
 	if activeStreamIdsForConn == nil {
-		return
+		return fmt.Errorf("active stream Ids for consumer %s is nil", consumerID)
 	}
 
 	if _, ok := activeStreamIdsForConn[streamID]; !ok {
-		return
+		return fmt.Errorf("no stream with id %s found for consumer %s", streamID, consumerID)
 	}
 
 	delete(activeStreamIdsForConn, streamID)
@@ -109,6 +109,8 @@ func (pc *ProducerClient) RemoveStream(consumerID string, streamID string) {
 	if len(activeStreamIdsForConn) == 0 {
 		delete(pc.activeConsumers, consumerID)
 	}
+
+	return nil
 }
 
 func (pc *ProducerClient) heartBeat(ctx context.Context) {
@@ -142,7 +144,7 @@ func (pc *ProducerClient) heartBeat(ctx context.Context) {
 				msg, err := pc.Conn.Request(v.HeartbeatRequestSub, data, pc.config.HeartBeatRequestTimeout)
 				if err != nil {
 					log.Ctx(ctx).Err(err).Msg("heartbeat request to consumer client timed out")
-					nonActiveStreamIds[c] = append(nonActiveStreamIds[c], v.getActiveStreamIds()...)
+					nonActiveStreamIds[c] = v.getActiveStreamIds()
 					continue
 				}
 
@@ -153,10 +155,7 @@ func (pc *ProducerClient) heartBeat(ctx context.Context) {
 					continue
 				}
 
-				nonActiveStreamIdsFromConsumer := getStringList(heartBeatResponse.NonActiveStreamIds)
-				if len(nonActiveStreamIdsFromConsumer) != 0 {
-					nonActiveStreamIds[c] = append(nonActiveStreamIds[c], nonActiveStreamIdsFromConsumer...)
-				}
+				nonActiveStreamIds[c] = getStringList(heartBeatResponse.NonActiveStreamIds)
 			}
 
 			pc.mu.RUnlock()
--- a/pkg/nats/stream/producer_client.go
+++ b/pkg/nats/stream/producer_client.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/nats-io/nats.go"
 	"github.com/rs/zerolog/log"
+	"github.com/samber/lo"
 )
 
 type ProducerClientParams struct {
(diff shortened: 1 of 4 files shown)


pkg/nats/stream/producer_client.go
```diff
@@ -132,7 +134,8 @@ func (pc *ProducerClient) heartBeat(ctx context.Context) {
 
 				msg, err := pc.Conn.Request(v, data, pc.config.HeartBeatRequestTimeout)
 				if err != nil {
-					log.Ctx(ctx).Err(err).Msg("error while sending heart beat request from NATS streaming producer client")
+					log.Ctx(ctx).Err(err).Msg("heartbeat request to consumer client timed out")
+					nonActiveStreamIds[c] = append(nonActiveStreamIds[c], activeStreamIds...)
```
wdbaruni, line 138: does this mean we close all streams the first time we fail to heartbeat? There can be an area of improvement here we should only close streams that fail N consecutive heartbeats, where N can equal to 3 by default

pkg/nats/stream/producer_client.go
```diff
@@ -111,17 +111,19 @@ func (pc *ProducerClient) heartBeat(ctx context.Context) {
...
+				// Create an empty slice for activeStreamIdsByReqSubj
+				activeStreamIdsByReqSubj := make(map[string][]string)
+				var activeStreamIds []string
 
 				if streamInfoMap, ok := pc.activeStreamInfo[c]; ok {
-					for _, streamInfo := range streamInfoMap {
-						activeStreamIds[streamInfo.RequestSub] = append(activeStreamIds[streamInfo.RequestSub], streamInfo.ID)
+					for streamId, streamInfo := range streamInfoMap {
+						activeStreamIds = append(activeStreamIds, streamId)
+						activeStreamIdsByReqSubj[streamInfo.RequestSub] = append(activeStreamIdsByReqSubj[streamInfo.RequestSub], streamInfo.ID)
```
wdbaruni, line 121: is `streamInfo.ID` the same as `streamId`? 

There are too many maps in `ProducerClient`, `activeStreamInfo` and `activeConnHeartBeatRequestSubjects`, but still they are not enough as we have to construct this `activeStreamIdsByReqSubj` with every heartbeat. Is there anything we can do to make things more readable? If having two maps already are not making things more readable or efficient as we still have to loop here, does it make sense to just have one `activeStreamInfo` type and have helper methods that return grouped by streams on-demand?

pkg/nats/stream/producer_client.go
```diff
@@ -143,7 +146,10 @@ func (pc *ProducerClient) heartBeat(ctx context.Context) {
 					continue
 				}
 
-				nonActiveStreamIds[c] = getStringList(heartBeatResponse.NonActiveStreamIds)
+				nonActiveStreamIdsFromConsumer := getStringList(heartBeatResponse.NonActiveStreamIds)
+				if len(nonActiveStreamIdsFromConsumer) != 0 {
+					nonActiveStreamIds[c] = append(nonActiveStreamIds[c], nonActiveStreamIdsFromConsumer...)
+				}
```
wdbaruni, line 152: why do we expect `nonActiveStreamIds` to have values for the same consumer from a previous loop iteration?

pkg/nats/stream/streaming_client_interactions_test.go
```diff
@@ -0,0 +1,183 @@
+package stream
```
wdbaruni, line 1: build will fail complaining about missing build target. Add `//go:build unit || !integration`

pkg/nats/stream/streaming_client_interactions_test.go
```diff
@@ -0,0 +1,183 @@
...
+func (s *StreamingClientInteractionTestSuite) createProducerClient() *ProducerClient {
+	clientManager, err := nats_helper.NewClientManager(s.ctx, s.natServer.ClientURL(), nats.Name("streaming-test"))
+	s.Require().NoError(err)
+
+	pc, err := NewProducerClient(s.ctx, ProducerClientParams{
+		Conn: clientManager.Client,
+		Config: StreamProducerClientConfig{
+			HeartBeatIntervalDuration:        2 * time.Second,
+			HeartBeatRequestTimeout:          1 * time.Second,
+			StreamCancellationBufferDuration: 2 * time.Second,
```
wdbaruni, line 72: feel free to adjust these values to speed up the test if they are too slow

pkg/nats/stream/producer_client.go
udsamani: Let me revisit it !

pkg/nats/stream/producer_client.go
udsamani: Not sure I understand. This is still happening in the same loop right ! We are just flattening the map. The loop is for different consumers!

pkg/nats/stream/producer_client.go
udsamani: Made it more readable!

pkg/nats/stream/producer_client.go
udsamani: Makes sense. I already have an issue opened for optimising this ! Will solve it as part of that !  #4026

pkg/nats/stream/producer_client.go
```diff
@@ -51,33 +69,34 @@ func (pc *ProducerClient) AddStream(
...
 func (pc *ProducerClient) RemoveStream(consumerID string, streamID string) {
 	pc.mu.Lock()
 	defer pc.mu.Unlock()
 
-	activeStreamIdsForConn, ok := pc.activeStreamInfo[consumerID]
-	if !ok {
+	activeStreamIdsForConn := pc.activeConsumers[consumerID].ActiveStreamInfo
+	if activeStreamIdsForConn == nil {
 		return
 	}
```
wdbaruni, line 101: can this return nil pointer exception if no entry for for `consumerID` is found?

pkg/nats/stream/producer_client.go
wdbaruni: What I mean is `nonActiveStreamIds[c] = append(nonActiveStreamIds[c], nonActiveStreamIdsFromConsumer...)` effectively the same as `nonActiveStreamIds[c] = nonActiveStreamIdsFromConsumer` because `nonActiveStreamIds[c]` is only populated once

wdbaruni (approved):

PR #4151, opened 2024-06-26
mark timed-out executions as failed instead of cancelled

wdbaruni:
For some reason, timed-out executions were marked as cancelled instead of failed, which is wrong. Also this resulted in compute node calling `OnCancelComplete` on the requester node, which is a noop. This means the requester node will only mark the execution as failed when the housekeeper kicks in which has a buffer of 2 minutes, instead of as soon as the failure is reported by the compute node.

Previously, this job will be marked as failed after 2-2:30 minutes:
```
bacalhau docker --timeout 10 run ubuntu sleep 120
```

With this change it will marked as failed in ~10 seconds


pkg/compute/executor.go
```diff
@@ -337,24 +337,7 @@ func (e *BaseExecutor) Run(ctx context.Context, state store.LocalExecutionState)
...
-			// There is prior discussion regarding this point here:
-			// https://github.com/bacalhau-project/bacalhau/pull/2705#discussion_r1283543457
-			//
-			// Moving forward we must avoid canceling executions via the context.Context. When pluggable executors
-			// become the default since canceling the context will simply result in the RPC connection closing (I think)
-			// The general solution here is to stop using contexts for canceling jobs and to instead make explicit calls
-			// the an executors `Cancel` method.
-			log.Ctx(ctx).Info().Msg("execution timeout exceeded canceling execution")
-			return nil
+			return NewErrExecTimeout(state.Execution.Job.Task().Timeouts.GetExecutionTimeout())
```
frrist, line 340: Might be worth leaving a small part of this comment here for context (heh) on why we are checking the `conext.DeadlineExceeded` error.

pkg/compute/errors.go
```diff
@@ -0,0 +1,33 @@
...
+	return fmt.Sprintf("Execution timed out after %s", e.ExecutionTimeout)
+}
+
+func (e ErrExecTimeout) Retryable() bool {
+	return true
+}
+
+func (e ErrExecTimeout) Details() map[string]string {
+	return map[string]string{
+		models.DetailsKeyHint: "Try increasing the task timeout or reducing the task size",
```
frrist, line 31: maybe say "increasing task resources" since it's not clear what it means to reduce the size of a task, least to me its not.

frrist (approved):

pkg/compute/errors.go
wdbaruni: addressed in https://github.com/bacalhau-project/bacalhau/pull/4151/commits/94c4824dae9dd31fc54c966d0d25e8e32b3014d5

PR #4147, opened 2024-06-26
Frrist/tests/migrate to v2 client

frrist:
Closes #4146, #4141, https://github.com/bacalhau-project/bacalhau/issues/4119, and #4143

PR #4154, opened 2024-06-26
Fix Python SDK.

udsamani:
After a refactor of our package `publicapi` we introduce a sub-package called `legacymodels` which changes our code generation. This PR just fixes that.

### DIFF
--- a/python/bacalhau_sdk/api.py
+++ b/python/bacalhau_sdk/api.py
@@ -3,11 +3,11 @@
 import json
 
 from bacalhau_apiclient.api import job_api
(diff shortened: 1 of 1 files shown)


wdbaruni:
Thanks for investigating this. So all the legacy api models are going away with https://github.com/bacalhau-project/bacalhau/pull/4147

I think we need to work on implementing the SDK against the new models and APIs instead

udsamani:
@wdbaruni  Created this issue https://github.com/bacalhau-project/bacalhau/issues/4162 !  Will fix it as part of this ! Closing the PR.

PR #4155, opened 2024-06-26
refactor: migrate exec command to use models

frrist:


cmd/cli/exec/exec.go
```diff
@@ -223,129 +220,27 @@ func PrepareJob(cmd *cobra.Command, cmdArgs []string, unknownArgs []string, opti
...
-	// Constraints for node selection
-	if err := prepareConstraints(options, job); err != nil {
-		return nil, err
+	job.Labels, err = options.JobSettings.Labels()
+	job.Task().Publisher = options.TaskSettings.Publisher.Value()
+	job.Task().ResultPaths = options.TaskSettings.ResultPaths
+	job.Task().Env = options.TaskSettings.EnvironmentVariables
+	job.Task().InputSources = options.TaskSettings.InputSources.Values()
+	if err != nil {
+		return nil, fmt.Errorf("parsing job labes: %w", err)
```
wdbaruni, line 229: Why you are not using `BuildJobFromFlags` similar to `docker run` and `wasm run`, and then override whatever you need to override?

wdbaruni (approved):
LGTM. Just minor comment

cmd/cli/exec/exec.go
frrist: I decided against this as the base version of the job spec is derived from a template and it was simpler to produce the templated job first then modify it.

PR #4156, opened 2024-06-26
Remove the model package

frrist:
finally!


cmd/cli/docker/docker_run_cli_test.go
```diff
@@ -177,15 +177,6 @@ func TestJobFlagParsing(t *testing.T) {
...
 		},
-		{
-			name:  "with concurrency",
-			flags: []string{"--concurrency=100", "image:tag"},
-			assertJob: func(t *testing.T, j *models.Job) {
-				assert.EqualValues(t, 100, j.Count)
-				defaultTaskAssertions(t, j.Task())
-			},
-			expectedError: false,
-		},
```
wdbaruni, line 188: why are we coupling `model` removal with supporting deprecated flags? Can we please do them separately, and only remove support of deprecated flags by 1.5.0?

pkg/node/config_defaults.go
```diff
@@ -43,7 +43,7 @@ func NewDefaultComputeParam(storagePath string) ComputeConfigParams {
 
 var DefaultRequesterConfig = RequesterConfigParams{
 	JobDefaults: transformer.JobDefaults{
-		TotalTimeout: model.NoJobTimeout,
```
wdbaruni, line 46: maybe you should've just left it and migrated it to `models.NoJobTimeout` instead of doing this in many places. I am also thinking we should change this to `-1` to indicate no timeout instead of max value

pkg/orchestrator/event_emitter.go
```diff
@@ -33,38 +32,38 @@ func NewEventEmitter(params EventEmitterParams) EventEmitter {
 
 func (e EventEmitter) EmitJobCreated(
 	ctx context.Context, job models.Job) {
-	event := model.JobEvent{
```
wdbaruni, line 36: can you please add a TODO to deprecate `models.JobEvent` and this event emitter? I don't think it is used anywhere except for logging, and it should be replaced by a jobstore watcher for history events

wdbaruni (changes requested):
Nice work. Just a minor comment about decoupling removal of `model` pkg from removing support of legacy cli flags.

Also I recommend we merge this after 1.4.0 as it is a big change and bugs might slip

pkg/node/config_defaults.go
frrist: I think somewhere we have stated that `0` implies no timeout, wdyt of that pattern?

pkg/orchestrator/event_emitter.go
frrist: I'd love to remove these now, if we don't think they are being used for anything can I remove them now?

cmd/cli/docker/docker_run_cli_test.go
frrist: I can revert this change and land it later, tracking via https://github.com/bacalhau-project/bacalhau/issues/4182

pkg/node/config_defaults.go
wdbaruni: as it stands, 0 means no timeout was defined and will be replaced by default value. I don't know how 0 timeout and 0 default timeout is handled in our system. Also 0 queue-timeout means no queueing enabled.

I'll be concerned about trying to change the meaning of 0 and -1 timeout as part of this change, but just having a `NoJobTimeout` const or var seemed a good approach to easily change things in the future

pkg/orchestrator/event_emitter.go
wdbaruni: My preference is to do it on a separate PR. This will allow you to merge this PR today, which Udit is blocked on for the python sdk work, and will simplify the review of removing the `JobEvent` models

pkg/node/config_defaults.go
frrist: Okay sounds good, will define a default constant and use everywhere.

pkg/orchestrator/event_emitter.go
frrist: Sounds good.

wdbaruni (approved):

PR #4170, opened 2024-06-29
Update version.go to fix running a server built from source

js-ts:
currently when you try to run a server using the binary built from  source and try to run any command it throws this error, this error can be fixed by  stripping the metadata from version before parsing

To run test jobs against the local cluster which uses the build from source binary without getting this error everytime a job is run this error needs to be fixed

bin/darwin/amd64/bacalhau serve --node-type compute --node-type requester

bin/darwin/amd64/bacalhau node list
```
Error: failed request: Unexpected response code: 403 ({
  "Error": "Client version is outdated. Update your client",
  "MinVersion": "1.4.0",
  "ClientVersion": "1.4.0-6-g081eabfba",
  "ServerVersion": "1.4.0-6-g081eabfba"
})
```


frrist (commented):
The problem here is that versions with a pre-release are considered less than versions without a prerelease all other fields equal. i.e. `v1.4.0` > `v1.4.0-abc123`. I believe the correct change here is to strip the pre-release tag from the client before doing the comparison. @wdbaruni wdyt?

pkg/publicapi/middleware/version.go
```diff
@@ -29,6 +30,17 @@ type VersionCheckError struct {
...
+func parseVersion(versionStr string) (*semver.Version, error) {
+	// Strip build metadata from the version string
+	re := regexp.MustCompile(`^(\d+\.\d+\.\d+)(?:-\d+-g[0-9a-f]+)?$`)
+	matches := re.FindStringSubmatch(versionStr)
+	if len(matches) < 2 {
+		return nil, fmt.Errorf("invalid version format")
+	}
+
+	return semver.NewVersion(matches[1])
+}
```
frrist, line 42: Please remove this method, suggestion on replacement below

pkg/publicapi/middleware/version.go
```diff
@@ -64,7 +76,7 @@ func VersionNotifyLogger(logger *zerolog.Logger, serverVersion semver.Version) e
 			}
 
 			// there is a single version header, attempt to parse it.
-			clientVersion, err := semver.NewVersion(cVersion[0])
+			clientVersion, err := parseVersion(cVersion[0])
```
frrist, line 79: We should instead make a client version with the pre-release tag removed via `clientVersion.SetPrerelease("")`

frrist (changes requested):

js-ts:
@frrist

pkg/publicapi/middleware/version.go
```diff
@@ -70,10 +79,18 @@ func VersionNotifyLogger(logger *zerolog.Logger, serverVersion semver.Version) e
...
+				notif.Message = fmt.Sprintf("error stripping pre-release tag from client version: %s", err)
+				return nil
+			}
+
 			// extract parsed client version for comparison
-			notif.ClientVersion = clientVersion.String()
+			notif.ClientVersion = strippedClientVersion.String()
 
-			diff := serverVersion.Compare(clientVersion)
+			diff := serverVersion.Compare(strippedClientVersion)
```
frrist, line 93: we don't want to strip the client version here, this method is for logging the version of the request received from the client. Please revert this change.

pkg/publicapi/middleware/version.go
```diff
@@ -29,11 +29,20 @@ type VersionCheckError struct {
...
 
+func stripPreRelease(version *semver.Version) (*semver.Version, error) {
+	newVersion, _ := semver.NewVersion(version.String())
+	_, err := newVersion.SetPrerelease("")
+	if err != nil {
+		return nil, err
+	}
+	return newVersion, nil
+}
+
```
frrist, line 40:
```suggestion
func stripPreRelease(versionStr string) (*semver.Version, error) {
	v, err := semver.NewVersion(versionStr)
	if err != nil {
		return version.Unknown, fmt.Errorf("failed to parse client version: %w", err)
	}
	strippedVersion, err := v.SetPrerelease("")
	if err != nil {
		return version.Unknown, fmt.Errorf("failed to strip pre-release version from client: %w", err)
	}
	return &strippedVersion, nil
}
```
please move this utility method to the bottom of the file.

pkg/publicapi/middleware/version.go
```diff
@@ -109,12 +126,20 @@ func VersionCheckMiddleware(serverVersion, minVersion semver.Version) echo.Middl
 				})
 			}
 
-			if clientVersion.LessThan(&minVersion) {
```
frrist, line 112: please use the suggested implementation of `stripPreRelease` here to both parse and strip the client version, i.e.
```go
// VersionCheckMiddleware returns a middleware that checks if the client version is at least minVersion.
func VersionCheckMiddleware(serverVersion, minVersion semver.Version) echo.MiddlewareFunc {
	return func(next echo.HandlerFunc) echo.HandlerFunc {
		return func(c echo.Context) error {
			clientVersionStr := c.Request().Header.Get(apimodels.HTTPHeaderBacalhauGitVersion)
			if clientVersionStr == "" ||
				clientVersionStr == version.DevelopmentGitVersion ||
				clientVersionStr == version.UnknownGitVersion {
				// allow the request to pass through
				return next(c)
			}

			clientVersion, err := stripPreRelease(clientVersionStr)
			if err != nil {
				return c.JSON(http.StatusBadRequest, map[string]string{
					"Error": "Invalid client version " + clientVersionStr,
				})
			}

			if clientVersion.LessThan(&minVersion) {
				// Client version is less than the minimum required version
				return c.JSON(http.StatusForbidden, VersionCheckError{
					Error:         "Client version is outdated. Update your client",
					MinVersion:    minVersion.String(),
					ClientVersion: clientVersion.String(),
					ServerVersion: serverVersion.String(),
				})
			}

			// Client version is acceptable, proceed with the request
			return next(c)
		}
	}
}
```

frrist (changes requested):

wdbaruni:
I don't think this is an issue anymore

PR #4193, opened 2024-07-02
`api/v1/orchestrator/jobs/{id}/executions` endpoint, move sorting to job store rather then having it at endpoint level

### LINKED ISSUE #3852: Sort executions in jobstore
The orchestrator's jobExecutions endpoint currently performs sorting of the executions within the API method, and we should move this to the jobstore.

udsamani:
## What is this PR about ?
With this PR we shifting the sorting for executions endpoint a the job store level rather then having at endpoint level.

## GitHub Keywords
closes #3852


pkg/jobstore/boltdb/store.go
```diff
@@ -266,6 +269,35 @@ func (b *BoltJobStore) getExecutions(tx *bolt.Tx, options jobstore.GetExecutions
...
+		sortFnc = func(a, b models.Execution) int { return util.Compare[int64]{}.Cmp(a.CreateTime, b.CreateTime) }
+	case "id":
+		sortFnc = func(a, b models.Execution) int { return util.Compare[string]{}.Cmp(a.ID, b.ID) }
+	case "state":
+		sortFnc = func(a, b models.Execution) int {
+			return util.Compare[models.ExecutionStateType]{}.Cmp(a.ComputeState.StateType, b.ComputeState.StateType)
+		}
+	default:
+		return nil, fmt.Errorf("invalid order_by")
+	}
```
wdbaruni, line 286: It seems we are using "modified_at" and "created_at" for sorting jobs, where "modified_at" is the default option if none specified. It will be great to have a unified standards for listing and sorting our entries.

Though one thing we are doing wrong with sorting jobs is we are using "create_at" even if they passed an invalid order_by value.

Lets try to unify the logic between these two, and try to keep backward compatibility so that "modify_time" would still work for now

Also with "less is more" principal, can we remove sorting by _id_ and _state_ if we are not using them internally? We are not doing a great work in indexing our entries today and we are sorting everything in-memory after retrieval. I want to reduce our requirements when we try to implement this the proper way

wdbaruni (changes requested):

pkg/jobstore/boltdb/store.go
wdbaruni: Also please add unit tests under `store_test.go`

udsamani: Going with `create_time` and `modify_time` as our docutmentation is using that https://docs.bacalhau.org/references/api/jobs.

Ideally we are using `created_at` and `modified_at` internally, and externally we have exposed `create_time` and `modified_time`. 

Just to comply to our documentation using it, and making sure internally also we use that.

Moreover I cannot get rid of `id` and `state` as it is something published in our documentation ? https://docs.bacalhau.org/references/api/jobs#job-executions ! If you still feel that we need to get rid of `id` and `state` we will have to take care of the documentation too.

wdbaruni: don't worry much about our documentation for now. These are relatively new APIs, and specifically the `executions` APIs are not widely used

I am okay with either `create_time` or `create_at` as long as we are consistent

udsamani: Makes sense. In that case, getting rid of `id` and `state` for `executions` do you want me to do the same for list jobs ?

wdbaruni (approved):

PR #4194, opened 2024-07-02
Fix Python SDK

### LINKED ISSUE #4164: Implement Python SDK for new models and api changes

udsamani:
This PR does the following

- Run `swag fmt` on `pkg/publicapi/orchestrator/` this formats all API comments on endpoints.
- Generate and update the swager `go generate pkg/swaggger/generate.sh`
- Build the apiclient `make build-python-apiclient`
- Fix the SDK meaning all files under `python/bacalhau_sdk`
- Update README.md for `python/` folder.

Github Keywords
close #4164


pkg/swagger/docs.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by swaggo/swag. DO NOT EDIT
+// Package swagger Code generated by swaggo/swag. DO NOT EDIT
```
wdbaruni, line 1: I know we just discussed the benefits and potential of generating the swagger definitions manually instead of using swaggo, but looking at this huge file it scares me to manage it manually. Specially that it includes specifications of things like `models.Job` and `models.Executions`. 

Is there a middle ground where the models definitions are auto-generated, but the API specs are manually generated and documented? We don't need to think about this now, but something for later when we decide to improve our clients generation

python/bacalhau_sdk/job_store.py
```diff
@@ -0,0 +1,152 @@
...
+    ApiListJobsResponse as ListJobsResponse,
+)
+from bacalhau_apiclient.models.api_get_job_response import (
+    ApiGetJobResponse as GetJobResponse,
+)
+
+from bacalhau_sdk.config import init_config
+
+
+class JobStore:
```
wdbaruni, line 31: Nice structure of resource based SDKs. Though can we call them `Jobs` similar to our [go sdk](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/publicapi/client/v2/api_jobs.go#L13). `JobStore` is in internal component related to persisting and retrieving job state, and can be confusing here

python/bacalhau_sdk/job_store.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Args:
+            request (PutJobRequest): A request to put a job to bacalhau network. It encapsulates the job model.
+
+        Returns:
+            PutJobResponse: Once job is successful put on bacalhau netowork, this returns the job details.
+        """
+        return self.orchestrator_service.put_job(request)
+
+    def stop(self, job_id: str, reason: str = None) -> StopJobResponse:
```
wdbaruni, line 46: I forgot about the standards in python, but should they be `stop` or `Stop`?

python/bacalhau_sdk/api.py
```diff
@@ -1,121 +1,44 @@
...
-        state_request = StateRequest(
-            client_id=get_client_id(),
-            job_id=job_id,
-        )
-        api_response = api_instance.results(state_request)
-    except ApiException as e:
-        print("Exception when calling JobApi->results: %s\n" % e)
-    return api_response
+    """Return Job Results"""
+    return orchestrator_service.job_results(id=job_id)
```
wdbaruni, line 34: can we just delete all these legacy methods? I really don't want to maintain this code if we don't need to, sepially that we broke `submit`, which is the most important one

scripts/generate_swagger.sh
```diff
@@ -2,6 +2,7 @@
 PATH_TO_PROJECT_ROOT=$(git rev-parse --show-toplevel)
 SWAGGER_DIR="${PATH_TO_PROJECT_ROOT}/pkg/swagger"
 PUBLIC_PATH="${PATH_TO_PROJECT_ROOT}/webui/public/swagger"
+DOCS_PATH="${PATH_TO_PROJECT_ROOT}/docs"
```
wdbaruni, line 5: can we revisit the structure of these directories? I really don't understand why we are involving `webui` in this. I think it was a wrong decision at some point and I would like to see if we can avoid that

Also does it doesn't make sense to have the swagger generation code under `/pkg/swagger` as `/pkg` should only include code related to bacalhau and not for external tooling

wdbaruni (commented):
LGTM. Just minor comments

python/bacalhau_sdk/job_store.py
udsamani: Python is snake case so `stop`.

scripts/generate_swagger.sh
udsamani: Makes sense. I am creating an issue to fix it in V2 SDK. #4200 .

clients/python/README.md
```diff
@@ -3,8 +3,8 @@ This page is the reference of the Bacalhau REST API. Project docs are available
 
 This Python package is automatically generated by the [Swagger Codegen](https://github.com/swagger-api/swagger-codegen) project:
 
-- API version: ${PYPI_VERSION}
-- Package version: 0.3.24.dev8
+- API version: ${VERSION}
+- Package version:
```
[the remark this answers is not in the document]
udsamani: It will be same as our git version. 

https://github.com/bacalhau-project/bacalhau/blob/main/scripts/get_current_version_for_builds.py

This is the script that gets used.

wdbaruni (approved):

Issue #4168, opened 2024-06-28
Test `TestStreamConsumerClientGoingDown` is flaky


wdbaruni:
Examples:
https://app.circleci.com/pipelines/github/bacalhau-project/bacalhau/13228/workflows/1a54ca69-849c-4b43-8bb5-5f0ecb22c3ec/jobs/177536/steps

https://app.circleci.com/pipelines/github/bacalhau-project/bacalhau/13228/workflows/1a54ca69-849c-4b43-8bb5-5f0ecb22c3ec/jobs/177537/tests

https://app.circleci.com/pipelines/github/bacalhau-project/bacalhau/13228/workflows/1a54ca69-849c-4b43-8bb5-5f0ecb22c3ec/jobs/177534/steps

udsamani:
Easy to reproduce the test flakines.


![Image](https://github.com/bacalhau-project/bacalhau/assets/47295083/dc7babb3-9793-4ff8-b39d-8975f64219a0)

PR #4199, opened 2024-07-03
Remove Un-necessary tries to mark execution complete when it has already been cancelled

### LINKED ISSUE #3698: Incorrect state when cancelling job
When calling `bacalhau cancel` on a job, the logs show the following info, but two errors to do with the state of the job.
It's almost like the job is being cancelled twice with the second entry complaining ...

```
19:13:12.103 | INF pkg/publicapi/middleware/logger.go:50 > 9128751d-7748-40a8-9c0c-c9c7ad178081] [Method:POST] [NodeID:node-0] [Referer:] [RemoteAddr:127.0.0.1] [RequestID:] [Size:702] [StatusCode:200] [URI:/api/v1/requester/cancel] [UserAgent:Go-http-client/1.1]
19:13:12.109 | INF pkg/executor/docker/handler.go:172 > killing the container [container:dc451521008e2437d6d54e8058d3dc5909d9031c6d717f188c573487535c73f8] [execution:e-50dd14d0-c873-43b7-a49a-bda5f86889ce] [job:9128751d-7748-40a8-9c0c-c9c7ad178081]
...
19:13:13.233 | INF pkg/executor/docker/handler.go:189 > removing container [container:dc451521008e2437d6d54e8058d3dc5909d9031c6d717f188c573487535c73f8] [execution:e-50dd14d0-c873-43b7-a49a-bda5f86889ce] [job:9128751d-7748-40a8-9c0c-c9c7ad178081]
19:13:13.242 | INF pkg/executor/docker/executor.go:217 > received results from execution [executionID:e-50dd14d0-c873-43b7-a49a-bda5f86889ce]
19:13:13.242 | INF pkg/compute/executor.go:195 > cleaning up execution [NodeID:node-1] [execution:e-50dd14d0-c873-43b7-a49a-bda5f86889ce] [job:9128751d-7748-40a8-9c0c-c9c7ad178081]
....
19:13:13.242 | ERR pkg/compute/executor.go:480 > Publishing execution e-50dd14d0-c873-43b7-a49a-bda5f86889ce failed error="execution e-50dd14d0-c873-43b7-a49a-bda5f86889ce is in state Cancelled but expected one of [Running]" [NodeID:node-1] [execution:e-50dd14d0-c873-43b7-a49a-bda5f86889ce] [job:9128751d-7748-40a8-9c0c-c9c7ad178081]
....
19:13:13.242 | ERR pkg/compute/executor.go:488 > Failed to update execution (e-50dd14d0-c873-43b7-a49a-bda5f86889ce) state to failed: execution e-50dd14d0-c873-43b7-a49a-bda5f86889ce is in terminal state Cancelled and cannot transition to Failed error="execution e-50dd14d0-c873-43b7-a49a-bda5f86889ce is in terminal state Cancelled and cannot transition to Failed" [NodeID:node-1] [execution:e-50dd14d0-c873-43b7-a49a-bda5f86889ce] [job:9128751d-7748-40a8-9c0c-c9c7ad178081]
```

udsamani:
This PR aims at the following

- Make sure we do no try to mark the execution as Completed if the user has already cancelled it.

closes #3698 #4168


wdbaruni:
I have realized bunch of other issues with our execution cancellation logic that are more serious. It seems we have a race condition where we can mark an execution as completed instead of cancelled.

Inside `endpoint`, we trigger the execution cancellation first, which is then carried in a go routine, and then update the state of the execution to cancelled:
https://github.com/bacalhau-project/bacalhau/blob/a766d1cc26fd9c135f35269ce434403016728b6c/pkg/compute/endpoint.go#L143-L154

https://github.com/bacalhau-project/bacalhau/blob/a766d1cc26fd9c135f35269ce434403016728b6c/pkg/compute/executor_buffer.go#L201-L216

If executor cancellation was fast, then it will actually mark it as completed instead of cancelled. I was able to reproduce this by injecting a sleep

What we should do is, first update the execution state, and then cancel the executor. 

In addition to that, `e.Wait` doesn't return an error or any indication that it was interrupted or cancelled, which lets the executor to think it was successful and then attempts to publish results and mark it as completed successfully. We should also make `Wait` return an error if the execution didn't complete all the way
https://github.com/bacalhau-project/bacalhau/blob/a766d1cc26fd9c135f35269ce434403016728b6c/pkg/compute/executor.go#L336-L338

pkg/compute/executor.go
```diff
@@ -300,7 +302,7 @@ func (e *BaseExecutor) Run(ctx context.Context, state store.LocalExecutionState)
 	stopwatch := telemetry.Timer(ctx, jobDurationMilliseconds, state.Execution.Job.MetricAttributes()...)
 	topic := EventTopicExecutionRunning
 	defer func() {
-		if err != nil {
+		if err.Error() != executor.ErrAlreadyCancelled.Error() {
```
wdbaruni, line 305: better to check using `!errors.Is(err, executor.ErrAlreadyCancelled)`

pkg/compute/executor.go
```diff
@@ -392,6 +394,17 @@ func (e *BaseExecutor) Run(ctx context.Context, state store.LocalExecutionState)
...
+	// by the user. In such case we should not mark it as completed.
+	currentExecutionState, err := e.store.GetExecution(ctx, execution.ID)
+	if err != nil {
+		return err
+	}
+	if currentExecutionState.State == store.ExecutionStateCancelled {
+		log.Ctx(ctx).Debug().Msgf("Execution has already been cancelled for %s", execution.ID)
+		return nil
+	}
+
```
wdbaruni, line 407: do we still need this check since cancelling the execution would trigger an `ErrAlreadyCancelled`

wdbaruni (commented):
LGTM. Just minor comments. How difficult to add a test case to verify the changes work as expected?

udsamani:
> LGTM. Just minor comments. How difficult to add a test case to verify the changes work as expected?

Let me take a stab at adding tests :)

pkg/compute/executor.go
udsamani: So the problem there, is that we return a new error, by creating one from the one we get from [result](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/compute/executor.go#L355) As a consequence of which it does not match the error type which we have created.

So this is what is happening

ErrAlreadyCanceled (at docker layer) -> Translated to string message and added to `RunCommanResult` -> And a new error type created.

We don't have a mapping of any error messages to specific type at computer/executor.go layer.

Do you have any other idea we can use. ?

pkg/test/compute/cancel_job_test.go
```diff
@@ -0,0 +1,119 @@
...
+		devstack.WithNumberOfComputeOnlyNodes(1),
+		devstack.WithComputeConfig(computeConfig),
+		devstack.WithRequesterConfig(requesterConfig),
+	)
+
+	nodeutils.WaitForNodeDiscovery(s.T(), stack.Nodes[0].RequesterNode, 2)
+
+	es, err := dockermodels.NewDockerEngineBuilder("ubuntu").
+		WithEntrypoint("bash", "-c", "sleep 100000").
+		Build()
```
wdbaruni, line 72: I would be careful about sleeping too much where failures would block or delay the test termination

pkg/test/compute/cancel_job_test.go
```diff
@@ -0,0 +1,119 @@
...
+	apiPort := apiServer.Port
+	api := clientv2.New(fmt.Sprintf("%s://%s:%d", apiProtocol, apiHost, apiPort))
+
+	submittedJob, err := api.Jobs().Put(ctx, &apimodels.PutJobRequest{
+		Job: job,
+	})
+	s.Require().NoError(err)
+
+	resolver := scenario.NewStateResolverFromAPI(api)
+	time.Sleep(time.Second * 10)
```
wdbaruni, line 95: We don't want to slow our test suites more than how slow they are already. You may want to look into `noop_executor` instead of docker where you can inject a job handler. You can use and block on a channel and have the job handler add a value to the channel to indicate the execution started, and then sleep

pkg/test/compute/cancel_job_test.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	s.Require().NoError(err)
+
+	resolver.Wait(ctx, submittedJob.JobID, func(js *scenario.JobState) (bool, error) {
+		s.Require().Len(js.Executions, 1)
+		if js.Executions[0].ComputeState.StateType == models.ExecutionStateCancelled {
+			return true, nil
+		} else {
+			return false, nil
+		}
```
wdbaruni, line 116: based on the issue that this PR is fixing, we are actually more interested in knowing the compute's local state and not the requester's state, which is what you are checking here.

My recommendation is to test using just a compute node and not a full devstack. There is already a base test suite that you can build on top, similar to how https://github.com/bacalhau-project/bacalhau/blob/7332b2a8890bb90539bc54bf18839e136a488e7f/pkg/test/compute/bid_accepted_test.go is implemented.

This will give you access to the compute node's endpoint to cancel the job after submission, and access to the compute node's execution store to validate the state.

However, if there are better tooling for devstack, then you can still access the compute node using `stack.Nodes[<index>]`

wdbaruni (changes requested):
some minor changes to the test suite

pkg/test/compute/cancel_job_test.go
udsamani: I can't do that right. The code changes are part of docker execution engine. If i don't use docker, i won't be testing the right thing. Here we are making changes to specific engine. 

Also my bad, i don't need that sleep there ! Will get rid of it !

wdbaruni (approved):
LGTM. Just a minor comment

pkg/nats/stream/streaming_client_interactions_test.go
```diff
@@ -1,21 +1,24 @@
-//go:build unit || !integration
+//go:build integration || !unit
```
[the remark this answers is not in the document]
udsamani: Done :)

pkg/executor/docker/executor_test.go
```diff
@@ -401,6 +402,54 @@ func (s *ExecutorTestSuite) TestDockerNetworkingFiltersHTTPS() {
...
+	s.Require().NoError(err)
+
+	select {
+	case err := <-errC:
+		s.Require().Nil(err)
+	case result := <-resultC:
+		s.Require().NotNil(result)
+		s.Require().Equal(executor.ErrAlreadyCancelled.Error(), result.ErrorMsg)
+	}
+}
```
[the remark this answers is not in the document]
udsamani: Done :)

PR #4223, opened 2024-07-09
Pagination Support for Job History

### LINKED ISSUE #4220: Support Pagination for fetching History from JobStore

udsamani:
This PR aims at the following

- Add support for pagination at JobStore layer for job histories

closes #4220

### DIFF
--- a/pkg/publicapi/endpoint/orchestrator/job.go
+++ b/pkg/publicapi/endpoint/orchestrator/job.go
@@ -289,7 +289,7 @@ func (e *Endpoint) jobHistory(c echo.Context) error {
 		return err
 	}
 
-	var limit uint32
+	limit := args.Limit
 	var offset uint32
 	var err error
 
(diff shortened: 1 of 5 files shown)


pkg/publicapi/endpoint/orchestrator/job.go
```diff
@@ -92,7 +92,8 @@ func (e *Endpoint) getJob(c echo.Context) error { //nolint: gocyclo
 			if response.History != nil {
 				continue
 			}
-			history, err := e.store.GetJobHistory(ctx, jobID, jobstore.JobHistoryFilterOptions{})
+			jobHistoryQueryResponse, err := e.store.GetJobHistory(ctx, jobID, jobstore.JobHistoryQuery{})
+			history := jobHistoryQueryResponse.JobHistory
```
wdbaruni, line 96: you should do nil check validation first

wdbaruni (approved):
LGTM. Just a minor comment

pkg/publicapi/endpoint/orchestrator/job.go
udsamani: Done.

frrist: I think we should check the error before assigning `history := jobHistoryQueryResponse.JobHistory` else this introduce the potential for panics if `GetJobHistory` returns `nil` and an `err`, which is a common practice.

Issue #4218, opened 2024-07-09
Replace `repo.version` and `update.json` with `system_metadata.yaml`


frrist:
- The `system_metadata.yaml` file MUST exist and MUST contain the `RepoVersion` field.
- The `system_metadata.yaml` MAY contain the following optional fields: 
  - `InstallationID`  
  - `LastUpdateCheck`

PR #4237, opened 2024-07-15
implement system_metadata.yaml replaces repo.version

frrist:
This change replaces `repo.version` with `system_metadata.yaml` based on the criteria defined in https://github.com/bacalhau-project/bacalhau/issues/4218. Further it includes a migration that does the following:
- Updates the repo version to `4` and sets it in `system_metadata.yaml`
- moves the installationID to `system_metadata.yaml` if one is present in the config.
- delete the `update.json` file, its replacement is `system_metadata.yaml`

Additionally the update check has been modified to use an `UpdateStore` interface for reading and writing the `LastUpdateTime` field from the `system_metadata.yaml` file. The FsRepo satasfies this interface.

Lastly the UpdateCheckStatePath field has been removed from the config. A migration for this is not possible as we no longer permit users setting the field


pkg/repo/legacy.go
```diff
@@ -0,0 +1,41 @@
...
+
+// writeLegacyVersion writes the repo version to LegacyVersionFile.
+func (fsr *FsRepo) writeLegacyVersion(version int) error {
+	repoVersion := Version{Version: version}
+	versionJSON, err := json.Marshal(repoVersion)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(filepath.Join(fsr.path, LegacyVersionFile), versionJSON, util.OS_USER_RW)
+}
```
wdbaruni, line 41: do we still need this?

pkg/repo/migration.go
```diff
@@ -34,9 +34,12 @@ func (m *MigrationManager) Add(migration Migration) error {
...
 func (m *MigrationManager) Migrate(repo FsRepo) error {
-	currentVersion, err := repo.readVersion()
+	currentVersion, err := repo.readLegacyVersion()
 	if err != nil {
-		return err
+		currentVersion, err = repo.readVersion()
+		if err != nil {
+			return err
+		}
 	}
```
wdbaruni, line 43: I am looking at `repo.readVersion()` implementation, and it falls back to `readLegacyVersion()` if no version was read. Should we just keep the fallback logic inside `readVersion()` instead of repeating it here as well?

pkg/repo/migrations/base_test.go
```diff
@@ -7,13 +7,14 @@ import (
...
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/suite"
+
+	"github.com/bacalhau-project/bacalhau/pkg/repo"
 )
 
 const (
-	expectedRepoVersion = repo.RepoVersion3
+	expectedRepoVersion = repo.Version3
 )
```
wdbaruni, line 18: should this be version 4?

pkg/repo/migrations/v3_4.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/bacalhau-project/bacalhau/pkg/repo"
+)
+
+var V3Migration = repo.NewMigration(
```
wdbaruni, line 11: please add comments about this migration does, similar to how we document previous migrations

pkg/repo/migrations/v3_4.go
```diff
@@ -0,0 +1,52 @@
...
+
+		_, fileCfg, err := readConfig(r)
+		if err != nil {
+			return err
+		}
+		if fileCfg.User.InstallationID != "" {
+			if err := r.WriteInstallationID(fileCfg.User.InstallationID); err != nil {
+				return err
+			}
+		}
```
wdbaruni, line 43: we are going to have a lot of fun with migrations when we change the config structure. Maybe we can keep a copy of the old config structs inside migrations pkg. Will need to figure this out later

pkg/repo/sysmeta.go
```diff
@@ -0,0 +1,141 @@
...
+	"os"
+	"time"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/bacalhau-project/bacalhau/pkg/storage/util"
+)
+
+type SystemMetadata struct {
+	RepoVersion     int       `yaml:"Version"`
```
wdbaruni, line 14: Looking at the yaml tag, `Version` can be confusing as it might indicate bacalhau version. Since it represents the version of the `DataDir`, maybe we should call it `DataDirVersion`

pkg/repo/sysmeta.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := os.WriteFile(fsr.join(SystemMetadataFile), metaBytes, util.OS_USER_RW); err != nil {
+		return fmt.Errorf("writing repo system metadata file: %w", err)
+	}
+
+	return nil
+}
+
+// updateOrCreateMetadata updates an existing metadata file or creates a new one if it doesn't exist.
+// The update is applied using the provided updateFunc.
+func (fsr *FsRepo) updateOrCreateMetadata(updateFunc func(*SystemMetadata)) error {
```
wdbaruni, line 89: Nice helper method

pkg/version/update.go
```diff
@@ -150,15 +156,12 @@ func RunUpdateChecker(
...
-			err = writeNewLastCheckTime(cfg.Update.CheckStatePath)
+			err = store.WriteLastUpdateCheck(time.Now())
 			log.Ctx(ctx).WithLevel(logger.ErrOrDebug(err)).Err(err).Msg("Completed update check")
 		}
 	}
 
-	// TODO(forrest) [efficiency]: rather than repeatedly reading a file, set the value in a cache and flush the cache
-	// to disk when this service shutwdown, no reason for this IO.
-	lastCheck, err := readLastCheckTime(cfg.Update.CheckStatePath)
+	lastCheck, err := store.ReadLastUpdateCheck()
```
wdbaruni, line 164: what changed your mind about the TODO statement?
> // TODO(forrest) [efficiency]: rather than repeatedly reading a file, set the value in a cache and flush the cache
	// to disk when this service shutwdown, no reason for this IO.

wdbaruni (approved):
Looks good. Just minor comments

pkg/repo/legacy.go
frrist: Yes for the case were we are mitigating a repo before version 3.

pkg/repo/migration.go
frrist: Yes good idea, will adjust.

pkg/repo/migrations/base_test.go
frrist: I don't think so, this variable is only used by the v2_3 migration test suite

pkg/repo/migrations/v3_4.go
frrist: will do.

PR #4238, opened 2024-07-16
feat: generate an instanceID

frrist:
- The ID is based on the host running the bacalhau node, and is deterministic across re-installs and node initalization. There is no guarantee setting the ID will succeed, its a best effort.

### DIFF
--- a/pkg/repo/instanceid.go
+++ b/pkg/repo/instanceid.go
@@ -1,0 +1,99 @@
+package repo
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/denisbrodbeck/machineid"
+)
+
+// GenerateInstanceID creates a unique, anonymous identifier for the instance of bacalhau.
+// It combines the machine ID and MAC address to ensure uniqueness across different
+// environments, including virtual machines and cloud instances.
+//
+// The function aims to produce consistent results on repeated calls on the same host,
+// but there are scenarios where the result may differ:
+//
+//  1. Network Configuration Changes: If network interfaces are added, removed, or their
+//     order changes, a different MAC address might be selected.
+//  2. Virtual Environments: In VMs or containers, network configurations might change
+//     between runs, affecting the selected MAC address.
+//  3. MAC Address Randomization: Some systems implement MAC address randomization for
+//     privacy, which could cause the function to return different results.
+//  4. Hardware Changes: Adding or removing network adapters could alter the result.
+//  5. System Updates: Major system updates might affect how machine IDs are generated
+//     or how network interfaces are enumerated.
+//
+// The function uses SHA-256 hashing to maintain user anonymity by not exposing actual
+// hardware identifiers. The resulting ID is a 64-character hexadecimal string.
+func GenerateInstanceID() (string, error) {
+	var elements []string
+
+	// Get machine ID, which provides a consistent identifier for the device
+	machineID, err := machineid.ID()
+	if err != nil {
+		return "", fmt.Errorf("failed to get machine ID: %w", err)
+	}
+	elements = append(elements, machineID)
+
+	// Get MAC address to add another layer of uniqueness, especially useful in virtualized environments
+	// or cloned disk images which share a machineID.
+	macAddr, err := getMacAddress()
+	if err != nil {
+		return "", fmt.Errorf("failed to get MAC address: %w", err)
+	}
+	elements = append(elements, macAddr)
+
+	// Combine all elements into a single string
+	combined := strings.Join(elements, "|")
+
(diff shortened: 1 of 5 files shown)


pkg/repo/instanceid.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/denisbrodbeck/machineid"
+)
+
+// GenerateInstanceID creates a unique, anonymous identifier for the instance of bacalhau.
+// It combines the machine ID and MAC address to ensure uniqueness across different
+// environments, including virtual machines and cloud instances.
```
wdbaruni, line 15: Maybe worth clarifying in the comment:
We only attempt to generate an instance id if no id was found the `system_metadata`, which can happen during new node initialization or if the DataDir was wiped. We attempt to generate deterministic id for analytics purposes and track the usage of the instance.
 
-- 

Now since it is best effort, rather than failing because of errors when trying to extract the machine id or mac address, why don't we fallback to `hash(uuid)`? This will be sufficient 99.999% of the time since we do persist it in the system_metadata

My other question is, if the user is running a node or a client against different networks with different `DataDir`, we will assign them the same InstanceID. Maybe this is the desired behaviour, but worth thinking about

wdbaruni (commented):
The diff is mixed with metadata changes, but I reviewed the instanceID commit. Looks good overall, but left a comment to fallback to uuid

pkg/repo/instanceid.go
frrist: There are several clarifying comments related to when we generate the instanceID, and how we persist it in the repo Open and Init methods. I think usage is well covered there. (It should be easier to grok now that I've fixed the mixed commits :grinning:)

> Now since it is best effort, rather than failing because of errors when trying to extract the machine id or mac address, why don't we fallback to hash(uuid)?

What value would including this random string (UUID) add? 
The purpose of the instanceID is to identify (anomalously) unique instances of bacalhau. I'd much prefer an empty value - which implies a value couldn't be derived - than a random string with no relationship to the actual instance. My preference is to leave the field empty if it couldn't be derived.

> This will be sufficient 99.999% of the time since we do persist it in the system_metadata

I've made a small change since your last review: We now generate and persist an instanceID when the repo is first initialized and **also each time the repo is opened**. This ensures the instanceID is associated with the machine running bacalhau, even if the repo is copied to a different machine. In this sense we are treating the `system_metadata.yaml` file a bit like a cache. In work to follow this, we can read the instanceID from the `system_metadata.yaml` file and include it in the job metadata and/or HTTP headers when an instance interacts with the network.

> My other question is, if the user is running a node or a client against different networks with different DataDir, we will assign them the same InstanceID. Maybe this is the desired behaviour, but worth thinking about

Right, I think this behavior makes sense. The same instance of bacalhau is free to interact with many networks. A measure of this interaction could be derived be counting the number of different networks an instance has interactions with.

wdbaruni: > What value would including this random string (UUID) add?
> The purpose of the instanceID is to identify (anomalously) unique instances of bacalhau. I'd much prefer an empty value - which implies a value couldn't be derived - than a random string with no relationship to the actual instance. My preference is to leave the field empty if it couldn't be derived.

How a _persisted_ UUID doesn't provide a relationship to the instance? Our goal is to identify the instance the jobs are submitted from (client) or executing the job (node). Assigning and persisting a UUID when you first initialize the client/node is more than enough and fulfils this purpose. What you are trying to do here with generating idempotent IDs is just to handle the edge cases where the repo is deleted and we still want to tag jobs with the same instance, which is great, but lets treat it as best effort and fallback to UUID instead of not tracking at all

Keep in mind this means we only attempt to generate an ID if no ID was found in the system_metadata.yaml, and don't attempt to override the value

wdbaruni: Can you address these comments so we can close the PR?

frrist: > this means we only attempt to generate an ID if no ID was found in the system_metadata.yaml, and don't attempt to override the value

Sounds good, will address.

wdbaruni:
Lets avoid having PRs open for too long. It gets more difficult to review things again after weeks, and I am sure it gets difficult for you as well

frrist:
I will close this and re-open when there is a strong need for it.

wdbaruni:
@frrist there is a strong need for having a persisted InstanceID in the job spec as we've removed the ClientID, and not persisting the InstallationID anymore

pkg/repo/instanceid.go
frrist: replaced by https://github.com/bacalhau-project/bacalhau/pull/4411

frrist:
superseded by https://github.com/bacalhau-project/bacalhau/pull/4411

PR #4294, opened 2024-07-31
Integration for Buildkite

### LINKED ISSUE #4297: Get a basic build pipeline running for golang

udsamani:
This PR aims at the following

- Create a buildkite-hosted-agent for bacalhau
- Create a bacalhau-golang pipeline
- Add buildkite scripts for different components

closes #4297


buildkite/scripts/test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+export LOG_LEVEL=DEBUG
+export TEST_BUILD_TAGS=$1
+export TEST_PARALLEL_PACKAGES=4
+
+make build-webui
+make test-and-report
```
wdbaruni, line 8: do we need to build the webui with each test?

docker/buildkite-hosted-agent-bacalhau/Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM buildkite/hosted-agent-base
```
wdbaruni, line 1: Nice. Much cleaner that circleci setup. Can we put everything related to buildkite under `/buidlkite` directory, including this Dockerfile?

buildkite/pipelines/bacalhau-golang.yaml
```diff
@@ -0,0 +1,64 @@
+cache:
+  paths:
+    - "/golangci-lint-cache"
+    - "/vendor-cache"
+    - "/node-modules-cache"
+  size: 20g
+  name: "vendor-module-cache"
```
wdbaruni, line 7: how is this cache used? We also want to make sure we build once, and then use the builds in later phases, such as running unit tests, integration tests, and releasing them if this build includes a release

wdbaruni: also do we pull the code once, or every step in the pipeline pulls the code again?

buildkite/pipelines/bacalhau-golang.yaml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - label: ":golang: Build linux amd64"
+        command: "./buildkite/scripts/build_tarball.sh linux amd64"
+        agents:
+          queue: "buildkite-hosted-linux-small"
+          
+      - label: ":golang: Build linux arm64"
+        command: "./buildkite/scripts/build_tarball.sh linux arm64"
+        agents:
+          queue: "buildkite-hosted-linux-small"
```
wdbaruni, line 27: how are steps in the same group executed? Do each require spinning up their own container, which means slower start and cost, or are they run within the same agent/container? I am double checking because we define the agents in each step, which implies they run on separate environments

buildkite/pipelines/bacalhau-golang.yaml
```diff
@@ -0,0 +1,64 @@
...
+    command: "./buildkite/scripts/test.sh unit"
+    key: "unit-test"
+    agents:
+      queue: "buildkite-hosted-linux-large"
+      
+  - label: ":testengine: Integration Test"
+    command: "./buildkite/scripts/test.sh integration"
+    key: "integration-test"
+    agents:
+      queue: "buildkite-hosted-linux-large"
```
wdbaruni, line 64: eventually we want to run on different environments, including windows, linux arm, linux amd, and darwin.

Also, separating unit and integration tests into separate steps was done before because of consistent failures in integration tests and to be able to retry only the failed steps in circleci instead of the whole tests again. We can continue to do the same here as well if it makes sense, but we should re-evaluate if this is adding unnecessary cost

buildkite/scripts/build_tarball.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+GOOS=$1 GOARCH=$2 make build-bacalhau-tgz
```
wdbaruni, line 3: I looked at the build logs https://buildkite.com/expanso/bacalhau-golang/builds/291#01910786-63c9-4a7e-b38d-d12c63b50496

some observations:
- It seems we are building the webui everytime. This is an issue with make target setup
- I see some failures at the end related to signing the builds because of missing /tmp/private.pem. Check https://github.com/bacalhau-project/bacalhau/blob/e71b30bf47fc6bfc9e19d4be976f9996e3426d88/.circleci/config.yml#L317 on how we were doing things with circleci

buildkite/pipelines/bacalhau-golang.yaml
udsamani: Every step in pipeline unfortunately pulls code :( But we use git mirroring which fastens up. 

If you want to do just once. You cant take advantage of parallel builds. 

Circle CI was the same too.

buildkite/pipelines/bacalhau-golang.yaml
udsamani: They are on different agents. Spinning up the agent is pretty fast. It stays in queue for max 5-6s. Group step only helps seeing things visually !

buildkite/scripts/build_tarball.sh
udsamani: Yess ! I want to have at least one green build with buildkite before I merge this. I am debugging those.

pkg/executor/docker/executor_test.go
```diff
@@ -381,6 +384,7 @@ func (s *ExecutorTestSuite) TestDockerNetworkingFiltersHTTP() {
 }
 
 func (s *ExecutorTestSuite) TestDockerNetworkingFiltersHTTPS() {
+	s.T().Skip("Skipping the test until buildkite is fixed.")
```
wdbaruni, line 387: nit: will be nice to only skip in buildkite and let them run when testing locally. Env variables can help with that

pkg/executor/docker/gateway/gateway.sh
```diff
@@ -29,12 +29,6 @@ while IFS= read -r BRIDGE_SUBNET; do
     iptables -A INPUT -p tcp --src "${BRIDGE_SUBNET}" --dport 8080 -j ACCEPT
 done < <(cat /etc/bacalhau/allowed-clients.txt)
 
-# Apply rate limits to the outbound connections. We just do this for all
-# interfaces rather than working out which is our Internet connection.
-while IFS= read -r IFACE; do
-    tc qdisc add dev "${IFACE}" root tbf rate 10mbit burst 32kbit latency 10sec
-done < <(ip --json address show | jq -rc '.[] | .ifname')
-
```
wdbaruni, line 37: do we need to remove these even if we are skipping the tests? If yes, then please add an issue to make sure we fix this way before 1.5

wdbaruni (approved):

pkg/executor/docker/gateway/gateway.sh
udsamani: Done. We don't need to remove this. So reverted it.

PR #4315, opened 2024-08-12
validate duplicate inputs

### LINKED ISSUE #2375: Multiple inputs fail badly when writing to same target
Running a job with multiple inputs, where they all write to the same folder can fail with an obscure error.

```
bacalhau \
    docker run \
    -i src=ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/,dst=/inputs/data \
    -i src=ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/,dst=/inputs/data \
ubuntu echo hello
```

fails (in logs/events) with 

```
failed to copy /inputs/data to volume: failed to write to ''/tmp/bacalhau-ipfs900595700/QmbyYag1VXo45qPG9UVfbF6hVe5aHj2pR61voFeuA7pTH8'':
      path already exists and overwriting is not allowed
```

We should investigate whether it is possible to validate the inputs earlier to avoid the users waiting for something that will certainly fail.

wdbaruni:
Fail fast when there are collisions in job inputs and result paths, such as multiple input sources are mounted to the same local path.

#### Before the fix
```
→ bacalhau \
    docker run \
    -i src=ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/,dst=/inputs/data \
    -i src=ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/,dst=/inputs/data \
ubuntu echo hello
failed to copy /inputs/data to volume: failed to write to ''/tmp/bacalhau-ipfs900595700/QmbyYag1VXo45qPG9UVfbF6hVe5aHj2pR61voFeuA7pTH8'':
      path already exists and overwriting is not allowed
```

#### After the fix
```
→ bacalhau \
    docker run \
    -i src=ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/,dst=/inputs/data \
    -i src=ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/,dst=/inputs/data \
ubuntu echo hello
building job spec: failed to create job: input source with alias 'ipfs://qmbrr4kuxmxqfzpnlusb1ksmdvftbucv1hvsdaauke4epj/' already exists
input source with target '/inputs/data' already exists

```

Closes #2375

### DIFF
--- a/pkg/models/input_source.go
+++ b/pkg/models/input_source.go
@@ -46,6 +46,7 @@ func (a *InputSource) Copy() *InputSource {
 	}
 	return &InputSource{
 		Source: a.Source.Copy(),
+		Alias:  a.Alias,
 		Target: a.Target,
 	}
 }
--- a/pkg/models/task.go
+++ b/pkg/models/task.go
@@ -139,12 +139,40 @@ func (t *Task) ValidateSubmission() error {
 		mErr = errors.Join(mErr, errors.New("publisher must be set if result paths are set"))
 	}
 
+	// Check for collisions in input sources
 	seenInputAliases := make(map[string]bool)
+	seenInputTargets := make(map[string]bool)
 	for _, input := range t.InputSources {
-		if input.Alias != "" && seenInputAliases[input.Alias] {
-			mErr = errors.Join(mErr, fmt.Errorf("input source with alias %s already exist", input.Alias))
+		if input.Alias != "" {
+			if seenInputAliases[input.Alias] {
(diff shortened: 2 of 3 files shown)


udsamani (approved):
LGTM :)

PR #4326, opened 2024-08-12
Release Pipeline Readiness for Go Binaries

udsamani:
This PR aims at the following
- Release Golang binaries.


buildkite/pipelines/bacalhau-golang.yaml
```diff
@@ -1,64 +1,64 @@
...
     key: "integration-test"
     agents:
       queue: "buildkite-hosted-linux-large"
+
+  - wait: ~
+
+  - label: ":rocket: Release CLI"
+    command: "./buildkite/scripts/release_cli.sh"
+    key: "release-cli"
+    if: build.tag != ""
```
wdbaruni, line 63: Help me understand this. We're only releasing if the build has a tag, but this pipeline is only triggered when committed right? Also what is the relationship between this step and the release pipeline? Some comments would be really helpful.

udsamani: So this pipeline gets triggered when one of the following condition 

`build.branch == "main" || !build.pull_request.draft || build.pull_request.labels includes "build/go" || build.pull_request.labels includes "build/golang" || build.tag != null` 

I ll add this as a comment :)

udsamani: I got rid of the release pipeline so that it does not confuse. This is a simpler way where we don't have to deal with artifact upload and artifact download in another job.

wdbaruni (approved):

Issue #4219, opened 2024-07-09
Decouple Bacalhau repo from the configuration file.


frrist:
- The configuration file MAY contain a field named `RepoPath` whose value configures where Bacalhau initializes/opens a repo. If no config file is provided, or one is provided and the field is not present, the default value `$HOME/.bacalhau` MUST be used.
- The configuration system MUST NOT contain the following fields (read: remove them):
  - `Node.User.KeyPath`
    - A Bacalhau repo MUST contain a file named `user.key` which stores this value. If the file is not present, Bacalhau MUST generate one on the user's behalf. If the file is present, it MUST be used. If the file contains invalid data, Bacalhau MUST fail to start.
  - `Node.User.InstallationID`
    - The value of this field MAY be stored in a file named `system_metadata.yaml`.
  - `Node.Metrics.EventTracerPath`
    - This value MUST be removed. (We may consider replacing it at a later date with a filesystem journal containing the events a Bacalhau node has seen.)
  - `Node.Update.CheckStatePath`
    - This value MUST be removed. The data previously contained by it MUST be kept in a file named `system_metadata.yaml`.
  - `Node.ExecutorPluginPath`
    - This value MUST be removed. Pluggable executors MUST be stored in `./bacalhau/compute/plugins/executors/`.
  - `Node.ComputeStoragePath`
    - This value MUST be removed. Compute storage MUST live in `/.bacalhau/compute/**`.
  - `Node.Compute.ExecutionStore.Path`
    - This value MUST be removed. Execution storage MUST live in `/.bacalhau/compute/executions/**`.
  - `Node.Requester.JobStore.Path`
    - This value MUST be removed. The job store MUST live in `/.bacalhau/orchestrator/state_bolt.db`.
  - `Node.Network.StoreDir`
    - This value MUST be removed. The network store MUST live in `./bacalhau/orchestrator/nats-store`.

The below defines the configuration schema that MUST be accepted by a Bacalhau node.

```yaml
# The directory used to store state and other persistent data.
# On compute nodes this is used to store execution data such as
# inputs, results and execution store. 
# On orchestrator nodes, the repo is also used to store the job store.
Repo: $HOME/.bacalhau

# What this node is named on the network. NOTE: this is not the node identity.
Name: "Bytey McByteface"

# How and where the node reports telemetry data
Telemetry:
  Logging:
    Level: Info
    Format: console
  Metrics:
    Endpoint: http://localhost:4318
  Traces:
[...]

PR #4330, opened 2024-08-15
drat: first pass at decoupling the repo and config paths

frrist:
closes: https://github.com/bacalhau-project/bacalhau/issues/4219

## Background:
At present the bacalhau config contains several fields that allow paths to location on disk to be set, these fields include, but are not limited to:
- KeyPath: https://github.com/bacalhau-project/bacalhau/blob/50831fa55c96660fdc573aab4a4441f10d9f6000/pkg/config/types/config.go#L14
- ExecutorPluginPath: https://github.com/bacalhau-project/bacalhau/blob/50831fa55c96660fdc573aab4a4441f10d9f6000/pkg/config/types/node.go#L21
- ComputeStoragePath: https://github.com/bacalhau-project/bacalhau/blob/50831fa55c96660fdc573aab4a4441f10d9f6000/pkg/config/types/node.go#L24
- StoreDir: https://github.com/bacalhau-project/bacalhau/blob/50831fa55c96660fdc573aab4a4441f10d9f6000/pkg/config/types/node.go#L123

If a user doesn't set a value for these fields in the config, bacalhau will set them to default values under the repo path: 
https://github.com/bacalhau-project/bacalhau/blob/50831fa55c96660fdc573aab4a4441f10d9f6000/pkg/repo/fs.go#L216-L229

This setup tightly couples the repo and config: in order for a config to be correctly populate it must first be modified by the repo to set paths accordingly in the event that are not provided. In order for a repo to be initialized it requires a config. 
This means the vanilla config returned by config.New() isn't actually usable by a bacalhau node until it has been passed to a repo. 

### Proposed Change
This draft removes usage of `*Path` fields defined in the bacalhau configuration structure. Meaning the config file no longer dictates the location on disk of various files and directories. Instead these paths are defined as constants declared with in the config package. When a service such as transport, compute, or requester require a path on disk they may retrieve it from the config package, at which point the value will be created, or returned if it already exists.

Exceptions:
The following `*Path` fields will still remain present in the config:
- All paths in the TLSConfiguration, these values are optional when running bacalhau.
- `Metrics.EventTracerPath` - we currently always set this field to `os.DevNull` - perhaps it would make sense to remove it entirely and its related event tracer since its never used.
- `AuthConfig.AccessPolicy` - bacalhau provides a verity of access policies in the form of .rego files, and users will likely want to fiddle with this value, further its optional.
- `AuthentiatorConfig.PolicyPath` - same story as `AccessPolicy`


cmd/cli/devstack/devstack.go
```diff
@@ -206,12 +207,39 @@ func runDevstack(cmd *cobra.Command, cfg types.BacalhauConfig, fsr *repo.FsRepo,
...
+	}
 	if IsNoop {
 		options = append(options, devstack.WithDependencyInjector(devstack.NewNoopNodeDependencyInjector()))
 	} else if ODs.ExecutorPlugins {
-		options = append(options, devstack.WithDependencyInjector(node.NewExecutorPluginNodeDependencyInjector(cfg)))
+		pluginPath, err := fsr.EnginePluginsDir()
+		if err != nil {
+			return err
+		}
+		options = append(options, devstack.WithDependencyInjector(
```
wdbaruni, line 229: every node in the devstack should have its own config and DataDir. Though I understand that might be a bigger change for this PR

cmd/cli/devstack/devstack.go
```diff
@@ -206,12 +207,39 @@ func runDevstack(cmd *cobra.Command, cfg types.BacalhauConfig, fsr *repo.FsRepo,
...
 		devstack.WithRequesterConfig(requesterConfig),
 	)
+	userKeyPath, err := fsr.UserKeyPath()
+	if err != nil {
+		return err
+	}
+	userKey, err := baccrypto.LoadUserKey(userKeyPath)
+	if err != nil {
+		return err
+	}
```
wdbaruni, line 217: do we really need to pass userKey? Isn't it deprecated?

cmd/cli/devstack/devstack.go
```diff
@@ -206,12 +207,39 @@ func runDevstack(cmd *cobra.Command, cfg types.BacalhauConfig, fsr *repo.FsRepo,
...
+				pluginPath,
+				executionDir,
+				userKey,
+			)))
 	} else {
-		options = append(options, devstack.WithDependencyInjector(node.NewStandardNodeDependencyInjector(cfg)))
+		options = append(options, devstack.WithDependencyInjector(
+			node.NewStandardNodeDependencyInjector(
+				cfg,
+				executionDir,
```
wdbaruni, line 240: This is where things can get confusing. we should just pass the config, and expose `cfg.ExecutionDir()`

cmd/cli/serve/util.go
```diff
@@ -28,25 +29,37 @@ import (
...
 	var executionStore store.ExecutionStore
-
 	if createExecutionStore {
-		executionStore, err = getExecutionStore(ctx, cfg.Compute.ExecutionStore)
+		computeDirPath, err := r.ComputeDir()
+		if err != nil {
+			return node.ComputeConfig{}, err
+		}
+		// TODO(forrest): would it be better to hand the bolt store a path and have it create the database there?
+		executionStore, err = getExecutionStore(ctx, cfg.Compute.ExecutionStore, filepath.Join(computeDirPath, "executions.db"))
```
wdbaruni, line 52: passing a file path should be fine. the file doesn't have to exist as boltdb will create it

pkg/node/node.go
```diff
@@ -188,15 +206,18 @@ func NewNode(
...
+		/*
+			// TODO calling `Keys` on the publishers takes ~10 seconds per call
+			// https://github.com/bacalhau-project/bacalhau/issues/4153
+			metrics.NodeInfo.Add(ctx, 1,
+				attribute.StringSlice("node_publishers", publishers.Keys(ctx)),
+				attribute.StringSlice("node_storages", storages.Keys(ctx)),
+				attribute.StringSlice("node_engines", executors.Keys(ctx)),
+			)
+
+		*/
```
wdbaruni, line 219: why commenting this out?

pkg/repo/fs.go
```diff
@@ -129,31 +128,31 @@ func (fsr *FsRepo) Open(c config.ReadWriter) error {
...
 
-	// modifies the config to include keys for accessing repo paths if they are not set.
-	// This ensures either user provided paths are valid to default paths for the repo are set.
-	fsr.EnsureRepoPathsConfigured(c)
-
-	cfg, err := c.Current()
-	if err != nil {
-		return err
+	// create files required by repo
+	if err := fsr.openRepoFiles(); err != nil {
```
wdbaruni, line 132: First, lets get rid of this package and move the logic under config. Maybe only migration logic should remain. Second, no files should be auto-created at an early level such as the repo or config. This will result in empty directories unnecessary created, such as executions dir created when you run a requester only node. Let `Node` type create the directories it needs only when trying to initialize a dependency

wdbaruni (changes requested):

cmd/cli/config/auto.go
```diff
@@ -116,8 +117,12 @@ func newAutoResourceCmd() *cobra.Command {
...
 }
 
-func autoConfig(ctx context.Context, cfg types.BacalhauConfig, settings *autoSettings) error {
-	pp := system.NewPhysicalCapacityProvider(cfg.Node.ComputeStoragePath)
+func autoConfig(ctx context.Context, r *repo.FsRepo, settings *autoSettings) error {
+	path, err := r.ExecutionDir()
+	if err != nil {
+		return err
+	}
+	pp := system.NewPhysicalCapacityProvider(path)
```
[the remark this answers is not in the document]
frrist: What creates `path.Join(DataDir, "compute", "executions")`? At present, the directory is created on the first call to `r.ExecutionDir()`.

cmd/cli/devstack/devstack.go
frrist: I agree with both points. Changing this behavior in the devstack is beyond the scope of this PR.

cmd/cli/devstack/devstack.go
frrist: Yes, its required by the authentication service, and removing it is out of scope here.

cmd/cli/serve/util.go
frrist: The directory `executions.db` exists in needs to be created before the path may be handed to boldb, what creates that path?

pkg/node/node.go
frrist: See the TODO above, this code takes ~10 seconds to run because we lack a way (that I am aware of) to disable bacalhau from attempting to retrieve s3 credentials for storages and publishers.

pkg/repo/fs.go
frrist: Does this imply config is responsible for creating the necessary files and directories - e.g. `orchestrator_store`, `compute_store`, `user_id.pem`, and `compute_store/executions`? Or that the config should return constant values for these paths that are not configurable? 

I'd take issue with the former as I don't think the config should be making changes to the file system. If the latter, what service is responsible for ensuring the paths exist?

cmd/cli/config/auto.go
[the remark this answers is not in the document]
wdbaruni: We can create what we need under `node.NewComputeNode(...)` and `node.NewRequesterNode(...)`

cmd/cli/devstack/devstack.go
frrist: Similar question as well: what creates the ExecutionDir? Asking since the executionDir provided here must exist on the filesystem, `NewStandardNodeDependencyInjector` (and many other constructors) expect the paths they are handed to exist on disk, if the path doesn't exist they error.

cmd/cli/serve/util.go
wdbaruni: Do you mean `computeDirPath`? It can be created in `node.NewComputeNode(...)` if we start initializing the executionStore inside `node.NewComputeNode()`, or create the parent path here

frrist: I think I could be in favor of `NewComputeNode` and `NewRequesterNode` being responsible for creating their respective paths. However implementing the code that that gets us there will be nontrivial as it requires re-writing much of our node construction logic. Here's a couple of reasons why:
1. `NewComputeNode` expects a `ComputeConfig`. In order to create a compute config, the ComputePath and ExecutionsPath both must already exist since the `ComputeConfig` contains an `ExecutionStore`, a `LocalPublisher`, and a `PhysicalResourcesProvider` all of which require initialized paths. The `NewComputeNode` method could be modified to create these fields internally, but then we'd lose the ability to mock them out in testing before building a compute node
2. The same is true for `NewRequesterNode`, it accepts a jobStore as input, so it must be initialized before the RequesterNode. Yes we could do it internally, but then we can't configure it differently for testing.
3. Before we initialize either a compute or a requester node, providers for engines, storages, and publishers must exist. Several of these providers require a path on disk to exist. The providers are parameters to `NewComputeNode`, and we frequently override these values with a `NoopProvider` during testing, meaning they can't (yet) be created internally by the compute node based on its config.

My thinking with this PR was: It would be far easier to make a code change where the repo creates these paths on demand, rather than the individual components. In the event we want to make these paths configurable in the future, we can simply add a field to the config and use it in-place of the path provided by the repo.

I am open to suggestion on ways of doing this differently than is done here. But at present, simply adding methods to the config that return paths won't work since they almost always must exist before a compute or requester node is constructed.

cmd/cli/config/auto.go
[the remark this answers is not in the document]
frrist: https://github.com/bacalhau-project/bacalhau/pull/4330#discussion_r1719012969

cmd/cli/devstack/devstack.go
wdbaruni: If it makes things easier, create the dirs and their parent path the first time you call ExecutionDir() if they don't exist, assuming we use the dir when we call this method and we don't call it for requester nodes as well and end up creating unnecessary dirs.

Otherwise we might need another explicit GetOrCreatExecutionDir() method

pkg/repo/fs.go
wdbaruni: As I explained, 'node' pkg should be the one initialzing and creating dependencies and the required paths. If it makes easier now during the migration to let config creat the dirs then go for it. No need for a separate repo pkg just for that purpose

cmd/cli/devstack/devstack.go
frrist: > create the dirs and their parent path the first time you call ExecutionDir() if they don't exist

This is what `r.ExecutionDir()` does, it either creates the directory if it doesn't exist, or returns its path if it does.

pkg/repo/fs.go
frrist: Do I understand the feedback correctly:
You want me to remove the methods I've added to the repo and instead add them to the config package?

cmd/cli/serve/util.go
wdbaruni: Move the methods to config, create dirs on demand, and delete repo pkg except for existing migrations

pkg/repo/fs.go
wdbaruni: I want our code to only rely on config for node configuration. Having both config and repo for a subset of the configuration is confusing and unnecessary. For now, just have the path getters with lazy path creation in config, delete repo, and in the future we can have config read only and move path creation to node package

frrist: You want the `pkg/repo` package deleted?

pkg/repo/migrations/v3_4.go
```diff
@@ -15,45 +19,84 @@ import (
 // - if an installationID is present in the config its value is persisted to system_metadata.yaml.
 // - removes update.json if the file is present.
 // - removes repo.version.
-var V3Migration = repo.NewMigration(
+// - creates a new directory .bacalhau/compute_store/executions which contains the content of ./bacalhau/execution_store
+// - removes ./bacalhau/execution_store
+// - iff a user has configured a user key path, the configured value is copied to .bacalhau/user_id.pem
+// - iff a user has configured a auth tokens path, the configured value is copied to .bacalhau/tokens.json
+var V3Migration = StagedMigration(
```
frrist, line 26: Given we haven't released since this migration was added I chose to update it to accommodate this change

pkg/repo/util.go
```diff
@@ -17,39 +16,31 @@ import (
 
 // initRepoFiles initializes all files required for a valid bacalhau repo.
 func initRepoFiles(cfg types.BacalhauConfig) error {
```
frrist, line 18: The repo remains responsible for initializing paths under it. The difference after this change is that a repo is _given_ a config, and it initializes itself based on the config values. I have kept the repo responsible for initializing paths for several reasons:
1. https://github.com/bacalhau-project/bacalhau/pull/4330#discussion_r1719012969
2. The config doesn't initialize paths because the config isn't guaranteed to have a repo path set, especially the case for testing. We are unable to create a default config with the right paths because such a config would require the correct repo path defined in it, and that's not possible.
3. This keeps the repo responsible for mutating the file system, and allows the config to fully dictate how the repo is initialized.

frrist:
@wdbaruni After giving this more thought, I must disagree with moving path management to the config package as done in this PR. The paths in question are constants derived from the repo path and are not user-configurable. Including them in the config package misaligns with its purpose of managing adjustable settings, not fixed operational constants. The config package should focus on user-configurable parameters, while the responsibility for file-system paths logically belongs to the repo package, which understands and manages the file-system structure.

A more maintainable design would have the repo manage all path-related data, keeping the config package clean and focused on its core role - providing user config. This approach will also future-proofs our system, allowing for easier reintroduction of configurable paths if needed, without requiring significant refactoring - the repo can use the path in the config if set, else continue to return a constant and default path.

I think of it this way: the config has a field named RepoPath that is passed to the repo, configuring where it initializes itself and its child paths. From there, services that only require configuration can simply accept the config. Services that require file-system access can accept a repo or a subset of it presented as an interface. For those that need both configuration and file-system access, they can accept both a config and a repo.

pkg/config/config.go
```diff
@@ -204,6 +204,9 @@ func (c *config) Current() (types.BacalhauConfig, error) {
 	if err := c.base.Unmarshal(&out, DecoderHook); err != nil {
 		return types.BacalhauConfig{}, err
 	}
+	if out.Repo == "" {
+		out.Repo = viper.GetString("repo")
+	}
```
wdbaruni, line 209: can you clarify this?

pkg/repo/util.go
```diff
@@ -17,39 +16,31 @@ import (
...
-	if err := initParentDir(cfg.Node.Compute.ExecutionStore.Path); err != nil {
-		return fmt.Errorf("failed to create executor execution store dir: %w", err)
+	if err := ensureDir(cfg.ComputeDir()); err != nil {
+		return fmt.Errorf("failed to create executor storage dir: %w", err)
 	}
-	if err := initParentDir(cfg.Node.Requester.JobStore.Path); err != nil {
-		return fmt.Errorf("failed to create orchestrator job store dir: %w", err)
+	if err := ensureDir(cfg.ExecutionDir()); err != nil {
+		return fmt.Errorf("failed to create executor storage dir: %w", err)
 	}
```
wdbaruni, line 33: You are still creating unnecessary directories. Why do we still need to create `ComputeDir` and `ExecutionDir` if the user is running an orchestrator node or just running a client? I have asked multiple times to initialize dirs in the getters until we move the logic in `node` package in the future if we see a need for that. Please lets stop wasting more time on this and take this as a requirement. Do initialize the dirs in the getters and lets move forward

pkg/repo/util.go
wdbaruni: Configs should NOT have an empty repo. Either a user provides a repo path, or we fall back to default one _and_ set it in the config, but we should have validation and fail if config ends with an empty repo

pkg/repo/migrations/v3_4.go
```diff
@@ -15,45 +19,84 @@ import (
 // - if an installationID is present in the config its value is persisted to system_metadata.yaml.
 // - removes update.json if the file is present.
 // - removes repo.version.
-var V3Migration = repo.NewMigration(
+// - creates a new directory .bacalhau/compute_store/executions which contains the content of ./bacalhau/execution_store
+// - removes ./bacalhau/execution_store
```
wdbaruni, line 23: please stick to the exact naming and directory structure as in https://www.notion.so/expanso/Rethinking-Configuration-435fbe87419148b4bbc5119d413786eb?pvs=4#4671ae7e54704844a2b3058da6f6e1ed

frrist: fixed in latest changes

pkg/repo/util.go
frrist: fixed in latest changes

pkg/config/config.go
frrist: added a comment in latest changes, in the event a user has provided a config without a data dir we need to set it here based on what the repo was initialized with.

pkg/repo/util.go
frrist: see latest changes

pkg/repo/fs.go
```diff
@@ -101,15 +101,11 @@ func (fsr *FsRepo) Init(c config.ReadWriter) error {
...
-	if cfg.Repo == "" {
-		cfg.Repo = fsr.path
-		c.Set("repo", fsr.path)
-	}
-
-	if err := initRepoFiles(cfg); err != nil {
-		return fmt.Errorf("failed to initialize repo: %w", err)
+	if cfg.DataDir == "" {
+		cfg.DataDir = fsr.path
+		c.Set("DataDir", fsr.path)
```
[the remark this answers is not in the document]
frrist: Before we can create a config we need to know the repo path. This is because we want to maintain backwards compatibility with existing behavior: When no config is provided check the repo, which may contain a config. See line 71

https://github.com/bacalhau-project/bacalhau/blob/50831fa55c96660fdc573aab4a4441f10d9f6000/cmd/util/repo.go#L58-L101

When we opt to stop searching the repo path for a config file we can remove this logic.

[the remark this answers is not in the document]
wdbaruni: I hear you, but I still don't see that a reason for this behaviour and for only setting `cfg.DataDir` with flag provided repo if the `cfg` has no `DataDir`

Take an example if a config was found in the repo, but that config had a `DataDir` configuration set to a different repo path. In the current implementation you will be respecting the `DataDir` in the config file instead of the flag provided `--repo/--data-dir` that the user provided

Lets decouple two things:
1. finding the config file: current implementation looks ok where we also look in the repo dir temporarily. To confirm though, the repo is either provided as a flag, env variable or default value which is currently `~/.bacalhau`, right? 
2. setting repo value in the config file: this should be treated the same as setting other config values that also have cli flags regardless how the config file was found and where it lives. We just need to make sure the logic for finding the repo path is the same when we first tried to look for the config file and when we are setting to set the default value for config

pkg/repo/util.go
```diff
@@ -16,33 +16,13 @@ import (
...
-		return fmt.Errorf("failed to create orchestrator dir: %w", err)
-	}
-	if err := ensureDir(cfg.ComputeDir()); err != nil {
-		return fmt.Errorf("failed to create executor storage dir: %w", err)
-	}
-	if err := ensureDir(cfg.ExecutionDir()); err != nil {
-		return fmt.Errorf("failed to create executor storage dir: %w", err)
-	}
-
+	panic("TODO")
```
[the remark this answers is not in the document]
frrist: Removed in [381af2d](https://github.com/bacalhau-project/bacalhau/pull/4330/commits/381af2d30586d0ab6890f43ae76babae4fa814ba)

pkg/repo/fs.go
```diff
@@ -96,17 +96,16 @@ func (fsr *FsRepo) Init(c config.ReadWriter) error {
...
 	cfg, err := c.Current()
 	if err != nil {
 		return err
 	}
 
-	if err := initRepoFiles(cfg); err != nil {
-		return fmt.Errorf("failed to initialize repo: %w", err)
+	// in the event a user has provided a config without a path we need to set it here based on what the repo was
+	// initialized with.
+	if cfg.DataDir == "" {
```
frrist, line 106: Issue to move away from this here: https://github.com/bacalhau-project/bacalhau/issues/4337

wdbaruni (approved):

PR #4386, opened 2024-09-09
fix TestSetupConfig on OSX

frrist:


cmd/util/repo_test.go
```diff
@@ -152,7 +152,22 @@ func setupTestCommand() (*cobra.Command, error) {
...
+			if err := os.MkdirAll(path, 0777); err != nil {
+				return "", err
+			}
+		case "plan9":
+			path = filepath.Join(path, "lib")
+			if err := os.MkdirAll(path, 0777); err != nil {
+				return "", err
+			}
+		}
+	}
```
wdbaruni, line 170: can we just use `$HOME/.config` for mac as well?

e.g.
```
func customUserConfigDir() string {
    if runtime.GOOS == "darwin" {
        home, err := os.UserHomeDir()
        if err != nil {
            log.Fatal(err)
        }
        return filepath.Join(home, ".config")
    }
    
    dir, err := os.UserConfigDir()
    if err != nil {
        log.Fatal(err)
    }
    return dir
}
```

frrist: That would mean placing the bacalhau config in a (somewhat) non-standard location on OSX (see [here](https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/os/file.go;l=543)). Will leave the decision up to you.

wdbaruni: I understand. Still a very common practice for apps to use `$HOME/.config` as their config path including `github`, `gcloud`, `nats`, `iterm2`, ... and will unify the experience with linux

frrist: Happy to go with whatever you prefer, let me know what you decide and I'll make the required changes.

wdbaruni: Sorry if it wasn't clear already, but I am asking to use `$HOME/.config` for mac as well

wdbaruni:
I dont think this is needed anymore

PR #4388, opened 2024-09-09
feat: implement analytics recoder

frrist:


pkg/analytics/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+	record := otellog.Record{}
+	record.SetTimestamp(time.Now().UTC())
+	record.AddAttributes(
+		otellog.String(EventKey, string(event)),
+		otellog.Map(PropertiesKey, properties...),
+	)
+	a.provider.Logger("bacalhau-analytics").Emit(ctx, record)
+}
+
+func (a *LogRecorder) EmitJobEvent(ctx context.Context, event EventType, j models.Job) {
```
wdbaruni, line 112: Remember we need to collect information about executions as well to understand how often failures happen before job is completed, how long jobs in the queue before execution is created, how evenly distributed node selection is, and many more

wdbaruni (commented):
Looks good overall

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+
+var _ Recorder = (*LogRecorder)(nil)
+
+type LogRecorder struct {
+	provider *sdklog.LoggerProvider
+}
+
+func New(ctx context.Context, opts ...Option) (*LogRecorder, error) {
+	config := &Config{
+		otlpEndpoint: "localhost:4317", // Default endpoint
```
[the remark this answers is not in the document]
frrist: Its the endpoint we will be sending data to, e.g. `telemetry.expanso.io` - this is a temporary solution for local development.

pkg/jobstore/boltdb/store.go
```diff
@@ -1000,6 +1009,7 @@ func (b *BoltJobStore) updateJobState(tx *bolt.Tx, request jobstore.UpdateJobSta
 	}
 
 	if job.IsTerminal() {
+		b.recorder.EmitJobEvent(context.TODO(), analytics.JobComplete, job)
```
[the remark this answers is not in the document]
frrist: Failed and Cancelled jobs are "complete", yea? What if we called it `Job_Terminal`? The nature of a jobs terminality can be gathered from inspecting its `State` which is included in the data we send.
By no means am I arguing against your proposal - just trying to think of a more specific name for this event which is recorded when the job reaches a terminal state.

pkg/analytics/util.go
```diff
@@ -0,0 +1,48 @@
...
+		otellog.String("task_name", t.Name),
+		otellog.String("task_engine", t.Engine.Type),
+		otellog.String("task_publisher", t.Engine.Type),
+		otellog.Slice("task_inputs", inputTypes...),
+		otellog.Int("task_env_count", len(t.Env)),
+		otellog.Int("task_meta_count", len(t.Meta)),
+		otellog.String("task_cpu", t.ResourcesConfig.CPU),
+		otellog.String("task_memory", t.ResourcesConfig.Memory),
+		otellog.String("task_disk", t.ResourcesConfig.Disk),
+		otellog.String("task_gpu", t.ResourcesConfig.GPU),
```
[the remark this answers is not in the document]
frrist: Meaning record an integer value representing the disk and memory requirements (in bytes) of the job, yeah?

pkg/analytics/util.go
```diff
@@ -0,0 +1,48 @@
...
+		inputTypes[i] = otellog.StringValue(s.Source.Type)
+	}
+
+	return []otellog.KeyValue{
+		otellog.String("task_name", t.Name),
+		otellog.String("task_engine", t.Engine.Type),
+		otellog.String("task_publisher", t.Engine.Type),
+		otellog.Slice("task_inputs", inputTypes...),
+		otellog.Int("task_env_count", len(t.Env)),
+		otellog.Int("task_meta_count", len(t.Meta)),
```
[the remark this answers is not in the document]
frrist: I think it could be helpful to get a count of all meta data - we know how many we expect to set, so any value over that could be interesting to collect. Additionally, we we see `0` here we know there is probably a bug in our code. 

I'll leave the decision up to you.

pkg/analytics/util.go
```diff
@@ -0,0 +1,48 @@
...
+	otellog "go.opentelemetry.io/otel/log"
+
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+)
+
+func makeJobAttributes(j models.Job) []otellog.KeyValue {
+	return []otellog.KeyValue{
+		otellog.String("id", j.ID),
+		otellog.String("name", j.Name),
+		otellog.String("namespace", j.Namespace),
```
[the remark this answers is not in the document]
frrist: Got it.
If name == id omit the name.
if name != id hash the name and include the digest.

pkg/analytics/analytics.go
frrist: I think we can achieve this with a new event type and method, yea?

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+func (a *LogRecorder) EmitEvent(ctx context.Context, event EventType, properties ...otellog.KeyValue) {
+	record := otellog.Record{}
+	record.SetTimestamp(time.Now().UTC())
+	record.AddAttributes(
+		otellog.String(EventKey, string(event)),
+		otellog.Map(PropertiesKey, properties...),
+	)
+	a.provider.Logger("bacalhau-analytics").Emit(ctx, record)
```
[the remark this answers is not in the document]
frrist: The logger name is used as the `Scope` field from an event - and an event emitted looks roughly like this:
```json
{
  "Timestamp": "2024-09-12T21:48:22.150123874Z",
  "ObservedTimestamp": "2024-09-12T14:48:22.150128342-07:00",
  "Severity": 0,
  "SeverityText": "",
  "Body": {
    "Type": "Empty",
    "Value": null
  },
  "Attributes": [
   ...<omitted>...
  ],
  "Scope": {
    "Name": "bacalhau-analytics",
    "Version": "",
    "SchemaURL": ""
  },
  "DroppedAttributes": 0
}
```

Eventually I image we could use the logger name to differentiate between events from clients and servers, or possibly something else. For now I'll define a constant for this field (suggestions?) and create it in the constructor.

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+	EventKey      = "event"
+	PropertiesKey = "properties"
+)
+
+func (a *LogRecorder) EmitEvent(ctx context.Context, event EventType, properties ...otellog.KeyValue) {
+	record := otellog.Record{}
+	record.SetTimestamp(time.Now().UTC())
+	record.AddAttributes(
+		otellog.String(EventKey, string(event)),
+		otellog.Map(PropertiesKey, properties...),
```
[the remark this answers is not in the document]
frrist: The pattern I've implement here forces us to be specific with the fields we extract from jobs/executions/evaluations/models, removes the possibility of encountering an error while extracting fields (cuz no json marshal), and also provides a type system for each event:
```json
 {
      "Key": "properties",
      "Value": {
        "Type": "Map",
        "Value": [
          {
            "Key": "id",
            "Value": {
              "Type": "String",
              "Value": "j-67bff92a-5938-4324-9a1d-5226371e41b1"
            }
          },
          {
            "Key": "name",
            "Value": {
              "Type": "String",
              "Value": "j-67bff92a-5938-4324-9a1d-5226371e41b1"
            }
          },
          {
            "Key": "namespace",
            "Value": {
              "Type": "String",
              "Value": "default"
            }
          },
          {
            "Key": "type",
            "Value": {
              "Type": "String",
              "Value": "batch"
            }
          },
          {
            "Key": "count",
            "Value": {
              "Type": "Int64",
              "Value": 1
            }
          },
          {
            "Key": "state",
            "Value": {
              "Type": "String",
              "Value": "Completed"
            }
          },
```
The alternative is to implement specific types for each event as was done in the POC, e.g.
```go
type Job struct {
	ID       string `json:"id"`
	Type     string `json:"type"`
	Image    string `json:"image"`
	CPUCores int    `json:"cpu_cores"`
	MemoryMB int    `json:"memory_mb"`
}

func stubJob() Job {
	return Job{
		ID:       uuid.NewString(),
		Type:     "docker",
		Image:    "ubuntu:latest",
		CPUCores: 2,
		MemoryMB: 4096,
	}
}
```

Which would you prefer?

pkg/analytics/analytics.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (a *LogRecorder) Stop(ctx context.Context) error {
+	defer func() {
+		if err := a.provider.Shutdown(ctx); err != nil {
+			log.Warn().Err(err).Msg("failed to shutdown analytics")
+		}
+	}()
+	if err := a.provider.ForceFlush(ctx); err != nil {
+		return fmt.Errorf("failed to flush analytics: %w", err)
+	}
```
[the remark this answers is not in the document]
frrist: Not sure what you mean. This function will return when `ForceFlush` and `Shutdown` return, or when the passed `ctx` times out.

cmd/cli/serve/serve.go
```diff
@@ -123,12 +124,29 @@ func serve(cmd *cobra.Command, cfg types.Bacalhau, fsRepo *repo.FsRepo) error {
...
+	} else {
+		recorder, err = analytics.New(ctx, analytics.WithNodeNodeID(nodeName),
+			analytics.WithInstallationID(installationID), analytics.WithNodeType(isRequesterNode, isComputeNode))
+		if err != nil {
+			// if we fail here for some reason, do so silently and use the noop recorder.
+			recorder = &analytics.NoopRecorder{}
+		}
+	}
+
+	requesterConfig, err := GetRequesterConfig(cfg, isRequesterNode, recorder)
```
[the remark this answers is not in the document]
frrist: I didn't use a global here since we'd like to only create a recorder if the configuration allows it.
Our logger, and other otel providers can all be configured with environment variables which makes it easier to configure them inside an `init()` method. 
I'll leave the decision here up to you.

pkg/analytics/analytics.go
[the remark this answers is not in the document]
wdbaruni: force flush means all buffered events will be flushed to the collector before it exists, or at least thats how it sounds like. If you want to do that, then you need to define a timeout and an aggressive one as pubilshing telemetry should be best effort

pkg/analytics/analytics.go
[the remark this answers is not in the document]
wdbaruni: Parsing and processing the first version is going to be a pain, and I don't think it will play nicely with posthog. My preference is having specific types and publish their json format. This simplies post processing as well when we just deserialize to the same type easily and process. lets just be consistent with using pascalcase everywhere

pkg/analytics/util.go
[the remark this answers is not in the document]
wdbaruni: name_set = name != id # helps understand if name is being set. don't care about its hash or how many names they have
namespace = hash(namespace) # helps with understanding how many namespaces a single network can have

pkg/analytics/util.go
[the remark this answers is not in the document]
wdbaruni: I am okay with whatever you decide

pkg/analytics/util.go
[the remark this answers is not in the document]
wdbaruni: Correct. Handle failure gracefuly and set zero values instead of propagating the error

pkg/jobstore/boltdb/store.go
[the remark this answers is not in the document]
wdbaruni: job completed means successful completion
https://github.com/bacalhau-project/bacalhau/blob/16538b3d7f16972213051c91098afebb1f77f799/pkg/models/job.go#L33-L35

cmd/cli/serve/serve.go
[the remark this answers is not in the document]
wdbaruni: I know you didn't use globals in your implementation, but it is a common practice and we are already doing it today with setting otel trace and metric providers

https://github.com/bacalhau-project/bacalhau/blob/246a2bdc0dce46233a23727386f55c9f602c929c/pkg/telemetry/metrics.go#L39

globals doesn't mean you set them inside an `init()`, but wherver you set them you then assign them to a global reference

pkg/analytics/analytics.go
wdbaruni: I am trying to simplify processing by having a single self contained event. better to have this in a single event

pkg/jobstore/boltdb/store.go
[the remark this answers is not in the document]
frrist: I meant completed in the sense that the job is no longer running - it's stopped, terminal, etc.. I'll just call this `Job`

cmd/cli/serve/serve.go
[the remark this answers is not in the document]
frrist: okay I will set it as the global provider.

pkg/analytics/analytics.go
frrist: okay I will create an event structure rather than use attributes.

pkg/analytics/analytics.go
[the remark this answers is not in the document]
frrist: sounds good, I will create an event structure rather than use attributes.

pkg/analytics/analytics.go
```diff
@@ -2,57 +2,114 @@ package analytics
...
+		opt(config)
 	}
 
 	// Create the file exporter
 	// TODO before merging we'll need to disable this
 	stdoutExporter, err := stdoutlog.New()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create stdout exporter: %w", err)
+		return fmt.Errorf("failed to create stdout exporter: %w", err)
 	}
```
wdbaruni, line 105: do you still need this expoerter?

pkg/analytics/analytics.go
```diff
@@ -2,57 +2,114 @@ package analytics
...
 
-	exporter, err := otlploggrpc.New(ctx,
-		otlploggrpc.WithEndpoint(config.otlpEndpoint), otlploggrpc.WithInsecure(),
+	exporter, err := otlploghttp.New(ctx,
+		otlploghttp.WithEndpoint(config.otlpEndpoint),
+		otlploghttp.WithTLSClientConfig(&tls.Config{}),
```
wdbaruni, line 109: not needed. just removing `WithInsecure()` is enough

pkg/analytics/analytics.go
```diff
@@ -71,46 +128,33 @@ func New(ctx context.Context, opts ...Option) (*LogRecorder, error) {
...
 
-func (a *LogRecorder) Stop(ctx context.Context) error {
-	defer func() {
-		if err := a.provider.Shutdown(ctx); err != nil {
-			log.Warn().Err(err).Msg("failed to shutdown analytics")
+func ShutdownAnalyticsProvider(ctx context.Context) error {
+	provider, ok := global.GetLoggerProvider().(*sdklog.LoggerProvider)
+	if ok {
+		if err := provider.ForceFlush(ctx); err != nil {
+			log.Trace().Err(err).Msg("failed to flush analytics log provider")
```
wdbaruni, line 144: You are still force flushing. Does it mean it won't delay shutdown?

pkg/node/node.go
```diff
@@ -292,6 +293,20 @@ func NewNode(
...
+		}
+		if err := analytics.SetupAnalyticsProvider(ctx,
+			analytics.WithNodeNodeID(config.NodeID),
+			analytics.WithNodeType(config.IsRequesterNode, config.IsComputeNode),
+			analytics.WithInstallationID(installationID),
+			analytics.WithVersion(version.Get()),
+		); err != nil {
+			log.Trace().Err(err).Msg("failed to setup analytics provider")
+		}
+	}
```
wdbaruni, line 309: will this cause issues with devstack? previously we only enabled analytics through `bacalhau serve`. Doing in node is fine, but we need to make sure all test paths have analytics disabled

pkg/node/node.go
```diff
@@ -311,6 +326,9 @@ func NewNode(
 		if apiServer != nil {
 			err = errors.Join(err, apiServer.Shutdown(ctx))
 		}
+		if err := analytics.ShutdownAnalyticsProvider(ctx); err != nil {
+			log.Trace().Err(err).Msg("failed to shutdown analytics provider")
+		}
```
wdbaruni, line 331: any potential race condition issues with devstack?

pkg/analytics/analytics.go
```diff
@@ -2,57 +2,114 @@ package analytics
...
+		} else {
+			c.attributes = append(c.attributes, attribute.String(NodeVersionKey, v.String()))
+		}
+	}
+}
+
+func SetupAnalyticsProvider(ctx context.Context, opts ...Option) error {
 	config := &Config{
-		otlpEndpoint: "localhost:4317", // Default endpoint
+		otlpEndpoint: "localhost:4318", // Default endpoint
```
wdbaruni, line 92: lets switch to `t.bacalhau.dev:4317` with todo to move to `t.bacalhau.org:4317`

pkg/analytics/models.go
```diff
@@ -0,0 +1,140 @@
...
+	TaskGPUCount         uint64   `json:"task_gpu_count"`
+	TaskNetworkType      string   `json:"task_network_type"`
+	TaskDomainsCount     int      `json:"task_domains_count"`
+	TaskExecutionTimeout int64    `json:"task_execution_timeout"`
+	TaskQueueTimeout     int64    `json:"task_queue_timeout"`
+	TaskTotalTimeout     int64    `json:"task_total_timeout"`
+}
+
+func NewJobTerminalEvent(j models.Job) *Event {
+	t := j.Task()
```
wdbaruni, line 91: Please add a TODO to include execution telemetry. Can be a follow up PR, but please lets do it before 1.5

pkg/analytics/models.go
```diff
@@ -0,0 +1,140 @@
...
+
+	otellog "go.opentelemetry.io/otel/log"
+
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+)
+
+type EventType string
+
+const (
+	JobEventType = "job"
```
wdbaruni, line 17: lets add versioning to event types so that change in pubished schema wouldn't break things

pkg/analytics/analytics.go
frrist: No, it's been removed in the latest commit.

pkg/node/node.go
frrist: I'll just move the logic to serve

pkg/analytics/analytics.go
frrist: `provider.ForceFlush` should respect context cancellation of the context provided to it, but I'll just remove this call.

pkg/node/node.go
wdbaruni: still, please make sure tests don't trigger telemetry. we still have tests that call `serve`

pkg/analytics/analytics.go
frrist: done.

pkg/analytics/models.go
frrist: done.

pkg/analytics/models.go
frrist: done, good idea.

pkg/node/node.go
frrist: addressed by moving to the server and disabling during testing.

pkg/telemetry/common.go
```diff
@@ -1,20 +1,37 @@
...
+		case *url.Error:
+			if strings.Contains(e.URL, analytics.DefaultOtelCollectorEndpoint) {
+				return
+			} else {
+				log.Err(err).Msg("Error occurred while handling spans")
+			}
+		default:
+			// fall through and log the error
+		}
 		log.Err(err).Msg("Error occurred while handling spans")
```
wdbaruni, line 35: you are still logging the error at the end and will log the error twice for non analytics errors

wdbaruni (approved):
There is a bug with logging errors. you can merge after addressing it

PR #4401, opened 2024-09-11
s3 publisher errors should root from base error

udsamani:
This PR aims at standardizing S3 Errors to use the BaseError Present


cmd/cli/job/executions.go
```diff
@@ -147,7 +148,7 @@ func (o *ExecutionOptions) run(cmd *cobra.Command, args []string, api client.API
 		},
 	})
 	if err != nil {
-		return err
+		return errors.New(err.Error())
```
frrist, line 151: please revert this, either do:
`return err`
or 
`return fmt.Errorf("<helpful message>: %w", err)`

cmd/cli/job/describe.go
```diff
@@ -82,13 +82,14 @@ func NewDescribeCmd() *cobra.Command {
...
 	jobID := args[0]
+
 	response, err := api.Jobs().Get(ctx, &apimodels.GetJobRequest{
 		JobID:   jobID,
 		Include: "executions,history",
 	})
 
 	if err != nil {
-		return fmt.Errorf("could not get job %s: %w", jobID, err)
+		return err
```
frrist, line 92: I assume this is being removed since the `Get` method's error already contains a detailed description?

cmd/cli/job/history.go
```diff
@@ -160,7 +161,7 @@ func (o *HistoryOptions) run(cmd *cobra.Command, args []string, api client.API)
 		},
 	})
 	if err != nil {
-		return err
+		return errors.New(err.Error())
```
frrist, line 164: similar comment as stated elsewhere.

cmd/util/download.go
```diff
@@ -33,7 +34,7 @@ func DownloadResultsHandler(
 		JobID: jobID,
 	})
 	if err != nil {
-		Fatal(cmd, fmt.Errorf("could not get results for job %s: %w", jobID, err), 1)
+		return errors.New(err.Error())
```
frrist, line 37: similar comment as stated elsewhere.

pkg/jobstore/boltdb/store.go
```diff
@@ -176,7 +178,7 @@ func (b *BoltJobStore) getJob(tx *bolt.Tx, jobID string) (models.Job, error) {
 
 	data := GetBucketData(tx, NewBucketPath(BucketJobs, jobID), SpecKey)
 	if data == nil {
-		return job, bacerrors.NewJobNotFound(jobID)
+		return job, apimodels.NewJobNotFound(jobID)
```
frrist, line 181: I think we can used `jobstore.NewErrJobNotFound` here, no?

pkg/publicapi/middleware/error_handler.go
```diff
@@ -0,0 +1,60 @@
...
+			err = c.NoContent(code)
+		} else {
+			err = c.JSON(code, apimodels.APIError{
+				HTTPStatusCode: code,
+				Message:        message,
+				RequestID:      requestID,
+			})
+		}
+		if err != nil {
+			log.Info().Msg("unable to send json response while handling error.")
```
frrist, line 56: include the error message in the log - and possibly the requestID - consider elevating the severity of this log to `Error` or `Warn`, e.g
```go
log.Error().Err(err).Str("request_id", requestID).Msg("unable to send json response while handling error")
```
Also ensure this `err` isn't being shadowed by the function parameter.

pkg/s3/error.go
```diff
@@ -0,0 +1,9 @@
+package s3
+
+import "github.com/bacalhau-project/bacalhau/pkg/models"
+
+const S3_PUBLISHER_COMPONENT = "S3PUB"
```
frrist, line 5: nit: tweak to name and variable pattern, please remove underscores at a minimum.
```suggestion
const ComponentName = "S3_PUB"
```
suggesting to drop "S3" from the name since it's implied by the package (s3) containing the variable.

pkg/jobstore/errors.go
```diff
@@ -6,154 +6,75 @@ import (
...
 
-func (e ErrInvalidExecutionState) Error() string {
-	if len(e.Expected) > 0 {
-		return fmt.Sprintf("execution %s is in unexpected state %s", e.ExecutionID, e.Actual)
+func NewErrInvalidExecutionState(id string, actual models.ExecutionStateType, expected ...models.ExecutionStateType) *models.BaseError {
+	var errorMessage string
+	if len(expected) > 0 {
+		errorMessage = fmt.Sprintf("execution %s is in unexpected state %s", id, actual)
+	} else {
+		errorMessage = fmt.Sprintf("execution %s is in state %s, but expected %s", id, actual, expected)
```
frrist, line 60: This seems wrong. Shouldn't it be:
```go
if len(expected) == 0 {
	errorMessage = fmt.Sprintf("execution %s is in unexpected state %s", id, actual)
} else {
	errorMessage = fmt.Sprintf("execution %s is in state %s, but expected %s", id, actual, expected)
}
```

pkg/models/error.go
```diff
@@ -41,6 +41,38 @@ type HasDetails interface {
 	Details() map[string]string
 }
 
+type HasCode interface {
+	// Details a code
+	Code() int
+}
```
frrist, line 47: Unused?

pkg/publicapi/client/v2/client.go
```diff
@@ -80,20 +96,36 @@ func (c *httpClient) write(ctx context.Context, verb, endpoint string, in apimod
...
-	if err != nil && resp != nil && resp.StatusCode == http.StatusUnauthorized {
-		return apimodels.ErrInvalidToken
-	} else if err != nil {
+
+	_, resp, err := c.doRequest(ctx, verb, endpoint, r) //nolint:bodyclose // this is being closed
+	defer resp.Body.Close()
+	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
```
frrist, line 89: Move the defer after the error check. If an error is returned, the response will likely be nil, and attempting to close a nil response will panic.

pkg/publicapi/apimodels/error.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Parse HTTP Resposne to APIError
+func FromHttpResponse(resp *http.Response) (*APIError, error) {
+
+	if resp == nil {
+		return nil, errors.New("response is nil, cannot be unmarsheld to APIError")
+	}
+
+	defer resp.Body.Close()
```
frrist, line 115: We are potentially closing the response body twice since callers of this method typically close the body after calling it. I'd suggest not closing it here and leaving it to the caller. Include a code comment stating such.

pkg/publicapi/client/v2/client.go
```diff
@@ -55,12 +55,28 @@ type httpClient struct {
...
-	} else if err != nil {
+
+	_, resp, err := c.doRequest(ctx, http.MethodGet, endpoint, r) //nolint:bodyclose // this is being closed
+	if err != nil {
 		return err
 	}
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		return apimodels.NewUnauthorizedError("invalid token")
+	}
```
frrist, line 66: Do you think it would be cleaner to do:
```go
return apimodels.NewAPIError(http.StatusUnauthorized, "invalid token")
```
instead?

The methods like `NewBadRequestError`, `NewForbiddenError`, etc. feel like unnecessary wrapping.

pkg/jobstore/errors.go
```diff
@@ -6,154 +6,75 @@ import (
...
-	JobID string
-}
-
-func NewErrJobAlreadyExists(id string) ErrJobAlreadyExists {
-	return ErrJobAlreadyExists{JobID: id}
-}
+const (
+	JOB_STORE_COMPONENT = "JBS"
+	BOLTDB_COMPONENT    = "BDB"
+)
```
frrist, line 12: What is the different between these components? `BOLTDB_COMPONENT` seems like an implementation detail.
Also, similar comment here regarding naming.

pkg/publicapi/client/v2/client.go
```diff
@@ -55,12 +55,28 @@ type httpClient struct {
 // and deserialize the response into a response object
 func (c *httpClient) Get(ctx context.Context, endpoint string, in apimodels.GetRequest, out apimodels.GetResponse) error {
 	r := in.ToHTTPRequest()
-	_, resp, err := requireOK(c.doRequest(ctx, http.MethodGet, endpoint, r)) //nolint:bodyclose // this is being closed
-	if err != nil && resp != nil && resp.StatusCode == http.StatusUnauthorized {
-		return apimodels.ErrInvalidToken
-	} else if err != nil {
+
+	_, resp, err := c.doRequest(ctx, http.MethodGet, endpoint, r) //nolint:bodyclose // this is being closed
```
frrist, line 59: remember to close the body, there is a potential for it to not get closed here if the status code isn't OK

pkg/publicapi/client/v2/client.go
```diff
@@ -55,12 +55,28 @@ type httpClient struct {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if apiError != nil {
+		return apiError
+	}
+
 	defer resp.Body.Close()
```
frrist, line 80: move this up as other comment suggests.

pkg/s3/error.go
```diff
@@ -0,0 +1,9 @@
...
+
+func NewErrBadS3Request(msg string) *models.BaseError {
```
frrist, line 7: nit on naming here and elsewhere, would prefer:
`NewBadS3RequestError`

and more generally:
`New<Type>Error`

frrist (commented):
good stuff, left a few comments/thoughts.

PR #4405, opened 2024-09-11
do not print usage and error and infer error codes in custom error handler

udsamani:
This PR aims at the following

- Do not print usage and error
- HTTP Status Code infer in custom error handler

### DIFF
--- a/cmd/cli/job/describe.go
+++ b/cmd/cli/job/describe.go
@@ -75,6 +75,10 @@ func NewDescribeCmd() *cobra.Command {
 			return o.run(cmd, args, api)
 		},
 	}
+
+	jobCmd.SilenceUsage = true
+	jobCmd.SilenceErrors = true
+
 	jobCmd.Flags().AddFlagSet(cliflags.OutputNonTabularFormatFlags(&o.OutputOpts))
 	return jobCmd
 }
--- a/cmd/cli/job/describe.go
+++ b/cmd/cli/job/describe.go
@@ -76,9 +76,6 @@ func NewDescribeCmd() *cobra.Command {
 		},
 	}
 
-	jobCmd.SilenceUsage = true
-	jobCmd.SilenceErrors = true
-
 	jobCmd.Flags().AddFlagSet(cliflags.OutputNonTabularFormatFlags(&o.OutputOpts))
 	return jobCmd
 }
--- a/cmd/cli/job/describe.go
+++ b/cmd/cli/job/describe.go
@@ -56,11 +56,13 @@ func NewDescribeOptions() *DescribeOptions {
 func NewDescribeCmd() *cobra.Command {
 	o := NewDescribeOptions()
 	jobCmd := &cobra.Command{
-		Use:     "describe [id]",
-		Short:   "Get the info of a job by id.",
-		Long:    describeLong,
-		Example: describeExample,
-		Args:    cobra.ExactArgs(1),
+		Use:           "describe [id]",
+		Short:         "Get the info of a job by id.",
+		Long:          describeLong,
+		Example:       describeExample,
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Args:          cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// initialize a new or open an existing repo merging any config file(s) it contains into cfg.
 			cfg, err := util.SetupRepoConfig(cmd)
@@ -75,9 +77,6 @@ func NewDescribeCmd() *cobra.Command {
 			return o.run(cmd, args, api)
 		},
 	}
-
-	jobCmd.SilenceUsage = true
-	jobCmd.SilenceErrors = true
 
 	jobCmd.Flags().AddFlagSet(cliflags.OutputNonTabularFormatFlags(&o.OutputOpts))
 	return jobCmd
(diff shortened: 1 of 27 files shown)


cmd/cli/job/describe.go
```diff
@@ -75,6 +75,10 @@ func NewDescribeCmd() *cobra.Command {
 			return o.run(cmd, args, api)
 		},
 	}
+
+	jobCmd.SilenceUsage = true
+	jobCmd.SilenceErrors = true
```
frrist, line 80: For all cases where a change is made like this, please do so in the `&cobra.Command{` block to keep things clean. So:
```go
jobCmd := &cobra.Command{
		Use:     "describe [id]",
		Short:   "Get the info of a job by id.",
		Long:    describeLong,
		Example: describeExample,
		SilenceUsage: true,
                SilenceErrors: true,
		Args:    cobra.ExactArgs(1),
			// initialize a new or open an existing repo merging any config file(s) it contains into cfg.
			cfg, err := util.SetupRepoConfig(cmd)
			if err != nil {
				return fmt.Errorf("failed to setup repo: %w", err)
			}
			// create an api client
			api, err := util.GetAPIClientV2(cmd, cfg)
			if err != nil {
				return fmt.Errorf("failed to create api client: %w", err)
			}
			return o.run(cmd, args, api)
		},
	}

Comment (udsamani) [reply cmd/cli/job/describe.go:80]: Aah thanks ! This saves so much !

File: pkg/jobstore/errors.go [hunk 2]
```diff
@@ -6,20 +6,29 @@ import (
 	"github.com/bacalhau-project/bacalhau/pkg/models"
 )
 
+const JobStoreComponent = "JobStore"
 const (
-	JOB_STORE_COMPONENT = "JBS"
+	InvalidJobState   models.ErrorCode = "InvalidJobState"
+	MultipleJobsFound models.ErrorCode = "MultipleJobsFound"
+	InvalidJobVersion models.ErrorCode = "InvalidJobVersion"
```
wdbaruni, line 13: nit: might make more sense to call them `ConflictingJobState` and `ConflictingJobVersion` as these issues usually happen due to concurrent updates or race conditions. Users don't provide conditions when they update jobs and will be a bit confusing to say their request was invalid

pkg/models/error.go
```diff
@@ -158,3 +184,30 @@ func (e *BaseError) Details() map[string]string {
...
 	return e.code
 }
+
+// HTTPStatusCode is a method that returns the httpStatusCode field of BaseError.
+// If no specific HTTP status code has been set, it returns 0.
+// This method can be used to retrieve the HTTP status code associated with the error,
+// which is useful when translating the error into an HTTP response.
+func (e *BaseError) HTTPStatusCode() int {
+	return e.httpStatusCode
+}
```
wdbaruni, line 194: this method should return `e.httpStatusCode` if it is not 0, otherwise it should call `inferHttpStatusCode`

pkg/models/error.go
```diff
@@ -158,3 +184,30 @@ func (e *BaseError) Details() map[string]string {
...
+
+func InferHTTPStatusCode(code ErrorCode) int {
+	switch code {
```
wdbaruni, line 197: this should be private and callers should only use `HTTPStatusCode()`

pkg/publicapi/middleware/error_handler.go
```diff
@@ -28,13 +31,17 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
 		// size accepted
 		code = e.Code
 		message = e.Message.(string)
+		errorCode = string(models.InternalError)
+		component = "Echo"
```
wdbaruni, line 35: `Echo` is an internal implementation details. Should we use `APIServer` or something like that?

cmd/cli/job/describe.go
wdbaruni, line 80: instead of doing this in each and every command, we can just silence the root command
https://github.com/bacalhau-project/bacalhau/blob/1ca678c752371f790e84220e6061e63d3aed5e32/cmd/cli/root.go#L110-L118

Tested this works:
```

func Execute(ctx context.Context) {
	rootCmd := NewRootCmd()
	rootCmd.SetContext(ctx)
	rootCmd.SilenceErrors = true
	rootCmd.SilenceUsage = true
	if err := rootCmd.Execute(); err != nil {
		util.Fatal(rootCmd, err, 1)
	}
}
```

wdbaruni (approved):
Minor comments and you should be able to merge after addressing them

cmd/cli/job/describe.go
udsamani: Acknowledge. Fixing it !

pkg/jobstore/errors.go
udsamani: Acknowledged fixing it

pkg/publicapi/middleware/error_handler.go
udsamani: Makes sense

pkg/models/error.go
udsamani: Acknowledged fixing it

cmd/cli/job/describe.go
udsamani: Actually i did try that and it does not work a root command layer.  You unfortunately have to do it at all layers.

pkg/models/error.go
udsamani: Done

PR #4406, opened 2024-09-11
Enhanced WebUI

wdbaruni:
Improving the embedding web ui and making it compatible with our latest APIs.

## How to test locally
```
# this will build the static web ui and the binary as well
make build

# using the built binary
bacalhau serve --orchestrator --compute  -c WebUI.Enable

# now the webui can be reached at http://localhost:8438/
```
## Main changes in core bacalhau
- Enable CORS for websockets as well to enable streaming logs
- Fix swagger generation for some APIs
@udsamani if you can please review these changes

## Main changes in the ui http server
- Better handling for 404 instead of always returning the home page
- Previously we were mutating the html files by replacing template placeholders with the http endpoint of the orchestrator node to allow the frontend to call the right node. Now the webui server exposes a `/_config` endpoint that the frontend calls when it loads to figure out the endpoints it needs to call
@frrist if you can please review these changes. Just the [webui/webui.go](https://github.com/bacalhau-project/bacalhau/pull/4406/files#diff-a3f7d48d48587b3914d504ac3e764aae8652ed9bd8615aea9ce723fa8558d192) file

## Main changes in the frontend
- Frontend design close to figma https://www.figma.com/design/PAtQCqnpJ78UCM4oKHT2Y1/expanso?node-id=0-1&t=EoXg2e4UAWX4MsQv-0
- Switched to nextjs and shadcn mainly to be consistent with Expanso Cloud implementation to enable code and component reuse
- Used generated swagger models and APIs instead of manually creating new models 
- Support for streaming job logs
- Bunch of other improvements
@aronchick a lot of code changes here, but are mainly generated code by openapi and other tools we are using. Still I would appreciate overall feedback on the component's implementations as I am not an expert in FE coding

## Remaining work for follow up PRs
- Better error handling for failed API requests
- Server disconnection handling and better reporting
- Download job results
- Job submission
- Node management (approve, delete, ...)
- Executions list and details view
- Implement tests

## Open questions
- Storybook was dropped to release this faster. Do we think it is valuable to add in the future? @aronchick 

## Screenshots

image

image

image

image

image


udsamani (approved):
Swagger API generation LGTM

webui/webui.go
```diff
@@ -1,81 +1,234 @@
 package webui
```
frrist, line 1: looks good to me

frrist (approved):

PR #4410, opened 2024-09-12
chore: add node name flag to serve command

### LINKED ISSUE #4407: Allow users to set the bacalhau node name when starting a bacalhau server.
The bacalhau node name value was moved from the config to `system_metadata.yaml`. With this change came the removal of the config flag `--name` that could be provided to the `server` command. We need to continue supporting this feature.

frrist:
- closes #4407


cmd/cli/serve/serve.go
```diff
@@ -50,6 +50,13 @@ var (
 `))
 )
 
+const (
+	NameFlagName        = "name"
+	NameFlagDescription = `The node's name.
+If unset, it will be read from .bacalhau/system_metadata.yaml, or automatically generated if no name exists.
+If set, the provided value will override the current name in .bacalhau/system_metadata.yaml.`
```
wdbaruni, line 57: it shouldn't. We want the same as the existing behaviour. A node `--name` is only applied the first time you instantiate a node, will be persisted afterwards, and then the persisted node name will be used and the flag will be ignored

In the future we might fail if the provided name doesn't match the one in system_metadata, but we are not sure if that is the right behaviour and no need for us to change the existing behaviour at this time

In other words, we only use this flag before calling `config.GenerateNodeID()`

wdbaruni (changes requested):

cmd/cli/serve/serve.go
frrist: If the user provides the name flag and a value is already present in `system_metadata.yaml` should we fail, ignore the flag, or ignore the flag and warn?

frrist: I've addressed this and added a warning if the flag is used and a value is already present

cmd/cli/serve/serve.go
```diff
@@ -110,19 +119,40 @@ func serve(cmd *cobra.Command, cfg types.Bacalhau, fsRepo *repo.FsRepo) error {
...
+				return fmt.Errorf("failed to write node name %s: %w", nodeName, err)
+			}
+			log.Info().Msgf("persisted node name %s", nodeName)
+		} else {
+			// the flag was provided with the name, persist it
+			log.Info().Msgf("persisting node name from flag %s", nodeName)
+			if err := fsRepo.WriteNodeName(nodeName); err != nil {
+				return fmt.Errorf("failed to write node name %s: %w", nodeName, err)
+			}
+			log.Info().Msgf("persisted node name %s", nodeName)
```
wdbaruni, line 147: nit: persisting node name logic is duplicated and can be improved. Also maybe too much logging

cmd/cli/serve/serve.go
```diff
@@ -110,19 +119,40 @@ func serve(cmd *cobra.Command, cfg types.Bacalhau, fsRepo *repo.FsRepo) error {
...
 		}
-		if err := fsRepo.WriteNodeName(nodeName); err != nil {
-			return fmt.Errorf("failed to write node name %s: %w", nodeName, err)
+	} else {
+		// warn the user if they passed a flag and a name already exists
+		if flagNodeName := cmd.PersistentFlags().Lookup(NameFlagName).Value.String(); flagNodeName != "" {
+			log.Warn().Msgf("--name flag with value %s ignored. name %s already exists", flagNodeName, nodeName)
```
wdbaruni, line 152: you should only warn if the provided name doesn't match the persisted one. As a user, I will run the node using `bacalhau serve --name node1`, and will always run that node using that command and not only the very first time. Think of deplying bacalhau using terrfaofrm or any other iac or fleet management logic, similar to how we deploy to our demo network. We have a start script that has all the flags, and the same scrip is called each time

wdbaruni (changes requested):

wdbaruni (approved):
Approved, but please address the comments before merging

cmd/cli/serve/serve.go
frrist: addressed.

cmd/cli/serve/serve.go
frrist: done.

PR #4411, opened 2024-09-12
Frrist/meta/instanceid

frrist:
Note: this PR should follow https://github.com/bacalhau-project/bacalhau/pull/4410 to avoid conflicts.


pkg/repo/fs.go
```diff
@@ -98,6 +98,11 @@ func (fsr *FsRepo) Init() error {
 	// TODO this should be a part of the config.
 	telemetry.SetupFromEnvs()
 
+	// never fail here as this isn't critical to node start up.
+	if err := fsr.WriteInstanceID(GenerateInstanceID()); err != nil {
+		log.Trace().Err(err).Msgf("failed to write instanceID")
```
wdbaruni, line 103: we should log as warning right?

pkg/repo/fs.go
```diff
@@ -120,6 +125,16 @@ func (fsr *FsRepo) Open() error {
 		}
 	}
 
+	// check if an instanceID exists persisting one if not found.
+	// never fail here as this isn't critical to node start up.
+	if instanceID, err := fsr.ReadInstanceID(); err != nil {
+		log.Trace().Err(err).Msgf("failed to read instanceID")
```
wdbaruni, line 131: same as above for log level. Also if repo doesn't have system_metadata for some reason, this will throw an error, right? Should we attempt to write the instanceID and recreate the metadata file if thats the case? Or we should handle file not found graceully when reading and return blank insanceID instead of error

wdbaruni (approved):
Minor comments

pkg/repo/fs.go
frrist: We should if we want to warn users we were unable to generate an instanceID, but I don't see any action they can take to right the problem, what do you think?

imo the instanceID is best effort, if for some reason we can't write it (which should be _really_ rare) then I don't see a point warning them as it doesn't affect the operation of bacalhau.

pkg/repo/fs.go
frrist: `fsr.Exists()` will fail if the repo doesn't have `system_metadata.yaml` as it reads the repo version from it.
The only case this will return an error is if a user managers to delete `system_metadata.yaml` after the `Exist` method runs but before this method runs.

In general, my goal with the trace logs here is to avoid notifying a user of any issues with an InstanceID as its presents isn't required for bacalhau to function.

PR #4412, opened 2024-09-13
Propogating centralized errors

udsamani:


pkg/compute/executor.go
```diff
@@ -302,8 +302,10 @@ func (e *BaseExecutor) Run(ctx context.Context, state store.LocalExecutionState)
 	stopwatch := telemetry.Timer(ctx, jobDurationMilliseconds, state.Execution.Job.MetricAttributes()...)
 	topic := EventTopicExecutionRunning
 	defer func() {
-		if err != nil && err.Error() != executor.ErrAlreadyCancelled.Error() {
-			e.handleFailure(ctx, state, err, topic)
+		if err != nil {
+			if baseErr, ok := err.(*models.BaseError); ok && baseErr.Code() == executor.ExecutionAlreadyCancelled {
+				e.handleFailure(ctx, state, baseErr, topic)
```
wdbaruni, line 307: should be `!=`. Some helper methods in `models` might simplify the error checks, such as 
```
// IsErrorWithCode checks if the given error is a BaseError (or its pointer)
// and if its code matches any of the provided expected codes.
func IsErrorWithCode(err error, expectedCodes ...ErrorCode) bool {
	var baseErr *BaseError
	if errors.As(err, &baseErr) {
		errCode := baseErr.Code()
		for _, code := range expectedCodes {
			if errCode == code {
				return true
			}
		}
	}
	return false
}
```

and then you can say:
```
	if !IsErrorWithCode(err, executor.ExecutionAlreadyCancelled) {
		e.handleFailure(ctx, state, err, topic)
	}
```

pkg/executor/docker/errors.go
```diff
@@ -0,0 +1,17 @@
+package docker
+
+import "github.com/bacalhau-project/bacalhau/pkg/models"
+
+const DOCKER_EXECUTOR_COMPONENT = "Executor/Docker"
```
wdbaruni, line 5: some errors will have `Docker` and others `Executor/Docker`, even though they are both related to Docker executor. Better to just use `Docker` for both

pkg/executor/docker/errors.go
```diff
@@ -0,0 +1,17 @@
...
+
+const (
+	ExectorSpecValidationErr = "ExectorSpecValidationErr"
+	ExecutionAlreadyStarted  = "ExecutionAlreadyStarted"
+	ExecutionNotFound        = "ExecutionNotFound"
```
wdbaruni, line 10: I think you can reuse the error codes in `/pkg/executor/types.go`

pkg/s3/errors.go
```diff
@@ -0,0 +1,15 @@
...
+const S3 = "S3"
+
+const (
+	S3BadRequest = "S3BadRequest"
+)
+
+func NewBadS3RequestError(message string) *models.BaseError {
+	return models.NewBaseError(message).
+		WithCode(S3BadRequest).
+		WithComponent(S3)
```
wdbaruni, line 14: using S3 as a component name can be a little ambiguous. Might be useful to understand if the failure happened with S3 storage source as it indicates an issue with job inputes, s3 publisher as it indiciates an issue with outputs, or S3 result signer as it indicates an issue with downloading results

pkg/executor/wasm/errors.go
```diff
@@ -0,0 +1,13 @@
+package wasm
+
+import (
+	"github.com/bacalhau-project/bacalhau/pkg/models"
+)
+
+const WasmExecutorComponent = "Executor/Wasm"
```
wdbaruni, line 7: I don't mind having a standard for naming nested components, but lets make sure it plays nicely with string representation of base errors. I keep thinking that `Docker:ImageNotFound`, `S3Publisher:BucketNotFound` and `S3InputSource:BucketNotFound` as potential candidates. Adding `/` as a separator between component type and component name is valid, but just think it through

pkg/compute/executor.go
udsamani: Done

pkg/executor/docker/errors.go
udsamani: Done.

pkg/executor/docker/errors.go
udsamani: Done.

pkg/executor/wasm/errors.go
udsamani: Removed at the moment. Just going with executor.

pkg/s3/errors.go
udsamani: Fixed.

wdbaruni (approved):
Minor comments. You can merge after addressing them

pkg/s3/errors.go
```diff
@@ -2,14 +2,21 @@ package s3
 
 import "github.com/bacalhau-project/bacalhau/pkg/models"
 
-const S3 = "S3"
+const S3_PUBLISHER = "S3Publisher"
+const S3_STORAGE = "S3Storage"
```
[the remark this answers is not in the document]
udsamani: Done

pkg/s3/types.go
```diff
@@ -37,7 +37,7 @@ type PreSignedResultSpec struct {
 
 func (c PreSignedResultSpec) Validate() error {
 	if c.PreSignedURL == "" {
-		return NewBadS3RequestError("invalid s3 signed storage params: signed url cannot be empty")
+		return NewS3PublisherError(S3BadRequest, "invalid s3 signed storage params: signed url cannot be empty")
 	}
```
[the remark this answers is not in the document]
udsamani: Done

pkg/s3/types.go
```diff
@@ -21,7 +21,7 @@ type SourceSpec struct {
 
 func (c SourceSpec) Validate() error {
 	if c.Bucket == "" {
-		return NewBadS3RequestError("invalid s3 storage params: bucket cannot be empty")
+		return NewS3PublisherError(S3BadRequest, "invalid s3 storage params: bucket cannot be empty")
 	}
 	return nil
 }
```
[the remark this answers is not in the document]
udsamani: Done

Issue #4085, opened 2024-06-14
installationID is not persisted


wdbaruni:
We don't persist installationID for neither client nor serve operations. This should be fixed as it is screwing up our metrics

frrist:
replaced with https://github.com/bacalhau-project/bacalhau/issues/4184

PR #4450, opened 2024-09-17
Frrist/metrics/include instance

frrist:
Decouples the metadata store from the repo, allow service that accept only a config to read the contents of metadata stored in the repo.
feat: client and server include ids in job meta
  - clients pass their instance and installation id in headers, this is
      added to the jobs metadata when submitted
  - server includes their instance and installation id in meta of job.
  - Now, given a job spec it may be determined which node submitted the job
    and which node receieved the submission


cmd/cli/serve/serve.go
```diff
@@ -247,14 +251,19 @@ func serve(cmd *cobra.Command, cfg types.Bacalhau, fsRepo *repo.FsRepo) error {
...
 	if !cfg.DisableAnalytics {
-		installationID, err := fsRepo.ReadInstallationID()
+		installationID, err := metastore.ReadInstallationID()
 		if err != nil {
 			log.Trace().Err(err).Msg("failed to read installationID")
 		}
+		instanceID, err := metastore.ReadInstanceID()
+		if err != nil {
+			log.Trace().Err(err).Msg("failed to read instanceID")
+		}
```
wdbaruni, line 261: trace are mainly for more granular and request level logging. It might be useful to use debug level here

cmd/util/api.go
```diff
@@ -34,13 +35,24 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
 
+	metastore := repo.NewMetadataStore(cfg.DataDir)
+	installationID, err := metastore.ReadInstallationID()
+	if err != nil {
+		return nil, err
+	}
+	instanceID, err := metastore.ReadInstanceID()
+	if err != nil {
+		return nil, err
+	}
```
wdbaruni, line 46: these should be best effort instead of failing the request

cmd/util/api.go
```diff
@@ -34,13 +35,24 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
-		apimodels.HTTPHeaderBacalhauBuildDate:  {bv.BuildDate.UTC().String()},
-		apimodels.HTTPHeaderBacalhauBuildOS:    {bv.GOOS},
-		apimodels.HTTPHeaderBacalhauArch:       {bv.GOARCH},
+		apimodels.HTTPHeaderBacalhauGitVersion:          {bv.GitVersion},
+		apimodels.HTTPHeaderBacalhauGitCommit:           {bv.GitCommit},
+		apimodels.HTTPHeaderBacalhauBuildDate:           {bv.BuildDate.UTC().String()},
+		apimodels.HTTPHeaderBacalhauBuildOS:             {bv.GOOS},
+		apimodels.HTTPHeaderBacalhauArch:                {bv.GOARCH},
+		apimodels.HTTPHeaderBacalhauBuildInstallationID: {installationID},
+		apimodels.HTTPHeaderBacalhauInstanceID:          {instanceID},
```
wdbaruni, line 55: should be `HTTPHeaderBacalhauInstallationID` instead of `HTTPHeaderBacalhauBuildInstallationID`. Also should we always send the headers or only if they have values?

pkg/node/requester.go
```diff
@@ -53,6 +53,11 @@ type Requester struct {
 	debugInfoProviders []models.DebugInfoProvider
 }
 
+type MetaStore interface {
+	ReadInstallationID() (string, error)
+	ReadInstanceID() (string, error)
+}
+
```
wdbaruni, line 60: why introducing an interface instead of just passing the actual metastore to `NewRequesterNode`?

pkg/node/requester.go
```diff
@@ -209,11 +215,21 @@ func NewRequesterNode(
...
 	}
 
+	instanceID, err := metaStore.ReadInstanceID()
+	if err != nil {
+		return nil, err
+	}
+	installationID, err := metaStore.ReadInstallationID()
+	if err != nil {
+		return nil, err
+	}
```
wdbaruni, line 225: This is a common pattern I am seeing in different places in the code where we should treat these as best effort instead of failing. My recommendation is to have the metastore handle the failures internally by just logging and returning an empty id. If there are cases where you need to fail, then you can introduce `MustReadInstanceID()` methods.

What would be nice is:
1. Have the metastore read the ids from disk only once the first time a getter is called, store the entries in-memory and then serve subsequent requests from memory
2. If a failure happens while reading from disk, then it should log only once, and then all future calls will just get an empty id without more logging

pkg/orchestrator/endpoint.go
```diff
@@ -62,6 +62,9 @@ func (e *BaseEndpoint) SubmitJob(ctx context.Context, request *SubmitJobRequest)
 		analytics.EmitEvent(ctx, analytics.NewEvent(analytics.SubmitJobEventType, submitEvent))
 	}()
 
+	job.Meta[models.MetaClientInstallationID] = request.ClientInstallationID
+	job.Meta[models.MetaClientInstanceID] = request.ClientInstanceID
```
wdbaruni, line 66: should we only set these if the request has these metadata?

pkg/orchestrator/transformer/job.go
```diff
@@ -115,6 +115,22 @@ func RequesterInfo(requesterNodeID string) JobTransformer {
...
+}
+
+func OrchestratorInstallationID(installationID string) JobTransformer {
+	f := func(ctx context.Context, job *models.Job) error {
+		job.Meta[models.MetaServerInstallationID] = installationID
+		return nil
+	}
+	return JobFn(f)
+}
+
```
wdbaruni, line 133: the same here

pkg/repo/sysmeta.go
```diff
@@ -77,13 +90,13 @@ func (fsr *FsRepo) ReadLastUpdateCheck() (time.Time, error) {
...
 	return fsr.updateExistingMetadata(func(m *SystemMetadata) {
 		m.LastUpdateCheck = lastUpdateCheck
 	})
 }
 
-func (fsr *FsRepo) ReadInstallationID() (string, error) {
+func (fsr *MetadataStore) ReadInstallationID() (string, error) {
 	sysmeta, err := fsr.readMetadata()
 	if err != nil {
 		return "", err
```
wdbaruni, line 102: installationID should no longer be in the meta store. It is fine to keep the method here if you prefer, but the source is not `system_metadata.yaml`

cmd/cli/serve/serve.go
frrist: changing to trace

cmd/util/api.go
frrist: Will modify to log a debug message, and only include in headers if found and non-empty.

cmd/util/api.go
frrist: will only include if value is present: https://github.com/bacalhau-project/bacalhau/pull/4450#discussion_r1767521569

pkg/node/requester.go
frrist: To prevent the receivers of the type from calling methods that can modify the contents of the store - this is a read only store given this interface. It will also make testing easier in the future.

pkg/repo/sysmeta.go
frrist: Keeping it in the store prevents it from being scattered in different files. This was the design we decided here: https://www.notion.so/expanso/Rethinking-Configuration-435fbe87419148b4bbc5119d413786eb?pvs=4#5d5f7298b9ef4ab4bcb1fd6a44ca8c30. Additionally, for users migrating from v1.4.0 to v1.5.0, the installationID will be moved to the metadata store from its previous location the config file.

I know we discussed this in our meeting; How would you like to proceed here?
We can continuing doing what we do now and keep it in the store, or we can modify the metadata store to check both locations - its internal file and the file produced by the installation script. The latter feels like unnecessary complexity and coupling with the implementation of the install script.

wdbaruni: We know the installationID is not persisted in v1.4.0 https://github.com/bacalhau-project/bacalhau/issues/4085. The discussion to set the installationID outside of repo happened few weeks ago after the design doc, and that why you already changed the install script to write to the new location.

Lets please continue with the implementation to just check the new location

frrist: iirc https://github.com/bacalhau-project/bacalhau/issues/4085 was about adding the installationID to the job metadata, not persisting it to the filesysystem

frrist: Does this mean we should change the migration here? https://github.com/bacalhau-project/bacalhau/blob/main/pkg/repo/migrations/v3_4.go#L57. i.e. where do we put an installationID for users that already have on in their config?

pkg/node/requester.go
frrist: I have modified this slightly such that the entire metadata struct is returned rather than reading individual fields.

pkg/orchestrator/transformer/job.go
frrist: done.

pkg/repo/sysmeta.go
frrist: I have removed the installationID bits from the migration

cmd/util/api.go
```diff
@@ -35,24 +36,25 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
+	}
+
+	sysmeta, err := repo.LoadSystemMetadata(cfg.DataDir)
+	if err == nil {
+		if sysmeta.InstanceID != "" {
+			headers[apimodels.HTTPHeaderBacalhauInstanceID] = []string{sysmeta.InstanceID}
+		}
+		if installationID := config.ReadInstallationID(); installationID != "" {
+			headers[apimodels.HTTPHeaderBacalhauInstallationID] = []string{installationID}
+		}
```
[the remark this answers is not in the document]
frrist: This is addressed by your change merged in here.

pkg/node/requester.go
```diff
@@ -215,23 +212,24 @@ func NewRequesterNode(
...
+		if installationID := config.ReadInstallationID(); installationID != "" {
+			jobTransformers = append(jobTransformers, transformer.OrchestratorInstallationID(installationID))
+		}
+		if sysmeta.InstanceID != "" {
+			jobTransformers = append(jobTransformers, transformer.OrchestratorInstanceID(sysmeta.InstanceID))
+		}
+	} else {
+		log.Debug().Err(err).Msg("failed to read system metadata from repo")
+	}
+
```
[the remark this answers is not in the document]
frrist: also addressed by your change merged here

wdbaruni (approved):

PR #4454, opened 2024-09-18
add hint to get job downloads

udsamani:


cmd/util/printer/job_progress_printer.go
```diff
@@ -109,7 +109,7 @@ func NewJobProgressPrinter(client clientv2.API, runtimeSettings *cliflags.RunTim
 
 // PrintJobProgress displays the job progress depending upon on cli runtime
 // settings
-func (j *JobProgressPrinter) PrintJobProgress(ctx context.Context, jobID string, cmd *cobra.Command) error {
+func (j *JobProgressPrinter) PrintJobProgress(ctx context.Context, job *models.Job, jobID string, cmd *cobra.Command) error {
```
wdbaruni, line 112: nit: you don't need to pass both `job` and `jobID`

udsamani: Acknowledged. My  bad missed that. Fixing it!

PR #4468, opened 2024-09-18
timeout job progress

udsamani:
Screenshot 2024-09-19 005144


wdbaruni (approved):
Looks good. General feedback with similar PRs is to include a sample output of how things look like with the change

udsamani:
> Looks good. General feedback with similar PRs is to include a sample output of how things look like with the change

Done

PR #4482, opened 2024-09-19
fix: lock when accessing internal map

### LINKED ISSUE #4481: Node terminated with panic
Happened multiple times that the orchestrator node died with panic even though no job was processing at that time

Sample logs
[panic.md](https://github.com/user-attachments/files/17062438/panic.md)

frrist:
- fixes #4481

### DIFF
--- a/pkg/lib/collections/hashed_priority_queue.go
+++ b/pkg/lib/collections/hashed_priority_queue.go
@@ -5,7 +5,7 @@ import "sync"
 type HashedPriorityQueue[K comparable, T any] struct {
 	identifiers map[K]struct{}
 	queue       *PriorityQueue[T]
-	mu          sync.Mutex
+	mu          sync.RWMutex
 	indexer     IndexerFunc[K, T]
 }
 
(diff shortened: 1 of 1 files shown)


udsamani (approved):

pkg/lib/collections/hashed_priority_queue.go
```diff
@@ -27,6 +27,8 @@ func NewHashedPriorityQueue[K comparable, T any](indexer IndexerFunc[K, T]) *Has
 // Contains will return true if the provided identifier (of type K)
 // will be found in this queue, false if it is not present.
 func (q *HashedPriorityQueue[K, T]) Contains(id K) bool {
+	q.mu.Lock()
+	defer q.mu.Unlock()
 	_, ok := q.identifiers[id]
```
wdbaruni, line 32: should we use `RWMutex`?

wdbaruni (approved):
Approved, but we should switch to `RWMutex`

pkg/lib/collections/hashed_priority_queue.go
frrist: done.

PR #4469, opened 2024-09-19
reduce logging noise

wdbaruni:
### Before
```
❯ bacalhau-1.5.0 serve --orchestrator --compute --data-dir v1.5.0/data-dir
17:49:24.837 | WRN cmd/cli/serve/serve.go:146 > --name flag with value  ignored. Name n-a4e0671b-2679-4bc0-86b7-160ac093d2dc already exists
17:49:24.879 | INF cmd/cli/serve/serve.go:218 > starting bacalhau...
17:49:24.891 | INF pkg/nats/logger.go:47 > Starting nats-server [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.891 | INF pkg/nats/logger.go:47 >   Version:  2.10.11 [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.891 | INF pkg/nats/logger.go:47 >   Git:      [not set] [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.891 | INF pkg/nats/logger.go:47 >   Name:     n-a4e0671b-2679-4bc0-86b7-160ac093d2dc [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.891 | INF pkg/nats/logger.go:47 >   Node:     KQ4D6Bsn [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.891 | INF pkg/nats/logger.go:47 >   ID:       NCEPXW5MIPEZJFEYIWHZ6UVM3UKTG3SAYEP5RZ3U33RZSZ4A6HBYAEMM [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.893 | INF pkg/nats/logger.go:47 > Starting JetStream [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.893 | INF pkg/nats/logger.go:47 > ---------------- JETSTREAM ---------------- [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
17:49:24.893 | INF pkg/nats/logger.go:47 >   Max Memory:      48.00 GB [Server:n-a4e0671b-2679-4bc0-86b7-160ac093d2dc]
[...]

Issue #4457, opened 2024-09-18
Running a job with input does not work


aronchick:
```
 bacalhau docker run \
        -i https://gist.githubusercontent.com/js-ts/ec6a58c9e0991ed27dce5ca11198a88c/raw/fe16cb212661a30c46309da60fdd055ba4648bc5/fake_logs.log \
        ubuntu cat /inputs/fake_logs.log

Job successfully submitted. Job ID: j-731504aa-3ca9-4714-bda8-f0825b7bd708
Checking job status... (Enter Ctrl+C to exit at any time, your job will continue running):

 TIME                 EXEC. ID    TOPIC              EVENT
 Sep 18 11:36:53.113  e-452a3caa  Running Execution  creating container: Error response from daemon: invalid moun
                                                     t config for type "bind": invalid mount path: 'v1.5.0/data-d
                                                     ir/compute/executions/j-731504aa-3ca9-4714-bda8-f0825b7bd708
                                                     /e-452a3caa-8438-482f-8c51-f59183c0a907/3745027211/fake_logs
                                                     .log' mount path must be absolute
 Sep 18 11:36:53.269  e-b09cb95c  Running Execution  creating container: Error response from daemon: invalid moun
                                                     t config for type "bind": invalid mount path: 'v1.5.0/data-d
                                                     ir/compute/executions/j-731504aa-3ca9-4714-bda8-f0825b7bd708
                                                     /e-b09cb95c-0f05-42a7-ab2d-df8a77ae0662/2398540161/fake_logs
                                                     .log' mount path must be absolute
 Sep 18 11:36:53.281              Scheduling         not enough nodes to run job. requested: 1, available: 1, sui
                                                     table: 0.
                                                     • Node n-a4e0671b: job already executed on this node more th
                                                     an once
Error: job failed

Job Results By Node:

• Node n-a4e0671b: creating container: Error response from daemon: invalid mount config for type "bind": invalid mount path:
                     'v1.5.0/data-dir/compute/executions/j-731504aa-3ca9-4714-bda8-f0825b7bd708/e-452a3caa-8438-482f-8c51-f59183c0a907/3745027211/fake_logs.log' mount
                     path must be absolute

• Node n-a4e0671b: creating container: Error response from daemon: invalid mount config for type "bind": invalid mount path:
[...]

PR #4485, opened 2024-09-19
fix: resolve tilda paths and validate datadir config

### LINKED ISSUE #4484: The `--repo` and `--data-dir` flag don't expand paths
### Bug Description
Running the command `bacalhau --repo=~/b_requester <command>` results in a config with a `DataDir` value of `~/b_requester` rather than an expanded path, e.g. `/home/$USER/b_requester`

frrist:
Running the command `bacalhau --repo=~/some/path <command>` results in a config with a `DataDir` value of `~/some/path` rather than an expanded path, e.g.` /home/$USER/some/path`. We must handle tilda paths gracefully by expanding them to an absolute path before accepting them as configuration. 

Further, we must either dis-allow relative paths, or expand them to absolute paths, before accepting them as configuration. In this PR we have chosen the latter. This is because the `DataDir` field is used as the root path for our [hard-coded config paths](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/types/paths.go#L1), and said path must be absolute when provided to docker as a volume mount point, else we get errors like: https://github.com/bacalhau-project/bacalhau/issues/4506

This PR adds path expansion when `DataDir` is a tilda path, and validation to enforces `DataDir` be an absolute path before accepting it as configuration. 

Below are some examples of valid, and invalid paths - enforced via unit tests:
### Valid
```
DataDir:  "/absolute/path",
DataDir:  "~/absolute/path",
DataDir:  "~",
DataDir:  "~/",
DataDir:  "~/.",
DataDir:  "/path/with/special/char$",
DataDir:  "/path/with space",
DataDir:  "/absolute/path/",
DataDir:  "~/~/path",
DataDir:  "~/path/with space",
DataDir:  "/very/long/path/" + strings.Repeat("a", 255),
DataDir:  "/path/~to/file",
DataDir:  "~/path/~to/file",
DataDir:  "/path/with?invalid",
```
### Invalid
```
DataDir: "",
DataDir: "not/absolute/path",
DataDir: "./not/absolute/path",
DataDir: "~not/absolute/path",
```

- closes #4484


pkg/config/config.go
```diff
@@ -208,6 +210,29 @@ func New(opts ...Option) (*Config, error) {
...
+	dataDirPath := c.base.GetString(types.DataDirKey)
+	if dataDirPath[0] == '~' {
+		log.Info().Msgf("configuration field 'DataDir' contains '~': (%s). Attempting to expand to the home directory...", dataDirPath)
+		expanded, err := homedir.Expand(dataDirPath)
+		if err == nil {
+			dataDirPath = expanded
+			c.base.Set(types.DataDirKey, dataDirPath)
+			log.Info().Msgf("successfully expanded data directory to %s", dataDirPath)
+		}
+	}
```
wdbaruni, line 225: I am missing the purpose of expanding `~`. Also this is going to generate a lot of unnecessary logs for users. In most cases when I define a dataDir I use `~` and I believe most of our users do as we keep saying `~/.bacalahu` is the default repo

pkg/config/config.go
```diff
@@ -208,6 +210,29 @@ func New(opts ...Option) (*Config, error) {
...
+
+	// validate the config
+	var cfg types.Bacalhau
+	if err := c.base.Unmarshal(&cfg); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+	if err := cfg.Validate(); err != nil {
+		return nil, fmt.Errorf("config invalid: %w", err)
+	}
+
```
wdbaruni, line 235: this seems a little weird to unmarshal `types.Bacalhau` just for validation purposes, drop it, and then unmarshal again when setting up cli commands to get `types.Bacalhau` again

pkg/config/config_test.go
```diff
@@ -16,13 +16,13 @@ import (
...
 func TestConfigWithValueOverrides(t *testing.T) {
-	overrideRepo := "overrideRepo"
+	overrideRepo := "/overrideRepo"
 	overrideName := "puuid"
 	overrideClientAddress := "1.1.1.1"
 	overrideClientPort := 1234
 
 	defaultConfig := types.Bacalhau{
-		DataDir: "defaultRepo",
+		DataDir: "/defaultRepo",
```
wdbaruni, line 25: if relative paths are no longer supported, you should still keep a test case and validate if returns a well formated error

pkg/config/types/bacalhau.go
```diff
@@ -26,6 +32,15 @@ type Bacalhau struct {
...
 }
 
+func (b Bacalhau) Validate() error {
+	var err error
+	if !filepath.IsAbs(b.DataDir) {
+		err = errors.Join(err, fmt.Errorf("DataDir (%q) must be an absolute path", b.DataDir))
+	}
+
+	return err
+}
```
wdbaruni, line 42: I understand that relative paths can be confusing and result in unpredictable behaviour, but that shouldn't be the case for `~`. My recommendation is to no attempt to expand it i the config constructor, and validate here that the path is either abs or relative to $HOME

pkg/config/types/default_config.go
```diff
@@ -142,13 +145,36 @@ const defaultBacalhauDir = ".bacalhau"
...
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
+	defer zerolog.SetGlobalLevel(currentLevel)
+
+	// Check if the BACALHAU_DIR environment variable is set
 	if repoDir, set := os.LookupEnv("BACALHAU_DIR"); set && repoDir != "" {
 		return repoDir
+	} else if set {
+		log.Warn().Msg("BACALHAU_DIR environment variable is set but empty. Falling back to default directories.")
+	} else {
+		log.Debug().Msg("BACALHAU_DIR environment variable is not set. Trying to use $HOME.")
```
wdbaruni, line 159: These are a noisy and unnecessary logs. We now log the config path and repo path that we ended up using, and that should be enough

pkg/config/types/default_config.go
```diff
@@ -142,13 +145,36 @@ const defaultBacalhauDir = ".bacalhau"
...
 		return filepath.Join(userHome, defaultBacalhauDir)
+	} else {
+		log.Warn().Err(err).Msg("$HOME is unset or inaccessible. Falling back to current working directory.")
+	}
+
+	// Fallback: attempt to use the absolute path of the default directory
+	path, err := filepath.Abs(defaultBacalhauDir)
+	if err == nil {
+		log.Info().Str("Directory", path).Msg("Bacalhau will initialize in current working directory.")
+		return path
```
wdbaruni, line 174: I don't think we should. If no data directory was explicitly provided and user has no home, then we should just fail with a clear error message. 

Meaning this method should just return an empty data dir and let the config validation take care of the rest.

To be honest, and I think I've asked this multipe times before, I don't know why we are still looking into `BACALHAU_DIR` to get the default repo path. Default is `~/.bacalhau`, and then viper should take of the rest and replacing it with `BACALHAU_DIR` if it was provided

wdbaruni (changes requested):
Overall feedback, we need to slow down on logging. We just reduce our logging here https://github.com/bacalhau-project/bacalhau/pull/4469, and this PR seems like a regression

frrist:
> Also I am not sure what the problem the PR is trying to solve. I see you are linking an issue, but that issue also doesn't have a description on why data dir with ~ is a problem

Sorry, I should have provided a more verbose description in the issue - at a high level we need to expand `~` paths to be absolute/valid paths before using them for filesystem operations in bacalhau.

> I am missing the purpose of expanding `~`

As an example, `~/something` isn't a valid path: **`~` is a convenience provided by a shell, not a universally valid path component**. So, without expansion, `~/something` is just a string that won't lead to the intended directory unless the we explicitly handle it, and we don't currently.

More generally this PR is:
1. Handling `~` paths by expanding them from, say, `~/some/path` to `/home/$USER/some/path`
2. Validating the `DataDir` path is an absolute path - specifying the exact path on the file system.

> I believe most of our users do as we keep saying ~/.bacalahu is the default repo

Yes, this is the intention, the goal is for this to work as expected, right now it doesn't.

https://github.com/bacalhau-project/bacalhau/issues/4457 was actually caused by this issue, for example try this:

## Terminal 1
`bacalhau --repo=~/bacalhau serve --node-type=compute,requester`

## Terminal 2
`bacalhau docker run -i https://gist.githubusercontent.com/js-ts/ec6a58c9e0991ed27dce5ca11198a88c/raw/fe16cb212661a30c46309da60fdd055ba4648bc5/fake_logs.log  ubuntu cat /inputs/fake_logs.log`

Observe this is the error returned:
```
Job successfully submitted. Job ID: j-0195016a-2b62-4763-a86b-654da0ad6283

 TIME                 EXEC. ID    TOPIC              EVENT                                                        
 Sep 20 15:18:48.189  e-6f0519b4  Running Execution  creating container: Error response from daemon: invalid moun 
                                                     t config for type "bind": invalid mount path: '~/bacalhau/co 
                                                     mpute/executions/j-0195016a-2b62-4763-a86b-654da0ad6283/e-6f 
                                                     0519b4-c77f-445c-95fb-717585781d15/2114722141/fake_logs.log' 
                                                      mount path must be absolute                                 
 Sep 20 15:18:48.402  e-872ea25a  Running Execution  creating container: Error response from daemon: invalid moun 
                                                     t config for type "bind": invalid mount path: '~/bacalhau/co 
                                                     mpute/executions/j-0195016a-2b62-4763-a86b-654da0ad6283/e-87 
                                                     2ea25a-8b91-4b57-9f4f-04f886f3fdd0/4278892486/fake_logs.log' 
                                                      mount path must be absolute                                 
 Sep 20 15:18:48.420              Scheduling         not enough nodes to run job. requested: 1, available: 1, sui 
                                                     table: 0.                                                    
                                                     • Node n-ae099dd0: job already executed on this node more th 
                                                     an once                                                      

```

Since `~/bacalhau` was never expanded before being persisted to the config, methods that rely on it, like [`ComputeDir()`](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/types/paths.go#L80) end up returning an invalid (`~`) path - resulting in this error from docker.

pkg/config/config.go
frrist: > I am missing the purpose of expanding ~

see description here: https://github.com/bacalhau-project/bacalhau/pull/4485#issuecomment-2364714820

> Also this is going to generate a lot of unnecessary logs for users

Okay, I will remove the logs.

> In most cases when I define a dataDir I use ~ and I believe most of our users do as we keep saying ~/.bacalahu is the default repo

Thats the intention, but we need to expand it to a valid path before using it internally for filesystem operations.

pkg/config/types/bacalhau.go
frrist: At some point we need to expand it before using it for a FS operation, where/when should that expansion happen?

pkg/config/types/default_config.go
frrist: okay I will delete these

wdbaruni (commented):
looks good overall. still some comments not addressed

pkg/config/config.go
```diff
@@ -208,6 +210,27 @@ func New(opts ...Option) (*Config, error) {
 			absoluteConfigPaths[i] = path
 		}
 	}
+
+	// allow the users to set datadir to a path like ~/.bacalhau or ~/something/idk/whatever
+	// and expand the path for them
+	dataDirPath := c.base.GetString(types.DataDirKey)
+	if dataDirPath[0] == '~' {
+		expanded, err := homedir.Expand(dataDirPath)
```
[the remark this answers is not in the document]
frrist: done.

pkg/config/config.go
```diff
@@ -208,6 +210,27 @@ func New(opts ...Option) (*Config, error) {
...
+
+	// validate the config
+	var cfg types.Bacalhau
+	if err := c.base.Unmarshal(&cfg); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+	if err := cfg.Validate(); err != nil {
+		return nil, fmt.Errorf("config invalid: %w", err)
+	}
+
```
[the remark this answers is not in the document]
frrist: yup, done.

wdbaruni:
few examples that should work, but aren't:
```
→ bacalhau  serve --orchestrator --compute --data-dir=./here
failed to setup config: path must be absolute (start with '/')
DataDir ("./here") must be an absolute path

→ bacalhau  serve --orchestrator --compute --data-dir here
failed to setup config: path must be absolute (start with '/')
DataDir ("here") must be an absolute path


# this works, which maybe it shouldn't
→ bacalhau config set --config config.yaml datadir ./here
10:46:06.327 | INF ../ProtocolLabs/workspace/bacalhau/cmd/cli/config/set.go:94 > Writing config to /Users/walid/bacalhau-versions/config.yaml

# but trying to fix it fail
→ bacalhau config set --config config.yaml datadir $(PWD)/here
failed to setup config: path must be absolute (start with '/')
DataDir ("./here") must be an absolute path
```

First, please do more through manual testing from your end and document them in the PR's description

Second, you need to differentiate between setting the data-dir from config file and from cli flags as it makes sense if users define relative paths as cli flags, but I'll that to product to decide @aronchick 

Lastly, since we are validating when unmarshalling the config, then we need to also validate before persisting config changes using `config set`. Otherwise we will corrupt the config file and only manual intervention will be required to fix it. I might even say you should _only_ validate after writing when using `config set` and shouldn't validate before

wdbaruni (changes requested):
few examples that should work, but aren't:
```
→ bacalhau  serve --orchestrator --compute --data-dir=./here
failed to setup config: path must be absolute (start with '/')
DataDir ("./here") must be an absolute path

→ bacalhau  serve --orchestrator --compute --data-dir here
failed to setup config: path must be absolute (start with '/')
DataDir ("here") must be an absolute path


# this works, which maybe it shouldn't
→ bacalhau config set --config config.yaml datadir ./here
10:46:06.327 | INF ../ProtocolLabs/workspace/bacalhau/cmd/cli/config/set.go:94 > Writing config to /Users/walid/bacalhau-versions/config.yaml

# but trying to fix it fail
→ bacalhau config set --config config.yaml datadir $(PWD)/here
failed to setup config: path must be absolute (start with '/')
DataDir ("./here") must be an absolute path
```

First, please do more through manual testing from your end and document them in the PR's description

Second, the error message `path must be absolute (start with '/')` might be specific to unix and won't play nicely with windows

Third, you need to differentiate between setting the data-dir from config file and from cli flags as it makes sense if users define relative paths as cli flags, but I'll that to product to decide @aronchick 

Lastly, since we are validating when unmarshalling the config, then we need to also validate before persisting config changes using `config set`. Otherwise we will corrupt the config file and only manual intervention will be required to fix it. I might even say you should _only_ validate after writing when using `config set` and shouldn't validate before

frrist:
Right now this PR enforces the `DataDir` path be [absolute, or tilda paths](https://github.com/bacalhau-project/bacalhau/pull/4485/files#diff-29cea1b5b831c8655c7155f43e2367cec73e66d1e338f8b3c7877a2f339b8811R19), so it's understandable that: `--data-dir=./here` and `--data-dir here` don't work as they are not absolute. The command with `datadir $(PWD)/here` should work - least it works for me:
```bash
$ bacalhau config set datadir $(pwd)/here
07:25:50.478 | INF pkg/repo/fs.go:91 > Initializing repo at /home/frrist/.bacalhau
07:25:50.479 | INF cmd/cli/config/set.go:94 > Writing config to /home/frrist/.bacalhau/config.yaml

$ cat ~/.bacalhau/config.yaml 
datadir: /home/frrist/workspace/src/github.com/bacalhau-project/bacalhau/here
```

> First, please do more through manual testing from your end and document them in the PR's description

The tests are fairly thorough: https://github.com/bacalhau-project/bacalhau/pull/4485/files#diff-29cea1b5b831c8655c7155f43e2367cec73e66d1e338f8b3c7877a2f339b8811R19 is there a specific case you have in mind that is not covered here? Or it this comment more related to the nature of absolute vs relative paths?

> Second, the error message path must be absolute (start with '/') might be specific to unix and won't play nicely with windows

Good point, I'll modify this instead say: 
`failed to setup config: path must be absolute DataDir ("./here") must be an absolute path` since windows and unix path separators are different. (but note there is validation in this PR for both OS types)

> Third, you need to differentiate between setting the data-dir from config file and from cli flags as it makes sense if users define relative paths as cli flags, but I'll that to product to decide aronchick

For the case when paths are passed via flags, I can modify this PR such that it attempts to expand them to absolute paths before accepting them as config. If they cannot be expanded then we should fail and not accept them as configuration.

> Lastly, since we are validating when unmarshalling the config, then we need to also validate before persisting config changes using config set

Ooo yeah, good catch - will add validation to set before persisting the values to the file.

wdbaruni:
> The tests are fairly thorough:

You don't think unit tests for the config are enough, right? We keep finding issues with each config change, so obviously we can do better. As I mentioned several times before, try to break the feature, think of edge cases, and document the scenarios you ran. I am more than happy with manual tests for now, and we can migrate those to bashtub after launch.

> failed to setup config: path must be absolute DataDir ("./here") must be an absolute path

The message still has duplication

> For the case when paths are passed via flags, I can modify this PR such that it attempts to expand them to absolute paths before accepting them as config. If they cannot be expanded then we should fail and not accept them as configuration.

Yeah that sounds right

frrist:
@wdbaruni can you review this, please? Needed for 1.5.0

wdbaruni:
i don't understand this. We are now allowing relative paths even in config files where the original goal was to no longer support relative paths, and expand them only if provided in cli flags as you mentioned below

> For the case when paths are passed via flags, I can modify this PR such that it attempts to expand them to absolute paths before accepting them as config. If they cannot be expanded then we should fail and not accept them as configuration.

relative paths in config files result in unpredictable behaviour as the data-dir will depend on where you are running the binary from and not where the config file is. The reason it makes sense to support them in cli flags is because the user will provide a path relative to where they are running the binary and that make sense.

Here is an exmaple of unpredictable behavour:
```
→ pwd
/Users/walid/bacalhau-versions

→ touch config.yaml

→ bacalhau --config config.yaml config set datadir ./bacalhau

# note the data dir /Users/walid/bacalhau-versions/bacalhau
Walid-MacBook-5: ~/bacalhau-versions
→ bacalhau serve --orchestrator  --config config.yaml
23:54:07.727 | INF ../ProtocolLabs/workspace/bacalhau/cmd/cli/serve/serve.go:102 > Config loaded from: [/Users/walid/bacalhau-versions/config.yaml], and with data-dir /Users/walid/bacalhau-versions/bacalhau

→ cd ..

# note the data dir is now /Users/walid/bacalhau
Walid-MacBook-5: ~
→ bacalhau serve --orchestrator --config bacalhau-versions/config.yaml
23:55:08.08 | INF ProtocolLabs/workspace/bacalhau/cmd/cli/serve/serve.go:102 > Config loaded from: [/Users/walid/bacalhau-versions/config.yaml], and with data-dir /Users/walid/bacalhau

**Comment by frrist** [comment]
I see 56e7fd507cc14f4aece7a875a6ea7acfe89ab9f2. It seems we needed to add a couple things:
- Validate the datadir of each config file loaded by the config package, if any file contains an invalid data dir, fail.
- Ensure the `datadir` path provided to `config set` is expanded and valid. If its not valid, fail
- Prevent users from editing the `datadir` of the config file within their bacalhau repo via `config set` to avoid confusion. 

Here is the new behavior from your commands above with this change:
```
> pwd
/home/frrist/parent/child

> touch config.yaml

> bacalhau --config config.yaml config set datadir ./bacalhau
16:29:31.975 | INF cmd/cli/config/set.go:122 > Writing config to /home/frrist/parent/child/config.yaml

> cat config.yaml 
datadir: /home/frrist/parent/child/bacalhau

> bacalhau serve --orchestrator  --config config.yaml
16:29:57.526 | INF cmd/cli/serve/serve.go:102 > Config loaded from: [/home/frrist/parent/child/config.yaml], and with data-dir /home/frrist/parent/child/bacalhau
16:29:57.526 | INF pkg/repo/fs.go:94 > Initializing repo at /home/frrist/parent/child/bacalhau

> cd ..

> bacalhau serve --orchestrator --config child/config.yaml
16:30:15.967 | INF cmd/cli/serve/serve.go:102 > Config loaded from: [/home/frrist/parent/child/config.yaml], and with data-dir /home/frrist/parent/child/bacalhau
```

Additionally, removing the foot gun of modifying the data dir path of the config in the default repo location:
```
> bacalhau config set datadir somepath
16:32:19.456 | INF pkg/repo/fs.go:94 > Initializing repo at /home/frrist/.bacalhau

Error: modifying the config key "datadir" within the bacalhau repo config "/home/frrist/.bacalhau/config.yaml" is not permitted.
Hint:  You are free to do so manually, but advised against it.
```

Ensure paths provided to `config set` are expanded:
```
> touch config.yaml

> bacalhau config set --config=config.yaml datadir ./this_will_expand_in_file
16:33:34.743 | INF cmd/cli/config/set.go:122 > Writing config to /home/frrist/config.yaml

> cat config.yaml 
datadir: /home/frrist/this_will_expand_in_file
```

Ensure config files cannot be loaded with invalid paths already present:
```
> echo 'datadir: ./bad/path' > config.yaml

> bacalhau serve --config=config.yaml 

Error: failed to setup config: config file "/home/frrist/config.yaml" with value "./bad/path" invalid. datadir must be an absolute path
```

Works with several config files:
```
> echo 'datadir: /home/frrist/good' > config.yaml

> echo 'datadir: home/frrist/bad' > override.yaml

> bacalhau serve --config=config.yaml --config=override.yaml 
Error: failed to setup config: config file "/home/frrist/parent/override.yaml" with value "home/frrist/bad" invalid. datadir must be an absolute path

> bacalhau serve --config=override.yaml --config=config.yaml 
Error: failed to setup config: config file "/home/frrist/parent/override.yaml" with value "home/frrist/bad" invalid. datadir must be an absolute path

> bacalhau serve --config=config.yaml
16:36:35.823 | INF cmd/cli/serve/serve.go:102 > Config loaded from: [/home/frrist/parent/config.yaml], and with data-dir /home/frrist/good
16:36:35.823 | INF pkg/repo/fs.go:94 > Initializing repo at /home/frrist/good

> bacalhau serve --config=override.yaml 

Error: failed to setup config: config file "/home/frrist/parent/override.yaml" with value "home/frrist/bad" invalid. datadir must be an absolute path
```

The --config flags still accept relative paths, and expand them:
```
> bacalhau serve --config=datadir=relative/path
16:38:48.631 | INF cmd/cli/serve/serve.go:102 > Config loaded from: [], and with data-dir /home/frrist/parent/relative/path
16:38:48.631 | INF pkg/repo/fs.go:94 > Initializing repo at /home/frrist/parent/relative/path
```

File: pkg/config/config.go [hunk 9]
```diff
@@ -212,6 +217,26 @@ func New(opts ...Option) (*Config, error) {
...
 	}
 
+	// log the resolved config paths
+	absoluteConfigPaths := make([]string, len(c.paths))
+	for i, path := range c.paths {
+		absoluteConfigPaths[i], err = filepath.Abs(path)
+		if err != nil {
+			absoluteConfigPaths[i] = path
+		}
+	}
```
wdbaruni, line 227: I don't think this is needed as we are replacing paths with abs paths in 119

pkg/config/config.go
```diff
@@ -212,6 +217,26 @@ func New(opts ...Option) (*Config, error) {
...
+	// allow the users to set datadir to a path like ~/.bacalhau or ~/something/idk/whatever
+	// and expand the path for them
+	dataDirPath := c.base.GetString(types.DataDirKey)
+	if dataDirPath != "" {
+		dataDirPath, err := ExpandPath(dataDirPath)
+		if err != nil {
+			return nil, fmt.Errorf("%s invalid: %w", types.DataDirKey, err)
+		}
+		c.base.Set(types.DataDirKey, dataDirPath)
+	}
```
wdbaruni, line 238: do we need to expand? we do validate in line 138

pkg/config/types/default_config.go
```diff
@@ -134,13 +137,33 @@ const defaultBacalhauDir = ".bacalhau"
 // 2. User's home directory with .bacalhau appended.
 // 3. If all above fail, use .bacalhau in the current directory.
 func DefaultDataDir() string {
+	// this method runs before root.go, so we set the level to info for these calls, then return it to previous value
+	currentLevel := zerolog.GlobalLevel()
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
+	defer zerolog.SetGlobalLevel(currentLevel)
+
+	// Check if the BACALHAU_DIR environment variable is set
```
wdbaruni, line 145: my concern with the warning logs you introduced here is there is no way for the user to avoid them, except by setting the env variable. If they run `--data-dir <path>` they will still get warnings if they don't have home directory which doesn't make sense.

This method is about figuring out the default path, and it shouldn't fail if it couldn't. The config validation should fail later

wdbaruni (changes requested):

pkg/config/config.go
frrist: whoops, I think this is residual from a merge, will remove this.

pkg/config/config.go
frrist: 138 is validating the contents of the config file(s) passed, ensuring iff repo dir is present in the file its an absolute path. 

The logic here is expanding values provided via flags, envvars, and --config flags, expanding them on the users behalf.

pkg/config/types/default_config.go
frrist: We only warn if the user set the env var to an empty value, a rare case, but possible. Would you prefer we remove the log line?

wdbaruni: Not true. You warn when you can't find the home directory and when you can't figure out a default repo

frrist: My previous comment was in the context of the line you commented on. True, for other cases we log a warning: If we cannot determine a users home directory, and thus the default location where bacalhau initializes/opens it repo this seems worth a warning. But it sounds like you rather we remove all these log lines, is that correct?

wdbaruni:
## Testing done
1. Unit tests
2. Manual scenarios as show below


### Default

```
# defaults
→ bacalhau serve --orchestrator --compute
12:16:00.093 | INF  > Config loaded from: [], and with data-dir /Users/walid/.bacalhau
```

### Env Var
```
# env var: tila path
Walid-MacBook-5: ~
→ export BACALHAU_DIR="~/.bacalhau2"

Walid-MacBook-5: ~
→ bacalhau serve --orchestrator --compute
12:16:30.991 | INF  > Config loaded from: [], and with data-dir /Users/walid/.bacalhau2

# env var: relative path
Walid-MacBook-5: ~
→ export BACALHAU_DIR=".bacalhau"

Walid-MacBook-5: ~
→ bacalhau serve --orchestrator --compute

Error: data dir path ".bacalhau" is not an absolute path
Hint:  Use an absolute path for the data directory

Walid-MacBook-5: ~
→ bacalhau serve --orchestrator --compute

Walid-MacBook-5: ~
→ export BACALHAU_DIR="bacalhau"

Walid-MacBook-5: ~
→ bacalhau serve --orchestrator --compute

Error: data dir path "bacalhau" is not an absolute path
Hint:  Use an absolute path for the data directory

Walid-MacBook-5: ~
→ export BACALHAU_DIR="./bacalhau"

Walid-MacBook-5: ~
→ bacalhau serve --orchestrator --compute

Error: data dir path "./bacalhau" is not an absolute path
Hint:  Use an absolute path for the data directory
```

### Cli Flag
```
Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau serve --orchestrator --compute --data-dir .
12:22:49.32 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test

→ bacalhau serve --orchestrator --compute --data-dir dir
12:23:14.236 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute --data-dir ./dir
12:23:19.186 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute --data-dir ~/bacalhau-versions/test/dir
12:23:33.105 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute --data-dir ~/bacalhau-versions/test/dir2
12:23:37.277 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir2

→ bacalhau serve --orchestrator --compute --data-dir /Users/walid/bacalhau-versions/test

12:24:38.836 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test
```

### Config Value
```
→ bacalhau serve --orchestrator --compute -c datadir=.

12:25:28.673 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test

→ bacalhau serve --orchestrator --compute -c datadir=dir

12:25:38.466 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute -c datadir=./dir

12:26:03.447 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute -c datadir=~/bacalhau-versions/test/dir

12:32:49.703 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute -c datadir=/Users/walid/bacalhau-versions/test

12:27:32.918 | INF  > Config loaded from: [], and with data-dir /Users/walid/bacalhau-versions/test
```

### Config File
```
→ bacalhau serve --orchestrator --compute -c config.yaml

Error: data dir path "." is not an absolute path
Hint:  Use an absolute path for the data directory

Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau serve --orchestrator --compute -c config.yaml

Error: data dir path "dir" is not an absolute path
Hint:  Use an absolute path for the data directory

Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau serve --orchestrator --compute -c config.yaml

Error: data dir path "./dir" is not an absolute path
Hint:  Use an absolute path for the data directory

→ bacalhau serve --orchestrator --compute -c config.yaml
12:39:52.24 | INF  > Config loaded from: [/Users/walid/bacalhau-versions/test/config.yaml], and with data-dir /Users/walid/bacalhau-versions/test/dir

→ bacalhau serve --orchestrator --compute -c config.yaml
12:40:12.412 | INF  > Config loaded from: [/Users/walid/bacalhau-versions/test/config.yaml], and with data-dir /Users/walid/bacalhau-versions/test/dir2

→ bacalhau serve --orchestrator --compute -c config.yaml
12:40:22.19 | INF  > Config loaded from: [/Users/walid/bacalhau-versions/test/config.yaml], and with data-dir /Users/walid/bacalhau-versions/test

```

### Config Set
```
Walid-MacBook-5: ~/bacalhau-versions/test
→  bacalhau --config config.yaml config set datadir .
DEPRECATED: use key=value instead of space-separated key value
Writing config to /Users/walid/bacalhau-versions/test/config.yaml

Walid-MacBook-5: ~/bacalhau-versions/test
→ cat config.yaml
datadir: /Users/walid/bacalhau-versions/test

Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau --config config.yaml config set datadir ~/.bacalhau
DEPRECATED: use key=value instead of space-separated key value
Writing config to /Users/walid/bacalhau-versions/test/config.yaml

Walid-MacBook-5: ~/bacalhau-versions/test
→ cat config.yaml
datadir: /Users/walid/.bacalhau

Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau --config config.yaml config set datadir=/Users/walid/bacalhau-versions
Writing config to /Users/walid/bacalhau-versions/test/config.yaml

Walid-MacBook-5: ~/bacalhau-versions/test
→ cat config.yaml
datadir: /Users/walid/bacalhau-versions

Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau --config config.yaml config set datadir=
Writing config to /Users/walid/bacalhau-versions/test/config.yaml

Walid-MacBook-5: ~/bacalhau-versions/test
→ cat config.yaml
datadir: /Users/walid/bacalhau-versions/test

Walid-MacBook-5: ~/bacalhau-versions/test
→ bacalhau --config config.yaml config set datadir=""
Writing config to /Users/walid/bacalhau-versions/test/config.yaml

Walid-MacBook-5: ~/bacalhau-versions/test
→ cat config.yaml
datadir: /Users/walid/bacalhau-versions/test
```

wdbaruni (approved):

PR #4486, opened 2024-09-20
push output to stdout

udsamani:
due to some weird reason, the output for docker run command is getting directed to stderr. With this we make sure that it goes to stdout. Hence making sure the piping works correctly.

Screenshot 2024-09-20 034811
Screenshot 2024-09-20 034841


wdbaruni:
Lets delay this PR. I am already working on changing the progress output as there are other related issues

wdbaruni:
fixed in #4488

PR #4498, opened 2024-09-24
introduce system.GlobalConfig + migrate installation id

wdbaruni:
After another look, we actually do persist the installationID in the `config.yaml` file in v1.4.0. The bug was that if no installationID existed, we kept on generating new ones with each run and not persisting it. So if `config.yaml` has an installation, we should migrate it to `~/.config/bacalhau/installation_id` which is what this PR does.

Few changes were required to achieve _and_ test this:
1. Moved `GetInstallationID()` from `config` to `system` package as the installationID is not really configurable
2. Faced import cycle that forced me to move `system/tracer.go` to `telemetry/tracer.go`, which is what is causing a lot of files to change in the PR
3. Introduced `system.GlobalConfig` interface to allow mocking the `ConfigDir()` and be able to test the migrations properly as we don't want the tests to actually update `~/.config/bacalhau`

### DIFF
--- a/pkg/node/requester.go
+++ b/pkg/node/requester.go
@@ -8,7 +8,6 @@ import (
 	"github.com/rs/zerolog/log"
 	"go.opentelemetry.io/otel/attribute"
 
-	"github.com/bacalhau-project/bacalhau/pkg/config"
 	legacy_types "github.com/bacalhau-project/bacalhau/pkg/config_legacy/types"
 	"github.com/bacalhau-project/bacalhau/pkg/lib/backoff"
 	"github.com/bacalhau-project/bacalhau/pkg/lib/ncl"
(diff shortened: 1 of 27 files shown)


pkg/node/requester.go
```diff
@@ -212,24 +212,21 @@ func NewRequesterNode(
 		translationProvider = translation.NewStandardTranslatorsProvider()
 	}
 
+	installationID := system.InstallationID()
+	var instanceID string
+	if sysmeta, err := fsr.SystemMetadata(); err == nil {
+		instanceID = sysmeta.InstanceID
+	}
```
frrist, line 219: we should still log (trace or debug)) in the event this fails, for our own sake when/if debugging.

pkg/repo/migrations/helpers.go
```diff
@@ -141,3 +143,18 @@ func copyFile(srcPath, dstPath string) error {
...
+		return fmt.Errorf("creating config dir: %w", err)
+	}
+
+	// Write installation ID to file
+	installationIDFile := filepath.Join(cfg.ConfigDir(), system.InstallationIDFile)
+	if err := os.WriteFile(installationIDFile, []byte(installationID), util.OS_USER_RW); err != nil {
+		return fmt.Errorf("writing installation ID file: %w", err)
+	}
+	return nil
+}
```
frrist, line 160: nit: I would prefer we accept the path as a parameter rather than the `GlobalConfig` interface.

frrist (approved):
LGTM. I will land the

frrist:
On second thought, I'll just merge this into my branch.

PR #4526, opened 2024-09-25
fix not persisting instanceID

wdbaruni:
We are failing to persist instanceID today when initializing the repo as `system_metadta.yaml` doesn't exist yet and we call the update method instead of updateOrCreate.

Also this PR sends the instanceID to the version update checker instead of clientID. No immediate change is needed in the update server as it doesn't check for the clientID


wdbaruni:
@frrist can you please review this? we need to merge it before v1.5.0

frrist (approved):

PR #4532, opened 2024-09-25
feat: add execution analytics

frrist:
- also ensure JobID is called job_id in all events for easier sql


pkg/analytics/executions.go
```diff
@@ -0,0 +1,95 @@
...
+	NodeNameHash string `json:"node_name_hash,omitempty"`
+	// hash of namespace execution is running in
+	NamespaceHash string `json:"namespace_hash,omitempty"`
+
+	// AllocatedResources
+	// NB: total execution resources for all tasks in execution
+	TotalCPUUnits    float64 `json:"total_cpu_units,omitempty"`
+	TotalMemoryBytes uint64  `json:"total_memory_bytes,omitempty"`
+	TotalDiskBytes   uint64  `json:"total_disk_bytes,omitempty"`
+	TotalGPUCount    uint64  `json:"total_gpu_count,omitempty"`
```
wdbaruni, line 26: lets include gpu types as well

pkg/analytics/executions.go
```diff
@@ -0,0 +1,95 @@
...
+
+	DesiredState        string `json:"desired_state,omitempty"`
+	DesiredStateMessage string `json:"desired_state_message,omitempty"`
+	ComputeState        string `json:"compute_state,omitempty"`
+	ComputeStateMessage string `json:"compute_state_message,omitempty"`
```
wdbaruni, line 31: as discussed yesterday, lets just include the error code with this event, and have a separate event with raw messages. The other event can just have the execution id and raw messages if you want to keep things simple. We should only publish those when the execution fails or rejected as we don't care about tracking those for successful executions

pkg/analytics/executions.go
```diff
@@ -0,0 +1,95 @@
...
+
+	PublishedResultType string `json:"published_result_type,omitempty"`
```
wdbaruni, line 33: isn't this the same as `publisher_type`? I mean you can publish it along with engine types and input source types that you publish with job event, but just be consistent about field naming

pkg/analytics/executions.go
```diff
@@ -0,0 +1,95 @@
...
+
+	//RunCommandResult
+	RunResultStdoutTruncated bool `json:"run_result_stdout_truncated,omitempty"`
+	RunResultStderrTruncated bool `json:"run_result_stderr_truncated,omitempty"`
+	RunResultExitCode        int  `json:"run_result_exit_code,omitempty"`
+	// TODO determine if this contains PII.
+	RunResultErrorMessage string `json:"run_result_error_message,omitempty"`
```
wdbaruni, line 40: except for exit code, all these include PII and we shouldn't collect them. Not even in the new event we talked about yesterday

pkg/analytics/executions.go
```diff
@@ -0,0 +1,95 @@
...
+		stderrTruncated = e.RunOutput.StderrTruncated
+		exitCode = e.RunOutput.ExitCode
+		errorMessage = e.RunOutput.ErrorMsg
+	}
+
+	event := ExecutionEvent{
+		ExecutionID:              e.ID,
+		JobID:                    e.JobID,
+		EvalID:                   e.EvalID,
+		Name:                     e.Name,
```
wdbaruni, line 70: I don't this is being used at the moment, but lets publish the hash value for now. Or better just publish `name_set` similar to job events

pkg/analytics/job_submit.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	// if we can't parse the resources use zero
+	taskResources, err := t.ResourcesConfig.ToResources()
+	if err != nil {
+		taskResources = &models.Resources{
+			CPU:    0,
+			Memory: 0,
+			Disk:   0,
+			GPU:    0,
+			GPUs:   nil,
```
wdbaruni, line 58: we need to capture types of GPUs used as well

pkg/jobstore/boltdb/store.go
```diff
@@ -1220,6 +1221,8 @@ func (b *BoltJobStore) updateExecution(tx *bolt.Tx, request jobstore.UpdateExecu
 		}
 	}
 
+	analytics.EmitEvent(context.TODO(), analytics.NewExecutionEvent(newExecution))
+
```
wdbaruni, line 1225: we should not publish the event for each execution update. The most valuable event when the execution is terminated as it will give us how long the execution ran, whether it succeeded or not, and reason of failure. Similar to `job.terminal` event.

If you want to publish another event for when the execution was created, then that is fine, but that is mainly going to be useful for long running jobs.

In summary, the must have is `bacalhau.execution_v1.terminal`, and if you want to publish the execution created event, it should be a separate `bacalhau.execution_v1.create`

wdbaruni (changes requested):
My overall preference as discussed before is to have a single self-contained event that includes job information, and summary of execution stats related the job, such as execution count, state counts, reasons for failures or rejection, number of nodes, ... 

The intention of that ask as explained before is to simplify processing of events and avoid having to do joins as much as possible.

Unfortunately that is not what is delivered here as we are publishing a dedicated event per execution, which is nice to have, but the job event now doesn't have reasons why the job actually failed. The message field in the job is not the actual reason as discussed.

Time is short to get this right before 1.5.0. Lets continue with this implementation, but please create a backlog item to enrich the job event

pkg/analytics/executions.go
frrist: Okay, I can include this data in the event.

fwiw, I don't believe we currently set the `AllocatedResources` field of an execution, rather we read it from the `Job` field currently included in the execution.

frrist: Upon second look, we actually never set the `GPUs` field of `Resources` when submitting a job. The `GPUs` field is included in the `Resources` for the purpose of sharing `NodeInfo` and `Labels`. 

Still, I'll include it anyways as you prefer.

pkg/analytics/executions.go
frrist: > lets just include the error code with this event

What field is the error code? Do we mean the `ExitCode` from `RunCommandResult`?

pkg/analytics/executions.go
frrist: `RunResultStdoutTruncated` and `RunResultStderrTruncated` are both bools, are we sure that's PII?
I will remove the `RunResultErrorMessage` field

pkg/analytics/job_submit.go
frrist: This information isn't included in a task.

frrist:
> My overall preference as discussed before is to have a single self-contained event that includes job information, and summary of execution stats related the job, such as execution count, state counts, reasons for failures or rejection, number of nodes, ...

To clarify, how would you envision this information being collected when a job terminates execution?

Would the approach be to query the BucketJobHistory bucket when a job reaches a terminal state—extracting events related to the job and its executions into a single structure? Assuming that’s the intended outcome, how would you prefer to gather fields not present in the job history (like type, count, task engine/publisher/inputs, etc.)? Should we also query the BucketJobs bucket to enrich the event with that information? Similarly, for execution fields not present in the execution history, would querying BucketJobExecutions also be needed?

> Unfortunately that is not what is delivered here as we are publishing a dedicated event per execution, which is nice to have, but the job event now doesn't have reasons why the job actually failed. The message field in the job is not the actual reason as discussed.
... please create a backlog item to enrich the job event.

I understand this PR may have missed your expectations, and I want to ensure we are aligned moving forward. Given the questions raised above, I think it would be really helpful if you could provide more specific direction or open an issue outlining the desired structure, fields, and means of collecting the data. This would allow us to hit the mark more accurately and meet the goals you've outlined.

pkg/jobstore/boltdb/store.go
frrist: Does this imply we don't want to track executions that are related to bidding?

pkg/analytics/executions.go
wdbaruni: No its is bacerrors.ErrorCode. Unfortunately I just realized we are not adding that error code to `models.Event` and not propagating it back to the orchestrator. We can do this in v1.5.1 I guess

pkg/analytics/executions.go
wdbaruni: oh you are right. I thought those were the truncated outputs

wdbaruni:
lets focus on getting this PR out and then we can discuss how to aggregate execution stats in the job event next week

frrist:
Sounds good, I believe I have addressed all feedback.

wdbaruni (approved):

PR #4533, opened 2024-09-26
fix python apiclient build

udsamani:
Since the WebUI build merge, the python apiclient build has been broken. This fixes it.


pkg/swagger/docs.go
```diff
@@ -1245,48 +1245,25 @@ const docTemplate = `{
...
-                "Minute",
-                "Hour"
-            ]
-        },
         "github_com_bacalhau-project_bacalhau_pkg_config_legacy_types.JobDefaults": {
             "type": "object",
             "properties": {
                 "executionTimeout": {
-                    "$ref": "#/definitions/github_com_bacalhau-project_bacalhau_pkg_config_legacy_types.Duration"
+                    "type": "integer"
```
wdbaruni, line 1252: these are not integers but durations. The values can be 10s for seconds or 10h for hours. The new type is now at https://github.com/bacalhau-project/bacalhau/blob/1ca678c752371f790e84220e6061e63d3aed5e32/pkg/config/types/duration.go#L9

pkg/config_legacy/types/compute.go
```diff
@@ -30,21 +30,21 @@ type JobTimeoutConfig struct {
 	// check.
 	JobExecutionTimeoutClientIDBypassList []string `yaml:"JobExecutionTimeoutClientIDBypassList"`
 	// JobNegotiationTimeout default timeout value to hold a bid for a job
-	JobNegotiationTimeout Duration `yaml:"JobNegotiationTimeout"`
+	JobNegotiationTimeout Duration `yaml:"JobNegotiationTimeout" swaggertype:"primitive,integer"`
```
wdbaruni, line 33: not sure why we are trying to change config_legacy types. They shouldn't be in use except in the migration path

wdbaruni (changes requested):

pkg/config_legacy/types/compute.go
udsamani: Until and unless we get rid of them from the repo, they will be generated, as they are in path of our swagger generation.

pkg/swagger/docs.go
udsamani: https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/time/time.go;l=632

udsamani: It is a limitation of our tool we use for swagger generation. https://github.com/swaggo/swag/issues/388

udsamani: But ideally 'Duration' can be casted to int64 as internally go uses the same.

udsamani: @wdbaruni  Let me know if this makes sense ? I can accordingly merge this ! Thanks !

wdbaruni: I understand. My problem is we are doing this change in the legacy config type instead of the one actually being used, and that we are defining the actual type of `types.Duration` in its usages instead of its definition or at least in a single place.

I have pushed a PR with these fixes that is based on this one https://github.com/bacalhau-project/bacalhau/pull/4552

If all looks good, feel free to merge it into this PR or directly into main

wdbaruni (approved):

Issue #4541, opened 2024-09-26
Shouldn't print to end users raw go logs


aronchick:
```
![CleanShot 2024-09-26 at 07 48 51](https://github.com/user-attachments/assets/917280a1-4bd9-44e6-b020-b66a4d9c8e68)
```

Should probably be:
```
Writing config to /Users/daaronch/.bacalhau/config.yaml
```

Could be worth getting a copy editor on all these!

PR #4553, opened 2024-09-30
use cmd.Print instead of logs with

wdbaruni:
### Example:
```
Walid-MacBook-5: ~/bacalhau-versions
→ bacalhau config --config config.yaml set api.port 1122
Writing config to /Users/walid/bacalhau-versions/config.yaml
```

closes #https://github.com/bacalhau-project/bacalhau/issues/4541


cmd/cli/config/set.go
```diff
@@ -90,8 +89,8 @@ This command supports two input formats for setting configuration values:
 	return setCmd
 }
 
-func setConfig(cfgFilePath, key string, value ...string) error {
-	log.Info().Msgf("Writing config to %s", cfgFilePath)
+func setConfig(cmd *cobra.Command, cfgFilePath, key string, value ...string) error {
+	cmd.Printf("Writing config to %s", cfgFilePath)
```
frrist, line 93: We should write this line right before `if err := v.WriteConfigAs(cfgFilePath); err != nil {` to ensure we only print it when we write the file.

frrist (approved):

cmd/cli/config/set.go
wdbaruni: is says `Writing` and not `Wrote`. Useful to know when it is intending to write so failures can make more sense

PR #4565, opened 2024-09-30
Slightly improve API error handling

jamlo:
This commit tries to fix the "FromHttpResponse" function. Returning 2 errors from a function can be confusing and uncommon in Golang, especially if the function only returns these 2 errors.

The new code will return one error, which is always an APIError, and it also improves the error messaging wording, adding more details in the output.

Several more tweaks can be done to the current error message handling if needed.


pkg/publicapi/apimodels/error.go
```diff
@@ -67,31 +66,37 @@ func (e *APIError) Error() string {
...
 // Parse HTTP Resposne to APIError
-func FromHttpResponse(resp *http.Response) (*APIError, error) {
-
+func GenerateAPIErrorFromHTTPResponse(resp *http.Response) *APIError {
 	if resp == nil {
-		return nil, errors.New("response is nil, cannot be unmarsheld to APIError")
+		return NewAPIError(0, "API call error, invalid response")
 	}
 
 	defer resp.Body.Close()
```
frrist, line 74: I will preface this with: I am aware you are not the one that introduced this :smiley: . But the diff here caused be to notice a couple things.

We are closing the response body both inside this method and outside of it, e.g. in the file edit below on [line 88](https://github.com/bacalhau-project/bacalhau/pull/4565/files#diff-202b9471f43edeb5de262a1accf8cbe2be89e7af88bed8b70dd7d8b5cb0abaebR83). Probably best to avoid closing the response within this method, and leave it up to the callers/reduce unexpected side affects. Mind making that change here or in a follow on?

pkg/publicapi/apimodels/error.go
```diff
@@ -67,31 +66,37 @@ func (e *APIError) Error() string {
...
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return nil, fmt.Errorf("error reading response body: %w", err)
+		return NewAPIError(
+			resp.StatusCode,
+			fmt.Sprintf("Unable to read API response body. Error: '%s'", err.Error()))
```
frrist, line 80: nit: we can use `%q` here (and elsewhere) to add double quotes around the string and safely escapes any special characters. i.e.
```suggestion
			fmt.Sprintf("Unable to read API response body. Error: %q", err.Error()))
```

pkg/publicapi/client/v2/client.go
```diff
@@ -73,18 +73,13 @@ func (c *httpClient) Get(ctx context.Context, endpoint string, in apimodels.GetR
...
 	}
 
-	var apiError *apimodels.APIError
 	if resp.StatusCode != http.StatusOK {
-		apiError, err = apimodels.FromHttpResponse(resp)
-		if err != nil {
-			return err
+		apiError := apimodels.GenerateAPIErrorFromHTTPResponse(resp)
+		if apiError != nil {
+			return apiError
```
frrist, line 79: nit: compact this and others:
```suggestion
		if apiError := apimodels.GenerateAPIErrorFromHTTPResponse(resp); apiError != nil {
		    // code and stuff
		}
```

frrist (commented):
Good stuff,

Few comments, address them if you'd like, but they are non-blocking.

frrist (approved):

udsamani (approved):

pkg/publicapi/apimodels/error.go
jamlo: Thanks for the suggestion. Code changed.

pkg/publicapi/client/v2/client.go
jamlo: Thanks for the suggestion. Code changed.

wdbaruni:
What is the status of this PR? @jamlo feel free to merge if all comments are addressed

jamlo:
@wdbaruni since this PR was created (through a fork) before I had access to the repo, I will close this one, and create another one from a branch to close the loop.

jamlo:
Closing this PR in favor of https://github.com/bacalhau-project/bacalhau/pull/4607

PR #4573, opened 2024-10-01
webui backend configurable

wdbaruni:
Right now the webui only connects to localhost, which doesn't work when accessing the webui from remote locaions, such as trying to connect to demo network's webui


pkg/config/types/generated_constants.go
```diff
@@ -47,6 +47,7 @@ const ComputeAllocatedCapacityGPUKey = "compute.allocatedcapacity.gpu"
 const ComputeAllowListedLocalPathsKey = "compute.allowlistedlocalpaths"
 const WebUIEnabledKey = "webui.enabled"
 const WebUIListenKey = "webui.listen"
+const WebUIBackendKey = "webui.backend"
```
udsamani, line 50: Isn't listen and backend same ?

udsamani (approved):

pkg/config/types/generated_constants.go
wdbaruni: no listen is the address the webserver is listening to, but the frontend directly calls the orchestrator node (backend) and needs to know its address

the right thing is to have the webui webserver call the orchestrator and not the frontend (javascript), but that is a bigger change and unfortunately how the legacy webui used to work

PR #4574, opened 2024-10-01
fix not propagating heartbeat timeout

wdbaruni:


frrist:
Good find. 

As far as the heartbeat bug, I am not sure if this will fix it, since https://github.com/bacalhau-project/bacalhau/blob/2618c101c3a2dfd6bd60d6a6975a893ee8bdd475/pkg/node/config_defaults.go#L111 also sets it via a merge: https://github.com/bacalhau-project/bacalhau/blob/2618c101c3a2dfd6bd60d6a6975a893ee8bdd475/pkg/node/config_requester.go#L80

frrist (approved):

wdbaruni:
without the fix it is 30s, and with the fix it picks the 1minute!  You are not using type.NodeManager configurations at all

frrist:
Yes it looks like we have made some mistakes in implementation. What I am trying to point out here is the default has always been 30seconds: https://github.com/bacalhau-project/bacalhau/blob/v1.4.0/pkg/config/configenv/production.go#L179
I am unsure changing the value to a min will addresses the issues we are facing.

PR #4575, opened 2024-10-01
propagate more missing configs

wdbaruni:


frrist (approved):
once build passes LGTM

PR #4577, opened 2024-10-02
End-to-End Testing: Initial setup

jamlo:
Since Bacalhau is a distributed system with multiple components, it is critical to have a reliable method for end-to-end testing. Additionally, it's important that these tests closely resemble a real production environment without relying on mocks.

This initial setup addresses those needs by running Bacalhau inside containers while also supporting Docker workloads within these containers (using Docker-in-Docker, or DinD).

Notes:
1. This is a first step towards allowing end-to-end tests that identically runs on local machines as well ass CI. 
2. This allows to have a real world setup for integration tests 
3. Docker networks will allow all sorts of test scenarios of network failures, latency, and segmentation.
4. In the next step, Minio container will be added to test S3 interactions
5. IPFS could also be included in a seperate container as well.
6. Currently we need to privileged flag in containers to be true, but this can be changed when we use https://github.com/nestybox/sysbox 
7. This setup can be wrapped around a suite of integration tests written in Go. No need to mess around with bash scripts to test. 
8. This allows for easier experimentation for new users of Bacalhau, they can run the whole thing locally in one container (taking advantage of nested containers ability)
9. Allows for easier fuzz testing setup for later on. 
10. This can be adapted to have multiple requester and computer nodes scenarios. No mocking, actual real setup.

### DIFF
--- a/test-integration/Dockerfile
+++ b/test-integration/Dockerfile
@@ -1,0 +1,10 @@
+# Use the docker:dind image as the base image
+FROM docker:dind
+
+# Install curl and bash
+RUN apk update && apk add --no-cache curl bash
+
+# Download and execute the Bash script from the given URL
+RUN curl -sSL https://get.bacalhau.org/install.sh | bash
+
+ENTRYPOINT ["dockerd-entrypoint.sh"]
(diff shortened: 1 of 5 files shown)


jamlo:
Closing this pull request since it was created from a fork. Will create a new PR from a local branch in this repo.

PR #4578, opened 2024-10-02
run bashtubs as part of the CI

udsamani:


test/config_with_file_override_and_flag.sh
frrist, line 1: Is there a reason we are removing these tests?

udsamani: Don't know how you got that it is being deleted. It is updating chmod permission to allow it to run

udsamani:
Discarding this as the merges to main completely broke the bash tests again.

PR #4579, opened 2024-10-02
fix: only collect OS type; fix method name; add readme

frrist:
### What:
- Fix NodeID option name (cosmetic change)
- Only collect OS type, instead of all details
- Add a README.md describing what is collected, and how to opt out of collection


pkg/analytics/README.md
```diff
@@ -0,0 +1,118 @@
+# What Data is shared by users of Bacalhau?
```
frrist, line 1: @seanmtracey @aronchick is here is README on the data being collected, along with steps to opt out. Feedback welcome.

cmd/cli/serve/serve.go
```diff
@@ -260,7 +260,7 @@ func serve(cmd *cobra.Command, cfg types.Bacalhau, fsRepo *repo.FsRepo) error {
 
 	if !cfg.DisableAnalytics {
 		err = analytics.SetupAnalyticsProvider(ctx,
-			analytics.WithNodeNodeID(sysmeta.NodeName),
+			analytics.WithNodeID(sysmeta.NodeName),
```
frrist, line 263: @wdbaruni is collecting the name of the node acceptable, just want to double check.

pkg/analytics/analytics.go
```diff
@@ -108,7 +108,7 @@ func SetupAnalyticsProvider(ctx context.Context, opts ...Option) error {
 
 	// Create a new resource with auto-detected host information
 	res, err := resource.New(ctx,
-		resource.WithOS(),
+		resource.WithOSType(),
```
frrist, line 111: @wdbaruni I have reduced the information collected here (from what was requested in the initial POC) to just the operating system type. Prior to this change, this field contained data like:
```
Linux cypress 6.8.0-40-generic #40~22.04.3-Ubuntu SMP PREEMPT_DYNAMIC Tue Jul 30 17:30:19 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```
Which is beyond the scope of the information we want to share.

wdbaruni (approved):

PR #4580, opened 2024-10-02
remove config layers

wdbaruni:
This PR removes layers we have from the configs the user defines, and the configs that are actually applied. 

Along the way, it solves these other problems and adds these improvements:
1. Move default config from `types.Default`  to `config.Default`
2. Fixed devstack to use a dedicated repo per node instead of a single repo and overriding the paths of some dependencies. For example, all nodes were still using the same user_id.pem, nats storage, and system_metadata.yaml
3. Devstack wasn't configuring nodes with different local publisher ports
4. Enabled labels for orchestrator node as labels are using for API filtering and to get runtime information about the nodes, regardless compute or not
5. We were still allowing users to override the path for local publisher and they shouldn't as users now should only worry about defining the data-dir
6. Fixed missing nats token for compute nodes. Previously they were using the token defined under the orchestrator section which users shouldn't have to define if running compute only node
7. Initializing a node mainly cares about passing config instead of having to initialize a repo first or having to create many depdencies such as jobstore and execution store
8. Weren't propagating LogDebugInfoInterval properly 
9. Brought back locality job admission control as it does work for ipfs and local store, and might be better to provide a deprecation message instead of removing it now
10. We were creating new directories for local and s3 publisher with each run, which is not necessary and result is bunch of clutter in the repo when node doesn't exit gracefully 
11. We were not setting node's GPU capacity correctly and GPU jobs weren't passing

What is left is fixing the unit and integration tests with new config structure


cmd/cli/devstack/devstack.go
```diff
@@ -83,34 +78,12 @@ func NewCmd() *cobra.Command {
...
-				// configured repo if `--stack-repo` was specified
-				repoPath, _ = os.MkdirTemp("", "")
-
-				// If we don't set the repo value in config, readers will be given
-				// a different path to the one we've just created. Presumably a default.
-				defer os.RemoveAll(repoPath)
+			// TODO: a hack to force debug logging for devstack
+			//  until I figure out why flags and env vars are not working
+			if err := logger.ConfigureLogging(string(logger.LogModeDefault), "debug"); err != nil {
+				return fmt.Errorf("failed to configure logging: %w", err)
```
frrist, line 84: They aren't working because a config hasn't been initialized, e.g. via  `util.SetupConfig`, or similar.

pkg/compute/bidder.go
```diff
@@ -36,7 +34,6 @@ type Bidder struct {
 	usageCalculator capacity.UsageCalculator
 	executor        Executor
 	callback        Callback
-	getApproveURL   func() *url.URL
```
frrist, line 39: Is this feature being removed, or just moved somewhere else?

pkg/config/defaults.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return cfg.MergeNew(overrides)
+}
+
+// defaultDataDir determines the appropriate default directory for storing repository data.
+// Priority order:
+// 1. User's home directory with .bacalhau appended.
+func defaultDataDir() string {
+	if userHome, err := os.UserHomeDir(); err == nil && userHome != "" {
+		if expandedUserHome, err := filepath.Abs(userHome); err == nil {
```
frrist, line 205: nit: I don't think we need to expand the home dir. `os.UserHomeDir` returns an absolute path.

pkg/node/compute.go
```diff
@@ -269,77 +293,70 @@ func NewComputeNode(
...
-		NodeID:           nodeID,
+		NodeID:           cfg.NodeID,
 		SemanticStrategy: semanticBidStrats,
 		ResourceStrategy: resourceBidStrats,
 		Store:            executionStore,
 		Callback:         computeCallback,
 		Executor:         bufferRunner,
-		GetApproveURL: func() *url.URL {
-			return apiServer.GetURI().JoinPath("/api/v1/compute/approve")
-		},
```
frrist, line 342: If we are removing this, is there any point including `NewExternalHTTPStrategy`?

pkg/node/node.go
```diff
@@ -398,21 +329,32 @@ func createAPIServer(config NodeConfig, userKey *baccrypto.UserKey) (*publicapi.
...
+		return nil, err
+	}
+
+	// TODO: revisit how we setup the transport layer for compute only, orchestrator only and hybrid nodes
+	config := &nats_transport.NATSTransportConfig{
+		NodeID:                   cfg.NodeID,
+		Host:                     cfg.BacalhauConfig.Orchestrator.Host,
+		Port:                     cfg.BacalhauConfig.Orchestrator.Port,
+		AdvertisedAddress:        cfg.BacalhauConfig.Orchestrator.Advertise,
+		AuthSecret:               cfg.BacalhauConfig.Orchestrator.Auth.Token,
```
frrist, line 344: This single field, is used by both compute and orchestrator, right?

frrist (commented):
Left a few questions/thoughts

cmd/cli/serve/serve.go
```diff
@@ -168,65 +164,23 @@ func serve(cmd *cobra.Command, cfg types.Bacalhau, fsRepo *repo.FsRepo) error {
...
-		return errors.Wrapf(err, "failed to configure compute node")
-	}
-
-	requesterConfig, err := GetRequesterConfig(cfg, isRequesterNode)
-	if err != nil {
-		return errors.Wrapf(err, "failed to configure requester node")
-	}
-
 	// Create node config from cmd arguments
+	// TODO: validate if this is necessary
```
frrist, line 168: Upon inspecting the method, I don't believe this method does anything useful. Probably best to remove.

pkg/compute/bidder.go
wdbaruni: feature removed ages ago

cmd/cli/devstack/devstack.go
wdbaruni: thats what I thought initially, but we do configure the logger in root cmd and I don't see logging flags in devstack command

pkg/node/compute.go
wdbaruni: no that is still functional. This approval uri was introduced with insulated jobs

pkg/config/defaults.go
wdbaruni: In your tilda path PR you were validating if `UserHomeDir()` is abs or not, and I assumed you had a reason for that. Should I remove this?
https://github.com/bacalhau-project/bacalhau/pull/4485/commits/0387640d61af302a01e2b06c6daefbe1768a24a4#diff-5e48b782fb0f0b0ef48ec6cfe35aa63eb7d2390c411a9be112652233ba5ccafe

pkg/node/node.go
wdbaruni: it is. I am setting AuthSecret to be `cfg.BacalhauConfig.Compute.Auth.Token` if this is a compute only node.

```
if cfg.BacalhauConfig.Compute.Enabled && !cfg.BacalhauConfig.Orchestrator.Enabled {
		config.AuthSecret = cfg.BacalhauConfig.Compute.Auth.Token
	}

```

udsamani (approved):
LGTM.

pkg/config/defaults.go
frrist: Like this method?:
```go
func AbsPathSilent(path string) string {
	expandedPath, err := homedir.Expand(path)
	if err != nil {
		log.Debug().Msgf("failed to expand path %s: %v", path, err)
		return path
	}
	absPath, err := filepath.Abs(expandedPath)
	if err != nil {
		log.Debug().Msgf("failed to resolve absolute path for %s: %v", path, err)
		return path
	}
	return absPath
}
```
Slightly different, am attempting to expand a path containing a tilda, if the path doesn't contain a tilda the call is a noop, we then ensure its an absolute path - mainly in the event Expand was a noop.

wdbaruni: No I was referring to this if statement you had in `DefaultDataDir()`
```
if userHome, err := os.UserHomeDir(); err == nil && filepath.IsAbs(userHome) {
```

PR #4581, opened 2024-10-03
webui: show gpu even when available capacity is zero

wdbaruni:


frrist (approved):
lgtm - confirmed this renderers correctly in the webui

PR #4582, opened 2024-10-03
fix: remove warning flag causing noise

frrist:
This change was introduced in #4580. @wdbaruni was this intentional, or for debugging?


wdbaruni (approved):

wdbaruni:
Looks like i forgot to remove it

Issue #4595, opened 2024-10-07
Deploy Bacalhau using Docker Compose


jamlo:
### Is your feature request related to a problem? Please describe.

Since Bacalhau is a distributed system with multiple components, it is critical to have a reliable method for end-to-end testing. Also, it's important that these tests closely resemble a real production environment without relying on mocks. 

Additionally, it is important to have an easy way for new users of Bacalhau to quickly experiment with the system on their own local machines, without the need, or the complexity, or the overhead to deploy on a public cloud.

---
### Describe the solution you'd like

The suggested solution is to deploy different components of Bacalhau using Docker Compose, while utilizing the Docker-in-Docker capability. 

This solution:
1. Allows new users to standup a full deployment of Bacalhau on their own machines within seconds, and experiment with it. 
2. Gives the ability to write Jobs that uses S3 Input/Output sources by utilizing the Minio component deployed.
3. Removes the need to pull/push images from Docker Hub, and instead utilize the Image Registry component deployed.
4. Allows to easily and quickly run tests against a real Bacalhau deployment, and quickly iterate and test different setups and scenarios. Please note that Docker compose itself can be put inside a container (using Docker-in-Docker)
5. By utilizing Docker networks, it allows us to simulate different scenarios of network failures, latency, and segmentation between Compute Nodes and Requester Nodes.
6. Easily scale to multiple requester and compute nodes.
7. Allows Bacalhau developers to easily run tests locally before pushing their commits.
8. Paves the road for Fuzz testing 
9. Paves the way for easier testing TLS NATS communication (authentication and authorization)

---
### Describe alternatives you've considered
Alternatives would be to standup different processes on a single host, where each process is a different component of Bacalhau

---
### Additional context**

Few TODOs to keep in mind for later:
1. Utilize [Sysbox](https://github.com/nestybox/sysbox)
2. Think about Fuzz Tests and best to integrate them
3. IPFS support
4. Think about how to increase the number or requester nodes and compute nodes when needed to test different scenarios.
5. Utilize TestContainers when building the test suite around this setup

PR #4596, opened 2024-10-07
Support running Bacalhau in Docker compose

jamlo:
This pull requests addresses issue #4595


test-integration/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+    environment: *common-env-variables
+    depends_on:
+      bacalhau-minio-node:
+        condition: service_healthy
+    privileged: true
+    command:
+      - /bin/bash
+      - -c
+      - |
+        bacalhau config set "node.network.authsecret" "$${NETWORK_AUTH_TOKEN}" && bacalhau serve --port=$${BACALHAU_API_PORT} --node-type requester
```
udsamani, line 71: I think with 1.5 this has now  changed. 

```suggestion
        bacalhau config set "orchestrator.auth.token" "$${NETWORK_AUTH_TOKEN}" && bacalhau serve --orchestrator -c api.port=$${BACALHAU_API_PORT} 
```

jamlo: @udsamani Thanks for pointing this out. 
Btw, is this new change backward compatible ? 
i.e., will the old command still work ? or will it fail with v1.5 ?

wdbaruni: yes they are backward compatible

test-integration/Dockerfile-ClientNode
```diff
@@ -0,0 +1,27 @@
...
+RUN apk add --no-cache ca-certificates
+
+# Copy a root ca into the image
+COPY certificates/generated_assets/bacalhau_test_root_ca.crt /usr/local/share/ca-certificates/bacalhau_test_root_ca.crt
+
+# Update CA certificates
+RUN update-ca-certificates
+
+# Download and execute the Bash script from the given URL
+RUN curl -sSL https://get.bacalhau.org/install.sh | bash
```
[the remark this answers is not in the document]
jamlo: Good point. 
During the testing, we will be changing the flow to utilize `docker compose build` to support building the binary and other needed assets locally as well as on CI (if we take the docker compose route with CI).

Though we can still keep this one (with some docs augmentation) to support new Bacalhau users that want to experiment with the system locally.

test-integration/docker-compose.yml
wdbaruni: actually let me take this back. if a config file had the old config keys and format, we migrate them to the new format when you first run a v1.5.0 bacalhau, but if you call `bacalhau config set old_key=value` using v1.5.0 binary, then we fail with:

```
Error: "node.network.authsecret" is not a valid config key
Hint:  Run 'bacalhau config list' for the complete list of valid config keys
```

test-integration/compute_node_image_setup.sh
```diff
@@ -0,0 +1,38 @@
...
+dockerd-entrypoint.sh &
+
+while [[ ${attempt} -le ${MAX_ATTEMPTS} ]]; do
+    echo "Checking if dockerd is available (attempt ${attempt} of ${MAX_ATTEMPTS})..."
+
+    # Try to communicate with Docker daemon
+    if docker info >/dev/null 2>&1; then
+        echo "dockerd is available! Now Starting Bacalhau as a compute node"
+        bacalhau config set "node.network.authsecret" "${NETWORK_AUTH_TOKEN}"
+        bacalhau serve --node-type=compute --orchestrators="nats://${REQUESTER_NODE_LINK}:4222"
```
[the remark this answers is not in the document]
jamlo: Changes done. I did not use `nats://${NETWORK_AUTH_TOKEN}:${REQUESTER_NODE_LINK}:4222` since the connection string still needs the username to be `nats://${NATS_USER}:${NETWORK_AUTH_TOKEN}@${REQUESTER_NODE_LINK}:4222` .

test-integration/compute_node_image_setup.sh
```diff
@@ -0,0 +1,38 @@
...
+# Start the first docker daemon
+dockerd-entrypoint.sh &
+
+while [[ ${attempt} -le ${MAX_ATTEMPTS} ]]; do
+    echo "Checking if dockerd is available (attempt ${attempt} of ${MAX_ATTEMPTS})..."
+
+    # Try to communicate with Docker daemon
+    if docker info >/dev/null 2>&1; then
+        echo "dockerd is available! Now Starting Bacalhau as a compute node"
+        bacalhau config set "node.network.authsecret" "${NETWORK_AUTH_TOKEN}"
```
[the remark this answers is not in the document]
jamlo: Changes done.

test-integration/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+    environment: *common-env-variables
+    depends_on:
+      bacalhau-minio-node:
+        condition: service_healthy
+    privileged: true
+    command:
+      - /bin/bash
+      - -c
+      - |
+        bacalhau config set "orchestrator.auth.token" "$${NETWORK_AUTH_TOKEN}" && bacalhau serve --orchestrator -c api.port=$${BACALHAU_API_PORT}
```
frrist, line 71: This can also be written as:
```bash
bacalhau -c api.port=$${BACALHAU_API_PORT} -c orchestrator.auth.token=$${NETWORK_AUTH_TOKEN} serve --orchestrator
```
but I don't have strong feelings on changing this; sharing for posterity.

Something that might be nice to add later: Provide a config file to bacalhau directly instead of flags, e.g.
```yaml
  command:
    - /bin/bash
    - -c
    - |
      cat <<EOF > config.yaml
      API:
        Port: $${BACALHAU_API_PORT}
      Orchestrator:
          Enabled: true
          Auth:
              Token: "$${NETWORK_AUTH_TOKEN}"
      EOF
      bacalhau --config=config.yaml serve
```
I find this declarative approach to be easier to grok and change later, but again no strong feelings on this.

test-integration/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+      REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY: /var/lib/registry
+      REGISTRY_HTTP_ADDR: "0.0.0.0:5000"
+      REGISTRY_HTTP_TLS_CERTIFICATE: "/certs/bacalhau-container-img-registry-node.crt"
+      REGISTRY_HTTP_TLS_KEY: "/certs/bacalhau-container-img-registry-node.key"
+    healthcheck:
+      test: [ "CMD-SHELL", "nc -zv localhost 5000" ]
+      interval: 1s
+      timeout: 5s
+      retries: 30
+      start_period: 2s
```
frrist, line 55: I believe we have dedicated health check endpoints for both compute and orchestrator nodes: `/api/v1/healthz` and `/api/v1/livez`, the latter probably being what we'd want here.

test-integration/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  bacalhau-requester-node:
+    image: bacalhau-requester-node-image
+    container_name: bacalhau-requester-node-container
```
frrist, line 59: nit: as we are moving away from the name/term "requester", replacing it with "orchestrator" perhaps we should adopt that name here. @wdbaruni wdyt?

test-integration/certificates/generate_leaf_certs.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+# Set variables
+ROOT_CA_CERT="generated_assets/bacalhau_test_root_ca.crt"
+ROOT_CA_KEY="generated_assets/bacalhau_test_root_ca.key"
```
frrist, line 5: Are these used anywhere? Is the intention to pass them to bacalhau at some point?

test-integration/compute_node_image_setup.sh
```diff
@@ -0,0 +1,38 @@
...
+dockerd-entrypoint.sh &
+
+while [[ ${attempt} -le ${MAX_ATTEMPTS} ]]; do
+    echo "Checking if dockerd is available (attempt ${attempt} of ${MAX_ATTEMPTS})..."
+
+    # Try to communicate with Docker daemon
+    if docker info >/dev/null 2>&1; then
+        echo "dockerd is available! Now Starting Bacalhau as a compute node"
+        bacalhau config set compute.auth.token="${NETWORK_AUTH_TOKEN}"
+        bacalhau serve --compute -c compute.orchestrators="nats://${REQUESTER_NODE_LINK}:4222"
```
frrist, line 21: Similar comments here regarding a declarative approach of defining a config file, but not a required change.

frrist (approved):

frrist:
One more thing that comes to mind: https://github.com/bacalhau-project/bacalhau/tree/main/ops/metrics
We have a docker-compose based metrics suite for bacalhau nodes (it spins up grafana, prometheus, jaeger, and an otel collector. In the future it may be worth unifying these together.

test-integration/docker-compose.yml
jamlo: I am all in for declarative approach. I would even argue to minimize imperative commands as much as possible, and make the config the contract between the user and the system.

Though for this PR, I would keep it as is for now, since I would prefer to change the whole of Docker Compose integration to declarative.

test-integration/certificates/generate_leaf_certs.sh
jamlo: The CA is being injected into every image truststore. Potentially all the certificates used by this bacalhau deployment will be signed by it. Currently we only use it to sign the Docker Registry TLS cert.

The key on our hand, is needed because we need to sign any new certificates with it.

test-integration/compute_node_image_setup.sh
jamlo: Same as this https://github.com/bacalhau-project/bacalhau/pull/4596#discussion_r1792216676

test-integration/docker-compose.yml
jamlo: The quoted code is for the healthcheck of the Docker Registry.
I am assuming you were referring to the checks for orchestrator and compute nodes `nc -zv localhost 1234`

test-integration/docker-compose.yml
jamlo: I will take a note of that, and if the direction is to move from the "requester" naming to "orchestrator", I will change it in the following PRs on this task.

One point to think about, if we are changing the naming, it would be good to have one commit tha changes everything in the project repo at the same time, this way we do not end up with 2 namings for a long time.

test-integration/docker-compose.yml
jamlo: I will take a note to use these healthz and livez endpoints in the nexr PRs.

PR #4604, opened 2024-10-08
remove exec command and job translation

wdbaruni:


cmd/cli/deprecated/exec.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewExecCommand() *cobra.Command {
+	cancelCmd := &cobra.Command{
+		Use:        "exec",
+		Deprecated: "exec was an experimental feature and no longer supported",
+		RunE: func(cmd *cobra.Command, cmdArgs []string) error {
+			return nil
```
frrist, line 12: might return an error here too

wdbaruni: not necessary and follows the same pattern as with other deprecated commands:

```
[bacalhau]→ bacalhau exec                                                                                        
Command "exec" is deprecated, exec was an experimental feature and no longer supported

```

PR #4619, opened 2024-10-13
validate swagger.json

wdbaruni:
update swagger and add new github action to detect if swagger is not up to date with the source


wdbaruni:
closes ENG-272

Issue #4597, opened 2024-10-07
Explore TestContainers to Utilize in Testing Strategy


jamlo:
### Is your feature request related to a problem? Please describe
Since we can utilize Docker Compose in running a complete deployment of Bacalhau (see #4595 ) ,  it would be beneficial if we can incorporate it in our testing strategy.

### Describe the solution you'd like:
[TestContainers](https://testcontainers.com/) is a widely used testing framework within the Container development space. It also has a first class support for [Docker Compose](https://golang.testcontainers.org/features/docker_compose/). 

In this issue, we will explore how can we utilize TestContainers and Docker Compose in our test suite.

### Describe alternatives you've considered
Alternatives would be to build from scratch a golang test suite to test containers and/or docker compose deployments, or to spin up Bacalhau components as separate processes on the host and test them.

### Additional context
1. Need to look into the DevStack and the Bashtubs tests that are currently being utilized.

PR #4626, opened 2024-10-15
Support Using TestContainers for our Integration Tests

jamlo:
This PR adds a test suite utilizing TestContainers. It support spinning up docker compose stacks with all bacalhau components, thus resembling production setup.

This is not a complete TestSuite, but a big initial step towards consolidating out testing and evaluating TestContainers usage.

1- Allow us to run integration tests against a docker compose
   deployment of bacalhau

2- Bacalhau compute/requestor configuration can be injected in the
   container before start. We do not need to rebuild it every time

3- Tests are separated into suites which allows the combination of
   similar tests together that will use the same docker compose
   deployment. Help speed up things

4- Docker compose deployment has a local registry as well as a minio
   server running. This expands out testing ability and scope.

5- Test Suite Works with Github Actions

Linear issue tracker: ENG-263
Github Issue: #4597


jamlo:
Closing For better PR wording.

PR #4627, opened 2024-10-15
Support TestContainers PoC

jamlo:
**This PR adds a test suite utilizing TestContainers. It support spinning up docker compose stacks with all bacalhau components, thus resembling production setup.**

This is not a complete TestSuite, but a big initial step towards consolidating out testing and evaluating TestContainers usage.

1. Allow us to run integration tests against a docker compose
deployment of bacalhau

2. Bacalhau compute/requestor configuration can be injected in the
container before start. We do not need to rebuild it every time

3. Tests are separated into suites which allows the combination of
similar tests together that will use the same docker compose
deployment. Help speed up things

4. Docker compose deployment has a local registry as well as a minio
server running. This expands out testing ability and scope.

5. Test Suite Works with Github Actions

Linear issue tracker: ENG-263
Github Issue: https://github.com/bacalhau-project/bacalhau/issues/4597


.github/workflows/testcontainers-integration-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+  test-build-and-run-testcontainers:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.23'
```
wdbaruni, line 21: Maybe there is a better way to get the go-version we are using in the project, but here is an approach we are using in `swagger-validation.yml`

https://github.com/bacalhau-project/bacalhau/blob/85cbe5b3fa662191c19b2599ee32232a89c6753f/.github/workflows/swagger-validation.yml#L20-L33

Also, I've previously noticed Post Setup Go takes minutes with `setup-go@v4` and I ended up updating to `v5` to solve it. It doesn't seem it is taking long in your case though

test_integration/common_jobs_scenarios_suite_test.go
```diff
@@ -0,0 +1,77 @@
...
+			"--wait=false",
+			"--id-only",
+			"/app/assets/job_specs/hello_world.yml",
+		})
+	s.Require().NoError(err, "Error running hello world from job spec")
+
+	jobID, err := extractJobIDFromShortOutput(result)
+	s.Require().NoErrorf(err, "error extracting Job ID after running it: %q", err)
+
+	completedIn, err := s.waitForJobToComplete(jobID, 30*time.Second)
```
wdbaruni, line 66: will be great to reuse the state resolvers we already have

test_integration/common_jobs_scenarios_suite_test.go
```diff
@@ -0,0 +1,77 @@
...
+	result, err := s.executeCommandInDefaultJumpbox(
+		[]string{
+			"bacalhau",
+			"job",
+			"run",
+			"--wait=false",
+			"--id-only",
+			"/app/assets/job_specs/hello_world.yml",
+		})
+	s.Require().NoError(err, "Error running hello world from job spec")
```
wdbaruni, line 61: Nice!

wdbaruni (approved):
Looks good overall. Nice work! Though taking 4 minutes to run just two scenarios seem too long and I hope we can improve things in upcoming iterations

test_integration/base_suite_test.go
```diff
@@ -0,0 +1,285 @@
...
+		tmpDir,
+		initialImagesNames,
+	)
+	s.Require().NoErrorf(err, "Error rendering Docker Compose file: %q", err)
+
+	defer func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			fmt.Printf("Error removing temporary directory: %v\n", err)
+		}
+	}()
```
frrist, line 104: should this be done before `renderDockerComposeFile`?

test_integration/base_suite_test.go
```diff
@@ -0,0 +1,285 @@
...
+		tc.RemoveVolumes(true),
+	)
+	s.Require().NoErrorf(err, "Error tearing down docker compose stack: %q\n", err)
+	s.Cancel()
+}
+
+func (s *BaseDockerComposeTestSuite) executeCommandInContainer(containerName string, cmd []string) (string, error) {
+	typeAssertedComposeStack, ok := s.ComposeStack.(interface {
+		ServiceContainer(context.Context, string) (*testcontainers.DockerContainer, error)
+	})
```
frrist, line 186: Since this type assertion is used in several places, do we think it's worth defining it as an interface definition, rather than inline?

test_integration/base_suite_test.go
```diff
@@ -0,0 +1,285 @@
...
+		jobState, err := extractJobStateType(jobDescriptionResultJson)
+		if err != nil {
+			return time.Since(startTime), err
+		}
+
+		if jobState == "Completed" {
+			return time.Since(startTime), nil
+		}
+
+		time.Sleep(1000 * time.Millisecond)
```
frrist, line 236: nit: `time.Second`

test_integration/base_suite_test.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	return timeout, fmt.Errorf("job did not finish within allowed time limit %s", timeout.String())
+}
+
+func (s *BaseDockerComposeTestSuite) renderDockerComposeFile(inputFilePath string, tmpDir string, imageReplacements map[string]string) (string, error) {
+	data, err := os.ReadFile(inputFilePath)
+	if err != nil {
+		return "", fmt.Errorf("error reading file: %v", err)
```
frrist, line 245: nit: should we instead use `%w`: https://go.dev/blog/go1.13-errors#wrapping-errors-with-w ?

test_integration/test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+
+	pruneReport, err := cli.ImagesPrune(ctx, filters.NewArgs())
+	if err != nil {
+		return fmt.Errorf("deleting docker images: failed to prune images: %v", err)
+	}
+	fmt.Printf(
+		"Pruned %d images, reclaimed space: %d bytes\n",
+		len(pruneReport.ImagesDeleted),
+		pruneReport.SpaceReclaimed,
+	)
```
frrist, line 151: I think the `testing.M` has a logger, perhaps we should use it?

test_integration/test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+
+	// Merge additional env vars with default ones
+	if additionalSetupEnvVars != nil {
+		for key, value := range additionalSetupEnvVars {
+			defaultEnvVars[key] = value
+		}
+	}
+
+	for key, value := range defaultEnvVars {
+		err := os.Setenv(key, value)
```
frrist, line 290: I believe `testing.M` has a method for setting env vars in the context of the test, should we use that instead?

test_integration/test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	return nil
+}
+
+func extractJobIDFromOutput(jobRunOutput string, s *suite.Suite) string {
+	s.Require().Regexpf(
+		`Job successfully submitted\. Job ID: j-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`,
+		jobRunOutput,
+		"JOb output did not signal a successful job submission: %q",
```
frrist, line 303: nit: `Job ...`

test_integration/test_utils.go
```diff
@@ -0,0 +1,362 @@
...
+	s.Require().Regexp(`^j-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`,
+		jobID,
+		"Extracted Job ID should match the expected format. Job ID found: "+jobID,
+	)
+
+	return jobID
+}
+
+func extractJobIDFromShortOutput(input string) (string, error) {
+	pattern := `j-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`
```
frrist, line 320: nit: define this pattern as a named constant

frrist (approved):
LGTM

.github/workflows/testcontainers-integration-tests.yml
jamlo: @wdbaruni thanks for the pointer. Resolved.

test_integration/common_jobs_scenarios_suite_test.go
jamlo: I thought about it while developing the test suite, my approach was to avoid using any internal Bacalhau packages in this suite to prevent leaking internal implementation details into these tests. Currently, the state resolvers are importing packages from within the main Bacalhau codebase. This concern was one of the primary reasons I introduced test_integration as a separate Go module, establishing a clear boundary. This approach effectively treats the entire Bacalhau deployment as a black box. 

With that said, I will be reviewing the state resolvers and see what I can extrtact from them to help us with the tests.

test_integration/README.md
```diff
@@ -0,0 +1,42 @@
+# Test Suite for [Project Name]
+
+This test suite is designed to perform integration testing for [Project Name] using real docker containers, simulating a production-like environment.
```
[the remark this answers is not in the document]
jamlo: :-) that's what happen when I let claud io write my readme

jamlo:
> Looks good overall. Nice work! Though taking 4 minutes to run just two scenarios seem too long and I hope we can improve things in upcoming iterations

@wdbaruni Thank you.
I agree that the timing needs to be significantly lowered. Currently, these are the steps that occur when the suite runs:

1. The bacalhau project is being compiled (1 to 2 minutes). This is a temporary step that will be totally removed when we choose our CI system, which will have a build step before these tests run, and the binary will be passed down. The integraion tests will not be compiling their own  bacalhau binary. 

2. The container base images build: This step is only done one in the whole test, but still I am thinking about a way to use prebuilt container images, pull them from an image registry, and only inject them with tests assets (bacalhau binary, other specs) when the container starts. Thus reducing the overall time.

3. Each Test suite run is designed to run against a dedicated doeck compose stack, in series. This way if we have 20 tests in a suite, they will be using the same docker compose deployment, and not standing 20 docker compose stacks. This will show more as we add more tests.

test_integration/go.mod
```diff
@@ -0,0 +1,201 @@
+module bacalhau/integration_tests
```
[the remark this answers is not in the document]
jamlo: I see these integration tests as a separate entity that treats a bacalhau deployment as a blackbox. Having it as a separate module is one aspect to define that boundary and prevent any code import from bacalhau; although code import can be still done, but will be easily caught :-)

test_integration/base_suite_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+var globalTestExecutionId string
+
+func TestMain(m *testing.M) {
+	globalTestExecutionId = strings.ToLower(strings.Split(uuid.New().String(), "-")[0])
```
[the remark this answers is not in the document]
jamlo: @frrist

The way the suite is design in 2 layers: 
1. Global layer
2. Suite specific layer

A global layer will have dynamically generated assets that will be used by all the suites, for example, base container images. On the other side, Suite specific layer will have suite common assets. The globalTestExecutionId helps in naming things, that will be used to debug tests especailly when running them locally. There an an identifier as well at the suite level that serves the same purpose.

test_integration/base_suite_test.go
jamlo: `renderDockerComposeFile` needs the common suite tmp dir, thats why I am defering its deletion till the end.

test_integration/test_utils.go
jamlo: Unfortunately `testing.M` does not have a logger (that I am aware of), only `testing.T` have. Thats why I ended up using `Printf `. 
I was thinking of using the `log` package in multiple places in the tests, which would include `TestMain`

test_integration/test_utils.go
jamlo: Unfortunately , `testing.M` does not have a built in mechanism to set env variables, unlike `testing.T`. But since We need those Env variables in the `TestMain` function, there is no running away from using `os.setEnv`.

test_integration/base_suite_test.go
jamlo: Good point, I will be doing that in my current task related to Bashtub switch.

test_integration/base_suite_test.go
jamlo: Good point. I would consider it when complex context is needed to be passed down, but to your point I should have used `%q` for clarity

test_integration/base_suite_test.go
frrist: understood, my point here was: if (in the rare case) `renderDockerComposeFile` returns an error, the `tmpDir` won't be cleaned up.

test_integration/test_utils.go
frrist: +1 to using a dedicated log package

test_integration/base_suite_test.go
jamlo: I got you, good catch.

PR #4631, opened 2024-10-18
Migrate BashTub Tests to TestContainers Based Tests

jamlo:
This PR migrates Bashtub Tests to the new TestContainers Based framework.

1. Since BashTub framework has not been supported for the last 7 years
2. Since having a Golang compiler protect us while running tests, as well as utilizing Golang Testing frameworks is a plus.
3. Since TestContainers have been successfully used for a PoC to Use them in our Tests

The bashtub Tests have been migrated to the new TestContainers Suite, which gives a better assurance since Tests are running in a local docker based cluster.

The only BashTub Tests that have not been migrated are 2 tests related to auth, which will be moved when the Auth project is started. 

The total Run time of the new migrated tests in the new framework is around 2 minutes and a half.


test_integration/1_orchestrator_basic_config_suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	nodeType := nodeInfo["NodeType"].(string)
+	s.Require().Equal("Requester", nodeType)
+
+	nodeLabels := nodeInfo["Labels"].(map[string]interface{})
+	s.Require().Equal("label1Value", nodeLabels["label1"].(string))
+	s.Require().Equal("label2Value", nodeLabels["label2"].(string))
+}
+
+func (s *OrchestratorBasicConfigSuite) TestStartingOrchestratorNodeWithConfigFile() {
```
wdbaruni, line 69: nit: test name can be more representative. You are not really starting a node here :)

test_integration/1_orchestrator_basic_config_suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+func (s *OrchestratorBasicConfigSuite) TestNodesCanBeListed() {
+	nodeListOutput, err := s.executeCommandInDefaultJumpbox([]string{"bacalhau", "node", "list", "--output=json"})
+	s.Require().NoErrorf(err, "Error listing nodes: %q", err)
+
+	marshalledOutput, jsonType, err := s.unmarshalJSONString(nodeListOutput)
+	s.Require().NoErrorf(err, "Error unmarshalling json string: %q. JSON output received: %q", err, nodeListOutput)
+	s.Require().Equalf("array", jsonType, "incorrect json output type")
+
+	marshalledOutputArray := marshalledOutput.([]interface{})
+	s.Require().Equalf(1, len(marshalledOutputArray), "Node count should be 1")
```
wdbaruni, line 57: can we have a helper method to unmarshall the output into a go type? In this case the output I believe is `[]models.NodeState`

test_integration/1_orchestrator_basic_config_suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	nodeInfo := marshalledOutputArray[0].(map[string]interface{})["Info"].(map[string]interface{})
+
+	nodeType := nodeInfo["NodeType"].(string)
+	s.Require().Equal("Requester", nodeType)
+
+	nodeLabels := nodeInfo["Labels"].(map[string]interface{})
+	s.Require().Equal("label1Value", nodeLabels["label1"].(string))
+	s.Require().Equal("label2Value", nodeLabels["label2"].(string))
```
wdbaruni, line 66: should we have these as constants somewhere?

test_integration/1_orchestrator_basic_config_suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+func (s *OrchestratorBasicConfigSuite) TestStartingOrchestratorNodeWithConfigFile() {
+	agentConfigOutput, err := s.executeCommandInDefaultJumpbox([]string{"curl", "http://bacalhau-orchestrator-node:1234/api/v1/agent/config"})
+	s.Require().NoErrorf(err, "Error getting orchestrator agent config: %q", err)
+
+	marshalledOutput, jsonType, err := s.unmarshalJSONString(agentConfigOutput)
+	s.Require().NoErrorf(err, "Error unmarshalling json string: %q", err)
+	s.Require().Equalf("object", jsonType, "incorrect json output type")
+
+	orchestratorEnabled := marshalledOutput.(map[string]interface{})["Orchestrator"].(map[string]interface{})["Enabled"].(bool)
+	s.Require().Truef(orchestratorEnabled, "Expected orchestrator to be enabled, got: %t", orchestratorEnabled)
```
wdbaruni, line 78: similar to my comment above, this can be unmarshalled to `config.Bacalhau`

test_integration/2_orchestrator_config_override_suite_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	s.BaseDockerComposeTestSuite.SetupSuite(rawDockerComposeFilePath, extraRenderingData)
+
+	_, err := s.executeCommandInDefaultJumpbox(
+		[]string{
+			"mkdir",
+			"-p",
+			fmt.Sprintf("/app/%s", s.SuiteRunIdentifier),
+		})
+	s.Require().NoErrorf(err, "Error creating a suite tmp folder in the jumpbox node: %q", err)
```
wdbaruni, line 41: the same logic needs to be repeated in all test suites. Does it make sense to introduce a SuiteBuilder type or something similar with sane defaults and allows overriding certain behaviour when needed?

test_integration/2_orchestrator_config_override_suite_test.go
```diff
@@ -0,0 +1,77 @@
...
+	s.Require().Equalf("array", jsonType, "incorrect json output type")
+
+	marshalledOutputArray := marshalledOutput.([]interface{})
+	s.Require().Equalf(1, len(marshalledOutputArray), "There should be only one node, but actual count is: %d", len(marshalledOutputArray))
+
+	nodeMembership := marshalledOutputArray[0].(map[string]interface{})["Membership"]
+	s.Require().Equalf("APPROVED", nodeMembership.(string), "Expected node to be approved, but actual is: %s", nodeMembership)
+
+	nodeConnection := marshalledOutputArray[0].(map[string]interface{})["Connection"]
+	s.Require().Equalf("CONNECTED", nodeConnection.(string), "Expected node to be connectted, but actual is: %s", nodeConnection)
```
wdbaruni, line 64: unless you are intentionally don't want to depend on bacalhau's code, you might want to use `models` package's constants

wdbaruni (commented):
This is looking great. My main feedback is around how can we make writing the tests less verbose, including setting up the tests and verifying the outputs. It is a bit difficult to understand what the tests are actually doing

Second is adding more comments would be super helpful to understand the magic that is happening here :)

test_integration/1_orchestrator_basic_config_suite_test.go
jamlo: Good catch. Fixed.

test_integration/1_orchestrator_basic_config_suite_test.go
jamlo: I am looking for a way to nicely access the returned response, in a "python style" non verbose manner, and I agree with you it is currently done in a very verbose way. 

On the other hand, I am intentionally shying away from importing any code from the Bacalhau main module, including the models. My thinking is that these integration tests should only interface with the system using the CLI or SDk, or API. I feel if we open that door, a lot of the main module code will leak into the tests, making them very toughly coupled, and not focused on testing the outer contract.

test_integration/1_orchestrator_basic_config_suite_test.go
jamlo: I though about it, but decided to make it in a way that each test "suite" has its out test files, making tests not interconnected as much as possible.

And since each test will have its own test files with some test data in them, I think having the results explicitly stated in the tests will make it easier to read, and not affected by a common constant that is used by many tests.

test_integration/1_orchestrator_basic_config_suite_test.go
jamlo: My thoughts are similar to the comment above regarding importing code from the main Bacalhau module. 
Please let me know if you anticipate an issue in the future with this approach.

test_integration/2_orchestrator_config_override_suite_test.go
jamlo: Good catch. The copy paste laziness caught up with me last week.
Code is fixed, and I actually removed some of it that is not needed any more.

test_integration/2_orchestrator_config_override_suite_test.go
jamlo: Exactly. My thoughts are similar to the comment above regarding importing code from the main Bacalhau module.
Please let me know if you anticipate an issue in the future with this approach.

frrist (approved):

.github/workflows/testcontainers-integration-tests.yml
```diff
@@ -36,3 +36,4 @@ jobs:
           echo "hello test" > webui/build/index.html
           cd test_integration
           go test -v -count=1 ./...
+          go test -p 1 -v ./...
```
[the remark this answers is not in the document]
jamlo: @frrist thanks for the catch, I was testing a local version few minutes ago and forgot to remove it.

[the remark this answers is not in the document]
jamlo: Fixed

test_integration/1_orchestrator_basic_config_suite_test.go
jamlo: @frrist Correct, each test suite has an set of config files for it. They can be found in the SetupSuite function.

test_integration/base_suite.go
```diff
@@ -1,110 +1,70 @@
...
+	// Merge Rendering Data
+	if renderingData != nil {
+		for key, value := range renderingData {
+			defaultRenderingData[key] = value
 		}
 	}
 
 	// Render Docker Compose File
-	tmpDir, err := os.MkdirTemp("", "bacalhau-test-"+s.suiteRunIdentifier)
+	tmpDir, err := os.MkdirTemp("", "bacalhau-test-"+s.SuiteRunIdentifier)
```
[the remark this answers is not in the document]
jamlo: Good point I will make that change in the next task for this Suite.

test_integration/1_orchestrator_basic_config_suite_test.go
wdbaruni: > I am intentionally shying away from importing any code from the Bacalhau main module

I agree with this intention, but unfortunately our current code is not structured to enable this just yet. I mean our API and SDK depend directly on the `models` package inside the main pkg. Besides, our go SDK is actually part of the main module where the server and all backend code lives. Meaning you can't use the SDK without importing the whole thing, which is totally wrong. 

We need to fix this, but we don't have to fix it today. Most likely we will need to have a shared module for model definitions that are shared by the SDK, API, CLI and backend.  So my recommendation for the integration tests is to use all types the SDK and API models are accessing regardless where they live and those will be migrated eventually outside of the core module

wdbaruni (approved):
Approved the PR. The changes to reduce the verbosity, reuse existing models and extract helpers can come in a follow up PR, but should be a quick follow up to set the golden standard for writing more of these tests :)

Issue #4605, opened 2024-10-08
lint is failing but build is still passing


wdbaruni:
This is an example where lint state is successful, but checking lint logs show there are multiple violations https://buildkite.com/expanso/bacalhau-golang/builds/1118#01926c15-3806-4d08-92dd-743385d1596b

PR #4642, opened 2024-10-21
Prevent False Positives in CI

jamlo:
Since all of our CI scripts are run with BASH, we need to make sure that any unexpected non zero exit code instruction in the script should fail the CI run. 

We need to add the snippet below to all our CI scripts:

```
set -eu -o pipefail
```

Note: After this change the lint tests are red, which is good, since they were failing before but the test was passing. There is an already created issue for that: #4605


.github/workflows/lint.yml
```diff
@@ -14,16 +14,34 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - run: echo "GO_VER=$(cat .tool-versions | grep 'golang ' | cut -d' ' -f2)" >> $GITHUB_ENV
-      - run: echo "NODE_VER=$(cat .tool-versions | grep nodejs | cut -d' ' -f2)" >> $GITHUB_ENV
+      - name: Set Go Version
+        shell: bash
+        run: |
+          set -euo pipefail
```
wdbaruni, line 20: it feels weird to me that we have to do this for every single step. Are we really sure it is required? Github actions do have a `continue-on-error` options which is false by default, which implies this shouldn't be necessery.

My doubt this might be an issue with buildkite or how we are integrating with buildkite, but I doubt it is the case with GA

If you decide it is needed, then maybe you can define the `env` or `job` level, but not at the individual `step` level

wdbaruni (changes requested):

.github/workflows/lint.yml
jamlo: @wdbaruni I know from the aesthetics aspects of it, adding `set -euo pipefail` may look not pleasant and repetitive, but I intentionally and explicitly added them to all the steps based on my personal experience how GA propagates these flags to each CI shell. There were multiple bugs with how they propagate those to the shell of the step, inclusing `continue-on-error` in the past, although they have mentioned that they were fixed.

If you think we can live can with this explicit declaration, I would really prefer to leave it since false positives have bitten me so much in the past, with GA, that it was worth adding them to each step, and 100% insure CI tests are red if the internal tests are failing.

PR #4646, opened 2024-10-22
fix: default publishers not working

wdbaruni:
We are still having issues with using server side default publishers where the client will fail any job submission where outputs are defined, but no publisher is explicitly selected. The root cause is that we are still calling `job.Validate()` in the client instead of just `job.ValidateSubmission()`. `Validate()` is a more through validation and should only be called after all transformations and defaults applied.

This bug has slipped through because we are using some builders in the client side to translate from cli flags to job and task models instead of directly instantiating the models. The builders were calling Validate which wasn't very obvious.

In addition to fixing the bug, this PR also removes all of these unecessery builders that are mainly used in few tests in favour of using the models directly for better clarity. 

### Current Behaviour
```
# set default publisher
→ bacalhau config set  jobdefaults.batch.task.publisher.type=local
Writing config to /Users/walid/.bacalhau/config.yaml

# run a job with outputs but no publisher defined
→ bacalhau docker run --output outputs:/outputs ubuntu -- /bin/sh -c "echo hello > outputs/text"

Error: building job spec: failed to create job: publisher must be set if result paths are set
```

### Expected Behaviour
```
→ bacalhau docker run --output outputs:/outputs ubuntu -- /bin/sh -c "echo hello > outputs/text"
Job successfully submitted. Job ID: j-aba087a0-0516-4b58-aff0-9824af583723

 TIME          EXEC. ID    TOPIC            EVENT
 14:55:19.550              Submission       Job submitted
 14:55:19.592  e-13e6591c  Scheduling       Requested execution on n-a1af1cf4
 14:55:19.761  e-13e6591c  Execution        Running
 14:55:23.323  e-13e6591c  Execution        Completed successfully

	bacalhau job describe j-aba087a0-0516-4b58-aff0-9824af583723

	bacalhau job executions j-aba087a0-0516-4b58-aff0-9824af583723

To download the results, execute:
	bacalhau job get j-aba087a0-0516-4b58-aff0-9824af583723

```

### Testing Done:
- Fixed and added more coverage to `DefaultPublisherSuite`


jamlo (approved):
LGTM.

PR #4647, opened 2024-10-22
Fix Go lint issues

jamlo:
Linear Issue: https://linear.app/expanso/issue/ENG-267/lint-is-failing-but-build-is-still-passing
Github Issue: #4605


cmd/cli/job/executions.go
```diff
@@ -148,10 +148,10 @@ func (o *ExecutionOptions) run(cmd *cobra.Command, args []string, api client.API
...
 		JobID: jobID,
 		BaseListRequest: apimodels.BaseListRequest{
-			Limit:     o.Limit,
-			NextToken: o.NextToken,
-			OrderBy:   o.OrderBy,
-			Reverse:   o.Reverse,
+			Limit:     o.ListOptions.Limit,
+			NextToken: o.ListOptions.NextToken,
+			OrderBy:   o.ListOptions.OrderBy,
+			Reverse:   o.ListOptions.Reverse,
```
wdbaruni, line 154: hmm .. how do we feel about enforcing this rule? I am okay with disabling the rule in .golangci.yml if you think we are better off without it

pkg/compute/store/boltdb/utils.go
```diff
@@ -6,14 +6,19 @@ import (
...
+
 // strToBytes converts a string to a byte slice
 func strToBytes(s string) []byte {
 	return []byte(s)
 }
 
 // uint64ToBytes converts an uint64 to a byte slice
 func uint64ToBytes(i uint64) []byte {
-	buf := make([]byte, 8) //nolint:gomnd
+	buf := make([]byte, uint64ByteSize) //nolint:gomnd
```
wdbaruni, line 21: I think `gomnd` was renamed to `mnd` in later versions and that is why //nolint is not working here

pkg/docker/errors.go
```diff
@@ -48,62 +48,62 @@ func NewDockerError(err error) (bacErr bacerrors.Error) {
...
-		return bacerrors.New(err.Error()).
+		return bacerrors.New("%s", err.Error()).
 			WithCode(Forbidden).
 			WithHTTPStatusCode(http.StatusForbidden).
 			WithHint(fmt.Sprintf("You don't have permission to perform this action. "+
 				"Supply the node with valid Docker login credentials using the %s and %s environment variables",
 				config_legacy.DockerUsernameEnvVar, config_legacy.DockerPasswordEnvVar))
 	case errdefs.IsDataLoss(err):
-		return bacerrors.New(err.Error()).
+		return bacerrors.New("%s", err.Error()).
```
wdbaruni, line 68: I hate we have to do this. Feel free to skip or disable rules that don't ad much value

pkg/node/node.go
```diff
@@ -278,12 +283,22 @@ func createAPIServer(cfg NodeConfig, userKey *baccrypto.UserKey) (*publicapi.Ser
...
+	givenPortNumber := cfg.BacalhauConfig.API.Port
+	if givenPortNumber < minPortNumber {
+		return nil, fmt.Errorf("invalid negative port number: %d", cfg.BacalhauConfig.API.Port)
+	}
+	if givenPortNumber > maxPortNumber {
+		return nil, fmt.Errorf("port number %d exceeds maximum allowed value (65535)", cfg.BacalhauConfig.API.Port)
+	}
+
+	safePortNumber := uint16(givenPortNumber)
+
```
wdbaruni, line 295: nit: might make sense to include these validations in `Validate()` method of `config.Bacalhau` or of `node.NodeConfig`

pkg/publicapi/client/v2/client.go
```diff
@@ -400,11 +408,6 @@ func (c *httpClient) interceptError(ctx context.Context, err error, resp *http.R
 				WithCode(bacerrors.TimeOutError).
 				WithRetryable()
 		}
-		if netErr.Temporary() {
-			return bacerrors.New("temporary network error").
-				WithCode(bacerrors.NetworkFailure).
-				WithRetryable()
-		}
```
wdbaruni, line 407: why was this removed?

pkg/storage/s3/storage.go
```diff
@@ -83,8 +84,21 @@ func (s *StorageProvider) GetVolumeSize(ctx context.Context, volume models.Input
...
+		}
+
+		// Check for overflow
+		// MaxUint64 - size = remaining space before overflow
+		//nolint:gosec // G115: negative values already checked
+		if object.size > 0 && uint64(object.size) > math.MaxUint64-size {
+			return 0, fmt.Errorf("total size exceeds uint64 maximum")
+		}
+
+		//nolint:gosec // G115: Already checked above
```
wdbaruni, line 101: nit: it should be safe to skip these as we are basically testing S3 SDK in this case

wdbaruni (approved):
Looks great. I am really sorry you had to go through this change. Just minor non-blocking comments

pkg/publicapi/client/v2/client.go
jamlo: @wdbaruni Temporary Error is deprecated and should not be used anymore, especially from a client perespective.

Golang Code:
```
	// Deprecated: Temporary errors are not well-defined.
	// Most "temporary" errors are timeouts, and the few exceptions are surprising.
	// Do not use this method.
	Temporary() bool 
```
Source: https://go-review.googlesource.com/c/go/+/340261/3/src/net/net.go#404

cmd/cli/job/executions.go
jamlo: @wdbaruni I tend to lean toward verbosity here, and in general, which is more of a personal choice.
But I'll add the exception to the definitions file to balance things out.

pkg/docker/errors.go
jamlo: Since `bacerrors.New` is using `fmt.Sprintf` underneath the hood,  it think it is better to use the formatted string approach since it more idiomatic GO, where it nicely keep the format of the original error message, and nicely deal with nil errors. I did the change to use `"%s", err` , instead of `"%s", err.Error()` to remove repetitive string conversion and use the builtin one in SprintF

cmd/cli/job/executions.go
jamlo: Change reverted

pkg/compute/store/boltdb/utils.go
jamlo: Replaced gomnd with mnd where appropriate

pkg/storage/s3/storage.go
jamlo: I would like to explicitly leave this if possible due to how we are dealing with a lot of `unint` conversation which may not be needed. 
I added this issue to not forget about it https://github.com/bacalhau-project/bacalhau/issues/4652

PR #4655, opened 2024-10-23
fix test flakiness by useing busybox instead of ubuntu images 

wdbaruni:
Flakiness seems related to buildkite taking too long to pull images uses in tests, where we mainly use ubuntu. 

Udit mentioned that there are certain times of the day where buildkite tests were failing more than others, and this might explain why. During peek hours, buildkite network bandwidth might become limited and takes longer to fetch resources

This PR  changes from ubuntu to busybox for most tests, but it is werid that buildkite is pulling images multiple times instead of reusing them

### DIFF
--- a/cmd/cli/docker/docker_run_test.go
+++ b/cmd/cli/docker/docker_run_test.go
@@ -354,7 +354,8 @@ func (s *DockerRunSuite) TestTruncateReturn() {
 			ctx := context.Background()
 			_, out, err := s.ExecuteTestCobraCommand(
 				"docker", "run",
-				"busybox:latest", "--", "perl", "-e", fmt.Sprintf(`print "=" x %d`, tc.inputLength),
+				"busybox:latest", "--", "sh", "-c",
+				fmt.Sprintf(`yes "=" | tr -d "\n" | head -c %d`, tc.inputLength),
 			)
 			s.Require().NoError(err, "Error submitting job. Name: %s. Expected Length: %s", name, tc.expectedLength)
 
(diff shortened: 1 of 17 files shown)


pkg/test/executor/docker_entrypoint_test.go
```diff
@@ -301,7 +301,7 @@ type dockerfilePermutation struct {
 func createDockerfile(d dockerfilePermutation) string {
 	ep := "\nENTRYPOINT [\"/bin/echo\"]"
 	cmd := "\nCMD [\"echo\", \"This is from CMD\"]"
-	baseDockerFile := "FROM alpine:latest"
+	baseDockerFile := "FROM busybox:latest"
```
jamlo, line 304: @wdbaruni looks like this test is tightly coupled with Ubuntu distribution since they are failing in BuildKite

cmd/cli/docker/docker_run_test.go
```diff
@@ -354,7 +354,7 @@ func (s *DockerRunSuite) TestTruncateReturn() {
 			ctx := context.Background()
 			_, out, err := s.ExecuteTestCobraCommand(
 				"docker", "run",
-				"ubuntu", "--", "perl", "-e", fmt.Sprintf(`print "=" x %d`, tc.inputLength),
+				"busybox:latest", "--", "perl", "-e", fmt.Sprintf(`print "=" x %d`, tc.inputLength),
```
jamlo, line 357: This test seems to be also expecting perl to be present, which is not on busybox

jamlo (commented):
It seems few tests are tightly coupled with Ubuntu.

cmd/cli/docker/docker_run_test.go
wdbaruni: Sorry I didn't complete the migration and should've marked the PR as draft. Fixed now

PR #4665, opened 2024-10-25
Improve Go-linting in CI

jamlo:
Currently Go linting is tightly coupled with pre-commits, which make things very complex and error prone. 

This commit removes Go linting from any pre-commit logic, and only have it as a first class citizen in Github Actions.

More tasks were created to simplify all the tooling about pre-commits and linting, but will tackle it one step at a time.


jamlo:
Closing in favor of https://github.com/bacalhau-project/bacalhau/pull/4672

PR #4671, opened 2024-10-28
feat: adds `agent config` command that returns the agent configuration

### LINKED ISSUE #4113: API to retrive node config
Not obvious as to why?

repro:
```
bacalhau serve --orchestrators 192.168.1.30 & 

bacalhau config list
```

Notice node.network.orchestrators is not set.

frrist:
- closes #4113

Example:
```shell
$ bacalhau agent config
```
```yaml
API:
    Host: 0.0.0.0
    Port: 1234
    Auth:
        Methods:
            ClientKey:
                Type: challenge
NameProvider: puuid
DataDir: /home/frrist/.bacalhau
Orchestrator:
    Enabled: true
    Host: 0.0.0.0
    Port: 4222
    NodeManager:
        DisconnectTimeout: 1m0s
    Scheduler:
        WorkerCount: 32
        QueueBackoff: 1m0s
        HousekeepingInterval: 30s
        HousekeepingTimeout: 2m0s
    EvaluationBroker:
        VisibilityTimeout: 1m0s
        MaxRetryCount: 10
Compute:
    Orchestrators:
        - nats://127.0.0.1:4222
    Heartbeat:
        InfoUpdateInterval: 1m0s
        ResourceUpdateInterval: 30s
        Interval: 15s
    AllocatedCapacity:
        CPU: 70%
        Memory: 70%
        Disk: 70%
        GPU: 100%
    AllowListedLocalPaths: []
WebUI:
    Listen: 0.0.0.0:8438
InputSources:
    ReadTimeout: 5m0s
    MaxRetryCount: 3
Publishers:
    Types:
        Local:
            Address: 127.0.0.1
            Port: 6001
Engines:
    Types:
        Docker:
            ManifestCache:
                Size: 1000
                TTL: 1h0m0s
                Refresh: 1h0m0s
JobDefaults:
    Batch:
        Task:
            Resources:
                CPU: 500m
                Memory: 1Gb
    Ops:
        Task:
            Resources:
                CPU: 500m
                Memory: 1Gb
    Daemon:
        Task:
            Resources:
                CPU: 500m
                Memory: 1Gb
    Service:
        Task:
            Resources:
                CPU: 500m
                Memory: 1Gb
JobAdmissionControl:
    Locality: anywhere
Logging:
    Level: info
    Mode: default
    LogDebugInfoInterval: 30s
UpdateConfig:
    Interval: 24h0m0s
```

## Summary by CodeRabbit

- **New Features**
	- Introduced a command to retrieve the agent's configuration, allowing users to access configuration details easily.
	- Added support for customizable output formats (JSON and YAML) for the configuration command.
  
- **Bug Fixes**
	- Enhanced error handling and sensitive data redaction in the configuration retrieval process.

- **Tests**
	- Implemented tests to ensure the correct functionality and output of the new configuration command, including validation of sensitive data redaction.


jamlo:
@frrist 
Nice. I believe this command will be very useful.

I have few remarks/concerns:
1. It looks like this PR has broken the `/api/v1/agent/config` endpoint.  The response is now the binary encoding of the config `{"Config":"QVBJOgogICAgSG9zdDogMC4wLjAuMAogICAgUG9yd.....` 
2. Are there any potential sensitive data that we might need to mask from that config? Maybe any connection strings with passwords in them, or similar items ?

jamlo (commented):
Nice. I believe this command will be very useful.

I have few remarks/concerns:

1. It looks like this PR has broken the `/api/v1/agent/config` endpoint. The response is now the binary encoding of the config `{"Config":"QVBJOgogICAgSG9zdDogMC4wLjAuMAogICAgUG9yd.....` . I know about that since some of the TestContainers tests are failing
2. Are there any potential sensitive data that we might need to mask from that config? Maybe any connection strings with passwords in them, or similar items ?
3. Should we add few tests to protect this feature?

frrist:
> It looks like this PR has broken the /api/v1/agent/config endpoint. The response is now the binary encoding of the config {"Config":"QVBJOgogICAgSG9zdDogMC4wLjAuMAogICAgUG9yd..... . I know about that since some of the TestContainers tests are failing

Yes, this is to be expected given the body of the request is now `[]byte`. My aim here is to simplify the response type, as it stand right now, the body contains a yaml string reflecting the servers current configuration. Clients need only print the body, rather than try to interpret the body. Open to alternatives here. I suppose one alternative may be to return an actual string, instead of a binary encoded one.

> Are there any potential sensitive data that we might need to mask from that config? Maybe any connection strings with passwords in them, or similar items ?

Yes, several fields of the config may contain tokens, for example the Compute config [here](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/types/compute.go#L18). My personal opinion is that config files shouldn't contain sensitive data - it should live somewhere else. That point aside, I am not sure excluding them from the (returned) config is the right choice, perhaps it may be simpler to require auth to query this endpoint? Again, open to alternatives. 

> Should we add few tests to protect this feature

100%. Can you point me to the appropriate package to writing such tests?

wdbaruni:
The API should just return json representation of the config exactly the same how it works today. No binary and no need for nesting inside a `Config` field unless we think more fields can be returned, which I doubt it is the case. Parsing into yaml representation should only happen in the CLI. It should just work with `curl localhost:1234/api/v1/agent/config`

I agree with Jamil about having to mask sensitive data. The only fields we have today are `Compute.Auth.Token` and `Orchestrator.Auth.Token`. If running in debug mode, then we can return the values as they are. Otherwise we should mask them. The maintainable approach would be to have a dedicated field type for sensitive data instead of treating them as plain strings, and let the masking take place in one location when mashalling those fields. But that is a lot of work for now and the quick solution would be to mask those fields in the API handler or when mashalling the root config type

jamlo:
> Yes, this is to be expected given the body of the request is now `[]byte`. My aim here is to simplify the response type, as it stand right now, the body contains a yaml string reflecting the servers current configuration. Clients need only print the body, rather than try to interpret the body. Open to alternatives here. I suppose one alternative may be to return an actual string, instead of a binary encoded one.
> 

I think it better to return the json representation of the YAML config, since it is much more standard API behavior. Although Json is a subset of YAML, and not the other way around, I believe the current structure of the config should convert to JSON with now issues.

> Yes, several fields of the config may contain tokens, for example the Compute config [here](https://github.com/bacalhau-project/bacalhau/blob/main/pkg/config/types/compute.go#L18). My personal opinion is that config files shouldn't contain sensitive data - it should live somewhere else. That point aside, I am not sure excluding them from the (returned) config is the right choice, perhaps it may be simpler to require auth to query this endpoint? Again, open to alternatives.

I think requiring authentication is for sure needed for this endpoint, regardless if it contains sensitive data (exposing configuration publicly is a big security attack vector.) 
Also, I think masking certain types as "redacted" will also minimize that attack vector, and if we set the boundary that no credentials can be even queried through API, a user should be on the node itself to see these creds.

> 100%. Can you point me to the appropriate package to writing such tests?

I was thinking of a basic unit test that lives beside the `cmd/cli/agent/config.go` file for now. And I will fix the TestContainers tests when the PR is merged.

frrist:
> The API should just return json representation of the config exactly the same how it works today
>I think it better to return the json representation of the YAML config

Sounds good, will make the change.

> the quick solution would be to mask those fields in the API handler or when mashalling the root config type
> I think masking certain types as "redacted" will also minimize that attack vector

Also sounds good, will change.

> I was thinking of a basic unit test that lives beside the cmd/cli/agent/config.go file for now. And I will fix the TestContainers tests when the PR is merged.

Sounds good to me.

frrist:
@wdbaruni @jamlo how do we want the client/CLI to display the config to the users? Should it marshal the JSON payload to YAML?

jamlo:
> @wdbaruni @jamlo how do we want the client/CLI to display the config to the users? Should it marshal the JSON payload to YAML?

I've noticed that some commands already have this flag:

```
--output format            The output format for the command (one of ["table" "csv" "json" "yaml"]) (default table)
```

Maybe in this situation `csv` and does not apply. But we can give the user the option to choose the format, and default to any format you see fit (I prefer YAML since it is easier to read).

frrist:
@jamlo in fe291ae106af3e6a14e1c307b77048f475b0aa50 I have (attempted to) fix the integration tests. I've been unable to verify this locally as the (new) integration tests fail on my Ubuntu host with the following error message:
```
2024/10/30 14:02:45 ====> Starting the whole test flow:  a0c4d041
2024/10/30 14:02:45 Error compiling the bacalhau binary:  compiling bacalhau: failed to start container: create container: Error response from daemon: client version 1.45 is too new. Maximum supported API version is 1.43
```

I suspect I either need to downgrade my client version or upgrade the version required by the testing framework. wdyt?

jamlo:
@frrist For the docker API version, you can configure that in the Tests when you run it locally [here](https://github.com/bacalhau-project/bacalhau/blob/9b2daf7174dcda4f75ea70533584eff7de4bc23b/test_integration/utils/test_utils.go#L147) . I've set the APi version since Github Actions does not have the latest version, so I pin it. 

Btw, I checked the TestContainers tests, and they are running fine on CI, except for one Test that is failing due to how I am calling curl (I should call it to only output the response body). 

Since these tests are not a blocker to merge yet, I will fix it when you merge the PR.

jamlo (approved):

Issue #4507, opened 2024-09-24
CLI Pagination: Display Full Bacalhau Command for Next Results


frrist:
Commands like `bacalhau node list` and `bacalhau job list` return a `next-token` to support pagination. For example:
```shell
$ bacalhau job list --wide --limit=1
 CREATED              ID                                      JOB     TYPE   STATE     
 2024-09-24 16:12:44  j-b7533ad5-967d-493c-967a-d93b7ccd2356  docker  batch  Completed 

To fetch more records use `--next-token Ok46MTox`
```

Instead of returning the token, the command should return the bacalhau command that can be run to fetch the next result, so:
```shell
$ bacalhau job list --wide --limit=1
 CREATED              ID                                      JOB     TYPE   STATE     
 2024-09-24 16:12:44  j-b7533ad5-967d-493c-967a-d93b7ccd2356  docker  batch  Completed 

To fetch more records use: 
  bacalhau job list --wide --limit=1 --next-token Ok46MTox
```

## Acceptance Criteria:
- implement this for all commands that support pagination: `node list`, `job list`, and any others

PR #4691, opened 2024-11-05
Display full command for next results in `job list` pagination

markkuhn:
**Issue**: https://github.com/bacalhau-project/bacalhau/issues/4507

## Changes

This PR changes the message shown to users when `bacalhau job list` is run and a next token is returned. It now includes the full command to run.

Previous response:
```
❯ bacalhau job list --limit=1
 CREATED   ID          JOB     TYPE   STATE
 13:17:51  66986ccd    docker  batch  Completed

To fetch more records use `--next-token Ok46MTox`
```

New response:
```bash
❯ bacalhau job list --limit=1
 CREATED   ID          JOB     TYPE   STATE
 13:17:51  66986ccd    docker  batch  Completed

To fetch more records use:
	bacalhau job list --limit 1 --next-token Ok46MTox
```

## Testing

- `make test-all`
- Ran `make build-dev` with changes and tested manually
- Added basic job list tests

## Summary by CodeRabbit

- **New Features**
	- Improved user feedback with a clearer, multi-line message for the `--next-token` flag in the command-line interface.
  
- **Tests**
	- Introduced an integration test suite for job listing functionality, ensuring reliability and accuracy of job listing commands.
	- Added tests for job listing and limiting job results, enhancing overall test coverage.


cmd/cli/job/list_test.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		_, out, err := s.ExecuteTestCobraCommand(args...)
+		require.NoError(s.T(), err, "Error submitting job")
+		jobID := system.FindJobIDInTestOutput(out)
+		jobIDs[i] = jobID
+	}
+
+	// Run 'bacalhau job list --limit 2 --output csv'
+	_, listOut, err := s.ExecuteTestCobraCommand("job", "list", "--wide", "--limit", "2", "--output", "csv")
+	require.NoError(s.T(), err, "Error running job list")
```
jamlo, line 75: Any particular reason you chose to use CSV output here instead of JSON? 

JSON will be easier to read, and you do not need to worry about CSV headers, etc

cmd/cli/job/list_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (s *ListSuite) TestJobList() {
+	// Submit a job
+	args := []string{
+		"docker", "run",
+		"--wait",
+		"alpine", "echo", "hello world",
+	}
+	_, out, err := s.ExecuteTestCobraCommand(args...)
+	require.NoError(s.T(), err, "Error submitting job")
```
jamlo, line 46: Minor comment for all assertions:
Since you already created the Test suite, you can use `s.Require().NoError...` , `s.Require().Equal...` .

pkg/system/utils.go
```diff
@@ -94,3 +94,34 @@ func IsDebugMode() bool {
...
+	// Extract the CSV lines
+	csvLines := lines[headerIndex:endIndex]
+	csvData = strings.Join(csvLines, "\n")
+
+	// Extract the remaining output
+	remainingLines := lines[endIndex:]
+	remainingOutput = strings.Join(remainingLines, "\n")
+
+	return csvData, remainingOutput, nil
+}
```
jamlo, line 127: Same comment about choosing CSV over JSON.
If we can use JSON, it is easier to read and less surprises.

jamlo (commented):
Nice .
I added few comments to the code.

Also, for the issue you referenced, it also mentions `bacalhau node list` , did you intentionally skip that ?

markkuhn:
> Nice . I added few comments to the code.
> 
> Also, for the issue you referenced, it mentions `bacalhau node list` , did you intentionally skip that ?

Skipped it for now as it seems `bacalhau node list` doesn't paginate yet so it will need more work.

cmd/cli/job/list_test.go
markkuhn: No particular reason. Switched to JSON

pkg/system/utils.go
```diff
@@ -94,3 +94,18 @@ func IsDebugMode() bool {
 	// TODO: #4535 we need to add a flag to the CLI to enable debug mode
 	return os.Getenv("DEBUG") == "true" || zerolog.GlobalLevel() <= zerolog.DebugLevel
 }
+
+// ExtractJSONOutput extracts JSON data from the output
+func ExtractJSONOutput(output string) (jsonData string, remainingOutput string, err error) {
+	start := strings.Index(output, "[")
```
jamlo, line 100: I can see you are looking for the first index of the `[` , and the last index of `]`. 
Few comments on that:
1. I know that the "--json" output for this command unfortunately adds extra lines to the end that are not even JSON, so I know the pain and reason to add this logic.
2. Since your function is generic, what if the JSON output is not a JSON array, bit it is a json object ? so it will not start with a `[` .
3. You are adding this extract method to `pkg/system/utils.go`, but only using it in the tests. The name `ExtractJSONOutput` in this common package gives the illusion it is generic, but it not.

My recommendation is to move this helper method to the test file itself since it only applies for this  specific test case. 
This way you do not need to change the logic to support so many edge cases for parsing JSON, and just have the logic to support this test.

Note for later: This is a reminder for us to fix "--json" output for all commands and only produce JSON with no extra data.

jamlo (approved):

PR #4699, opened 2024-11-06
fix swagger generation

wdbaruni:
Swagger validation step was failing but the Github Workflow was still passing. This is because we were missing `set -euo pipefail` in the generate_swagger.sh script.

The PR fixes the validation step, and updates swagger schemas

## Release Notes

- **New Features**
  - Introduced a new property `DisableAnalytics` for enhanced user control over data sharing.
  - Added support for specifying communication protocols with the new `SupportedProtocols` property.
  - Expanded execution state management with new states: `ExecutionStateRunning` and `ExecutionStatePublishing`.
  - Introduced new TLS configuration types for improved security: `ComputeTLS` and `OrchestratorTLS`.

- **Removed Features**
  - The health check endpoint `/api/v1/healthz` has been removed, affecting clients relying on this functionality.

- **Documentation Updates**
  - Swagger API documentation has been updated to reflect the new and removed features.


scripts/generate_swagger.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+set -euo pipefail
+
 PATH_TO_PROJECT_ROOT=$(git rev-parse --show-toplevel)
 SWAGGER_DIR="${PATH_TO_PROJECT_ROOT}/pkg/swagger"
 WEBUI_PATH="${PATH_TO_PROJECT_ROOT}/webui/lib/api/schema"
```
markkuhn, line 7: In the case that swagger isn't installed can we add
```
# Check if 'swag' is installed and install it if not
if ! command -v swag &> /dev/null; then
    echo "'swag' not found. Installing 'swag'..."
    go install github.com/swaggo/swag/cmd/swag@latest
    if [[ $? -ne 0 ]]; then
        echo "Failed to install 'swag'. Exiting."
        exit 1
    fi
    echo "'swag' installed successfully."
fi
```

markkuhn (approved):

scripts/generate_swagger.sh
wdbaruni: Yeah good suggestion. Sorry I merged before setting this comment

PR #4726, opened 2024-12-06
Allow to pass URL as the API Host for CLI

jamlo:
When the CLI tries to connect to the orchestrator API over HTTP, the current configuration enforces the user to specify the PORT (if not default) and if they should connect over TLS or not.

In this change, the user can specify the Host as a URL (including scheme and port), and the right connection will be made.

This is a backward compatible change, the current behavior is still valid. Though if the Host is a URL, it will take precedence over other params (port and useTLS).

Check tests in this PR for more examples.

Example, now `BACALHAU_API_HOST` can be defined as:
```
'http://bacalhau-orchestrator-node:5555' or
'https://bacalhau-orchestrator-node:4444' or
'http://1.2.3.4:3333' or
'https://1.2.3.4' or
'https://1.2.3.4:7777' 

etc...
```

Linear: https://linear.app/expanso/issue/ENG-400/allow-api-host-to-accept-a-url

## Release Notes

- **New Features**
  - Enhanced URL validation and processing for API client configuration.
  - Introduced a Docker Compose setup for orchestrator and jumpbox nodes with URL support.
  - Added HTTP URL host configuration with authentication and labeling.

- **Tests**
  - Comprehensive unit tests for URL parsing functionality.
  - New test suite for validating CLI acceptance of URL as host parameter.

These updates improve the robustness and usability of the API client and related services.


cmd/util/api.go
```diff
@@ -97,3 +105,72 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
+		port = fmt.Sprintf("%d", defaultPort)
+	} else if strings.Count(host, ":") == 1 {
+		// IPv4 or hostname with port
+		hostParts := strings.Split(host, ":")
+		processedHost = hostParts[0]
+		port = hostParts[1]
+	} else {
+		// IPv4 or hostname without port
+		processedHost = host
+		port = fmt.Sprintf("%d", defaultPort)
```
wdbaruni, line 160: can we use [net.SplitHostPort](https://pkg.go.dev/net#SplitHostPort) to simplify this logic?

wdbaruni (approved):
Looks good. Minor comment about using `net` pkg

PR #4813, opened 2025-01-16
Add Agent license inspect command

jamlo:
# New Command: `bacalhau agent license inspect`

This PR introduces a new command that allows users to inspect the license information of a Bacalhau orchestrator node without exposing the license itself. The command provides a secure way to verify license status, capabilities, and metadata while maintaining the confidentiality of the underlying license file.

### Features
- New command: `bacalhau agent license inspect`
- Supports multiple output formats (default, JSON, YAML)
- Displays key license information:
  - Product name
  - License ID
  - Customer ID
  - Expiration date
  - Version
  - Capabilities
  - Custom metadata

### Add License Manager for Node Orchestration

Introduce a LicenseManager component that handles license validation for orchestrator nodes. Key features:

- Validates node count against licensed limits
- Simple API for license verification and capability checks

Usage example:
```golang
// Initialize license manager (config is bacalhau License config)
// IF any issues, other than expiry date validity, appears during initialization, it will fail
licenseManager, err := licensing.NewLicenseManager(config)

// Get current license claims, it will be nil if no license is configured for the orchestrator
licenseClaims := licenseManager.License()

// Then these helper functions can be called on the licenseClaims struct
licenseClaims.IsExpired() // Returns only a boolean
licenseClaims.MaxNumberOfNodes() // Returns only a number
```

### Security Considerations
- Does not expose the raw license file or cryptographic material
- Only returns parsed, necessary information for verification
- Maintains license confidentiality while providing essential details

### License File Structure
The license file, fed to the orchestrator node config, uses a JSON format to support future extensibility and dynamic configuration. The structure is intentionally simple:

```json
{
    "license": "your_license_token_here"
}
```

We chose JSON format for the license file because:
- It allows for easy addition of future configuration options
- Provides a structured way to include additional metadata if needed

### Configuration
To add a license to an orchestrator, you need to configure your orchestrator node. Here's a sample configuration example:

```yaml
NameProvider: "uuid"
API:
  Port: 1234
Orchestrator:
  Enabled: true
  Auth:
    Token: "your_secret_token_here"
  License:
    LocalPath: "/path/to/your/license.json"
Labels:
  label1: label1Value
  label2: label2Value
```

Key configuration points:
- The `Orchestrator.License.LocalPath` field specifies the path to your license file
- If no license is configured, the command will return an error message saying that no license was configured 
- If the license if expired, the inspect command will return the same license details, but will not that it is expired.

### Example Usage
```bash
# Default format
$ bacalhau agent license inspect

# JSON format
$ bacalhau agent license inspect --output=json

# YAML format
$ bacalhau agent license inspect --output=yaml
```

### Example Output

For `bacalhau agent license inspect`:

```bash
Product      = Bacalhau
License ID   = 2d58c7c9-ec29-45a5-a5cd-cb8f7fee6678
Customer ID  = test-customer-id-123
Valid Until  = 2045-07-28
Version      = v1
Expired      = false
Capabilities = max_nodes=1
Metadata     = someMetadata=valueOfSomeMetadata
```

For `bacalhau agent license inspect --output=yaml`:

```yaml
capabilities:
  max_nodes: "1"
customer_id: test-customer-id-123
exp: 2384889682
iat: 1736889682
iss: https://expanso.io/
jti: 2d58c7c9-ec29-45a5-a5cd-cb8f7fee6678
license_id: 2d58c7c9-ec29-45a5-a5cd-cb8f7fee6678
license_type: standard
license_version: v1
metadata:
  someMetadata: valueOfSomeMetadata
product: Bacalhau
sub: test-customer-id-123
```

### Documentation
- Added command documentation with usage examples
- Included field descriptions in help text

Linear: https://linear.app/expanso/issue/ENG-498/license-path-configuration

## Summary by CodeRabbit

- **New Features**
	- Added a new CLI command to inspect agent license information.
	- Introduced a new API endpoint to retrieve agent license details.
	- Implemented license configuration support for orchestrator nodes.

- **Configuration**
	- Added a new configuration option for specifying local license file path.
	- Enhanced configuration to support orchestrator settings with license metadata.

- **API Enhancements**
	- Created a new method to retrieve license information via API client.
	- Updated Swagger documentation to include license-related endpoints.

- **Testing**
	- Added comprehensive integration tests for license inspection scenarios, including expired licenses.
	- Included test cases for various license configuration states and error handling.
	- Enhanced tests for validating license output formats and error messages.


pkg/publicapi/endpoint/agent/endpoint.go
```diff
@@ -138,3 +142,50 @@ func (e *Endpoint) config(c echo.Context) error {
...
+
+	// Create validator and validate license
+	validator, err := license.NewOfflineLicenseValidator()
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("failed to create license validator: %s", err))
+	}
+
+	claims, err := validator.ValidateToken(licenseFile.License)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("failed to validate license: %s", err))
```
wdbaruni, line 185: I would have the logic to load, parse and validate the license to be owned by a dedicated type that should be used by the API and by other components in the orchestrator as well, such as when introducing warnings when running with an invalid, expired and under provisioned license 

the other issue I see is that this API should return the license the orchestrator is running with and not necessary the current one deployed on disk. Meaning if the license file was updated after start the orchestrator, the orchestrator will still be running using the old license until it is restarted or if we implement hot license reload. Regardless, the API should return the current orchestrator's view of the license even if it was stale

In summary, I would prefer a dedicated license provider or something similar that loads the license at startup, unmarshall it, validate it, and keep it in-memory for other components to query and inspect

wdbaruni (commented):
Just one comment. The rest looks great

pkg/publicapi/endpoint/agent/endpoint.go
jamlo: @wdbaruni Added a new LicenseManager with description how it can be used in the PR comment.

pkg/licensing/license_manager.go
```diff
@@ -0,0 +1,100 @@
...
+	return &LicenseManager{
+		bacalhauConfig:    config,
+		licenseConfigured: licenseConfigured,
+		licenseToken:      licenseToken,
+		licenseValidator:  validator,
+	}, nil
+}
+
+// ValidateLicense validates the current license token and returns the license claims
+func (l *LicenseManager) ValidateLicense() (*license.LicenseClaims, error) {
```
wdbaruni, line 64: The validation should happen in the constructor. This method should be more like get claims or something
 similar

pkg/node/node.go
```diff
@@ -85,6 +86,7 @@ type Node struct {
 	ComputeNode    *Compute
 	RequesterNode  *Requester
 	CleanupManager *system.CleanupManager
+	LicenseManager *licensing.LicenseManager
```
wdbaruni, line 89: We don't need to store the manager in the node struct

pkg/publicapi/endpoint/agent/endpoint.go
```diff
@@ -138,3 +143,30 @@ func (e *Endpoint) config(c echo.Context) error {
...
+//
+//	@ID			agent/license
+//	@Summary	Returns the details of the current configured orchestrator license.
+//	@Tags		Ops
+//	@Produce	json
+//	@Success	200	{object}	apimodels.GetAgentLicenseResponse
+//	@Failure	500	{object}	string
+//	@Router		/api/v1/agent/license [get]
+func (e *Endpoint) license(c echo.Context) error {
+	if e.licenseManager == nil {
```
wdbaruni, line 157: Let's add null validation in the constructor and fail early

wdbaruni (changes requested):

pkg/publicapi/endpoint/agent/endpoint.go
jamlo: @wdbaruni since `licenseManager` is a pointer, I think it is better to validate for null each time to avoid panics if another entity manipulated it somehow.

pkg/licensing/license_manager.go
jamlo: @wdbaruni License validation is time dependent, i.e. the license has an expiry date. If we called validate in the constructor once and only return the claims after that, then the license will be valid for as long as the instance is alive. Currently we instantiate a license manager instance when. node started. It will read the license file once, add the token to memory, a whenever `ValidateLicense` or `ValidateLicenseWithNodeCount` are called, it will check the license signature and expiry date. 

Having the the validation in the constructor will defeat that purpose, unless we are instantiating the manager each time we want to validate the license and we need to reading that license file each time.

pkg/licensing/license_manager.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"strconv"
+
+	"github.com/bacalhau-project/bacalhau/pkg/config/types"
+	"github.com/bacalhau-project/bacalhau/pkg/lib/license"
+)
+
+// LicenseManager handles license management and validation for the node
+type LicenseManager struct {
+	bacalhauConfig    *types.Bacalhau
```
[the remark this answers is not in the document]
jamlo: @wdbaruni I passed the config and not just the license path to support the ability for having multiple sources of the license in the future, maybe I can pass the License section of the config and not just the path itself, or the whole config.

Regarding allowing the NewNode to decide which instance of the manager to instantiate based on the license configuration, isn't that leaking the detials of the license manager to the node? Why would the NewNode need to know about the inner details of the license working? especially that for now we have only the license path as a configuration, but when we add more configurations, it might be a mess.

pkg/node/node.go
jamlo: @wdbaruni I added it to the node since it will be a "Node Wide" activity to check for the license. I am assuming that the license manager will be used in many places in the node, so making it easier on the caller side.

I can remove it, but each caller that is not instantiated in the NewNode needs to figure out how to get this manager.

pkg/publicapi/endpoint/agent/endpoint.go
wdbaruni: That is unnecessary complexity since the variable is immutable. We still need to validate at the constructor level even if you feel we need to validate here as well

pkg/node/node.go
wdbaruni: The fields we keep and expos at the node level are mainly for testing purposes and to make them accessible during certain tests. We should minimize those instances and just use the Node to instantiate and inject dependencies

pkg/licensing/license_manager.go
wdbaruni: I see. I think the method names are a little confusing. As a user of the license manager, I would expect to have these methods and behaviours:
- `License` which will get all the license information, even if it was expired. For example callers to `bacalhau agent license` should get a license, see when it has expired, and what claims it had. We shouldn't just fail
- `Expired` return true if the license has expired. Or maybe just `IsValid` and return true if the license is still valid
- `Claims` return the claims, and empty claims if expired
- `NodeCount` returns the number of nodes it supports. You can delay this until we implement orchestrator warnings

pkg/licensing/license_manager.go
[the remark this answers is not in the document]
wdbaruni: Yeah we shouldn't pass the whole config. Either pass the dedicated license section of the config, or have a separate params type for the license manager for indirection, or just pass the path for now. My preference is to just pass the config and we can decide in the future when more details needs to be sent to the manger

NewNode is where dependencies instantiated and BacalhauConfig is parsed to know what actual implementations we should instantiate. 
- If you were to have two implementations for the LicenseManager, then NewNode is where you decide which one to create. It is part of its responsibility so others parts of the system are simplified. 
- If you were to have a single implementation that handles noop license as well, then thats a different story

pkg/licensing/license_manager.go
```diff
@@ -0,0 +1,100 @@
...
+		// Will not inspect the license itself.
+		licenseData, err := os.ReadFile(licensePath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read license file: %w", err)
+		}
+
+		// Verify the JSON structure
+		var licenseFile struct {
+			License string `json:"license"`
+		}
```
wdbaruni, line 48: I don't like having inline structs for such behaviour. First, why can't we have license content at the root of the json object instead of having a `license` field? If it is a must, then lets introduce a `LicenseWrapper` type and not make it inline

jamlo: @wdbaruni 

> First, why can't we have license content at the root of the json object instead of having a license field?
Not sure what you meant here, but the license content is already at the root of the json object. The Json file structure of the license:

```json
{
  "license": "the-license-jwt-token"
}
```

> If it is a must, then lets introduce a LicenseWrapper type and not make it inline
Hmm, It is not used anywhere else, do you think it is still worth it ? 
The unit tests already cover all the edge cases for the format.

wdbaruni: if the file only contains the license jwt token and nothing else, why do you need a json wrapping the token?

pkg/node/node.go
jamlo: LicenseManager removed from Node.

pkg/licensing/license_manager.go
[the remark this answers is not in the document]
jamlo: Only the license config is being passed right now.

pkg/licensing/license_manager.go
jamlo: @wdbaruni I would rather have a JSON object as the structure of a license even if it has one item in it. It will easily allow for any future additions or migrations, compared to just an opaque string.

pkg/publicapi/endpoint/agent/endpoint.go
jamlo: Done. Null validation in the constructor now.

pkg/lib/license/license_validator_test.go
```diff
@@ -456,27 +476,225 @@ func TestNewOfflineLicenseValidator(t *testing.T) {
...
+		LicenseType:    "standard",
+		LicenseID:      "license_1",
+		CustomerID:     "customer_1",
+		Capabilities:   map[string]string{"max_nodes": "invalid"},
+	})
+
+	claims, err := validator.Validate(token.Raw)
+	require.ErrorContains(t, err, "license validation error: token is malformed: token contains an invalid number of segments")
+	assert.Nil(t, claims)
+}
```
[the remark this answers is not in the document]
jamlo: N/A for test

test_integration/common_assets/nodes_configs/16_orchestrator_config_with_expired_license.yaml
```diff
@@ -0,0 +1,12 @@
+NameProvider: "uuid"
+API:
+  Port: 1234
+Orchestrator:
+  Enabled: true
+  Auth:
+    Token: "i_am_very_secret_token"
+  License:
+    LocalPath: "/bacalhau_integration_tests/common_assets/licenses/test-license-valid-but-expired.json"
```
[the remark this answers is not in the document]
jamlo: This path is for inside the containers test.

test_integration/common_assets/nodes_configs/16_orchestrator_config_with_expired_license.yaml
```diff
@@ -0,0 +1,12 @@
+NameProvider: "uuid"
+API:
+  Port: 1234
+Orchestrator:
+  Enabled: true
+  Auth:
+    Token: "i_am_very_secret_token"
```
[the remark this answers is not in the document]
jamlo: N/A test

wdbaruni (approved):

PR #4841, opened 2025-02-08
Fix race conditions when starting dind containers

wdbaruni:
Fixes race conditions when multiple dind containers are started at the same time, which is usually the case with docker-compose

## Summary by CodeRabbit

- **Chores**
	- Enhanced the container startup process for improved reliability.
	- Introduced a random initial delay to help prevent simultaneous startup issues.
	- Implemented a robust retry mechanism with better error handling and cleanup for smoother operation.

### DIFF
--- a/docker/bacalhau-dind/entrypoint.sh
+++ b/docker/bacalhau-dind/entrypoint.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
+
 # Check for privileged mode by testing iptables access
 if ! iptables -L >/dev/null 2>&1; then
     echo "ERROR: This container must be run with --privileged flag"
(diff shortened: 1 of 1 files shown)


docker/bacalhau-dind/entrypoint.sh
```diff
@@ -1,25 +1,81 @@
...
+HOSTNAME=$(hostname)
+INITIAL_DELAY_MS=$(awk -v seed="$(echo $HOSTNAME | cksum | cut -d' ' -f1)" 'BEGIN{srand(seed);print int(rand()*2000)}')
+echo "Adding initial startup delay of ${INITIAL_DELAY_MS} milliseconds..."
+sleep "$(awk "BEGIN{print ${INITIAL_DELAY_MS}/1000.0}")"
+
+MAX_RETRIES=5
+ATTEMPT=1
+TIMEOUT=45
+
+start_docker_daemon() {
```
aronchick, line 22: @wdbaruni stupid question - isn't the whole point of docker in docker is that you use the daemon that's already on the host?

wdbaruni: Actually Docker-in-Docker (using docker:dind) runs a separate Docker daemon inside the container, completely isolated from the host's daemon. 

What you're thinking of is Docker-outside-of-Docker (DooD), where you mount the host's Docker socket (`/var/run/docker.sock`) to use the existing host daemon. 

The `docker:dind` image specifically sets up a new, isolated daemon - that's why it needs the `--privileged` flag. If you want to use the host's daemon instead, you'd use the regular `docker` image and mount the socket.

Quick example:
- DinD (new daemon): `docker run --privileged docker:dind`
- DooD (host daemon): `docker run -v /var/run/docker.sock:/var/run/docker.sock docker`

PR #4846, opened 2025-02-11
adding how to run docker with Bacalhau

aronchick:


aronchick:
@wdbaruni @jamlo almost sure this is wrong :) can you fix? Thanks!

README.md
```diff
@@ -83,6 +84,55 @@ bacalhau get 63d08ff0..... # make sure to use the right job id from the docker r
...
+
+You can find all the containers we support [here](https://github.com/bacalhau-project/bacalhau/pkgs/container/bacalhau).
+
+To use a container, you can use the following command with Docker:
+
+```bash
+docker run ghcr.io/bacalhau-project/bacalhau:latest
+```
+
+This will start a container with Bacalhau already installed.
```
wdbaruni, line 99: Reading this sounds like it will start bacalhau in a container, but it will just call `bacalhau` without any flags or subcommands as that is the default entrypoint. It will just print the help output and exit

The actual use should be:
```
docker run ghcr.io/bacalhau-project/bacalhau:latest <sub-command>

# checking the bacalhau version
docker run ghcr.io/bacalhau-project/bacalhau:latest version

# listing jobs
docker run ghcr.io/bacalhau-project/bacalhau:latest job list

# listing nodes
docker run ghcr.io/bacalhau-project/bacalhau:latest node list

# running an orchestrator
docker run ghcr.io/bacalhau-project/bacalhau:latest serve --orchestrator

# running a compute node with config
docker run ghcr.io/bacalhau-project/bacalhau:latest serve --compute -c <path>

...
```



or if you can do `docker run -it --entrypoint bash ghcr.io/bacalhau-project/bacalhau:latest`, and then call bacalhau commands internally

wdbaruni (changes requested):

README.md
aronchick: So this is a good point - however, in thinking about simplification, there's kind of the following.

What do we think the majority of folks will do with this? And then how do we make it so the default flags are correct?

I _THINK_ in priority order, what we would want to enable would be:

1. Run a container as a compute node
2. Run a container as a compute node with orchestrator specified (no auth)
3. Run a container as a compute node with orchestrator specified (auth)
4. Run a container as a compute node with full new config
5. Run a container as a CLI container
6. Run an orchestrator

So, by this logic, I think the defaults should be to have an entrypoint.sh script, and then execute using the defaults. It should have the following in it:
```
/etc/bacalhau/config.yaml - a file with the default config we have
/etc/node-info - a file with node-info that are read and put into labels automatically
/entrypoint.sh - a script that loads these items in by default, as well as pings any telemetry if we have it enabled
```

Then executing 1-6 above looks like:
```
1. docker run ghcr.io/bacalhau-project/bacalhau:latest # joins the public network? not sure
2. docker run -e BACALHAU_COMPUTE_ORCHESTRATORS=<IP> ghcr.io/bacalhau-project/bacalhau:latest # joins an IP address network
3. docker run -e BACALHAU_COMPUTE_ORCHESTRATORS=<IP> BACALHAU_COMPUTE_AUTH_TOKEN=<TOKEN> ghcr.io/bacalhau-project/bacalhau:latest # access the orchestrator with a token
4. docker run -v $(pwd)/config.yaml:/etc/bacalhau/config.yaml ghcr.io/bacalhau-project/bacalhau:latest
5. docker run ghcr.io/bacalhau-project/bacalhau:latest job list --api-host=<IP>
OR
docker run -e API_HOST=<IP> ghcr.io/bacalhau-project/bacalhau:latest job list 
```

And all of the above could also have a node-info injected with `-v $(pwd)/node-info:/etc/node-info`. We could also enable picking up `-e BACALHAU_LABEL='[a=1, b=2]'` etc.

Thoughts?

PR #4888, opened 2025-03-17
Bacalhau Auth v2: Basic Auth, API Keys, and SSO Support

jamlo:
# Enhanced Authentication System - Draft PR Message

*Note: More documentation coming.*

## Overview - More Details Coming
This PR introduces a new authentication system while maintaining backward compatibility with existing authentication methods. The implementation adds support for API keys, username/password authentication, and SSO login via OAuth2 Device Code flow.

### Environment Variable Authentication
- Added support for the following environment variables:
  - `BACALHAU_API_KEY` - Sent as a Bearer token in the Authorization header
  - `BACALHAU_API_USERNAME` and `BACALHAU_API_PASSWORD` - Used for Basic Authentication

### Authentication Rules
- Either `BACALHAU_API_KEY` OR the combination of `BACALHAU_API_USERNAME` and `BACALHAU_API_PASSWORD` can be defined, but not both simultaneously
- If no environment variables are defined, the system falls back to the previous authentication method (when supported by the server)

### SSO Authentication
- Introduced a new command: `bacalhau auth sso`
- Enables Single Sign-On via OAuth2 Device Code flow
- Configuration is handled server-side; CLI automatically detects and uses the appropriate flow
- Fully backward compatible with existing authentication methods

---

## Summary by CodeRabbit

- **New Features**
  - Introduced new CLI commands for password hashing and SSO login with an OAuth2 device flow for improved authentication.
  - Added an API endpoint that returns node OAuth2 configuration details.
  - Enhanced authentication flows to support both legacy and modern OAuth2 mechanisms, with updated environment guidance.
  - New configuration options for basic and API key authentication are now available.

- **Tests**
  - Expanded unit and integration test suites to cover CLI, API, and authentication flows, ensuring robust security and reliability.


cmd/util/api.go
```diff
@@ -79,26 +67,68 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
+		log.Debug().Msg("Using SSO authentication credentials")
+	} else {
+		resolvedAuthToken, err = ReadTokenFn(legacyAuthTokenFilePath, baseURL)
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to read access tokens – API calls will be without authorization")
+		}
+	}
+
+	newAuthenticationFlowEnabled := apiKeyOrBasicAuthFlowEnabled || existingSSOCredential != nil
+	skipAuthentication := cmd.Use == "sso" || cmd.Use == "version" || cmd.Use == "alive"
```
wdbaruni, line 117: shouldn't we avoid resolving the credentials if we are in these commands? Also wouldn't it be better to create a separate `GetUnauthenticatedClient` helper method and use it when needed, such as in `alive` command instead of handling it hear?

cmd/util/api.go
```diff
@@ -79,26 +67,68 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
+	apiKeyOrBasicAuthFlowEnabled, credentialScheme, credentialString, err := resolveAuthCredentials(
+		apiAuthAPIKey,
+		basicAuthUsername,
+		basicAuthPassword,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("authentication error: %v", err)
+	}
+
+	legacyAuthTokenFilePath, err := cfg.AuthTokensPath()
```
wdbaruni, line 82: am I reading this correctly that the legacy auth logic has higher precedence than sso tokens?

cmd/util/api.go
```diff
@@ -79,26 +67,68 @@ func GetAPIClientV2(cmd *cobra.Command, cfg types.Bacalhau) (clientv2.API, error
...
 	}
 
 	return clientv2.NewAPI(
 		&clientv2.AuthenticatingClient{
-			Client:     clientv2.NewHTTPClient(base, opts...),
-			Credential: existingAuthToken,
+			Client:                       clientv2.NewHTTPClient(baseURL, opts...),
+			Credential:                   resolvedAuthToken,
+			NewAuthenticationFlowEnabled: newAuthenticationFlowEnabled,
+			SkipAuthentication:           skipAuthentication,
```
wdbaruni, line 129: nit: New will be old at one point :) 
but its fine as I think this is temporary and will be removed in next release when we drop support for legacy auth

pkg/authz/basic_auth_authorizer.go
```diff
@@ -0,0 +1,142 @@
...
+	if subtle.ConstantTimeCompare([]byte(user.Password), []byte(password)) != 1 {
+		return types.AuthUser{}, false, errors.New("invalid basic auth credentials")
+	}
+
+	// Authentication successful
+	return user, true, nil
+}
+
+func NewBasicAuthAuthorizer(
+	nodeID string,
```
wdbaruni, line 59: why do we need to the nodeID in auth?

pkg/authz/capabilities.go
```diff
@@ -0,0 +1,139 @@
...
+const (
+	ResourceTypeNode  ResourceType = "node"
+	ResourceTypeJob   ResourceType = "job"
+	ResourceTypeAgent ResourceType = "agent"
+	ResourceTypeOpen  ResourceType = "open"
+)
+
+// GetRequiredCapability determines the required capability for a specific resource type and HTTP method
+func (c *CapabilityChecker) GetRequiredCapability(resourceType ResourceType, method string) string {
+	isReadOperation := method == http.MethodGet
```
wdbaruni, line 30: should we add head as well? we don't have handler for head today, and we should just make sure we either return 404 or unaotharized access with `read:<resource>` capability required instead of saying `write:<resource>`

pkg/authz/capabilities.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		return "write:job"
+	case ResourceTypeAgent:
+		if isReadOperation {
+			return "read:agent"
+		}
+		return "write:agent"
+	default:
+		// If no resource type matched, default to requiring node admin for safety
+		return "write:node"
```
wdbaruni, line 50: I don't like this default behaviour

pkg/authz/capabilities.go
```diff
@@ -0,0 +1,139 @@
...
+// GetDefaultEndpointPermissions returns the default endpoint to permission mapping
+func GetDefaultEndpointPermissions() map[string]string {
+	return map[string]string{
+		"/api/v1/auth":    "open",
+		"/api/v1/version": "open",
+
+		"/api/v1/agent":            "agent",
+		"/api/v1/agent/alive":      "open",
+		"/api/v1/agent/version":    "open",
+		"/api/v1/agent/authconfig": "open",
```
wdbaruni, line 134: what is the difference between `/auth` and `agent/authconfig`?

pkg/authz/entry_point_authorizer.go
```diff
@@ -0,0 +1,335 @@
...
+	apiKeyAuthorizer    Authorizer
+	jwtAuthorizer       Authorizer
+}
+
+// validateUser validates a user configuration and returns an error if it's invalid
+func (a *entryPointAuthorizer) validateUser(user types.AuthUser) error {
+	// Validation: Check that user has a non-empty Alias
+	if user.Alias == "" {
+		return fmt.Errorf("user has empty Alias, Alias is mandatory")
+	}
```
wdbaruni, line 46: I thought alias is optional?

pkg/config/types/auth.go
```diff
@@ -34,3 +37,28 @@ type AuthenticatorConfig struct {
 	Type       string `yaml:"Type,omitempty" json:"Type,omitempty"`
 	PolicyPath string `yaml:"PolicyPath,omitempty" json:"PolicyPath,omitempty"`
 }
+
+type AuthUser struct {
+	Alias        string       `yaml:"Alias,omitempty" json:"Alias,omitempty"`
+	Username     string       `yaml:"Username,omitempty" json:"Username,omitempty"`
+	Password     string       `yaml:"Password,omitempty" json:"Password,omitempty"`
```
wdbaruni, line 44: Are you using a single field for plain and hashed password? That will be nice

pkg/node/node.go
```diff
@@ -54,10 +54,27 @@ type NodeConfig struct {
...
+	//	c.BacalhauConfig.API.BuiltInAuth.BasicAuth.Users.SrcFilePath != "" ||
+	//	len(c.BacalhauConfig.API.BuiltInAuth.BasicAuth.Admins.UserConfig) > 0 ||
+	//	len(c.BacalhauConfig.API.BuiltInAuth.BasicAuth.Users.UserConfig) > 0 {
+	//	builtInAuthDefined = true
+	//}
+	//
+	//if authDefined && builtInAuthDefined {
+	//	mErr = errors.Join(mErr, fmt.Errorf("both Auth and BuiltInAuth are defined in the configuration; only one can be used at a time"))
+	//}
+
```
wdbaruni, line 77: do we need this?

pkg/node/node.go
```diff
@@ -269,6 +286,18 @@ func createAPIServer(cfg NodeConfig, userKey *baccrypto.UserKey) (*publicapi.Ser
...
+	if len(cfg.BacalhauConfig.API.Auth.Users) > 0 || cfg.BacalhauConfig.API.Auth.Oauth2.ProviderID != "" {
+		// If new auth configuration detected, use new authorizer
+		chosenAuthorizer, err = authz.NewEntryPointAuthorizer(ctx, cfg.NodeID, cfg.BacalhauConfig.API.Auth)
+		if err != nil {
+			return nil, fmt.Errorf("error initializing users: %s", err.Error())
+		}
+	} else {
+		chosenAuthorizer = authz.NewPolicyAuthorizer(authzPolicy, userKey.PublicKey(), cfg.NodeID)
+	}
+
```
wdbaruni, line 300: what is the differnce between this and the `NewStandardAuthenticatorsFactory` in the factories file?

pkg/publicapi/client/v2/client.go
```diff
@@ -482,6 +482,14 @@ type AuthenticatingClient struct {
...
 	// flow using the passed Auth API.
 	Authenticate func(context.Context, *Auth) (*apimodels.HTTPCredential, error)
+
+	// NewAuthenticationFlowEnabled is true when new auth flow is detected.
+	// This is kept here for backward compatibility reasons only
+	NewAuthenticationFlowEnabled bool
+
+	// SkipAuthentication is true when new auth flow is detected.
+	// This is kept here for backward compatibility reasons only
+	SkipAuthentication bool
```
wdbaruni, line 492: these are a bit confusing. Do you need two booleans? Can the names be more accurate?

pkg/publicapi/endpoint/agent/endpoint.go
```diff
@@ -171,3 +172,23 @@ func (e *Endpoint) license(c echo.Context) error {
...
+//	@ID			agent/authconfig
+//	@Summary	Returns the OAuth2 configuration of the node.
+//	@Tags		Ops
+//	@Produce	json
+//	@Success	200	{object}	apimodels.GetAgentNodeAuthConfigResponse
+//	@Failure	500	{object}	string
+//	@Router		/api/v1/agent/authconfig [get]
+func (e *Endpoint) nodeAuthConfig(c echo.Context) error {
+	cfg, err := e.bacalhauConfig.Copy()
+	if err != nil {
```
wdbaruni, line 187: do you need to copy?

pkg/sso/sso.go
```diff
@@ -0,0 +1,100 @@
...
+// This is our own representation of the device code response
+type DeviceCodeResponse struct {
+	DeviceCode              string `json:"device_code"`
+	UserCode                string `json:"user_code"`
+	VerificationURI         string `json:"verification_uri"`
+	VerificationURIComplete string `json:"verification_uri_complete"`
+	ExpiresIn               int    `json:"expires_in"`
+	Interval                int    `json:"interval"`
+	Message                 string `json:"message"`
+}
```
wdbaruni, line 24: not sure if it matters, but all our APIs are following title case

pkg/publicapi/endpoint/agent/endpoint.go
```diff
@@ -53,6 +53,7 @@ func NewEndpoint(params EndpointParams) (*Endpoint, error) {
 	g.GET("/debug", e.debug)
 	g.GET("/config", e.config)
 	g.GET("/license", e.license)
+	g.GET("/authconfig", e.nodeAuthConfig)
```
wdbaruni, line 56: you might want to verify how some of the commands will work against a compute node. Meaning calling `bacalhau agent version` for example against a compute only node

wdbaruni (approved):
LGTM. Just minor comments. Feel free to merge after addressing them

pkg/config/types/auth.go
jamlo: Correct, we are using the same field for plain passwords and hashed passwords (bcrypt ones)

pkg/sso/sso.go
jamlo: @wdbaruni these are forced on us by the RFC https://datatracker.ietf.org/doc/html/rfc8628#section-3.2

cmd/util/api.go
jamlo: No, SSO tokens have higher precedence compared to legacy tokens.

pkg/authz/entry_point_authorizer.go
jamlo: It is optional now.
Made it mandatory for quick testing while working on the PR.

pkg/node/node.go
jamlo: This was a TODO item for me to add some validation, which was added now.

pkg/publicapi/endpoint/agent/endpoint.go
jamlo: Integration Test Added

pkg/authz/capabilities.go
jamlo: `/auth` is the legacy API endpoint to get a session token and auth methods. 
`agent/authconfig` is new API to get the auth configuration of a node, including oauth2

pkg/authz/capabilities.go
jamlo: I'll be changing that

pkg/publicapi/client/v2/client.go
jamlo: TODO: Fix comments

pkg/publicapi/endpoint/agent/endpoint.go
jamlo: Todo

pkg/node/node.go
jamlo: Todo

pkg/authz/basic_auth_authorizer.go
jamlo: This is for future node specific JWT tokens (still thinking about the best way to add it)

cmd/util/api.go
jamlo: Todo

cmd/util/api.go
jamlo: Todo

Issue #4224, opened 2024-07-11
Bad error on "failed validation"


aronchick:
We're still passing raw go failures through.
```
the job provided is invalid: task Run over share validation failed: artifact validation failed: missing artifact target
```

Job:
```
Name: Run Over Share
Namespace: science
Type: ops
Count: 1
Tasks:
  - Name: Run over share
    InputSources:
      - Source:
          Type: localDirectory
          Params:
            SourcePath: /mnt/azureshare
            ReadWrite: true
          Target: /azureshare
    Engine:
      Type: docker
      Params:
        # the docker container that will download videos and perform inference
        Image: docker.io/bacalhauproject/python-runner:0.0.4
        EnvironmentVariables:
          - COLOR=blue
          - COMMAND="print('hello world')"
    Network:
      # this job requires networking to download content from buckets
      Type: full
    Resources:
      # dependent on compute nodes, this is based on e2-standard-8
      CPU: "1"
      Memory: "1GB"
      Disk: "1GB"

```

wdbaruni:
still not generating good errors with v1.5.0
```
bacalhau job run run-over-share.yaml
Error: the job provided is invalid: task Run over share validation failed: artifact validation failed: missing artifact target
```

PR #4942, opened 2025-04-18
Improve job validation error messages (#4224)

Olg-K:
Our job spec verification sometimes return pretty cryptic error messages to the user. This is a quick fix to make some of those messages more readable.

Ideally we should plan a long-term solution for this. Maybe introduce a JSON schema for Job Spec and use one of the existing schema validators.

Additionally, we can look into job spec validation on client side and fail early if given yaml/json is invalid. We won't be able to check every rule (some require compute node info, like supported engines/hardware) but we can totally verify that the job satisfies the schema. For example https://github.com/bacalhau-project/bacalhau/issues/4224 is a schema error so can be caught before even sending any requests to the orchestrator.

### Before
```
Error: the job provided is invalid: task Task1 validation failed: artifact validation failed: missing artifact target
```

### After
```
Error: the job provided is invalid: task Task1 validation failed: invalid input sources: input source is missing a target
```

Addresses https://github.com/bacalhau-project/bacalhau/issues/4224

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation error messages for input sources and tasks to be more specific and consistent.
- **Tests**
  - Added a new test case for missing input source targets.
  - Updated existing test cases to match revised error messages.

### DIFF
--- a/pkg/models/input_source.go
+++ b/pkg/models/input_source.go
@@ -57,10 +57,10 @@ func (a *InputSource) Validate() error {
 		return nil
 	}
 	mErr := errors.Join(
-		validate.NotBlank(a.Target, "missing artifact target"),
+		validate.NotBlank(a.Target, "input source is missing a target"),
 	)
 	if err := a.Source.Validate(); err != nil {
-		mErr = errors.Join(mErr, fmt.Errorf("invalid artifact source: %w", err))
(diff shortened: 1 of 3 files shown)


wdbaruni (approved):
LGTM. If we start returning `bacerrors` at a certain layer, we would get this long chain of wrapped errors and hopefully the error message, and optionally a hint, can be more readable

PR #4956, opened 2025-04-22
Ignore webui build artifacts

Olg-K:
Exclude some WebUI build artifacts from from being under version control. These files are generated during build and do not need to be in the repository.

The changes here are the result of the following
```
git rm -rf webui/build/*; echo "webui/build/*" >> .gitignore
```

## Summary by CodeRabbit

- **Chores**
  - Updated the ignore list to exclude all files in the web UI build directory from version control.
  - Removed all web UI build artifacts, including HTML, JavaScript, CSS, and manifest files.
- **User Impact**
  - The web UI and its related static assets are no longer available in the build output.


Olg-K:
Requires more work due to build artifacts being used by some of the integ tests. They are being used via `go:embed` and changing requires changes in the webui webserver.

Issue #3922, opened 2023-10-24
localDirectory storage doesn't create directoreis/files if they don't exist


wdbaruni:
A common use case for `localDirectory` storage spec is to mount local files in ReadWrite access so they can be used for checkpointing. The problem is `localDirectory` will fail if the file doesn't already exist, even if `ReadWrite` was true.

We either need a better way to allow long running executions to persist their state locally across retries without asking users to define `localDirectory` InputSource, or/and we should create directories/files if they are missing and if `ReadWrite` was true

PR #4961, opened 2025-04-23
Add automatic volume creation for local input source

Olg-K:
## Local Storage Provider Enhancements

* `localDirectory` input source renamed to `local`. The old name remains valid and supports all new functionality
* Local storage provider now can create empty files/directories when source paths don't exist, instead of failing
* Added new `CreateAs` property to Local InputSource Params to control file/directory creation behaviour
* Documentation updates are in https://github.com/bacalhau-project/docs/pull/108
## Background

Currently, when a local input source (file or directory) specified in a Job spec doesn't exist on a compute node, the node will not bid for execution. This change allows compute nodes to bid if they can create an empty directory or file at the specified path. When such a bid is accepted, the compute node will create the empty file/directory and mount it normally.

## `CreateAs` Parameter

A new optional `CreateAs` property can now be used in `InputSource` `Params` with the following values:
- `nocreate` (default): The driver will not attempt to create anything
- `dir`: The driver will create an empty directory
- `file`: The driver will create an empty file

Both files and directories are created with `-rwx------@` permissions.
**Important: To enable automatic creation, `ReadWrite` must be set to `true`.**

Example:
```yaml
InputSources:
  - Source:
      Type: "local"
      Params:
        SourcePath: "/tmp/empty_file"
        ReadWrite: true
        CreateAs: "file"
    Target: "/config_file"
```

Addresses https://github.com/bacalhau-project/bacalhau/issues/3922

### Revision History
Originally `CreateAs` had a different default value - `infer`.  With `infer` strategy the driver would attempt to determine whether to create a file or directory based on the source path. After further considerations we decided that creating empty file/directory as the **input** is rather a niche use-case and it should not be the default, therefore `infer` value was completely removed. It is still possible to make Bacalhau create empty input volumes by specifying either `file` or `dir` explicitly. 

## Summary by CodeRabbit

- **New Features**
  - Introduced configurable strategies for creating local storage paths, allowing explicit control over creation as file, directory, or no creation.
  - Added detailed error handling and validation for local storage creation strategies.
  - Enabled local storage provider to create volumes if they do not exist when permitted by job specifications.

- **Bug Fixes**
  - Enhanced compatibility and support for legacy local storage naming conventions.

- **Refactor**
  - Unified and modernized the local storage implementation, consolidating previous local directory logic.
  - Deprecated the old local directory storage type in favor of a new, streamlined local storage type.
  - Updated storage source naming to use a new local storage type with deprecation of the old local directory type.
  - Updated tests and internal references to replace old local directory storage with the new local storage package.

- **Tests**
  - Added comprehensive unit tests for local storage creation strategies and provider logic.
  - Removed outdated tests for the deprecated local directory storage provider.


pkg/storage/local/storage.go
```diff
@@ -63,7 +67,14 @@
...
-		return 0, errors.New("volume does not exist")
+		// if the volume does not exist but we can create it, we should return 0.
+		// we only know that it's going to be an empty file/directory
+		if source.ReadWrite {
+			return 0, nil
+		}
+		// if the volume does not exist and we can't create it, return an error
+		return 0, bacerrors.Newf("volume %s does not exist and read-write access is not allowed", source.SourcePath).
+			WithHint("If you want the job to create the volume, set the input Source ReadWrite property to true")
 	}
```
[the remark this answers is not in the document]
Olg-K: Also fixed and unit tests added.

pkg/storage/local/storage.go
```diff
@@ -133,5 +150,81 @@
...
+
+	// create the file
+	fileHandle, err := os.Create(filePath)
+	if err != nil {
+		return bacerrors.Wrapf(err, "could not create source file at %s", filePath)
+	}
+	defer fileHandle.Close()
+	err = os.Chmod(filePath, util.OS_USER_RWX)
+	if err != nil {
+		return bacerrors.Wrapf(err, "could not set permissions for source file at %s", filePath)
```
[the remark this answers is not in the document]
Olg-K: Great suggestion, although there is no need to open the file for writing. At this stage we only need to create an empty file and do not write anything to it.

Fixed.

pkg/storage/local/storage.go
```diff
@@ -45,7 +47,9 @@ func (driver *StorageProvider) HasStorageLocally(_ context.Context, volume model
 	}
 
 	if _, err := os.Stat(source.SourcePath); errors.Is(err, os.ErrNotExist) {
-		return false, nil
+		// If the volume does not exist, we should look into whether we can create it.
+		// In order to be able to create it, ReadWrite must be true.
+		return source.ReadWrite, nil
 	}
 	return true, nil
 }
```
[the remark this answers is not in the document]
Olg-K: This is a good suggestion. Even though `CreateAs` will be the same for every compute node, they might have different state of their respective file systems (e.g. SourcePath exists on one node and doesn't on another). Which will affect the outcome of the bidding. This is a legitimate scenario, for example if a user wants to only run a job on nodes that already have a certain file. In this case they would use `nocreate` to exclude nodes that do not have the file.

Fixed and added a unit test for this case.

pkg/storage/local/create_strategy.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Create as directory
+	Dir CreateStrategy = "dir"
+
+	// Create as file
+	File CreateStrategy = "file"
+
+	// Don't create anything
+	NoCreate CreateStrategy = "nocreate"
+)
```
wdbaruni, line 25: nit: should this be `noCreate` or `NoCreate`?

pkg/storage/local/storage.go
```diff
@@ -63,7 +68,16 @@ func (driver *StorageProvider) GetVolumeSize(_ context.Context, _ *models.Execut
...
+		// if the volume does not exist but we can create it, we should return 0.
+		// we only know that it's going to be an empty file/directory at the start of the job
+		if source.ReadWrite && source.CreateAs != NoCreate {
+			return 0, nil
+		}
+
+		// if the volume does not exist and we can't create it, return an error
+		return 0, bacerrors.Newf("volume %s does not exist and creation is not allowed", source.SourcePath).
+			WithHint("If you want the job to create the volume, set ReadWrite to true and CreateAs to one of [%s]",
+				strings.Join(PermissiveCreateStrategies(), ", "))
```
wdbaruni, line 80: nit: for the error messages, should we use `source path` or `local path` instead of `volume`?

wdbaruni (approved):
LGTM: just couple of minor comments

pkg/storage/local/create_strategy.go
Olg-K: I just wanted to keep all lowercase for simplicity.

pkg/storage/local/storage.go
Olg-K: `volume` term is used (somewhat) consistently across all input providers (and also docs) so I wanted to keep it here as well.

pkg/storage/local/create_strategy.go
wdbaruni: I thought I left a comment about supporting case insensitivity, but I might've missed that. Similar to all other specs, we should be case insensitive when parsing and decoding the job spec. Meaning `NoCreate`, `noCreate`, `nocreate` should all be treated the same.
https://github.com/bacalhau-project/bacalhau/blob/5de7ab88d80859d384f51c92d2f64cbb6fadab94/pkg/models/constants.go#L58

For readability though when printing things and in our docs, `nocreate` is not really readable.

Olg-K: ok, I made the spec decoding case-insensitive and also made sure that whenever `nocreate` is shown to the user it looks like `noCreate`.
```
Hint: CreateAs must be one of [dir, file, noCreate]
```

